PR #17, opened 2021-05-14
Add available facets as filters

dhmacs:
This PR introduces the business logic to display all facets accessible from the Algolia index. This will be useful for @sterlinghirsh to debug the Algolia index and see if all necessary facets are visible in the collection.

This PR does **NOT** contain: 
- Facet specific implementations (like the price slider), the UI should be considered a draft
- Localization of facets

## QA

1. Visit the Vercel preview
2. Visit a few collections that contain products and verify that facets are dynamically populated based on the collection

> ⚠️ Count badges next to facet values are displayed for debug purposes and might display wrong values. When the filtering behavior will be throughly debugged these badges will be removed.


libs/algolia/useFacets.ts
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
```
davidrans, line 1: I think we usually avoid `import *` to reduce the number of potential naming conflicts, and make it easier to see where things are being used, but I'm not too familiar with TS conventions or best practices when importing third party libs like this so feel free to ignore.

dhmacs: `import * as React from 'react'` is recommended by React core team ([see this](https://twitter.com/dan_abramov/status/1308739731551858689?s=20)).

That being said, if you are referring to named exports vs default exports then the reason I do this is to make it easier to see that I'm using a React primitive vs a custom primitive, but that's just my opinion

libs/algolia/useFacets.ts
```diff
@@ -0,0 +1,23 @@
...
+export function useFacets(): Facet[] {
+   const { searchResult } = useSearchContext();
+
+   const facets = React.useMemo<Facet[]>(() => {
+      return Object.keys(searchResult.facetsByName).map((facetName) => {
+         return {
+            name: facetName,
+            itemsByValue: searchResult.facetsByName[facetName],
+         };
+      });
```
davidrans, line 19: Minor, but I think you could simplify this to:
```suggestion
      return searchResult.facetsByName.map((facet, facetName) => {
         return {
            name: facetName,
            itemsByValue: facet,
         };
      });
```

features/collection/views/DefaultCollectionView/CollectionFilters.tsx
```diff
@@ -1,22 +1,64 @@
...
+            })}
          </VStack>
       );
    }
 );
+
+function filterFacet(facet: Facet): boolean {
+   return (
+      !FACET_BLOCKLIST.includes(facet.name) &&
+      Object.keys(facet.itemsByValue).length > 1
```
davidrans, line 49: Why are we filtering out facets with only one item?

davidrans: Oh, I see, facets with only one value would either filter out everything or nothing :+1:

davidrans:
Looks good to me!
CR :coffee:

ardelato:
dev_block ☁️  not all facets show up initially and formatting of the facet name does not get applied to all facets. 

<details>
	<summary>1. Not all facets are initially rendered on the page </summary>

## Steps to Reproduce
1. Go to the Nintendo Switch collection (Parts->Game Console->Nintendo Game Console->Nintendo Switch)
2. Take note that there are only four facets to choose from:
    - Item Type
    - Condition code
    - Price
    - Price range
3. Now go to the Price range facet and pick either the `10:25` or the `25:50` options 
   **Note:** It didn't seem to work with the `0:10` option
4. Now notice their is a 5th facet option _`Parts or_kits`_

https://user-images.githubusercontent.com/22064420/118312066-a671a000-b4a5-11eb-8265-1895c9ac7848.mp4


</details>

<details>
	<summary>2. It seems the facet name formatting does not apply to post loaded facets </summary>

I was only able to see this occur on the Nintendo Switch since it seemed to be the only one that later rendered another facet option and that had an `_` in the facet name.

1. Go to the Nintendo Switch collection (Parts->Game Console->Nintendo Game Console->Nintendo Switch)
2. Take note that there are only four facets to choose from:
    - Item Type
    - Condition code
    - Price
    - Price range
3. Now go to the Price range facet and pick either the `10:25` or the `25:50` options 
   **Note:** It didn't seem to work with the `0:10` option
4. Now notice their is a 5th facet option _`Parts or_kits`_
5. Notice that the facet title has a `_` which should have been removed with the following code

https://github.com/iFixit/react-commerce/blob/aae86050ce028648e3a99ba5568868ffcf204dae/features/collection/views/DefaultCollectionView/CollectionFilters.tsx#L53-L64

</details>

<details>
	<summary>3. More of question, not all item types are listed? </summary>

I noticed that if I select all the options under `Item Type` that not all the items are listed. After taking a look at the Shopify Collection https://ifixit-test.myshopify.com/admin/collections/261649498202 ,  I saw that there are some products without an `Item Type` therefore, the question is what should happen when there is no `Item Type` tag? CC: @sterlinghirsh 


https://user-images.githubusercontent.com/22064420/118313151-3cf29100-b4a7-11eb-9a6f-4b9be696ce09.mp4

No Item Type Product?

</details>

## Steps Taken:
1. Clicked on the preview link from the Vercel comment https://github.com/iFixit/react-commerce/pull/17#issuecomment-841250248
2. Went to the Parts page
3. Then went to the Nintendo Switch page (Parts->Game Console->Nintendo Game Console->Nintendo Switch)
4. Using https://ifixit-test.myshopify.com/admin/collections/261649498202 as context I checked if all the facets were rendering based on each product tags and variant condition codes

features/collection/views/DefaultCollectionView/CollectionFilters.tsx
```diff
@@ -1,22 +1,64 @@
...
+
+function formatFacetName(facetName: string): string {
+   let name = facetName;
+   if (name.startsWith('options.')) {
+      name = name.replace('options.', '');
+   }
+   if (name.startsWith('named_tags.')) {
+      name = name.replace('named_tags.', '');
+   }
+   name = name.replace('_', ' ');
```
sterlinghirsh, line 61: This will replace all the underscores.
```suggestion
   name = name.replace(/_/g, ' ');
```

dhmacs:
Thanks for the feedback @ardelato! Since you spotted quite a few issues on filters I decided to fix them right now instead of later. I've just rewritten the search state logic, and now filters should work as expected. Moreover the work I've done should simplify server side rendering of collections.

> 1. Not all facets are initially rendered on the page

The source of these inconsistencies was the `facetingAfterDistinct` flag enabled. I found out that when variants of the same product have different facets, this option causes inconsistent results if used along with `distinct` flag. 
Removing `facetingAfterDistinct` means that the number displayed next to filter items does not show the product count, rather it shows the variants count. This should not be an issue since the mockup doesn't include those numbers, and we're only using them for debug purposes right now, they'll be removed eventually.

> 2. It seems the facet name formatting does not apply to post loaded facets

Fixed as suggested by @sterlinghirsh 

---

**⚠️ Note: CI is currently not working due to a billing issue on Vercel**

sterlinghirsh:
Looks like Kyle got the CI issue fixed, so I just amended the commit message and force pushed to retrigger CI.

ardelato:
un_dev_block 👍

k0rvusk0r4x:
Using similar steps as [here](https://github.com/iFixit/react-commerce/pull/17#issuecomment-841448118), noted:

- The fifth facet "Part or kit" is now visible without filtering first.
- The underscore present in this facet is no longer there.
- While the worksin facet is not shown in the same way on this list, going to Parts -> Phone -> iPhone shows the model of iPhone as a choice, and we can also see other facets.

QA 😺

sterlinghirsh:
CR :zap: looks great

davidrans:
CR :coffee:

PR #24, opened 2021-06-21
feat: Add price filter

### LINKED ISSUE #18: Implement collection price filter
## Problem

The user would like to filter the collection by price. Two common scenarios are:
- Search for products in a given price range (e.g. "my budget for this gift is around $30.”) 
- Filter out expensive options or cheap options.

## Solution

**UPDATED**

Chakra UI does not provide a range slider out of the box, and implementing one from scratch takes time, so we are moving toward a different UX as shown below.

Screen Shot 2021-06-17 at 11 04 54

This isn't necessarily worse, on the contrary it might even be a UX improvement as sliders generally present many UX challenges as mentioned [here](https://baymard.com/blog/slider-interfaces). Just to name a few:
- Linear sliders do not account for product distribution across the price range (This makes the slider extremely sensitive when users are trying to set the most commonly used values)
- Even though sliders look simple, they require a lot of thoughtful considerations to avoid a bad UX

> **Previous solution**
> 
> Implement a range slider selector that allows the user to pick a price range:
> 
> Screen Shot 2021-06-17 at 10 53 32
>

dhmacs:
This PR adds a price range filter to the collection page.

A few other improvements are also introduced by this PR:

- Simplified filter ids: now root filter ids are the facet names, so its easier to clear filters by facet name
- Added clear buttons below filter sections
- Refactored lib folder

## QA

2. Visit any collection page (other than `/parts` and `/tools`)
3. Verify that when a price range is selected products are filtered accordingly
4. Verify that a custom range can be inserted and products are filtered accordingly

Closes #18


masonmcelvain:
CR 👍🏻

deltuh-vee:
QA 🎬 
- [x] Products are filtered when a price range is checked
- [x] Products are filtered when a custom price range is included
deploy_block ☝️ 
1. Should there be any validation for a min price range that is greater than the max?
CleanShot 2021-06-21 at 14 50 58

2. Should predetermined price ranges and custom price ranges be exclusive? So you can’t use the checkboxes and the custom ranges at the same time.
CleanShot 2021-06-21 at 14 52 05

sterlinghirsh:
CR :zap:

masonmcelvain:
CR ✌🏻 through https://github.com/iFixit/react-commerce/pull/24/commits/4c20cbd70c9e533af9df8df12d91deb9ff6dde9b

sterlinghirsh:
CR :zap:

> Should predetermined price ranges and custom price ranges be exclusive? So you can’t use the checkboxes and the custom ranges at the same time.

I don't have a strong preference. I don't know why you'd use both in practice though.

ardelato:
QA 👍 

- [x] Products are filtered when a price range is checked
- [x] Products are filtered when a custom price range is included
- [x] Min is not allowed to be greater than Max in Price Range

## Steps Taken:
1. Visited the Vercel PR preview
2. Visited `/collections/nintendo-switch`
3. Verified that when a price range is selected products are filtered accordingly 
4. Verified that a custom range can be inserted and products are filtered accordingly
5. Verified min could not be greater than max price range

dhmacs:
> CR ✌🏻 through [4c20cbd](https://github.com/iFixit/react-commerce/commit/4c20cbd70c9e533af9df8df12d91deb9ff6dde9b)
> 
> > Should predetermined price ranges and custom price ranges be exclusive? So you can’t use the checkboxes and the custom ranges at the same time.
> 
> I could also see an argument for wanting to see products that are $150, $400, and $50-$100, though I don't know why you'd want to do that. In the case where a checkbox fell within the range, then that checkbox is essentially useless. I think this is the current behavior right?

@sterlinghirsh I'm implementing the exclusiveness behavior since predetermined ranges and custom ranges are in AND, so it makes sense that if one is selected the other one is not. I'm currently fixing a few bugs on the implementation but I should be able to push an update soon

dhmacs:
I had to do some code refactor to account for the fact that React doesn't batch state updates unless they're directly triggered by an event handler (here they're not since we're debouncing the input).

@deltuh-vee now all your concerns should be addressed 👍

ardelato:
QA 👍 

- [x]  Products are filtered when a price range is checked
- [x]  Products are filtered when a custom price range is included
- [x]  Min is not allowed to be greater than Max in Price Range
- [x]  You can use the price checkboxes or the price range but not both 👍 

## Steps Taken:
1. Repeated the same steps as https://github.com/iFixit/react-commerce/pull/24#issuecomment-866255023

masonmcelvain:
CR ✌🏻

sterlinghirsh:
CR :zap:

sterlinghirsh:
un_deploy_block :zap:

PR #30, opened 2021-06-23
fix: Fix collection list view layout bug

### LINKED ISSUE #25: Products with single lined product descriptions are justified differently
# Description:

While QAing https://github.com/iFixit/react-commerce/pull/24 I noticed a product for the `/collections/nintendo-switch` had a considerable amount of spacing between the product image and product text. 

After inspecting it, it seems that because the product description does not wrap, the flex property `justify-content: space-between` adds spacing between the two DOM elements. This causes it to look misaligned as compared to the other products in the list.

|Current Structure	|
|:---:|
|Justify Content broken if product description does not wrap	|

|Without `justify-content: space-between`	|
|:---:|
|With out space between justification|

|With More text added to product description	|
|:---:|
|With loreum added|

# Repo Steps:
1. Go to `/collections/nintendo-switch` 
2. Scroll down to find **Nintendo Switch Digitizer Front Panel**
3. Notice there is more spacing between the text and image for the product
4. Inspect the product description and add more text
5. Now notice the spacing is reduced
6. Alternatively, uncheck the `justify-content: space-between` property for the product to see the spacing also get removed

dhmacs:
This PR closes #25 bug.

### DIFF
--- a/features/collection/views/DefaultCollectionView/CollectionProducts.tsx
+++ b/features/collection/views/DefaultCollectionView/CollectionProducts.tsx
@@ -72,12 +72,7 @@ export function CollectionProducts({ viewType }: CollectionProductsProps) {
             const isDiscounted =
                hit.compare_at_price != null && hit.compare_at_price > hit.price;
             return (
-               <HStack
-                  key={hit.handle}
-                  justify="space-between"
-                  spacing={4}
-                  py={4}
-               >
+               <HStack key={hit.handle} spacing={4} py={4}>
                   <Img
                      boxSize="160px"
                      flexShrink={0}


GrantMatejka:
CR :corn:

mlahargou:
CR 📱

ardelato:
QA 👍 

## Steps Taken:
2. Went to the `/collections/nintendo-switch` page
3. I then compared this Pull's preview with the one where the issue was originally reported and confirmed the spacing is gone

|Before	| After	|
|:---:|:---:|
|Before|After| 
4. I also changed the window size to confirm that it was still responsive and it did not break the layouts.

PR #38, opened 2021-07-12
Add collection page related news stories

### LINKED ISSUE #14: Add "Related News Stories" section into collection page
https://valkyrie.ifixit.com/wp-json/wp/v2/posts

dhmacs:
This PR adds related news stories to the collection page.

## QA

2. Visit any collection page (other than `/parts` and `/tools`)
3. Verify that related news stories are rendered
4. Verify that related news stories are consistent with the collection

Closes #14


config/env.ts
```diff
@@ -37,6 +37,12 @@ export const WP_BASIC_AUTH_PASSWORD = checkEnv(
 
 function checkEnv(env: string | null | undefined, envName: string): string {
    if (env == null) {
+      if (process.browser) {
+         if (envName.startsWith('NEXT_PUBLIC')) {
+            console.warn(`environment variable "${envName}" is not defined`);
```
masonmcelvain, line 42: I'm confused why we are emitting a warning that says '`envName` is not defined' when we know envName is a string, and we are checking that it begins with 'NEXT_PUBLIC'. Perhaps this was a copy/paste mistake?

masonmcelvain:
CR 👍🏻 deploy_block ✋🏻 on a small question

erinemay:
QA 👍 
General:
 - Links to related news stories open to the correct corresponding iFixit news page
 - Links to news stories are according to post tags that match the collection name
 - If no post is tagged with the collection name, it appears that random posts are filled in

### Spot checked several collections pages:
MAC (https://react-commerce-git-add-collection-page-related-ne-19e48b-ifixit.vercel.app/collections/mac)
- Only two stories were visible, but upon inspection of the news stories in Valkyrie.cominor.com, I found only 2 posts with the tag: "Mac"

<details><summary>Image with only two related stories on MAC page</summary>

image


</details>

PHONE (https://react-commerce-git-add-collection-page-related-ne-19e48b-ifixit.vercel.app/collections/phone)
- Only 4 stories are tagged with "phone" in valkyrie.cominor.com
- 3 of those 4 stories randomly appear at the bottom of the /phone collection page. These mix upon refresh and it appears that the choosing of the 3 is random.

HOUSEHOLD (https://react-commerce-git-add-collection-page-related-ne-19e48b-ifixit.vercel.app/collections/household)
- No news stories have the tag: "household"
- ❓ What is the expected behavior of collections that don't correspond to any news story tags? It appears to be totally random. @dhmacs is there any selection going on here I should test for or is it just randomly picking posts when no tags match?

GAME-CONSOLE (https://react-commerce-git-add-collection-page-related-ne-19e48b-ifixit.vercel.app/collections/game-console)
 - One related story is shown

<details><summary>Image with only one related story on GAME-CONSOLE page</summary>

image

</details>

deploy_block 🌵  on some questions:
@Annika29, not sure who is the right person to talk to about design decisions, but I have some questions:
- Is it expected that upon implementation of this collection page, tags will be filled in on more stories? OR is it expected that will we see many collection pages with one or two related stories until more stories are written and tagged? It seems odd that on an iFixit page for macs, there are only 2 related articles. 
- Is there any other logic that could be inserted or a note like "No related news stories were found, but here are some good ones"? It seems odd to me that these are listed as related news stories on the Nindendo Game Console collection page
image

config/env.ts
dhmacs: @masonmcelvain `envName` is the name of the environment variable, the null check is actually on the `env` value as you can see in line 39.

To be clear, the purpose of this function is to output clear errors when an environment variable is missing. 
When this is used server side we throw an error, so it fails the tests. 
When this is used client side we need to distinguish between variables that should be available client side (those starting with `NEXT_PUBLIC`) and variables that are available only server side. In the former case we should output at least a warning if the variable is not defined, in the latter case we should not output anything.

dhmacs:
> * ❓ What is the expected behavior of collections that don't correspond to any news story tags? It appears to be totally random. @dhmacs is there any selection going on here I should test for or is it just randomly picking posts when no tags match?

@erinemay I'm just showing posts returned by the API, I'm not doing any kind of selection. From the UI perspective, if no posts are returned by the API then I don't render the section. It's up to the API to decide what to return.

.env.production
```diff
@@ -1,5 +1,8 @@
...
 NEXT_PUBLIC_ALGOLIA_API_KEY=118b51a67eefbe4d3c445d1e1090afac
 NEXT_PUBLIC_SHOPIFY_DOMAIN=ifixit-test.myshopify.com
-NEXT_PUBLIC_SHOPIFY_STOREFRONT_ACCESS_TOKEN=bc6c5d0972b3ad5d565c99768e6065bf
-NEXT_PUBLIC_IFIXIT_API_ORIGIN=https://www.cominor.com
\ No newline at end of file
+NEXT_PUBLIC_SHOPIFY_STOREFRONT_ACCESS_TOKEN=d08db995952e88a37804b54a284e5130
+NEXT_PUBLIC_IFIXIT_API_ORIGIN=https://www.cominor.com
+WP_ORIGIN=https://valkyrie.ifixit.com
+WP_BASIC_AUTH_USER=ifixit
+WP_BASIC_AUTH_PASSWORD=iFixitIsTheNews!
```
djmetzle, line 8: We should avoid keeping secrets plaintext in repo.

dhmacs: @djmetzle are these credentials protecting sensitive API? Since posts are public I thought these were like public access tokens.

djmetzle: No, that is there to add an extra security guard over `wp-admin`.

That secret is public now.

dhmacs: Oh I'm really sorry for that, I think these needs to be changed 😬 This would also mean that to deploy this on Shopify I would need a public endpoint without credentials, since we can only fetch from APIs client side on the Shopify theme @Annika29

danielbeardsley: Have we rotated this secret @djmetzle ?

features/collection/loaders/index.ts
```diff
@@ -148,3 +156,87 @@ const collectionQuery = /* GraphQL */ `
...
+      return rawStories.map<News>((rawStory) => {
+         return {
+            id: rawStory.ID,
+            title: rawStory.post_title,
+            date: rawStory.post_date,
+            category: rawStory.categories[0]?.name,
+            image: rawStory.featured_image
+               ? { url: rawStory.featured_image.md }
+               : null,
+            permalink: rawStory.permalink || '',
```
danielbeardsley, line 232: Do these end up being correct? When our app embeds the WP content into our site (we access the API from the backend) we post-process the response and rewrite the urls (among other things). Do these urls make sense given the meant-to-be-private wordpress site?

I think we'd probably want this code to be pinging the equivalent API on our site so we can do the post-processing on the response. Thoughts @jarstelfox ?

jarstelfox: > I think we'd probably want this code to be pinging the equivalent API on our site so we can do the post-processing on the response. Thoughts @jarstelfox ?

That was my immediate thought.

I think we will need to set up an API on our side to proxy the data. CC @Annika29 @MilesRobinson

dhmacs: @jarstelfox that would be great, so we bypass basic auth and CORS issues 👍

jarstelfox: Awesome, I’ll see if I can get to it tomorrow or Monday so I can unblock you. Sorry about that!

cdcline:
I think this is dev_blocked?

Annika29:
Yes, thanks Chris! dev_block 😁 until @jarstelfox can set up an API on our side to proxy the data

.env.production
djmetzle: No, i haven't. Code ref: https://github.com/iFixit/ifixit/blob/877045a67c5d316d10dce1ca4dcbdf4e0fc2664c/Conf/varnish/production.vcl#L583

Annika29:
Issue on the iFixit side to get this unblocked: https://github.com/iFixit/ifixit/issues/38857

features/collection/loaders/index.ts
jarstelfox: Sorry about the delay. I have a pull open now.

sterlinghirsh:
un_dev_block ⚡

features/collection/loaders/index.ts
jarstelfox: Hi all I forgot to say this is working and live: https://ifixit.com/api/2.0/related_posts?data={%20%22tags%22:%20[%20%22repair%22%20],%20%22per_page%22:%205%20}

just hit https://ifixit.com/api/2.0/related_posts with `?data=<some_valid_json>`

dhmacs:
Hey @Annika29, can I have some instructions on how to use the new endpoint?

Annika29:
@jarstelfox added this above:
>Hi all I forgot to say this is working and live: https://ifixit.com/api/2.0/related_posts?data={%20%22tags%22:%20[%20%22repair%22%20],%20%22per_page%22:%205%20}

>just hit https://ifixit.com/api/2.0/related_posts with ?data=<some_valid_json>

but I'm sure he can provide more details if you have specific questions 😊

sterlinghirsh:
@Annika29 @jarstelfox We're using this from the product collection page, generally on a specific device. If we have a device like "iphone 7" how would we just include that in the tags? Must stories match all tags or some of them? Do we know if stories are tagged with specific devices? If so is there any specific formatting to those tags?

jarstelfox:
> If we have a device like "iphone 7" how would we just include that in the tags? Must stories match all tags or some of them? Do we know if stories are tagged with specific devices? If so is there any specific formatting to those tags?

I think this describes most behavior:

```
Notes :

- tags are required, accepts an array
- per_page is optional and defaults to 3 accepts numeric value
- posts by tags are returned in random order
- for empty tags [] will return random posts
```

See:
https://github.com/iFixit/valkyrie/issues/446
https://github.com/iFixit/valkyrie/issues/485
https://github.com/iFixit/valkyrie/pull/453


> If we have a device like "iphone 7" how would we just include that in the tags?

I think `['iphone' ,'iphone 7']` would be correct here.

> Must stories match all tags or some of them?

```
- posts by tags are returned in random order
- for empty tags [] will return random posts
```

I think this implies we do our best to match tags in order and worst case is we show random posts.

>Do we know if stories are tagged with specific devices?

That's a content question. I suspect no. CC @Annika29 

>If so is there any specific formatting to those tags?

No, I think we told authors to try to keep the tags as a single word. i.e. `iphone` instead of `iphone 7`.

Annika29:
>Do we know if stories are tagged with specific devices?
>That's a content question. I suspect no. CC @Annika29

It honestly depends on the post. I don't know what method (if there is one) that the content team uses to add tags to posts. Here's an example of some posts and the corresponding tags.

Screen Shot 2021-08-12 at 2 28 06 PM

sterlinghirsh:
Interesting. It looks a little inconsistent but I'm sure we can change the tags if necessary. So then on collection pages for devices, let's just include the name of the device and its parent in the list of tags for now. We'll probably want to add brand and maybe more tags later, but we don't have that data in place yet.

cdcline:
Looks like a lot of merge conflicts :(

dhmacs:
@cdcline @masonmcelvain the PR is ready to be reviewed again 👍

config/env.ts
masonmcelvain: Just saw this- thanks for the explanation, that makes perfect sense :)

features/collection/loaders/index.ts
```diff
@@ -205,21 +203,15 @@ interface PostImage {
    'rp4wp-thumbnail-post': string;
 }
 
-async function loadRelatedNews(tags: string[]): Promise<News[]> {
-   const base64Credentials = convertToBase64(
-      `${WP_BASIC_AUTH_USER}:${WP_BASIC_AUTH_PASSWORD}`
+async function loadRelatedNews(tags: string[]): Promise<Post[]> {
```
masonmcelvain, line 206: Minor: We could rename this method to `loadRelatedPosts` since we seem to be using that terminology now.

masonmcelvain:
CR ✌🏻 Looks great!

features/collection/loaders/index.ts
```diff
@@ -30,13 +33,16 @@ export async function loadCollection(
       const ancestorsHandles = getAncestorHandles(hierarchy).reverse();
       ancestors = await loadCollections(ancestorsHandles, client);
    }
+
+   const news = await loadRelatedNews([collection.title]);
```
danielbeardsley, line 37: Doing the await here means this whole component doesn't render until we've got the response from the API request. When doing async requests that result in data being displayed on the page, it's better to use a hook; something like [`useAsync`](https://usehooks.com/useAsync/).

danielbeardsley: Especially since we're already in a nested request here. Using `useAsync` for both of these means we could render the component immediately (even if blank / empty) and not have to wait for two requests, one after the other.

danielbeardsley:
I'm pretty new to this repo, but I think there's a significant performance gotcha here. I dunno how prevalent this pattern of `await` before we render the component is in this repo, but I feel like it's a bad pattern unless I'm missing something.

dev_block :+1:  on this discussion.

features/collection/loaders/index.ts
jarstelfox: It’s also potentially worth using https://github.com/async-library/react-async as it makes managing the render state quite easy

dhmacs: @danielbeardsley @jarstelfox I'm not sure what you mean here.. This request is made server side, when Next.js method `getServerSideProps` is called to get data for the page. You can't use hooks inside `getServerSideProps`.

Anyway If we want to skip SSR for related posts because we want the page to be served faster, we can lazy load them client side.

jarstelfox: @dhmacs Thanks for the clarification. I had missed this was part of SSR. I am less concerned about the performance issues here. I am not sure of this changes @danielbeardsley's mind.

dhmacs: @danielbeardsley let me know if you have more questions, so that we can move this forward

sterlinghirsh:
un_dev_block :zap: we're going to move forward with it

masonmcelvain:
CR ✌🏻

PR #62, opened 2021-08-04
Add collection empty states

### LINKED ISSUE #49: Show empty state when collection does not have products

dhmacs:
This PR adds an empty state when the collection is empty or when the search product no results. Moreover we're adding also an initial loading state, tough this would not be visible when the collection is rendered server side (right now we're still rendering the collection client side).

## QA

2. Visit an empty collection page (e.g `/parts`)
3. Verify that the collection empty state is rendered
    Screen Shot 2021-08-04 at 09 55 25
4. Visit a collection that has products (e.g. `/collections/macbook`) and filter so that there is no result (e.g. type random characters on the search bar like "dfsd")
5. Verify that the "No results found" empty state is rendered
    Screen Shot 2021-08-04 at 09 58 19 

Closes #49


deltuh-vee:
QA 🎬 
Verified empty state is rendered for empty collection pages ✅ 
Verified No Results Found empty state is rendered for filters and searches that match 0 products ✅

cdcline:
CR 👍

masonmcelvain:
CR ✌🏻 so pretty!

Issue #55, opened 2021-07-23
Decide how to resolve filter animation bug caused by virtualization


ardelato:
While on the `/collections/macbook` page, I selected the `Item Types` filter which expanded. I then collapsed the filter but noticed that some filters returned to their position immediately while other filters scrolled back into position?

I used the Animation Chrome Dev Tool to slow down and track the animations in the page.

https://user-images.githubusercontent.com/22064420/126719450-81302bbd-8bb3-479b-9709-ee292d5eb3ad.mp4

--- 

> This is a known bug. It happens because elements that are further down the list are unmounted due to virtualization. When they are scrolled back into view they are just mounted in place without animation.

https://github.com/iFixit/react-commerce/pull/46#issuecomment-885745735

More investigation is needed as it may require a custom virtualization library to fix the bug

PR #68, opened 2021-08-19
Improve filter virtualization UX

### LINKED ISSUE #42: Fix virtual list row wrong item measurement when inside a modal
## The problem

The items are missing the divider because the row height is set 1px shorter than it should be.

Screen Shot 2021-07-21 at 14 52 45

## Possible solution

We're still not sure about the root cause of this behavior, which happens only within a modal.
One solution that we could explore is to fix the row height.

dhmacs:
This PR closes #42 and #55

## QA

2. Visit any collection that contains products (e.g. "parts" collection)
3. Verify that the bug described by #42 is fixed
4. Verify that the bug described by #55 is fixed


ardelato:
QA 👍 

- [x] Mobile Filter modal now has visible dividers between filters
- [x] Filter animation issue with virtualization has been fixed 

## Steps Taken:
2. I confirmed that the animation issue seen https://github.com/iFixit/react-commerce/issues/55 has been resolved

https://user-images.githubusercontent.com/22064420/130104217-e1a9943b-1143-43a0-ab7c-8e61cca36fa5.mp4

3. I then checked out the Vercel preview on mobile and confirmed there were now visual dividers between the filters



4. I confirmed the filters still worked and confirmed no issues across different browsers and devices
- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Android Phone  
- [x] Android Tablet
- [x] iPhone

features/collection/components/FilterList/FilterRow.tsx
```diff
@@ -67,16 +153,21 @@ export const FilterRow = React.memo(
                         color="gray.500"
                         style={{
                            transform: `rotate(${isExpanded ? '45' : '0'}deg)`,
-                           transition: 'transform 300ms',
+                           transition: `transform ${TOGGLE_ANIMATION_DURATION_MS}ms`,
```
masonmcelvain, line 156: Why did we change this from 300 -> 250, and on line 143 we left the background color transition at 300ms?
`transition="background-color 300ms"`

features/collection/components/FilterList/RangeFilter/RangeFilterList.tsx
```diff
@@ -49,23 +61,31 @@ export function RangeFilterList({
...
+                  //       );
+                  //       toggle(item.id, { clearFacets: facetToBeCleared });
+                  //    }}
+                  // >
+                  //    <Text fontSize="sm">
+                  //       {renderItem
+                  //          ? renderItem(item, index, values)
+                  //          : item.value}
+                  //    </Text>
+                  // </Checkbox>
```
masonmcelvain, line 88: delete?

features/collection/components/FilterList/index.tsx
```diff
@@ -29,7 +29,145 @@ const FACET_BLOCKLIST = [
...
+      }
+   }, [areRefined, facets, send, state.context.items]);
+
+   const data = React.useMemo<ItemData>(() => {
+      return [state, send];
+   }, [state, send]);
+
+   const getSize = React.useCallback(
+      (index: number): number => {
+         return state.context.sizeMap[state.context.items[index].name] || 41;
```
masonmcelvain, line 118: Can we make 41 a constant with a descriptive name? Seems like we use it in a few places.

features/collection/components/FilterList/virtualAccordion.machine.ts
```diff
@@ -0,0 +1,106 @@
+// The statechart can be visualized here: https://stately.ai/viz?id=1a78f924-f329-4b6c-af20-c0952399409b
```
masonmcelvain, line 1: This is such a cool tool!

sterlinghirsh:
CR :zap:

masonmcelvain:
CR ✌🏻 We CR'ed this as a group and had a couple minor comments. You could just address these in your next pull, but if you decide to address them now we can carry the QA forward. We thought the state machine visualization was a great tool.

jande141:
CR :t-rex:

zdmitchell:
Very cool! CR 🎸

features/collection/components/FilterList/index.tsx
dhmacs: You're right, seems kind of a magic number.. It is actually the estimated default height of the row, so we could use something like `ROW_ESTIMATED_HEIGHT`

features/collection/components/FilterList/FilterRow.tsx
dhmacs: @masonmcelvain because they're two different animations.. normally I would stick to 300ms, but for the slide-up slide-down I opted for a faster time since it helps hiding some tricks that makes the animation work

dhmacs:
> CR ✌🏻 We CR'ed this as a group and had a couple minor comments. You could just address these in your next pull, but if you decide to address them now we can carry the QA forward. We thought the state machine visualization was a great tool.

I'll address the comments in the next pull 👍

PR #120, opened 2021-11-03
Use new collection type names

### LINKED ISSUE #116: Refactor Strapi collection types

dhmacs:
This PR closes #116 

I'm still keeping the name "collections" in many parts of the codebase mostly to avoid conflicts with other branches that are ready to be reviewed after this PR. When those branches will be merged too then I'll refactor all references to "collection"

## QA

2. Visit the parts collection page
3. Verify that product list page works with the new Strapi instance

### DIFF
--- a/lib/api/strapi/operations/getCollection.graphql
+++ b/lib/api/strapi/operations/getCollection.graphql
@@ -1,6 +1,14 @@
 query getCollectionPageData($whereCollection: JSON, $whereStoreSettings: JSON) {
    ...LayoutProps
-   collections(limit: 1, where: $whereCollection) {
+   global {
+      newsletterForm {
+         title
+         subtitle
+         inputPlaceholder
+         callToActionButtonTitle
+      }
+   }
+   productLists(limit: 1, where: $whereCollection) {
       id
       handle
       title
@@ -24,6 +32,14 @@ query getCollectionPageData($whereCollection: JSON, $whereStoreSettings: JSON) {
                parent {
                   title
                   handle
+                  parent {
+                     title
+                     handle
+                     parent {
+                        title
+                        handle
+                     }
+                  }
                }
             }
          }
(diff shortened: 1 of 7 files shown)


lib/api/strapi/operations/getCollection.graphql
```diff
@@ -24,6 +32,14 @@ query getCollectionPageData($whereCollection: JSON, $whereStoreSettings: JSON) {
                parent {
                   title
                   handle
+                  parent {
```
evannoronha, line 35: Why are we looking farther up the hierarchy now than we were previously?

evannoronha, line 35: Why are we looking farther up the hierarchy now than we were previously?

evannoronha (approved):
CR 🔥 

LGTM

erinemay:
QA 🦑 The parts page "works" as far as:
- It gets loaded successfully
- There is product information populated
- Search and filtering works as expected


I'm noticing several things, and I want to make sure these are outside the scope of this PR.
Deploy_block 🔱 on confirming that the following is current expected behavior.

<details><summary>There's no clear organization to the model buttons.</summary>

I would expect some kind of ordering or rationale why specific models are displayed at the top.

image

</details>

<details><summary>Most of the parts collections linked in the model section are empty</summary>

image

</details>


<details><summary>Price displayed here is not the same as the price displayed on the product page.</summary>

The first part listed on the page is the `MacBook Pro 13" (A2251, 2020) 2.3 GHz Logic Board with Paired Touch ID Sensor`
When I open the link to the actual product page on ifixit.com, it appears that the current available product is priced differently.

image

image


</details>

lib/api/strapi/operations/getCollection.graphql
dhmacs: @evannoronha I was extending this to allow for more nesting and see if the query breaks (spoiler alert: if you add even more it does). During today meeting we decided to just show 2 or 3 ancestors in the breadcrumb, so eventually I'll simplify this

GrantMatejka:
CR 🌽

erinemay:
Update from @dhmacs. This PR is addressing the page as a whole loading and working. It does. 

un_deploy_block 👍 

Collection/model behavior noted above is currently expected.

sterlinghirsh:
CR :zap:

dhmacs:
> There's no clear organization to the model buttons.
> I would expect some kind of ordering or rationale why specific models are displayed at the top.
> 
> image

@erinemay Yeah you're right, the most straightforward thing to do is to sort alphabetically, but that might not be the best UX. Another option is to have the most recent devices displayed first. This would required to add something like a `releasedAt` field in the product list collection type. This is something that needs to be defined

PR #127, opened 2021-11-05
Add support for markdown in product list description

### LINKED ISSUE #112: Render collection description as markdown

dhmacs:
This PR closes #112 

## Changelog

- Adds `snarkdown` library to render markdown description

## QA

2. Visit a few product list pages and verify that the description renders correctly


GrantMatejka:
@dhmacs snarkdown is great because it's so small, but I was wondering how you felt about a package like react-markdown so we could avoid directly setting html?
Or is this not going to be used enough to need it?

ardelato:
QA 👍 

## Steps Taken:
2. I then went to https://react-commerce-git-add-description-markdown-rendering-ifixit.vercel.app/collections/macbook-13-core-duo
3. I confirmed the markdown was properly rendered

|Before	| After	|
|:---:|:---:|
|Before markdown|	Markdown works|

dhmacs:
> @dhmacs snarkdown is great because it's so small, but I was wondering how you felt about a package like react-markdown so we could avoid directly setting html? Or is this not going to be used enough to need it?

@GrantMatejka We're already running over optimal budget (>200kb) in the product list page, so I'm trying hard to avoid importing heavy packages. It's hard to justify 32kb of JS if `snarkdown` can do the same with 1kb. 
What's the concern with directly setting html?

GrantMatejka:
@dhmacs I'd rather not directly set it, but I understand the need to stay small

ardelato:
QA 👍 

## Steps Taken:
1. Checked out this branch on Cominor
2. Git merged `origin/master`

3. Ran `make`

ardelato:
Sorry about that I missed clicked 😓

dhmacs:
> @dhmacs I'd rather not directly set it, but I understand the need to stay small

Are you thinking about this in term of security issues? Anyway if you are willing to trade kb for this I can make the change 👍

GrantMatejka:
I'm fine with it, just wondering. Thanks for talking it through with me, I appreciate it
CR 🌽

sterlinghirsh:
CR :zap: I'm fine with the snarkdown approach. I don't really expect security issues coming from Strapi. But come to think of it, do we even need to render it on the client? If we're server-rendering it and the content never changes, it wouldn't need to render on the client anyway right?

dhmacs:
> CR :zap: I'm fine with the snarkdown approach. I don't really expect security issues coming from Strapi. But come to think of it, do we even need to render it on the client? If we're server-rendering it and the content never changes, it wouldn't need to render on the client anyway right?

You're absolutely right. I can use snarkdown to convert markdown to html server side and just pass that to the client 👌

Issue #130, opened 2021-11-12
Bug: Able to have conflicting price range filters at the same time


erinemay:
### Description
Users can get to a state where they are filtering by conflicting price ranges. This causes no product results to be shown.

### Expected behavior
When the user enters numbers into the manual price range entry field, any price range checkbox options should be deselected.

### Steps to reproduce

1. Go to https://react-commerce.vercel.app/collections/parts
2. Expand the price range filter.
3. Select the 0-10 price range filter.
4. Enter 50-200 into the manual entry field.
5. Observe the checkbox you selected is not deselected as a result.

Resulting URL:
https://react-commerce.vercel.app/collections/parts?price_max=200&price_min=50&price_range=0-10

Screenshot:
![image](https://user-images.githubusercontent.com/72166715/141540975-9fe7e90f-00d0-4d18-b047-a43b610a0dba.png)

PR #132, opened 2021-11-18
Fix price filter bugs

### LINKED ISSUE #129: Price range error message out of view
### Description
When a user enters amounts into the price range entry boxes that are inappropriate (max < min), the error message is cut off. 

### Reproduction steps:

1. Go to https://react-commerce.vercel.app/collections/parts.
2. Expand the price range filter and scroll to the bottom to the text entry fields.
3. Enter a minimum price that is higher than the maximum price.
4. Observe error state. You can see the top of the error message text, but not the full message.

image

CC: @mlahargou

dhmacs:
This PR closes #129 and #130

## QA

Until we setup the staging servers we need to clone the repo and test it locally.

1. Clone this repo
2. Make sure you are running on node v14 (if you are using nvm run `nvm use` in the project working directory)
3. Install dependencies `npm install`
4. Rename `backend/.env.example` to `backend/.env` (this will enable db seed)
5. Run dev server: `npm run dev`
6. Make sure that bug described by #129 is fixed
7. Make sure that bug described by #130 is fixed

### Access to local strapi instance

Email: `strapi@ifixit.com`
Password: `Password1`


ardelato:
QA 👍 

- [x] Error message for price range where `min` value > `max` value is no longer cut off
- [x] Conflicting price facet filter and price range filter longer occurs 

--- 

deploy_block ☁️  on two issues I noticed, but they are also on the master branch so I'll leave it up to you @dhmacs on whether to address them in this pull or in another pull.

   <details>
	<summary>1. Unsetting the <code>max</code> price range does not update the filter like when setting it </summary>
   
   It seems that setting a value in the max price range automatically triggers the page and filter to update. However, when you remove the value from the max price range, the filter does not get removed. This does not occur with the min price range.

   

https://user-images.githubusercontent.com/22064420/142512101-e9cf6554-2f29-47a4-81d7-75ee482cac2b.mp4
   
   </details>

   <details>
	<summary>2. Price Range value immediately refreshes the page instead of waiting, resulting in bad UX </summary>
   
  While entering a value into the price range fields, the page automatically refreshes as soon as a value is entered into the field. This is leading to a bad user experience, since I might not have finished typing out the value, yet the page is already refreshing making it disorienting.  

   I expected that the page would be refreshed with the filter in one of these cases:
   - After ~3+ seconds since I last entered a digit 
  - After I leave the focus of the field
  - After pressing <kbd>enter</kbd>
   

https://user-images.githubusercontent.com/22064420/142512707-58c18762-fdc6-4619-9f4b-68fa14fd9b31.mp4


   </details>


## Steps Taken:
1. Followed the setup steps from the Pull Body
2. Then confirmed the bugs listed in the Pull Body were fixed
- [x] Error message for price range where `min` value > `max` value is no longer cut off
Price Range Error not cut off

- [x] Conflicting price facet filter and price range filter longer occurs 

https://user-images.githubusercontent.com/22064420/142513849-50207029-b525-4c1e-a597-461bc65934e8.mp4

cdcline:
CR 👍

dhmacs:
@ardelato 
> 1. Unsetting the `max` price range does not update the filter like when setting it

I'll try to fix this now

> 2. Price Range value immediately refreshes the page instead of waiting, resulting in bad UX
> While entering a value into the price range fields, the page automatically refreshes as soon as a value is entered into the field. This is leading to a bad user experience, since I might not have finished typing out the value, yet the page is already refreshing making it disorienting.
> 
> I expected that the page would be refreshed with the filter in one of these cases:
> 
> * After ~3+ seconds since I last entered a digit

Wouldn't it feel slow though? 🤔

GrantMatejka:
CR 🌽 up to https://github.com/iFixit/react-commerce/pull/132/commits/c81f5d9aa7e1aaa24b3d4a3aecf19e5a34ab93b7

I think a debounce timer is a good idea, but I don't think the time needs to be as high as 3s
But idk if it needs to be implemented in this pull, up to your discretion @dhmacs

ardelato:
@dhmacs 

>> After ~3+ seconds since I last entered a digit

> Wouldn't it feel slow though? 🤔

It might feel slow if we use if we use a time based event for updating the page, since getting the timing right would be a hassle. In this case, ~3+ seconds may be slow to some users while it may be fast to the other users.

If we go with one of the other options, then it won't feel as slow. The current issue is really there is no control over when the page reloads. If the page just updated the products listed without refreshing the entire page, then it would be okay to leave the updating functionality as is. However, since the entire page gets refreshed then I was thinking of changing the updating functionality to trigger with a different event.

dhmacs:
@ardelato I've fixed the max price unsetting issue.

Since we'd might want to thinker a little bit with the UX issue, I'd say let's move it into a separate issue ok?

PS: While fixing the issue I've also found out that the `price_min` and `price_max` params are not parsed when you reload the page, so I'm gonna open a new issue for that too

GrantMatejka:
CR 🌽

ardelato:
QA 👍 

- [x] Removing the value in the max price range field now removes the filter 👍 

## Steps Taken:
1. Repeated steps from https://github.com/iFixit/react-commerce/pull/132#issuecomment-973455275

GrantMatejka:
CR 🌽

ardelato:
QA 👍 

## Steps Taken:
1. Repeated steps from https://github.com/iFixit/react-commerce/pull/132#issuecomment-973455275

masonmcelvain:
CR ✌🏻

masonmcelvain:
un_deploy_block ✌🏻 

> @ardelato I've fixed the max price unsetting issue.
Since we'd might want to thinker a little bit with the UX issue, I'd say let's move it into a separate issue ok?
PS: While fixing the issue I've also found out that the price_min and price_max params are not parsed when you reload the page, so I'm gonna open a new issue for that too

Looks like the issues got made:
https://github.com/iFixit/react-commerce/issues/137
https://github.com/iFixit/react-commerce/issues/136

PR #139, opened 2021-11-23
Fix no safe guard against long search query

### LINKED ISSUE #101: Bug: No safe-guard against long search query
# Description:

While QAing https://github.com/iFixit/react-commerce/pull/97 I realized that I did not test a long entry for the search field in https://github.com/iFixit/react-commerce/pull/94. This is an issue since the search query is added to the url as a parameter, therefore, if the search query is long it seems to cause an internal server error.

https://user-images.githubusercontent.com/22064420/136464932-07170a7b-e324-469c-921a-0a4a4a855a35.mp4

# Reproduction Steps:
1. Go the product search field
2. Enter an absurd amount of text
3. Reload the page
4. Notice there `500` error that is displayed

# Expected Behavior

I am not sure what should be the expected behavior on this (i.e. limit the amount of characters we can search, not have the search text added to the url, etc.) however, I do expect that we should safe-guard or prevent this from occurring.

dhmacs:
This PR closes #101 

## Changelog

- Prevents user from inserting long search queries into the search field

## QA

2. Visit parts product list page and verify that the product list search input is bounded

### DIFF
--- a/frontend/features/productList/components/sections/FilterableProductsSection/SearchInput.tsx
+++ b/frontend/features/productList/components/sections/FilterableProductsSection/SearchInput.tsx
@@ -16,6 +16,8 @@ interface SearchInputProps {
    className?: string;
 }
 
+const MAX_SEARCH_QUERY_LENGTH = 100;
+
 export const SearchInput = chakra(({ className }: SearchInputProps) => {
    const inputRef = React.useRef<HTMLInputElement>(null);
    const [query, search] = useSearch();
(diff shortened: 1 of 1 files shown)


GrantMatejka:
CR 🌽

deltuh-vee:
QA 🎬 
A search query longer than 100 characters cannot be entered.

frontend/features/productList/components/sections/FilterableProductsSection/SearchInput.tsx
```diff
@@ -38,7 +53,7 @@ export const SearchInput = chakra(({ className }: SearchInputProps) => {
                bg="white"
                placeholder="Search"
                tabIndex={0}
-               onChange={(event) => search(event.currentTarget.value)}
+               onChange={onSearchChange}
```
evannoronha, line 56: Why the added complexity instead of just using a `maxLength` prop?

evannoronha (approved):
CR 🔥  ... with a question.

frontend/features/productList/components/sections/FilterableProductsSection/SearchInput.tsx
dhmacs: That's right, much more elegant solution! I've also removed the callback memoization as probably isn't doing that much

deltuh-vee:
Carry over QA 🎬

jande141:
CR :poultry_leg:

GrantMatejka:
CR 🌽

PR #133, opened 2021-11-18
Add product links to featured product list section

dhmacs:
This PR closes #110 

## Changelog

- Adds product links to FeaturedProductList section products
- Refactors product cards to match new design
- Renames some `collection` folder names to now use `productList` term instead

## QA

Until we setup the staging servers we need to clone the repo and test it locally.

1. Clone this repo
2. Make sure you are running on node v14 (if you are using nvm run `nvm use` in the project working directory)
3. Install dependencies `npm install`
4. Rename `backend/.env.example` to `backend/.env` (this will enable db seed)
5. Run dev server: `npm run dev`
6. Login to strapi at `localhost:1337/admin` using the credentials specified below
6. Navigate to the `parts` product list and add a `FeaturedProductList` section (e.g. Macbook parts)
7. Visit the parts collection at `localhost:3000/collections/parts`
8. Verify that products within the featured product list correctly link to product pages:
    Screenshot 2021-11-18 at 11 08 48

### Access to local strapi instance

Email: `strapi@ifixit.com`
Password: `Password1`

PR #149, opened 2021-12-09
fix: Use `npm` target instead of `postinstall` script

andyg0808:
This allows us to install the top-level package without having to convince Rush not to run postinstall scripts. For our iFixit monorepo, the subprojects will be installed by Rush, rather than with `npm` as here.

This will require `npm`-based users of this repo to switch to `npm run npm` as the install command.

# QA
Please confirm that `npm run npm` sets everything up correctly.

@dhmacs, feel free to push back if you don't like this idea. I realize it's a bit invasive, and I'm game to try something different if you'd rather.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ You can run the backend both using SQLite and using Postgres with docker compose
 This command will install both backend and frontend dependencies:
 
 ```sh
-npm install
+npm run npm
 ```
 
 ### Dev server
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "start:frontend": "cd frontend && npm run start",
     "start": "cross-env FORCE_COLOR=1 npm-run-all -l -p start:*",
     "test": "echo \"Error: no test specified\" && exit 1",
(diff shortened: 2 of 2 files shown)


rickisXP:
QA :dumpling: 
Confirmed that `npm run npm` works using at least node version 14.

Steps taken:
1) Repeated steps from https://github.com/iFixit/react-commerce/pull/133#issue-1057170131, replaced `npm install` with `npm run npm`

andyg0808:
I found a better solution to this: PNPM has the ability to edit a `package.json` on the fly before it's used to install anything. I added an edit to delete the `postinstall` script before trying to install, which eliminates the problem this was intended to solve.

dhmacs:
@andyg0808 I don't have the full picture of the work you are doing, I guess that eventually this repo will be merged into the monorepo right?

andyg0808:
@dhmacs We're planning to keep this repo separate from the monorepo, if possible, but to support building and running it from within the monorepo. That's currently looking like pulling it in as a Git submodule and integrating it into our [Rush](https://rushjs.io/)-based monorepo tooling.

Issue #124, opened 2021-11-04
Parts page: Add tests


mlahargou:
I chatted with Erin today about what test cases we think should be added to the Parts page. We came up with a list of desired tests.  Many of these tests can probably use jest or something similar to run tests on the individual component. Most will probably require an E2E Cypress test.

We'd also like the QA team to bring up any other test cases that they see fit when manually testing this page.

https://react-commerce-git-collection-page-seo-improvements-ifixit.vercel.app/collections/parts

1) Filters
    - Click the filter title 
      - [x] Assert that the filter options are expended
    - Click a filter option 
      - [x] Assert that the filter was added to the current filter list
      - [x] Assert that the products update according to that filter
    - Click a different filter and option 
      - [x] Assert that both filters are in the current filter list
    - Add a descending price filter
      - [x] Assert that the proper warning is displayed
    - Add an ascending price filter
      - [x] Assert that the price filter is added to the current filter list
      - [x] Assert that the products displayed conform to the specific price filter
    - Click the x to remove a single filter
      - [x] Assert that the filter is removed from the current filter list
    - Click "Clear All"
      - [x] Assert that the remaining two filters are removed
2) Search
    - Search for a common word like "watch"
      - [x] Assert that the remaining products all conform to the search specification above
    - Search for a nonexistent word like "asdfasdfasdf"
      - [x] Assert that the "No results found" component is displayed
3) Grid/List view
    - Click the grid view button
      - [x] Assert that products are displayed as a grid view
    - Click the list view button
      - [x] Assert that products are displayed as a list view
4) Results view
    - Jest tests might be the best way to test these ProductCard components
    - Body
      - [x] Assert these are truncated to the desired length
      - [ ] Do we want a "show more" button?
    - Reviews
      - [ ] Ensure the stars component accurately "fills" the correct number of stars
      - [x] Ensure the number of reviews accurately displays the data that the component is provided
    - Price
      - [x] Ensure we display the price data the component is provided
    - View button
      - [x] Ensure the button directs the user to the link that the component was provided 

cc @davidrans @erinemay

batbattur:
Added the grid/list view test in https://github.com/iFixit/react-commerce/pull/381.

batbattur:
Added the search test in https://github.com/iFixit/react-commerce/pull/409.

dhmacs:
[...]

PR #168, opened 2022-02-04
test: Add filters test

dhmacs:
This PR adds filters test according to #124 

Tests have been implemented following [guiding principles](https://testing-library.com/docs/guiding-principles) of Testing Library, and this led also to accessibility improvements (this is why there are a few changes in the filters section)

## Changelog

- Update dependencies
- Refactor newsletter tests
- Add product list filters test

## QA

This is mostly a PR that adds tests. These tests are run on CI, so if they pass it means that filters are still working. 
We might want to do a smoke test to verify that everything else still works.


jordycosta:
>CI tests passed ✅ 
Product list filters are working 

The tools [hyperlink](https://react-commerce-git-add-product-list-tests-ifixit.vercel.app/store/tools) on the **Vercel** preview app is giving a `500: Internal Server Error.` ❕ 
Is this expected?

dhmacs:
> The tools [hyperlink](https://react-commerce-git-add-product-list-tests-ifixit.vercel.app/store/tools) on the **Vercel** preview app is giving a `500: Internal Server Error.` ❕ Is this expected?

@jordycosta 
That's because the tool product list has not been created on Strapi. I'll put a note that we should throw a 404 instead, but that's a fix out of the scope of this PR

k0rvusk0r4x:
CI passed with the new tests.

QA 😺 

I would like to eventually either have react tests set up so we can easily manually stress/reliance test them, as well as making sure they break when they are supposed to for new tests. Even better if we could get them stress tested through CI at some point, but that is not for this pull, and may not be for this repo. @mlahargou, @BaseInfinity, @batbattur Think we can add this to the list to get to at some point?

dhmacs:
@k0rvusk0r4x 

> I would like to eventually either have react tests set up so we can easily manually stress/reliance test them, as well as making sure they break when they are supposed to for new tests. 

I'm not sure what do you mean by that.. Would you like to run the tests manually? If so, you can follow instructions on the readme to run tests on your machine

masonmcelvain:
CR ✌🏻 
I'm not super familiar with cypress/jest syntax, but I think I understand what's being asserted at least. @mlahargou said he'd take a look at the test cases when he has a chance

frontend/test/cypress/integration/product-list/filters.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+      user
+         .findByRole('region', { name: /products/i })
+         .as('section')
+         .should('have.attr', 'data-search-state', 'search');
+      return user
+         .get('@section')
+         .should('have.attr', 'data-search-state', 'idle');
+   }
+
+   it('should help user filter', () => {
```
mlahargou, line 43: It seems like there are many different test cases in this single `it()`. Do you think we should split each test case into their own functions?

frontend/test/cypress/integration/product-list/filters.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+      user
+         .findByLabelText(/set max price/i)
+         .type(TEST_MAX_PRICE.toString(), getVirtualListTypeOptions());
+
+      user.findByText(/max should be higher than min/i).should('exist');
+
+      user
+         .findByLabelText(/set min price/i)
+         .type(
+            `{selectall}{backspace}${TEST_MIN_PRICE}`,
```
mlahargou, line 119: This is cool.

frontend/test/cypress/integration/product-list/filters.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+   });
+
+   function getVirtualListClickOptions(): Partial<Cypress.ClickOptions> {
+      return {
+         // Cypress won't click the button because it detects that there's a parent
+         // element with the pointer-events: none style.
+         // Since this it's expected behavior for a virtual list while scrolling (
+         // https://github.com/bvaughn/react-window/issues/128#issuecomment-460163077 ),
+         // we need to force the click.
+         force: true,
```
mlahargou, line 18: That's interesting. Nice find.

frontend/test/cypress/integration/product-list/filters.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+         .as('section')
+         .should('have.attr', 'data-search-state', 'search');
+      return user
+         .get('@section')
+         .should('have.attr', 'data-search-state', 'idle');
+   }
+
+   it('should help user filter', () => {
+      user
+         .findByRole('button', { name: /expand item type/i, expanded: false })
```
mlahargou, line 45: Out of curiosity, why did we decide to go with text instead of selectors to find elements? I'm thinking that the text is more likely to change than the selector as we develop/maintain this.

frontend/test/cypress/integration/product-list/filters.spec.ts
dhmacs: Good point. I started that way, but then I've realized that the tests detailed [here](https://github.com/iFixit/react-commerce/issues/124) depend on each other (e.g.  `Click "Clear All"` assumes that the filters selected by the previous points are still there).
So either we keep it like this, or we change the tests so that they don't depend on each other.
I don't have strong opinions on this, so if you think the latter is better I can refactor

frontend/test/cypress/integration/product-list/filters.spec.ts
dhmacs: I've iterated over this a few times (initially I was using `data-testid` everywhere), and I settled on the approach described [here](https://testing-library.com/docs/queries/about#priority) because to me it seems the best tradeoff. This approach tries to mimic how users interact with the page (and it does so in a way that takes into account also accessibility). 
Although I agree with you that text is more likely to change, it's probably better to update tests accordingly (I think this makes sure that we are intending to change expectation over the copy, and we're not breaking the semantic and accessibility by mistake).
The `name` regex tries to get an elements from their [accessible name](https://www.w3.org/TR/accname-1.1/), so this should prevent us from breaking accessibility.

Let me also say, that we probably need to do more work to make sure that the page is completely accessible, since I haven't performed any test with the screen reader yet, but I think this is a good start.

frontend/test/cypress/integration/product-list/filters.spec.ts
mlahargou: I don't have super strong opinions either. I lean towards separate tests for each test case 🤷

dhmacs: @mlahargou do you want me to do the refactor then?

mlahargou: Works for me!

mlahargou:
CR 📱

PR #182, opened 2022-03-01
Allow specifying DB engine in database config

djmetzle:
We're running a mysql engine RDS instance in production for strapi.

CC @dhmacs @lithobraking @danielbeardsley

### DIFF
--- a/backend/config/database.js
+++ b/backend/config/database.js
@@ -4,7 +4,7 @@ module.exports = ({ env }) => {
   if (env("NODE_ENV") === "production") {
     return {
       connection: {
-        client: "mysql",
+        client: env("DATABASE_ENGINE", "postgres"),
         connection: {
           host: env("DATABASE_HOST", "127.0.0.1"),
           port: env.int("DATABASE_PORT", 5432),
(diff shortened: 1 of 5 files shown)


danielbeardsley (approved):
Looks right, but I can't say for certain there isn't more needed.

backend/config/database.js
```diff
@@ -4,7 +4,7 @@ module.exports = ({ env }) => {
   if (env("NODE_ENV") === "production") {
     return {
       connection: {
-        client: "postgres",
+        client: "mysql",
```
djmetzle, line 7: For now, lets just allow overriding Postgres from the ENV?
```suggestion
        client: env("DATABASE_ENGINE", "postgres"),
```

djmetzle:
Kewl, so that passed all the tests. :+1:

danielbeardsley:
CR :+1:

jarstelfox:
CR 🌵

dhmacs:
We've encountered [a few issues](https://github.com/mysqljs/mysql/pull/1962) with the integration between Strapi and mysql, so we're going to stick with postgres for the time being

PR #234, opened 2022-03-25
Show only public products

### LINKED ISSUE #233: Show only public search results
> we should be including a filter in our search to say public=1. As of a couple months ago ish, there are disabled products in our index that are intended to be searchable by admins, but they are all marked public=0. Not sure if there's anything we need to change about indexing to make sure public=1 is always used on this page.

_Originally posted by @sterlinghirsh in https://github.com/iFixit/react-commerce/issues/220#issuecomment-1072866829_

dhmacs:
fixes #233 

## Changelog

- Use `public=1` when searching with Algolia

## QA

2. Visit all parts product list 
3. Verify that products marked with `public=0` in Algolia are not visible in product list search results

### DIFF
--- a/frontend/components/product-list/sections/FeaturedProductListSection.tsx
+++ b/frontend/components/product-list/sections/FeaturedProductListSection.tsx
@@ -30,21 +30,11 @@ import NextLink from 'next/link';
 import * as React from 'react';
 
 export interface FeaturedProductListSectionProps {
-   // handle: string;
-   // title: string;
-   // description: string;
-   // imageSrc?: string;
-   // imageAlt?: string;
    productList: ProductListPreview;
    algoliaIndexName: string;
 }
 
 export function FeaturedProductListSection({
-   // handle,
-   // title,
-   // description,
-   // imageAlt,
-   // imageSrc,
    productList,
    algoliaIndexName,
 }: FeaturedProductListSectionProps) {
@@ -140,7 +130,8 @@ export function FeaturedProductListSection({
                   apiKey={ALGOLIA_API_KEY}
                   initialIndexName={algoliaIndexName}
                   filtersPreset={
-                     productList.filters ?? `device:${productList.deviceTitle}`
+                     productList.filters ??
+                     `device:${productList.deviceTitle} AND public=1`
                   }
                   productsPerPage={3}
                >
--- a/frontend/components/product-list/sections/FilterableProductsSection/index.tsx
+++ b/frontend/components/product-list/sections/FilterableProductsSection/index.tsx
@@ -200,7 +200,7 @@ export const FilterableProductsSection = React.memo(() => {
                            {hits.map((hit) => {
                               return (
                                  <ProductListItem
-                                    key={hit.handle}
+                                    key={hit.objectID}
                                     product={hit}
                                  />
                               );
--- a/frontend/models/product-list/index.ts
+++ b/frontend/models/product-list/index.ts
@@ -134,8 +134,8 @@ export async function createProductListSearchContext({
 }: CreateProductListSearchContextOptions): Promise<SearchContext<ProductSearchHit> | null> {
    const filtersPreset =
       filters && filters.length > 0
-         ? filters
-         : `device:${JSON.stringify(deviceTitle)}`;
+         ? `${filters} AND public=1`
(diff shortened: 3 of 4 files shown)


frontend/components/product-list/sections/FeaturedProductListSection.tsx
```diff
@@ -140,7 +130,8 @@ export function FeaturedProductListSection({
                   apiKey={ALGOLIA_API_KEY}
                   initialIndexName={algoliaIndexName}
                   filtersPreset={
-                     productList.filters ?? `device:${productList.deviceTitle}`
+                     productList.filters ??
+                     `device:${productList.deviceTitle} AND public=1`
```
davidrans, line 134: If `productList.filters` is set, do we need to ensure that it gets a `AND public = 1` also?

davidrans (approved):
CR :coffee:

frontend/components/product-list/sections/FeaturedProductListSection.tsx
davidrans: I can't think of any

deltuh-vee:
QA 🎬 
Products marked with public=0 in Algolia are not visible in product list search results

frontend/components/product-list/sections/FeaturedProductListSection.tsx
davidrans: Oh I just misread the code. We tack `public=1` on _after_ that conditional is evaluated. Carry on :D

frontend/components/product-list/sections/FeaturedProductListSection.tsx
```diff
@@ -140,7 +130,8 @@ export function FeaturedProductListSection({
                   apiKey={ALGOLIA_API_KEY}
```
evannoronha, line 130: @sctice-ifixit do you think it's a problem that we're exposing a key that has read access to the whole index? I don't believe we care _that_ much about the private data in Algolia, but filtering private data out client side seems like a no-go.

evannoronha (approved):
CR 🔥 

Although [this](https://github.com/iFixit/react-commerce/pull/234/commits/85190c935766e81ff59cd84ecaf57f8c968e8690#r835668921) should be addressed in a follow-up issue.

frontend/components/product-list/sections/FeaturedProductListSection.tsx
sctice-ifixit: > We tack `public=1` on _after_ that conditional is evaluated.

It doesn't look that way to me. I would also expect the filter for `public` to be part of the API key generated on the server side. Otherwise, clients are free to specify public/non-public as they like.

sctice-ifixit (commented):
CR :flags: but I think there are larger issues with the setup here that we should still address.

frontend/components/product-list/sections/FeaturedProductListSection.tsx
sctice-ifixit: Yeah, we should be generating a key server-side that has filters like `public=1` locked in. That will also make sure that we don't need to worry about specifying it as part of a product list page's filters.

frontend/components/product-list/sections/FeaturedProductListSection.tsx
sterlinghirsh: I agree @sctice-ifixit but I think it's not using the securely generated key yet.

frontend/components/product-list/sections/FeaturedProductListSection.tsx
dhmacs: Hey @sctice-ifixit, just to be clear, are you referring to [this](https://www.algolia.com/doc/guides/building-search-ui/going-further/api-keys-security/react/) ? I've only found this key generated via API, is there also a way to generate the key using the Algolia dashboard? I've tried in the API keys section but it seems like I can't add filters restrictions when using the dashboard.

sctice-ifixit: We talked about this in person, but just for posterity, I'm referring to [secured API keys](https://www.algolia.com/doc/api-reference/api-methods/generate-secured-api-key/), which are generated on the backend using a search-only API key without any network communication.

PR #248, opened 2022-04-08
Docker: add script to build and push the image

danielbeardsley:
This is meant to be run during a post-merge webhook or action.

### DIFF
--- a/backend/build-docker-and-push
+++ b/backend/build-docker-and-push
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+set -eu
+
+GIT_SHORT_REF=$(git rev-parse --short HEAD)
+DOCKER_TAG="strapi:$GIT_SHORT_REF"
+# iFixit AWS ECR for strapi
+ECR_TAG="884681002735.dkr.ecr.us-east-1.amazonaws.com/$DOCKER_TAG"
+
+echo "Buidling docker image"
+docker build -t "$DOCKER_TAG"
+docker tag "$DOCKER_TAG" "$ECR_TAG"
+
+echo "Logging into ECR..."
+$(aws ecr get-login --no-include-email --region us-east-1)
+docker push "$ECR_TAG"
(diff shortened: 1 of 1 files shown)


backend/build-docker-and-push
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -eu
+
+GIT_SHORT_REF=$(git rev-parse --short HEAD)
+DOCKER_TAG="strapi:$GIT_SHORT_REF"
+# iFixit AWS ECR for strapi
+ECR_TAG="884681002735.dkr.ecr.us-east-1.amazonaws.com/$DOCKER_TAG"
+
+echo "Buidling docker image"
```
sterlinghirsh, line 10:
```suggestion
echo "Building docker image"
```

sterlinghirsh:
CR :zap:

sctice-ifixit:
CR :flags:

deltuh-vee (changes requested):

cdcline:
CR 👍

backend/build-docker-and-push
```diff
@@ -0,0 +1,16 @@
...
+
+set -eu
+
+GIT_SHORT_REF=$(git rev-parse --short HEAD)
+DOCKER_TAG="strapi:$GIT_SHORT_REF"
+# iFixit AWS ECR for strapi
+ECR_TAG="884681002735.dkr.ecr.us-east-1.amazonaws.com/$DOCKER_TAG"
+
+echo "Building docker image"
+docker build -t "$DOCKER_TAG" .
```
djmetzle, line 11: I think we should probably use Kaniko here? To avoid docker-in-docker?

andyg0808: Do we currently plan to run this from Docker? If not, I hear you, but I think we should wait to solve that until it's an issue.

andyg0808 (approved):

backend/build-docker-and-push
djmetzle: Oh, im not sure what context this runs in is all. Just want to make sure we avoid coupling to docker _if_ we're going to run this inside docker.

andyg0808:
un_deploy_block :+1:
If Docker-in-Docker's an issue, I suspect either QA will find it, or it'll be obvious enough when we go to use this.

PR #254, opened 2022-04-14
Add main category and category to blocklist

### LINKED ISSUE #191: The Tools page filters are confusing
The filters on https://react-commerce-ifixit.vercel.app/store/tools are confusing and unintuitive. Examples:

- "Merchandise" means any product for sale, but the options for filtering include clothing, bags, and merchandise. The merchandise subfilter is only iFixit "gear" (the term we use now) that does not fit into the other two categories. If the intent is for this category to be iFixit branded, non-tool or part products, we should go back to using "gear" or choose something that similarly expresses this meaning or shorten the word to "Merch." (I've no idea why the abbreviation "merch" is understood to mean branded products like our T-shirts while the full word "merchandise" does not mean that at all, but what is important is that the customers know where to find what they are looking for.)
- "Color" as a main filter should apply to every product as literally everything has a color. Thus, I, as a customer, would expect all product of a certain color to come up if I select it. Toolkits, for example, should all be listen if I select "black" because all of their cases are black. Yet, if you select black, a bit of our branded "gear" and a few other things are listed, but nothing else. If "color" is intended to apply only to branded "gear," it should be a subfilter under that filter.
- "Apparel size" seems odd to have as a main filter. I would expect this at a clothing or department store like Target, not a store that mostly sells tools and parts. It should be a subfilter under "Clothing," which is currently under "Merchandise."

Windows browsers

dhmacs:
fixes #191 

## Changelog

- Add `Main Category` and `Category` to facet blocklist

## QA

3. Verify that neither `Main Category`, nor `Category` are displayed in the facets list

### DIFF
--- a/frontend/components/product-list/sections/FilterableProductsSection/ProductListFilters/index.tsx
+++ b/frontend/components/product-list/sections/FilterableProductsSection/ProductListFilters/index.tsx
@@ -25,6 +25,8 @@ const FACET_BLOCKLIST = [
    'facet_tags.Price',
    'inventory_management',
    'quantity_available',
+   'category',
+   'facet_tags.Main Category',
 ];
 
 const Sizer = chakra(AutoSizer);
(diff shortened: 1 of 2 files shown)


jordycosta:
QA 🟢 

`Main Category` & `Category` no longer show in the facet block-list on Parts ✔️ 

---

### Testing Procedure:
1. Navigate to the Vercel [preview](https://react-commerce-git-add-facets-to-block-list-ifixit.cominor.com/)
2. Select `Parts`
3. Ensure that the two filters listed above are no longer present

---

### Screenshot:

facet

andyg0808 (approved):

andyg0808:
CR :goat: through 11c895c

sctice-ifixit:
CR :rose:

PR #255, opened 2022-04-15
Use iFixit API for default device images

### LINKED ISSUE #247: Use iFixit API for default device images
We probably won't ever get custom images for every PLP, but we have a device image we can use. When a Device's Parts PLP has a missing image (including child device PLPs), look up the image from iFixit's api using the deviceTitle field.

For example, MacBook_Pro would use `https://www.ifixit.com/api/2.0/wikis/CATEGORY/MacBook_Pro`. Docs here: https://www.ifixit.com/api/2.0/doc/Wikis#get-a-specific-wiki but the short version is that there's a field called `image` which has various sizes of images we can use.

image

There is also a `children` field, which contains an array of immediate children of this device. Each of these has a similar `image` field. These children don't necessarily match the children in Strapi, but there's usually significant overlap, so we may want to check that before doing a secondary lookup to get another missing image.

This may end up sort of slow if we have to do many requests, so in the future we may optimize this more, but this should work well enough to get some data in here. We will also likely be using more info from this api in the future, at least for the main device on the page.

dhmacs:
fixes #247 

## QA

3. Verify that `/Store/Parts/Kindle_Tablet` has an hero image, even though the Strapi product list for `Kindle_Tablet` has no image set


frontend/next.config.js
```diff
@@ -21,6 +21,8 @@ module.exports = withTM({
          'valkyrie.cdn.ifixit.com',
          'cart-products.cdn.ifixit.com',
          'assets.cdn.ifixit.com',
+         'www.cominor.com',
+         'https://guide-images.cdn.ifixit.com',
```
andyg0808, line 25: Why does this one have a protocol and the others don't?

frontend/models/product-list/index.ts
```diff
@@ -48,6 +49,11 @@ export async function findProductList(
       return null;
    }
    const productListImageAttributes = productList.image?.data?.attributes;
+   let deviceWiki: DeviceWiki | null = null;
+   if (productList.deviceTitle) {
+      deviceWiki = await fetchDeviceWiki(productList.deviceTitle);
+   }
```
andyg0808, line 55: I don't know if it's worth it, but this eliminates a mutation:
```suggestion
   const deviceWiki = !productList.deviceTitle ? null : await fetchDeviceWiki(productList.deviceTitle);
```

frontend/models/product-list/index.ts
```diff
@@ -58,14 +64,16 @@ export async function findProductList(
...
+            ? getImageFromStrapiImage(productListImageAttributes, 'large')
+            : deviceWiki
+            ? getDeviceImage(deviceWiki)
+            : null,
       ancestors: createProductListAncestors(productList.parent),
       // Strapi sort order is case sensitive, so we need to improve on it in memory
       children: sortProductListChildren(
          filterNullableItems(
-            productList.children?.data.map(createProductListChild)
+            productList.children?.data.map(createProductListChild(deviceWiki))
```
andyg0808, line 76: Curious why this is a higher-order function instead of
```ts
            productList.children?.data.map(child => createProductListChild(child, deviceWiki))
```

Not that I think my version's better, just curious about the style choice.

andyg0808 (approved):

andyg0808:
CR :goat: through ea474d0
deploy_block :+1: on question about protocol.

frontend/models/product-list/index.ts
dhmacs: Not a specific reason for it.. I've just opted for the first refactor that I had in mind.. Maybe I just liked the idea of not having an arrow function there 😅 If you want I can refactor it as you're proposing, just let me know 👍

andyg0808: Nope; totally fine the way it is. Thanks for satisfying my curiosity!

andyg0808 (approved):

andyg0808:
CR :goat: through f18751e

deltuh-vee:
dev_block ✊ 
Looks like the E2E tests failed.

dhmacs:
@deltuh-vee @andyg0808 CI issue fixed 

un_dev_block 👍

frontend/models/product-list/index.ts
```diff
@@ -74,6 +82,48 @@ export async function findProductList(
...
+   const child = deviceWiki.children?.find(
+      (c: any) => c.title === childDeviceTitle
+   );
+   if (child?.image?.original) {
+      return {
+         url: child.image.original,
+         alternativeText: null,
+      };
+   }
+   return null;
```
masonmcelvain, line 124: It might be nice to reuse `getDeviceImage` here, but this is fine, not worth changing.

masonmcelvain (approved):
CR ✌🏻

frontend/models/product-list/index.ts
```diff
@@ -74,6 +82,48 @@ export async function findProductList(
...
+   } catch (error) {
+      return null;
+   }
+}
+
+function getDeviceImage(deviceWiki: DeviceWiki): ProductListImage | null {
+   return deviceWiki.image?.original == null
+      ? null
+      : {
+           url: deviceWiki.image.original,
```
sctice-ifixit, line 106: The `original` size is the largest version of the image that we have. If we can get away with it, we probably want to use the smallest size that we can without having to scale the image up in any view. I would guess that's one of `standard`, `medium`, or `large`, going from smallest to largest.

sctice-ifixit (commented):
dev_block :cactus: It's looking great, but I left a few suggestions for consideration/discussion.

frontend/models/product-list/index.ts
dhmacs: @sctice-ifixit I'm using this because the resize is done by Next Image.. `original` is not the image that ends up being used client side, just the one used by Next.js to create the image set

frontend/models/product-list/index.ts
dhmacs: I like your suggestion. The reason I haven't done it this way is that I would have to change the assumption about the typescript type that gets passed to the function.. `DeviceWiki` is supposed to be the root type returned by the api (even if now it's typed as `any`, it would be nice to autogenerate it in the future).. The child type is not exactly the same as `DeviceWiki` root and that's the reason why I'm not trying to reuse the function here...

masonmcelvain: Ah thanks for the clarification, I didn't realize there was a distinction 👍🏻

dhmacs:
I'm debugging an issue in the production deploy dev_block 🚫

dhmacs:
I'm sorry I have to increase a little bit the scope of this PR to fix issues related to dependencies in `backend` folder. To fix this I'm removing `backend` from pnpm workspaces, and let it be installed separately with npm, so that the correct package versions are installed.

un_dev_block ✨

jordycosta:
QA 🟢 

A hero image for `/Store/Parts/Kindle_Tablet` is displayed ✔️ 

---

### Testing Procedure:
1. Visit the Vercel [Preview](https://react-commerce-git-use-ifixit-api-for-default-dev-813790-ifixit.cominor.com)
2. Select **Parts**
3. Navigate to and click on **Amazon Kindle Parts**
4. Ensure that the page has a hero image

---

### Screen Captures:

hero_image

masonmcelvain (approved):
CR ✌🏻

andyg0808 (approved):

andyg0808:
CR :goat: through 3dba2e3

andyg0808:
One note for the future: please feel free to rebase your history before pushing in order to drop testing commits. We generally try to keep a clean Git history with helpful commit messages, because we use it extensively as documentation for why the code is the way it is.

dhmacs:
@andyg0808 gotcha 👍 un_deploy_block 🏎️

PR #262, opened 2022-04-21
Refactor filters tests

### LINKED ISSUE #261: Fix tests errors

dhmacs:
Fixes #261 

This is a quick refactor to make tests more resilient to rendering delays. In general usage of `.wait(..)` should be avoided as much as possible, so down the road I'd like to further investigate how to improve upon this. 

## QA

We're only refactoring tests, so QA is not really needed here.

qa_req 0


frontend/test/cypress/integration/product-list/filters.spec.ts
```diff
@@ -54,12 +47,11 @@ describe('product list filters', () => {
...
       });
 
-      waitForSearchCompletion();
-
       user
          .findByRole('button', { name: /collapse item type/i, expanded: true })
          .click(getVirtualListClickOptions());
 
+      user.wait('@search');
       user.wait(500);
```
ardelato, line 55: We should use the new constant `DEBOUNCE_MS` here as well

frontend/test/cypress/integration/product-list/filters.spec.ts
```diff
@@ -93,13 +85,13 @@ describe('product list filters', () => {
...
 
-      waitForSearchCompletion();
+      user.wait('@search');
 
       user
          .findByRole('button', { name: /collapse device$/i, expanded: true })
          .click(getVirtualListClickOptions());
 
-      user.wait(500);
+      user.wait(300);
```
ardelato, line 94: Why are we waiting for only 300ms here instead of 500ms?

ardelato (approved):
CR 👍

davidrans (approved):
CR :coffee:

frontend/test/cypress/integration/product-list/filters.spec.ts
dhmacs: Good question! It's because they have different meanings.. Debounce is for when I'm actually using debounce in the ui, here instead I'm just waiting an arbitrary amount of time, and I'm trying to make it shorter.. I'd like to improve this eventually

frontend/test/cypress/integration/product-list/filters.spec.ts
dhmacs: I'm trying to reduce the waiting amount so that we spare a little bit of time.. I'd like to improve the way this is handled eventually

PR #271, opened 2022-04-27
fix(strapi): Update dependencies and Dockerfile

### LINKED ISSUE #268: Bug: Strapi admin dashboards not loading
The admin dashboards for Strapi instances running on [branch-name].govinor.com aren't working for some reason. To recreate:
1. visit main.govinor.com
2. receive error message "Warning: An error occured while requesting the API" and an infinite loading screen

Despite the admin dash being inaccessible, the Next.js frontend still consumes the Strapi instance without any problems.

[The references to this problem](https://forum.strapi.io/t/warning-an-error-occurred-while-requesting-the-api/14313) I've found so far seem to focus on running Strapi in dev, but the common suggestion is to build Strapi with yarn instead of npm before running it. It could easily be something else, but this is a starting point!

dhmacs:
fixes #268 

## Changelog

- Refactor Dockerfile
- Use Yarn to install dependencies
- Update Strapi to v4.1.8

## QA

1. Open https://fix-broken-admin-dashboard.govinor.com/admin
2. Verify that you are able to login and the Strapi dashboard loads correctly


backend/Dockerfile
```diff
@@ -1,19 +1,16 @@
...
-COPY ./config/ ./config/
-COPY ./database/ ./database/
-COPY ./public/ ./public/
-COPY ./favicon.ico ./
-
-RUN npm run build
-
-CMD ["npm", "start"]
\ No newline at end of file
+FROM node:16
```
djmetzle, line 1: Worth pointing out:
Switching from alpine to [buster](https://github.com/nodejs/docker-node/blob/6e8f32de3f620833e563e9f2b427d50055783801/16/buster/Dockerfile#L1) is going to build much larger (read: slower) docker images. We should try to prefer `alpine` when we can? We dont need to block this PR, but i hope we can switch back to alpine in the near future.

dhmacs:
dev_block 🔴

backend/Dockerfile
dhmacs: Good point 👍 I'm still working on the PR, so I'll try to use also alpine

backend/. dockerignore
```diff
@@ -0,0 +1,6 @@
+.tmp/
```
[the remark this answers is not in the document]
dhmacs: Ouch, I'll fix it 👍

dhmacs:
un_dev_block 🏎️

backend/Dockerfile
```diff
@@ -1,19 +1,16 @@
...
-COPY ./public/ ./public/
-COPY ./favicon.ico ./
-
-RUN npm run build
-
-CMD ["npm", "start"]
\ No newline at end of file
+FROM node:16-alpine
+# Installing libvips for sharp compatibility
+RUN apk update && apk add vips-dev
```
djmetzle, line 3: You can skip the `update` with the `-U/--update-cache` command. And once that's complete, you can clear out the APK cache in that layer to lighten up the docker image:
```suggestion
RUN apk --update-cache add vips-dev \
 && rm -rf /var/cache/apk/*
```

erinemay:
What's the standard login for https://fix-broken-admin-dashboard.govinor.com/admin/auth/login?

rickisXP:
Login info is found [here](https://github.com/iFixit/react-commerce#working-with-strapi)

deltuh-vee:
QA 🎬 
I'm able to login and the dashboard still loads without any issues.

backend/Dockerfile
```diff
@@ -1,19 +1,17 @@
...
+# Installing libvips for sharp compatibility
+RUN apk --update-cache add vips-dev \
+ && rm -rf /var/cache/apk/*
+ARG NODE_ENV=production
+ENV NODE_ENV=${NODE_ENV}
+WORKDIR /opt/
+COPY ./package.json ./
+COPY ./yarn.lock ./
+ENV PATH /opt/node_modules/.bin:$PATH
+RUN yarn config set network-timeout 600000 -g
```
andyg0808, line 11: That seems pretty large. Is this just a testing line, or should it stay in here?

backend/docker-compose.yaml
```diff
@@ -13,10 +13,16 @@ services:
...
       ports:
          - '${HOST_PORT:-1337}:1337'
       volumes:
-         - uploads:/srv/app/public/uploads
+         - uploads:/opt/app/public/uploads
+         - ./config:/opt/app/config
+         - ./src:/opt/app/src
+         - ./package.json:/opt/package.json
+         - ./yarn.lock:/opt/yarn.lock
+         - ./.env:/opt/app/.env
```
andyg0808, line 25: Why are all of these directories using volumes? It looks like this will end up putting all the code in volumes, instead of using it from the container. And I think, after the first boot, no further changes will take effect on the volumes; I believe they get copies of the data that's present on first boot, but then remain unchanged after that.

andyg0808 (approved):

andyg0808:
CR :goat: through b9596e1
deploy_block :+1:

.github/workflows/main.yml
```diff
@@ -10,10 +10,19 @@ jobs:
...
+           run: docker ps -a
          - name: Install PNPM
-           run: corepack enable && corepack prepare pnpm@7.0.0-rc.5 --activate
+           run: corepack enable && corepack prepare pnpm@6.32.11 --activate
          - name: Install workspaces
            run: pnpm install
+         - name: Check again docker status
+           run: docker ps -a
+         - name: Check docker logs
+           run: docker logs backend_strapi_1
```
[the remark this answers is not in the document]
dhmacs: Oh gosh, I forget them there 😓

backend/Dockerfile
dhmacs: @andyg0808 
This Dockerfile basically comes from [this article](https://blog.dehlin.dev/docker-with-strapi-v4) which is the community article currently recommended by the Strapi team.

This particular line seems to be there to make Yarn more resilient on slow networks:
 
Screenshot 2022-05-02 at 10 45 18

Not sure if we need it, but I don't see any harm in keep it right?

backend/docker-compose.yaml
dhmacs: @andyg0808 
I think this is configured this way because it can support running Strapi in dev mode (although we probably won't need it). Do you want to get rid of this?

rickisXP:
QA :star: 
- Login on `https://fix-broken-admin-dashboard.govinor.com/admin/auth/login` works as expected
- Dashboard and menus load as expected

backend/Dockerfile
lithobraking: If the worst thing about this is that it's just an unconventional timeout length and we can afford to leave this be without incurring any real performance penalty, then I probably wouldn't worry about changing it for now.

backend/docker-compose.yaml
lithobraking: Do we anticipate any performance problems with keeping this configuration? If not, we can probably leave it alone I think

dhmacs: @lithobraking from Docker docs:

image

Since the dashboard is currently broken I would merge this as is.. I can always open a PR later for Docker optimizations

lithobraking: @dhmacs Agreed. We can worry about it later.

masonmcelvain (approved):
CR ✌🏻 under the assumption that we just want to get the dashboard working, and that performance can wait

backend/Dockerfile
andyg0808: Yup, totally on board with that.

PR #273, opened 2022-05-04
Improve filters test resilience

dhmacs:
This PR makes tests more resilient to performance mismatches between CI and local dev. To achieve this I've been experimenting with increasing waiting time and I've found out that with `2000ms` tests don't fail anymore in CI.
For context, the culprit of these issues is the filters virtual list, which dynamically creates and removes filters and that makes it tricky to perform assertions, because you have to make sure that animations have completed and the dom element that you're trying to assert has been added.

## QA

Nothing to QA, just verify that the tests have passed in CI.


masonmcelvain (approved):
CR ✌🏻

dhmacs:
dev_block 🚫

dhmacs:
By mistake I've pushed a commit that was meant for another branch

un_dev_block 👍

masonmcelvain (approved):
CR ✌🏻

dhmacs:
@erinemay if by that you mean reproducing the same conditions that we have in CI, I haven't yet found a practical way of doing it. I'm playing with waiting times because I'm trying not to spend too much time to find the perfect solution, because it would not be needed later when we remove the virtual list from the filters.

erinemay:
> I'm playing with waiting times because I'm trying not to spend too much time to find the perfect solution, because it would not be needed later when we remove the virtual list from the filters.

Fair. I'll un_deploy_block 👍

danielbeardsley:
CR :+1:

mlahargou:
CR 📱  Is there any sort of `waitForElementToDisappear` or something like that that we could use?

For our WebDriver tests, we use that feature all over. For instance, if we have some sort of loading animation, we can do something like `waitForElementToDissapear('#loading-indicator')`. 

Maybe something like `.should("not.be.visible")` which should automatically wait?

dhmacs:
@mlahargou 

> CR 📱 Is there any sort of `waitForElementToDisappear` or something like that that we could use?

I think that this is the default behaviour of Cypress, only it has a default timeout of 4 seconds. To be honest it's still not completely clear to me why this timeout is not enough sometimes, and adding an additional delay makes the test pass. 
This is also difficult to debug because on my machine the test pass, so my intuition is that somehow depends also on the performance of the machine that runs the tests. 
We have a new mockup for filters that hopefully would also help simplifying the tests.

PR #279, opened 2022-05-06
Use Algolia secured API key

### LINKED ISSUE #257: Create Secured API Key for querying Algolia index

dhmacs:
fixes #257 

## Changelog

- Refactored logic to account for Algolia secured key generation
- Refactored env variables
- Added github secret for `ALGOLIA_API_KEY`
- Added Vercel env variable for `ALGOLIA_API_KEY`

## QA

2. Visit a few product list pages and verify visible product parts reflect the current product list


masonmcelvain (approved):
CR ✌🏻

dhmacs:
`ALGOLIA_API_KEY` env variable needs to be set in AWS Amplify too

cc @djmetzle @sterlinghirsh

frontend/models/product-list/index.ts
```diff
@@ -59,6 +64,17 @@ export async function findProductList(
          ? await fetchDeviceWiki(productList.deviceTitle)
          : null;
 
+   const algoliaApiKey = createProductListAlgoliaKey({
+      appId: ALGOLIA_APP_ID,
+      apiKey: ALGOLIA_API_KEY,
+      filters:
+         productList.filters && productList.filters.length > 0
```
danielbeardsley, line 71: Looks like these filters are already represented in the function being called. It seems like we meant to delete this `filters` param after moving the code to the function but forgot to.

danielbeardsley:
Well done @dhmacs ! I like that we've collected the key generation into fewer places!

I may be missing something, but it seems this still generates the API key on the client side, allowing unfettered access to our index. I was expecting key generation to happen in `getServerSideProps` or `getStaticProps` or something and then we'd only ship *that* key to the client.

dev_block :+1: on a response to this concern.

frontend/models/product-list/index.ts
dhmacs: > Looks like these filters are already represented in the function being called

@danielbeardsley What do you mean by this?

danielbeardsley: Maybe I'm missing it, but it looks like this exact same expression (`filters: ...`) is in the function being called `createProductListAlgoliaKey()`

dhmacs:
> I may be missing something, but it seems this still generates the API key on the client side, allowing unfettered access to our index. I was expecting key generation to happen in `getServerSideProps` or `getStaticProps` or something and then we'd only ship _that_ key to the client.

@danielbeardsley it should be generated only within `getServerSideProps` 🤔 Can you share the line where the key is generated client side?

frontend/models/product-list/index.ts
dhmacs: Oh man, you're right! Now I see what you mean 😅  Fixing it  👨‍💻

danielbeardsley:
Sorry, this shows that I'm new to this project. I mistook `frontend/models/product-list/index.ts` as a front-end component. Is there pattern to our files / functions that makes it obvious which are called only on the server-side?

dhmacs:
> Sorry, this shows that I'm new to this project. I mistook `frontend/models/product-list/index.ts` as a front-end component. Is there pattern to our files / functions that makes it obvious which are called only on the server-side?

That's a concern I have too! Most of the code in `models` folder is meant to be server side only, although it's not really a rule.

[Remix](https://remix.run/) uses a `.server` suffix to enforce server only code. Maybe it's something that we could do too, and it would be nice to enforce this at compile time if possible.

dhmacs:
un_dev_block 🟢

danielbeardsley:
CR :

lithobraking:
CR ✨ through 8544fbf

djmetzle:
Key is added to Amplify :+1:

k0rvusk0r4x:
Product list pages reflect list as expected.

QA 😺

PR #285, opened 2022-05-10
Update AMEX logo seed image

### LINKED ISSUE #229: Use a different American Express logo for footer
The payment options grid in the footer currently uses this logo for American Express:

Screenshot from 2022-03-22 13-56-45
Screenshot from 2022-03-22 13-56-28

We talked it over and reached a consensus that we're not exactly in love with this logo, and were thinking we should use something like this one instead:

amex cropped

Let's see how it looks if we isolate the text from the background in the above image and use that.

dhmacs:
fixes #229 
 
## Changelog

- Replace AMEX logo seed image

## QA

3. Verify that the AMEX logo shown is the [full version of the blue box logo](https://github.com/iFixit/react-commerce/issues/229#issuecomment-1076505437) (although adapted B&W)


lithobraking:
CR ✨

erinemay:
qa 👍 Is the correct label. looks good on desktop and mobile
image

PR #287, opened 2022-05-11
Add guides redirects

### LINKED ISSUE #281: Redirect /Store/Guide/x links
## Problem
We used to support product list pages for specific guides showing all the products used in that guide. The urls took the form `/Store/Guide/{guideid}`. These weren't super userful, since the products were already advertised on the guide itself, and they only worked on the US store. We will no longer support those links, and https://github.com/iFixit/ifixit/issues/42690 has us drop them and add a redirect to the relevant guide page. We will need to maintain that redirect behavior on the next.js app, since random links to these pages still exist in the wild.

## Do This
When a request comes in for `/Store/Guide/113862`, redirect the user to `/Guide/_/113862` . Ideally we'd redirect to `/Guide/iPhone+XS+Screen+Replacement/113862`, which is the canonical link, but I'd rather keep the redirect logic simple and avoid having to hit the guide api just to get the title. The guide page already has logic to redirect to the canonical, so I don't really mind the double redirect in this rare case.

dhmacs:
fixes #281 
 
## Changelog

- Add redirect rule
- Re export env variables to support AWS Amplify lambda poor env variables support

## QA

2. Navigate to `/Store/Guide/113862` and verify that you are redirected to a guide page

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -7,11 +7,26 @@ const withTM = require('next-transpile-modules')([
 ]);
 
 module.exports = withTM({
+   env: {
+      ALGOLIA_API_KEY: process.env.ALGOLIA_API_KEY,
+      NEXT_PUBLIC_ALGOLIA_APP_ID: process.env.NEXT_PUBLIC_ALGOLIA_APP_ID,
+      NEXT_PUBLIC_IFIXIT_ORIGIN: process.env.NEXT_PUBLIC_IFIXIT_ORIGIN,
+      NEXT_PUBLIC_STRAPI_ORIGIN: process.env.NEXT_PUBLIC_STRAPI_ORIGIN,
+   },
    async rewrites() {
       return [
          {
             source: '/uploads/:name',
             destination: `${process.env.NEXT_PUBLIC_STRAPI_ORIGIN}/uploads/:name`,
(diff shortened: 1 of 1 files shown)


djmetzle (approved):
Thanks @dhmacs 
CR :gear:

lithobraking:
CR ✨

erinemay:
QA 👍 
Checked with
- /Store/Guide/113862 (above)
- /Store/Guide/143265 (random one)

Confirmed each of these pages redirect to the guide page. 

deploy_block 👍 
>When a request comes in for `/Store/Guide/113862`, redirect the user to `/Guide/_/113862` . Ideally we'd redirect to `/Guide/iPhone+XS+Screen+Replacement/113862`, which is the canonical link, but I'd rather keep the redirect logic simple and avoid having to hit the guide api just to get the title. The guide page already has logic to redirect to the canonical, so I don't really mind the double redirect in this rare case.


https://github.com/iFixit/react-commerce/issues/281#issue-1228283145
@sterlinghirsh, not sure if y'all talked about it outside of GitHub, but I want to make sure you're ok with the double redirect implementation here.

sterlinghirsh:
un_deploy_block :zap: yep this is fine

PR #294, opened 2022-05-12
Fix product list pagination url issue

### LINKED ISSUE #293: Bug: Collection pagination links have `[handle]` in href URL
## Problem
On collection pages other than `/Store/Parts` or `/Store/Tools`, the pagination links at the bottom of each page show a url that says something like `/Store/Parts/[handle]?handle=iPhone&p=4` for an example. The javascript is correct and when clicked the button takes you to the correct page, but the URL that's displayed in the HTML is incorrect. 

Screenshot 2022-05-11 101223

## Steps to Reproduce
1. Go to https://next.ifixit.com/Store/Parts
2. Navigate to iPhone collection page
3. Inspect pagination buttons at bottom of the page
4. Observe that the href for the fourth page button is `href="/Store/Parts/[handle]?handle=iPhone&p=4"`

## Do This
The URL _should_ be more like `/Store/Parts/iPhone?p=4`.

CC: @ifixitmaxb @sterlinghirsh @dhmacs

dhmacs:
fixes #293 

## QA

2. Verify that the problem described by #293 is fixed

### DIFF
--- a/frontend/components/product-list/sections/FilterableProductsSection/ProductListPagination.tsx
+++ b/frontend/components/product-list/sections/FilterableProductsSection/ProductListPagination.tsx
@@ -8,6 +8,7 @@ import {
 import { usePagination } from '@lib/algolia';
 import { useRouter } from 'next/router';
 import queryString from 'query-string';
+import { ParsedUrlQuery } from 'querystring';
 import * as React from 'react';
 import {
    HiChevronDoubleLeft,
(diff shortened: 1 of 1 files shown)


masonmcelvain (approved):
CR ✌🏻 

It's interesting that next.js puts the dynamic route parameters in the [router](https://nextjs.org/docs/api-reference/next/router#router-object).query object. Seems like distinguishing them in their own object would be easier, but maybe there's a use case for lumping them all together that I'm not aware of.

jordycosta:
QA 🟢 

The _href_ for pagination links no longer contain `[handle]`

---

### Testing Procedure:
1. Navigate to the Vercel [preview](https://react-commerce-git-fix-product-list-pagination-url-ifixit.cominor.com/Store/Parts/iPhone?p=4)
2. Inspect element
3. Ensure that the href elements for pagination links match #293 **Do This**

---

### Screen Captures:
href

lithobraking (approved):
CR ✨

dhmacs:
> It's interesting that next.js puts the dynamic route parameters in the [router](https://nextjs.org/docs/api-reference/next/router#router-object).query object. Seems like distinguishing them in their own object would be easier, but maybe there's a use case for lumping them all together that I'm not aware of.

@masonmcelvain you're absolutely right, they should improve their router api!

Issue #296, opened 2022-05-12
Strapi: The docker image is 3GB, this is very big


danielbeardsley:
Do This
===
Find a way to trim down the docker image while it's being built

djmetzle:
Interesting. There's a tool we can use to figure out whats going on...

djmetzle:
Im peeking at this with [`dive`](https://github.com/wagoodman/dive), and the image looks like its around 1.2GB, not 3GB. Not sure where we're seeing that?

djmetzle:
I figured out how to shave a significant amount of bytes off of the image

danielbeardsley:
> Not sure where we're seeing that?

That's good to hear!

Docker spit it out after building the image. Something like:

"Finished building and pushing, size: 2834" (which I assumed was MB).

Maybe AWS squashed some of the layers or compressed it or something. Or that size wasn't what I thought it was.

PR #297, opened 2022-05-13
Clear yarn cache during docker build

djmetzle:
The strapi docker image is kind of big-ish at 1.2GB. We can cut that almost _in half_ by cleaning out the yarn package cache.

Clearing the cache brings the image down to ~650MB, which isnt the smallest image, but is a relatively normal size.

qa_req 0

CC @danielbeardsley 

Closes: https://github.com/iFixit/react-commerce/issues/296

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -9,7 +9,7 @@ COPY ./package.json ./
 COPY ./yarn.lock ./
 ENV PATH /opt/node_modules/.bin:$PATH
 RUN yarn config set network-timeout 600000 -g
-RUN yarn install
+RUN yarn install && yarn cache clean --all
 WORKDIR /opt/app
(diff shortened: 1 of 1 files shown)


djmetzle:
Clearing the cache brings the size down from 1.2GB to 646MB for me, testing.
image

I'm not well versed in `yarn` usage. Perhaps @dhmacs or @danielbeardsley have suggestions to improve from here?

danielbeardsley:
CR :+1: 

Nice!

dhmacs:
CR ⚡  Looks neat @djmetzle ! I don't have other suggestions related to Yarn, cleaning the cache is the only thing that comes to mind. I would have to spend some time analyzing the Docker image to find out possible improvements. I think this already is a big win though 💪

PR #310, opened 2022-05-28
Next: support strapi images

### LINKED ISSUE #303: Images from Strapi not loading
On https://next.ifixit.com/Store/Parts, we have manually set images in Strapi for sub categories like Mac. It appears that the response from Strapi is only giving the path as `/uploads/imac_d8ebc98f54.png`. We pass that to next.js, which results in `https://next.ifixit.com/_next/image?url=%2Fuploads%2Fimac_d8ebc98f54.png&w=384&q=75` which gives a 500 response with the message `"url" parameter is valid but upstream response is invalid`.

I tried adding https://strapi.ifixit.com to the front of the url. Visiting it directly works https://strapi.ifixit.com/uploads/imac_d8ebc98f54.png but next.js refuses to render it. https://next.ifixit.com/_next/image?url=https%3A%2F%2Fstrapi.ifixit.com%2Fuploads%2Fimac_d8ebc98f54.png&w=384&q=75 gives a 400 error with the message `"url" parameter is not allowed`

danielbeardsley:
Support loading and using images from strapi in next.js

We may want to eventually make a `StrapiImage` component that perhaps uses the images from strapi directly instead of re-compressing them, but for now, this allows us to use the images in the `next/image` component.

There's actually a lot of decisions to make about how to deal with images. As I've done more research I've realized there's more complexity in hosting / serving images through next.js than I thought. For now, this gets us closer.

See #314 for the further research / decisions

Specifically:
* Makes it possible to use the strapi images in next/image
* Sets one-year cache headers on all strapi uploads (see commit message)

qa_req 0 The vercel deployment points at an empty strapi so it's not effective at testing this. I think for now, CI should be our QA and we'll see if the config change works live.

Closes #303


andyg0808 (approved):

andyg0808:
CR :goat: through c9c5495

PR #318, opened 2022-06-02
Search facets  hide irrelevant ones

### LINKED ISSUE #299: Hide facets when corresponding Wiki Info is set
## Problem
We show some facets when it's not appropriate. For example, on https://next.ifixit.com/Store/Parts/iPhone, we show the Device Type facet, even though we should already know that the user is looking for parts with device type = iPhone. Since we're fetching data from our api anyway, let's use the that info field to hide extraneous facets.

## Do This
For each facet in product_en.facet_tags, hide it if its name exists in the wiki info. For example, iPhone's wiki info (bottom of https://www.ifixit.com/api/2.0/wikis/CATEGORY/iPhone ) includes a record for Device Type, so hide `facet_tags.Device Type`.

danielbeardsley:
Hide Facets that are already specified by the Device Parts page we are on.

Context: Assume we are on the `iPhone 7` parts list page
The "Device Brand" facet and its "iPhone" option is useless and
confusing cause *all* products here are already compatible with iPhones.

Here we hide any facets that come from the wiki info of the current
device and its ancestors.

Closes #299

### DIFF
--- a/frontend/components/product-list/sections/FilterableProductsSection/ProductListFilters/index.tsx
+++ b/frontend/components/product-list/sections/FilterableProductsSection/ProductListFilters/index.tsx
@@ -189,27 +189,23 @@ function useFilteredFacets(wikiInfo: WikiInfoEntry[]) {
    const user = useAuthenticatedUser();
    const isProUser = user.data?.discountTier != null;
    const facets = useFacets();
-   const infoNames = React.useMemo(
-      () => new Set(wikiInfo.map((info) => `facet_tags.${info.name}`)),
-      [wikiInfo]
-   );
-   const sortedFacets = React.useMemo(() => {
-      return facets.slice().sort((a, b) => a.name.localeCompare(b.name));
-   }, [facets]);
-   const usefulFacets = React.useMemo(() => {
-      const facets = sortedFacets
+   const infoNames = React.useMemo(() => {
+      return new Set(wikiInfo.map((info) => `facet_tags.${info.name}`));
+   }, [wikiInfo]);
+   const [userFacets, refinedFacets] = React.useMemo(() => {
+      const usefulFacets = facets
+         .slice()
+         .sort((a, b) => a.name.localeCompare(b.name))
          .filter(isUsefulFacet)
          .filter((facet) => !infoNames.has(facet.algoliaName));
-      return isProUser ? facets.filter(isAvailableToProUsers) : facets;
-   }, [sortedFacets, isProUser, wikiInfo]);
-   const refinedFacets = React.useMemo(() => {
-      return usefulFacets.filter(hasMatchingOptions);
-   }, [usefulFacets]);
-   const displayedFacets = React.useMemo(() => {
-      return refinedFacets.length > 0 ? refinedFacets : usefulFacets;
-   }, [usefulFacets, refinedFacets]);
+      const userFacets = isProUser
+         ? usefulFacets.filter(isAvailableToProUsers)
+         : usefulFacets;
+      const refinedFacets = userFacets.filter(hasMatchingOptions);
+      return [userFacets, refinedFacets];
+   }, [facets, isProUser, infoNames]);
    return {
-      facets: displayedFacets,
+      facets: refinedFacets.length > 0 ? refinedFacets : userFacets,
       areRefined: refinedFacets.length > 0,
    };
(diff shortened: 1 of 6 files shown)


frontend/components/product-list/sections/FilterableProductsSection/index.tsx
```diff
@@ -48,7 +51,12 @@ export enum ProductViewType {
 
 const PRODUCT_VIEW_TYPE_STORAGE_KEY = 'productViewType';
 
-export const FilterableProductsSection = React.memo(() => {
+type SectionProps = {
+   productList: ProductListType;
+};
+
+export const FilterableProductsSection = React.memo((props: SectionProps) => {
+   const { productList } = props;
```
andyg0808, line 59: Is there a reason not to do this?
```suggestion
export const FilterableProductsSection = React.memo(({ productList }: SectionProps) => {
```

frontend/components/product-list/sections/FilterableProductsSection/ProductListFilters/index.tsx
```diff
@@ -181,17 +185,23 @@ function itemKey(index: number, data: ItemData): string {
...
       return facets.slice().sort((a, b) => a.name.localeCompare(b.name));
    }, [facets]);
    const usefulFacets = React.useMemo(() => {
-      const facets = sortedFacets.filter(isUsefulFacet);
+      const facets = sortedFacets
+         .filter(isUsefulFacet)
+         .filter((facet) => !infoNames.has(facet.algoliaName));
       return isProUser ? facets.filter(isAvailableToProUsers) : facets;
-   }, [sortedFacets, isProUser]);
+   }, [sortedFacets, isProUser, wikiInfo]);
```
andyg0808, line 204: Might be good to do this, so the dependencies are directly indicated:
```suggestion
   }, [sortedFacets, isProUser, infoNames]);
```

andyg0808 (approved):

andyg0808:
CR :goat: through 5f48ae7
I'm not an expert, but this looks pretty reasonable.

frontend/components/product-list/sections/FilterableProductsSection/index.tsx
danielbeardsley: Ha! Yes, this is the code I typed originally, but thanks to auto-formatting it added another level of indentation to the whole body of the component. This makes the code slightly harder to work with so I decided that a simple extraction line is just as easy to read and doesn't indent the body any more.

frontend/components/product-list/sections/FilterableProductsSection/ProductListFilters/index.tsx
danielbeardsley: Good point, but I went with refactoring this whole bit instead.

dhmacs:
@danielbeardsley I'm working on a new PR that is basically switching the filters implementation to use the new Algolia React hooks https://github.com/iFixit/react-commerce/tree/use-new-algolia-hooks . This by itself will also solve many of the outstanding issues related to filters, so I'd pause any work related to filters and wait for this branch to be merge. I'm sorry that you've already put some work into this, I should have assigned all these issue to myself to lock them.

deploy_block 🚫

danielbeardsley:
I could be wrong, but it doesn't appear your in-progress branch touches the code I'm mutating here (yet). Given that this is in review and much smaller, I'm not inclined to block this change on a much larger refactor. Also, algolia isn't going to know about our wikis, their wiki info, nor how that should affect which facets to show, so using the new hooks won't solve this particular issue.

un_deploy_block :+1:

dhmacs:
@danielbeardsley The code you're touching is going to be dropped entirely on my branch. 😁  It's only there because I haven't yet dropped it. For example the filter modal is replaced with a drawer. Also the way facets are sourced is changing. Everything Algolia related is basically moving to the new Algolia React hooks. 
The work on my branch is almost done, just need to polish it (including dropping the old implementation files), I'll open a PR for it early next week.

deploy_block ⛔

sterlinghirsh:
CR :zap: but block remains.

erinemay:
Feel free to remove the `external_block` label if this PR persists and you want it tested.

masonmcelvain:
Closing in favor of #331

PR #309, opened 2022-05-26
Strapi Uploads: persist in s3

danielbeardsley:
Make the S3 bucket configurable via an env variable.

Closes #309

PR #319, opened 2022-06-03
Fix Govinor deployment seed

dhmacs:
This PR fixes a Govinor deployment issue introduced by #309 that breaks the seed of new govinor preview instances.

The bug is due to Govinor usage of `.env.example` to populate env variables, but since Govinor deploys are not required to use S3, we're commenting out env variables related to that so they're not used by Govinor.
This is just a quick fix as probably a better way to solve it would be to build env variable management into Govinor.

# QA

1. Open the Govinor Strapi preview deployment for this branch
2. Verify that product lists have been seeded

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -1,5 +1,8 @@
-AWS_REGION="us-west-1"
-S3_BUCKET="some-bucket-name"
+# This file is used by govinor to generate preview deployment for the branch
+# Every env variable in this file will be passed to the govinor preview deployment.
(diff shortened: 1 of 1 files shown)


sterlinghirsh:
CR :zap: if this is a quick temporary fix. Seems like we should use s3 on govinor too if possible.

deltuh-vee:
QA 🎬 
ProductList is seeded on Govinor for this branch.

PR #315, opened 2022-06-01
frontend: Update `graphql`

andyg0808:
This is only a minor version bump and the docs don't seem to show any breaking changes. But it fixes a peer dependency issue:

```
@graphql-codegen/cli
└─┬ @graphql-tools/prisma-loader
  └─┬ @graphql-tools/url-loader
    └─┬ subscriptions-transport-ws
      └── ✕ unmet peer graphql@"^15.7.2 || ^16.0.0": found 15.5.3
```

## QA Notes
Not sure that there's much we need to do; I think this should be used enough that CI should be good.

Connects https://github.com/iFixit/ifixit/issues/43113

PR #320, opened 2022-06-03
Images for imageless devices

### LINKED ISSUE #304: Get images for ProductList children that aren't children of the device
## Problem
We have some instances where we set up ProductList children in a way that skips steps of the Device tree. For example, on Lenovo IdeaPad Parts https://next.ifixit.com/Store/Parts/Lenovo_IdeaPad we have a child IdeaPad 110-14AST. However, on the Lenovo IdeaPad Device page https://www.ifixit.com/Device/Lenovo_IdeaPad the immediate child is Lenovo IdeaPad 100 Series, and the grandchild is Lenovo IdeaPad 110-14AST. This difference is on purpose, since a we have guides for many IdeaPad 100 Series devices, but the only one we sell parts for is Lenovo IdeaPad 110-14AST, and we want to minimize extra steps in the navigation.

The way image loading is performed, we can't load the images except for the device's immediate children.

## Do This
After checking strapi and the current device's children, use the multiple-device lookup api defined in https://github.com/iFixit/ifixit/issues/43064 to fetch devices. If there are a very high number if devices to check, this may still have to be split up if we hit a length limit on urls or something.

danielbeardsley:
Fill in images for some devices on Parts ProductList pages that were missing images.

### More specifically:
The list of child-nodes we show on a given product list page (parent device) sometimes don't match up with the list of device children in our ifixit.com hierarchy. When we query ifixit.com for the parent device, we only get pictures for the immediate children. For those nodes in strapi which aren't immediate children of the parent-device in our device hierarchy, the ifixit.com device api thus won't give us pictures. For these nodes without pictures, query their information (in bulk) via the new `topic_images` API so we can get images for them.

Closes #304

### DIFF
--- a/frontend/lib/ifixit-api/devices.ts
+++ b/frontend/lib/ifixit-api/devices.ts
@@ -22,3 +22,32 @@ export async function fetchDeviceWiki(
       return null;
    }
 }
+
+export type MultipleDeviceApiResponse = {
+   images: Record<string, string>;
+};
+
+export type GuideImageSize =
+   | 'mini'
+   | 'thumbnail'
+   | '140x105'
+   | '200x150'
+   | 'standard'
+   | '440x330'
+   | 'medium'
+   | 'large'
+   | 'huge';
+
+export function fetchMultipleDeviceImages(
+   deviceTitles: string[],
+   size: GuideImageSize
+): Promise<MultipleDeviceApiResponse> {
+   const apiUrl = new URL(`${IFIXIT_ORIGIN}/api/2.0/wikis/topic_images`);
+   apiUrl.searchParams.set('size', size);
+   deviceTitles.forEach((deviceTitle) => {
+      apiUrl.searchParams.append('t[]', deviceTitle);
+   });
+   return fetch(apiUrl.toString())
+      .then((response) => response.json())
+      .catch(() => ({ images: {} }));
+}
--- a/frontend/lib/ifixit-api/devices.ts
+++ b/frontend/lib/ifixit-api/devices.ts
@@ -1,0 +1,24 @@
+import { IFIXIT_ORIGIN } from '@config/env';
+import { getDeviceHandle } from '@models/product-list';
+
+export type DeviceWiki = Record<string, any>;
+
+export async function fetchDeviceWiki(
+   deviceTitle: string
+): Promise<DeviceWiki | null> {
+   const deviceHandle = getDeviceHandle(deviceTitle);
+   try {
+      const response = await fetch(
+         `${IFIXIT_ORIGIN}/api/2.0/wikis/CATEGORY/${deviceHandle}`,
+         {
+            headers: {
+               'Content-Type': 'application/json',
+            },
+         }
+      );
+      const payload = await response.json();
+      return payload;
+   } catch (error: any) {
+      return null;
+   }
+}
--- a/frontend/lib/ifixit-api/devices.ts
+++ b/frontend/lib/ifixit-api/devices.ts
@@ -1,5 +1,4 @@
 import { IFIXIT_ORIGIN } from '@config/env';
-import { getDeviceHandle } from '@models/product-list';
 
 export type DeviceWiki = Record<string, any>;
 
@@ -50,4 +49,11 @@ export function fetchMultipleDeviceImages(
    return fetch(apiUrl.toString())
       .then((response) => response.json())
       .catch(() => ({ images: {} }));
+}
+
+/**
+ * Convert product list device title to a URL friendly slug
+ */
+export function getDeviceHandle(deviceTitle: string): string {
+   return deviceTitle.replace(/\s+/g, '_');
 }
--- a/frontend/models/product-list/index.ts
+++ b/frontend/models/product-list/index.ts
@@ -143,7 +143,7 @@ async function fillMissingImagesFromApi(
       if (imageFromDevice != null) {
          child.image = {
             url: imageFromDevice,
-            alternativeText: null,
+            alternativeText: child.deviceTitle,
          };
       }
    });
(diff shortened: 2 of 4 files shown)


frontend/models/product-list/index.ts
```diff
@@ -134,6 +121,35 @@ function getDeviceImage(deviceWiki: DeviceWiki): ProductListImage | null {
...
+   if (childrenWithoutImages.length === 0) {
+      return productListChildren;
+   }
+   const deviceTitlesWithoutImages = childrenWithoutImages.map(
+      (child) => child.deviceTitle
+   ) as string[]; // cast is safe cause we filter nulls above,
+                  // typescript just doesn't understand
+   const imagesResponse = await fetchMultipleDeviceImages(
+      deviceTitlesWithoutImages,
+      'standard'
```
sterlinghirsh, line 139: Looks like this is going to be working at about 80x60 so we can use a smaller size here.
```suggestion
      'thumbnail'
```

frontend/models/product-list/index.ts
```diff
@@ -134,6 +121,35 @@ function getDeviceImage(deviceWiki: DeviceWiki): ProductListImage | null {
...
+   );
+   childrenWithoutImages.forEach((child) => {
+      const imageFromDevice = imagesResponse.images[child.deviceTitle as string];
+      if (imageFromDevice != null) {
+         child.image = {
+            url: imageFromDevice,
+            alternativeText: null,
```
sterlinghirsh, line 146: Might as well make this the device title.
```suggestion
            alternativeText: child.deviceTitle,
```

sterlinghirsh:
CR :zap: dev_block :zap: this looks great, though I left a few suggestions.

danielbeardsley:
Good suggestions, implemented!

 un_dev_block :+1:

sterlinghirsh:
CR :zap:

sterlinghirsh (dismissed):

frontend/lib/ifixit-api/devices.ts
```diff
@@ -0,0 +1,53 @@
+import { IFIXIT_ORIGIN } from '@config/env';
+import { getDeviceHandle } from '@models/product-list';
```
dhmacs, line 2: What do you think about moving the implementation of `getDeviceHandle` in this new file so we avoid a circular dependency?

dhmacs (changes requested):
LGTM 👍 I've added only a small suggestion

frontend/lib/ifixit-api/devices.ts
danielbeardsley: Good call, Fixed!

sterlinghirsh:
CR :zap:

sterlinghirsh (approved):

dhmacs (approved):
CR 🚀

erinemay:
dev_block ❓ 

@danielbeardsley I'm seeing a 404 anywhere past the root page when I check out the vercel preview, similar to https://github.com/iFixit/react-commerce/pull/315#issuecomment-1145085143. Is there something broken with the Vercel build/preview process? 

I'm still seeing the master development branch looks ok: 
Master: https://react-commerce.cominor.com/Store/Parts ✔️ works
This preview: https://react-commerce-git-images-for-imageless-devices-ifixit.vercel.app/Store/Parts ❌ 404's

Both root pages work:
https://react-commerce.cominor.com/ ✔️
https://react-commerce-git-images-for-imageless-devices-ifixit.cominor.com/ ✔️


Notes: 
- Any other pages, like `/Store/Parts/Android` are also 404ing on this branch.
- Replacing `vercel.app` with `cominor.com` in the URL doesn't change the behavior noted.

danielbeardsley:
Hmm..   I don't see any reason for the 404s in this changeset, but I'll look again... 

It seems the only way to 404 is if the whole `productlist` evaluates to null. But this change doesn't affect that return value. I'll investigate this more.

sterlinghirsh:
Looks like we have this set up to show a 404 visually, even when there's a 500. Looking in the js console shows a 500. Let's not pretend it's a 404 when there's an error, at least in dev.

danielbeardsley:
> Looks like we have this set up to show a 404 visually,

Good job noticing that! That seems.... annoying. I'll look into it.

danielbeardsley:
Works locally just fine. I don't know why it's failing in Vercel. It looks like we 404 if we get nothing back from strapi for the productList api. Still investigating.

dhmacs:
@erinemay @danielbeardsley Strapi deployment was down because this branch was still affected by the #309 issue. I've removed the S3 env variable and now it's up an running again 👍

erinemay:
> @erinemay @danielbeardsley Strapi deployment was down because this branch was still affected by the #309 issue. I've removed the S3 env variables from the instance and now it's up an running again 👍

un_dev_block 👍 Thanks you!

erinemay:
QA ☂️ 
Spot checked a few collections (including Lenovo Ideapad from the issue) on [this branch](https://react-commerce-git-images-for-imageless-devices-ifixit.vercel.app/Store/Parts/Lenovo_IdeaPad) and [production](https://next.ifixit.com/Parts/Lenovo_IdeaPad). Confirmed all the images show up and accurately reflect their corresponding device images on ifixit.com.

Note: There isn't data parity between production, main development branch, and individual branches. So, there were quite a few more devices listed in https://next.ifixit.com/Parts/Lenovo_IdeaPad than in https://react-commerce-git-images-for-imageless-devices-ifixit.vercel.app/Store/Parts/Lenovo_IdeaPad. Confirmed that the difference is due to the data in [Strapi](https://main.govinor.com/admin/content-manager/collectionType/api::product-list.product-list?page=1&pageSize=10&sort=title:ASC&plugins[i18n][locale]=en) and not due to a UI change from this branch.

PR #323, opened 2022-06-06
Store/Parts and Store/Tools: Move pages to top level

andyg0808:
This will make deploy much easier, because we won't need to try to sort out the existing product pages' paths from the new NextJS paths. These paths are currently unused, although the PHP app will serve pages from them (we just don't use them officially).

We'll want to eventually add a redirect in the PHP app to redirect the old `Store/Parts` and `Store/Tools` pages to the new `Parts/` and `Tools/` URLs which will be served by NextJS.


danielbeardsley (approved):
CR :+1:

sterlinghirsh:
This is nearly identical to https://github.com/iFixit/react-commerce/pull/325 though I notice there are some differences. I'm going to dev_block :zap: both of them and let's figure out which is correct or merge both changesets into a single PR.

frontend/models/product-list/index.ts
```diff
@@ -174,7 +174,7 @@ export function getProductListPath(
 ): string {
    switch (productList.type) {
       case Enum_Productlist_Type.Tools: {
-         return `/Store/Tools/${productList.handle}`;
+         return `/Tools/${productList.handle}`;
       }
       case Enum_Productlist_Type.Marketing: {
          return `/Store/${productList.handle}`;
```
masonmcelvain, line 180: I think this is the only difference between this pull and #325

https://github.com/iFixit/react-commerce/pull/325/files#diff-3cc5a6bae5107749a587f5a63495efd258b9c9b13c80141b605d0e996d5ab0b8

masonmcelvain: Not true, @andyg0808 pointed out another class of differences [here](https://github.com/iFixit/react-commerce/pull/325#pullrequestreview-998937888)

masonmcelvain:
I think we'll use #325, sorry for the duplicate work @andyg0808, but thanks for taking this on

PR #325, opened 2022-06-07
Routing: Exclude `/Store` prefix from urls

### LINKED ISSUE #322: Change base parts and tools urls to exclude /Store
[`routing--exclude-store-prefix-from-urls`](https://github.com/iFixit/react-commerce/compare/routing--exclude-store-prefix-from-urls)

Related to https://github.com/iFixit/ifixit/pull/43268

## Problem
We want to make our urls less ambiguous. Our site currently uses /Store as a prefix for product pages and collection pages. This makes routing complex, since we have to check certain path elements to see if they look like productcodes. This is even worse when we have to do this routing at the cloudfront layer.

## Do This
Change collection urls in next.js to be like /Parts, /Parts/iPhone. Same for Tools. /Store by itself can remain.

masonmcelvain:
All `/Store/Parts/*` pages are now at `/Parts/*`. Same for `/Tools`.

`Store/[handle]` pages still exist. We'll redirect them to `/Tools/[handle]` or `/Parts/[handle]` in the future, but not in this pull.

## QA
Visit the preview. Make sure that there are no more /Store/Parts/* pages and no more /Store/Tools/* pages. Make sure we don't link to those kinds of pages anywhere either. Thanks!

CC @sterlinghirsh 

Closes #322


jordycosta:
QA 🟢 

All models on the preview exclude `/Store` in the URL ✔️ 

---

### Testing Procedure:
1. Navigate to the Vercel [preview](https://react-commerce-git-routing-exclude-store-prefix-f-84b881-ifixit.vercel.app/)
2. Click on both Parts & Tools _(open in separate tabs)_
3. Ensure that all models exclude `/Store` in the URL and follow the spec shown above

> Note:  Product pages will still have the /Store prefix because the new ink of pixel pages only replace product list pages

sterlinghirsh:
This is nearly identical to https://github.com/iFixit/react-commerce/pull/323 though I notice there are some differences. I'm going to dev_block :zap: both of them and let's figure out which is correct or merge both changesets into a single PR.

andyg0808 (approved):
I think there are a bunch more cases of `/Store/Tools` in the seed data; I don't know if they need to be updated or not.

andyg0808:
CR :goat: through c929aed
but deploy_block :+1: on possible issue with `/Store/Tools`.

masonmcelvain:
> I think there are a bunch more cases of /Store/Tools in the seed data; I don't know if they need to be updated or not.

Oh I see that difference between our pulls now, good catch @andyg0808. I only changed strings like `[tools](/Store/Tools)` to `[tools](/Tools)` and left product page links (like https://pnoble.cominor.com/Store/Tools/P5-Pentalobe-Screwdriver-Retina-MacBook-Pro-and-Air/IF145-090) as is, since product pages will still have the /Store prefix.

masonmcelvain:
There is also `/Store/Kits` in `menus.json`, which I left as is. Wasn't sure if that needed to be changed as well @sterlinghirsh

masonmcelvain:
I'm also now noticing that there are some legacy product page links in the seed data, like https://www.ifixit.com/Store/iPad/iPad-3rd-Gen-LTE-GSM. I don't think simply dropping `/Store` will work on those unfortunately, not sure how to reconcile those

sterlinghirsh:
/Store/Kits should just be /Kits. Legacy product page links in the seed data can remain as-is for now. We can change them eventually but we can use them to test redirects anyway. Product links should remain as-is.

masonmcelvain:
un_deploy_block ✌🏻 un_dev_block ✌🏻 I'm gonna let's use this PR. Sorry for the duplicate work @andyg0808

masonmcelvain:
QA ✌🏻 carry forward, did a smoke test and everything still works. Only changed one line in the seed data

andyg0808 (approved):

andyg0808:
CR :goat: through 38657da

PR #326, opened 2022-06-09
Use new algolia hooks

### LINKED ISSUE #137: Improve range filter input UX
From https://github.com/iFixit/react-commerce/pull/132#issuecomment-973455275 :

> While entering a value into the price range fields, the filtering side bar automatically refreshes as soon as a value is entered into the field. This is leading to a bad user experience, since I might not have finished typing out the value, yet the filtering side bar is already refreshing making it disorienting.  
> 
> I expected that the page would be refreshed with the filter in one of these cases:
>   - After ~3+ seconds since I last entered a digit 
>   - After I leave the focus of the field
>   - After pressing <kbd>enter</kbd>
> 
> https://user-images.githubusercontent.com/22064420/142512707-58c18762-fdc6-4619-9f4b-68fa14fd9b31.mp4

### LINKED ISSUE #165: Fix filter list not correctly re-rendered when there are no results
## Steps to reproduce

- [ ] Expand a filter (e.g. color)
- [ ] Expand price and set min price to `10000`

The color accordion item is not correctly re rendered

### LINKED ISSUE #225: Too much empty space in filter container and search results field
The container for product the filters is longer than necessary when the filters are all condensed. This leads to a state during scrolling when the user is looking at an empty container on the left. 

<details><summary>Tons of whitespace with empty search results</summary>

image

</details>

<details><summary>Some whitespace in the filter box</summary>

image

</details>

**To do with the space in the filter box:**
Not entirely sure what to do here, but I did a quick check on some other sites. I didn't find much inspiration since they see any blank space as ad territory....

Amazon, Ebay, and Carparts all leave the excess space empty (not an empty container, but just empty background). None of their filter containers are sticky, though. Best Buy's filter container is sticky, and they just fill the remaining space with ads. 

cc: @jeffsnyder

### LINKED ISSUE #239: Facet dividers should be between facets only
Horizontal facet dividers should be between facets only. A horizontal divider line above the top facet or below the bottom facet implies there's some division happening. This feels like a bottom facet was supposed to load but didnt:

image

### LINKED ISSUE #298: Page crash when specifying facet value
## Problem
I recently added "worksin" as a facet on product_en. Going through the navigation, I can get to https://next.ifixit.com/Store/Parts/iPhone_7?item_type=motherboards&worksin=i-phone-7-a-1660-verizon-sprint-china . However, if I specify another value for the facet, the page says 404, but reports a 500 in the JS console. In my case, I used the urlencoded formatting of the worksin value that we use for display: https://next.ifixit.com/Store/Parts/iPhone_7?item_type=motherboards&worksin=iPhone%207%20A1660%20Verizon%2FSprint%2FChina but any random value for worksin seemed to trigger this error.

## Do This
Handle an unexpected facet value gracefully. Don't crash the page, just show a no-results page and allow me to remove that filter like any other filter.

Ideally, using the display version of the facet value should still work. Not sure if there's a reason that won't do.

### LINKED ISSUE #301: Accept literal filter values in query params
Related to https://github.com/iFixit/react-commerce/issues/298

## Problem
We want to be able to link to a product list page with a filtered view like https://next.ifixit.com/Store/Parts/iPhone_7?item_type=motherboards&worksin=iPhone%207%20A1660%20Verizon%2FSprint%2FChina . Specifically, the worksin value will be part of the filter query. However, the way facet filtering works with urls right now, that won't work.

Currently, the query is first executed with no filters to build a list of possible facet values. Then, each facet value is kebab-cased, and mapping is created from kebab-case to the real value. Then, the url query parameters are searched for matching facet handles (expecting the url param values to be kebab-case). Once a parameter is found, we search for the real value in the kebab-case mapping. Once a match is found, we re-execute the query with the real value included in the filter.

This has a few issues:
1. We have to double-execute the query to first build the mapping from kebab-case to normal case. This is not ideal from a performance perspective.
2. Algolia has a limit on how many facet values it will return (default 100, max 1000). If the value in the query param doesn't happen to be at the top of that list, we won't find a match at all.
3. This strategy prevents us from using filterOnly facets (i.e. ones that don't list their possible values in the algolia response). Ideally, we can use filterOnly facets for things like worksin that will have thousands of unique values.

## Do This
Allow a way to set a literal filter value in a query param without requiring the kebab-case mapping. The query params won't look as nice, but I don't think that's a major issue.

Ideally, make it so we don't have to double-execute the query.

We may have to manually create a list of query params that should be turned into filters. This could augment the list derived from the algolia response. I don't imagine there will be too many filterOnly cases like this. Alternatively, we may be able to query Algolia on the server side for a complete list of facets, cache that response, and use that to check for url query params.

This relates to #298 because I suspect we crash when trying to look up a value from a query param that doesn't exist in the kebab-case mapping.

dhmacs:
closes #301 
closes #298 
closes #137 
closes #165 
closes #225 
closes #239 

This PR drops the custom Algolia implementation in favour of the new [React InstantSearch Hooks](https://www.algolia.com/doc/guides/building-search-ui/what-is-instantsearch/react-hooks/) library.
The motivation behind this change is that issue #301 requires a deep change of the custom implementation, so refactoring to the new React InstantSearch hooks library is an easier refactor and brings long term advantages including not having to maintain our custom implementation and faster deployment of new features using the readily available hooks.

Since filters component had to be refactored I took this opportunity to implement the [new design](https://www.figma.com/file/jRG4EyQXRDMGfyLqut08KP/iFixit-Shopify-Working-Files?node-id=3245%3A29545). 

## QA

2. Verify that you can filter using encoded url (e.g. https://react-commerce-git-use-new-algolia-hooks-ifixit.cominor.com/Parts?filter%5Bworksin%5D%5B0%5D=iPhone%207%20A1660%20Verizon%2FSprint%2FChina) as suggested in #301 (note that the syntax slightly differs in that params a enclosed in an array e.g. `filter[worksin]`
3. Verify that all other issues enumerated above are fixed

qa_req 2
cr_req 2

## Notes

:warning:  that currently this branch is using an Algolia app from my personal account, because we're waiting for Algolia to enable the "Facet display" feature on iFixit account.

deploy_block 🚨


frontend/components/product-list/sections/FilterableProductsSection/CurrentRefinements.tsx
```diff
@@ -0,0 +1,74 @@
...
+                                 }}
+                              />
+                           </Tag>
+                        );
+                     })}
+                  </React.Fragment>
+               );
+            })}
+            <Button
+               onClick={() => clearRefinements.refine()}
```
danielbeardsley, line 59:
```suggestion
               onClick={clearRefinements.refine}
```

jordycosta:
QA 🟢 

- [X] Can filter using encoded url
- [X] No results page instead of a page crash
- [X] UX timing improved
- [X] Facet list re-renders for no results found
- [X] No more unnecessary empty space in the facet list
- [X] Facet dividers are now only between facets

---

### Testing Procedure:
1. Navigate to the Vercel [preview](https://react-commerce-git-use-new-algolia-hooks-ifixit.vercel.app/)
2. Ensure that each issue listed above has been resolved here

danielbeardsley (changes requested):
Still going through it, but I'll use regular comments so you don't have to wait till the end.

dev_block :+1:  for the moment on this pull switching the algolia app to one that's.... not us?

frontend/components/product-list/ProductListView.tsx
```diff
@@ -47,78 +60,115 @@ export function ProductListView({
...
+                                    refinementList: routeState.filter,
+                                    range: routeState.range,
+                                 },
+                              };
+                           },
+                        },
+                        router: history({
+                           getLocation() {
+                              if (typeof window === 'undefined') {
+                                 return new URL(url!) as unknown as Location;
```
danielbeardsley, line 96: This doesn't seem to be valid JS syntax that I'm aware of:

image

frontend/components/product-list/sections/FilterableProductsSection/FacetsDrawer.tsx
```diff
@@ -0,0 +1,318 @@
...
+   isVisible: boolean;
+};
+
+function ClearAllButton({ isVisible }: ClearAllButtonProps) {
+   const { refine } = useClearRefinements();
+   if (!isVisible) {
+      return null;
+   }
+   return (
+      <Button w="50%" variant="outline" onClick={() => refine()}>
```
danielbeardsley, line 226:
```suggestion
      <Button w="50%" variant="outline" onClick={refine}>
```

frontend/components/product-list/sections/FilterableProductsSection/Pagination.tsx
```diff
@@ -0,0 +1,142 @@
...
+   });
+   const isMounted = useIsMounted();
+   const visibleNumberOfPages = isMounted ? responsiveVisibleNumberOfPages : 3;
+
+   const createPageUrl = (page: number) => {
+      try {
+         const fullUrl = createURL(page - 1);
+         const url = new URL(fullUrl);
+         return url.pathname + url.search;
+      } catch (error) {
```
danielbeardsley, line 38: Hmm, what kind of errors were we expecting here?

danielbeardsley:
The facets panel on the left seems like it always has scroll bars:
image

frontend/components/product-list/sections/FilterableProductsSection/RangeInput.tsx
```diff
@@ -0,0 +1,215 @@
...
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+   }, []);
+
+   const range = useRange(props);
+   const appliedRangeMin = range.start[0];
+   const appliedRangeMax = range.start[1];
+
+   const refineMinValue = React.useCallback(
+      (value: number | undefined) => {
+         const minAllowedValue = range.range.min ? range.range.min + 1 : 0;
```
danielbeardsley, line 38: I don't get the `min - 1` here...  The result is that if the cheapest product is $4 and you type in `4` in the `min` part of the range it'll be immediately changed to `5`. This feels like a bug.

frontend/components/product-list/sections/FilterableProductsSection/RangeInput.tsx
```diff
@@ -0,0 +1,215 @@
...
+            ]);
+         }
+      },
+      [appliedRangeMax, range]
+   );
+
+   const refineMaxValue = React.useCallback(
+      (value: number | undefined) => {
+         const maxAllowedValue = range.range.max
+            ? range.range.max - 1
```
danielbeardsley, line 62: Pretty sure this makes for buggy behavior:

https://user-images.githubusercontent.com/26855/172952580-3d774a9e-db3a-4fdc-a76f-9da5ca8a151b.mp4

frontend/components/product-list/sections/FilterableProductsSection/RangeInput.tsx
```diff
@@ -0,0 +1,215 @@
...
+      ? value
+      : Math.min(Math.max(value, min ?? -Infinity), max ?? Infinity);
+}
+
+type UseDebouncedValueProps = {
+   refine: (value: number | undefined) => void;
+   currentAlgoliaValue: number | undefined;
+};
+
+function useDebouncedValue({
```
danielbeardsley, line 180: It would be useful to have a function doc here explaining the feature this adds on top of `useDebounce()`

frontend/components/product-list/sections/FilterableProductsSection/RefinementList.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+const RefinementListItem = React.memo(function RefinementListItem(
+   props: RefinementListItemProps
+) {
+   const { onChange, ...item } = props;
+   const [isRefined, setIsRefined] = React.useState(item.isRefined);
+
+   React.useEffect(() => {
+      setIsRefined(item.isRefined);
+   }, [item.isRefined]);
```
danielbeardsley, line 75: I'm having trouble seeing why we need the `isRefined` state if we already have `item.isRefined`?

frontend/components/product-list/sections/FilterableProductsSection/SearchInput.tsx
```diff
@@ -1,76 +1,88 @@
...
 
 const MAX_SEARCH_QUERY_LENGTH = 100;
+const DEBOUNCE_INTERVAL_MILLIS = 300;
 
-export const SearchInput = chakra(({ className }: SearchInputProps) => {
+export const SearchInput = forwardRef<SearchInputProps, 'div'>((props, ref) => {
+   const searchBox = useSearchBox();
    const inputRef = React.useRef<HTMLInputElement>(null);
-   const [query, search] = useSearch();
+   const [searchQuery, setSearchQuery] = React.useState(searchBox.query);
```
danielbeardsley, line 25: Given [the docs](https://www.algolia.com/doc/api-reference/widgets/search-box/react-hooks/#hook), it's hard to tell how they expect us to use the hook.

frontend/components/product-list/sections/FilterableProductsSection/SearchInput.tsx
```diff
@@ -1,76 +1,88 @@
...
+   const debouncedQuery = useDebounce(searchQuery, DEBOUNCE_INTERVAL_MILLIS);
+   const lastDeboucedQueryRef = React.useRef<string | null>(null);
+
+   React.useEffect(() => {
+      if (debouncedQuery !== searchBox.query) {
+         if (lastDeboucedQueryRef.current === debouncedQuery) {
+            setSearchQuery(searchBox.query);
```
danielbeardsley, line 32: I'm struggling to understand this part... where we update the input box to the value from algolia. In other cases the updates go the other way (inputbox -> algolia).

frontend/components/product-list/sections/FilterableProductsSection/useFacets.tsx
```diff
@@ -0,0 +1,9 @@
+import { useDynamicWidgets } from 'react-instantsearch-hooks-web';
+
+export function useFacets() {
+   const { attributesToRender } = useDynamicWidgets({
+      maxValuesPerFacet: 200,
```
danielbeardsley, line 5: This seems unnecessarily high. Probably anything about 30-40 is on the edge of what's useful. Thoughts?

frontend/pages/Parts/[handle].tsx
```diff
@@ -62,38 +58,47 @@ export const getServerSideProps: GetServerSideProps<PageProps> = async (
...
-      algoliaIndexName: ALGOLIA_DEFAULT_INDEX_NAME,
-      urlQuery: context.query,
-   });
-
-   if (searchContext == null) {
-      return {
-         notFound: true,
-      };
-   }
+   const protocol = context.req.headers.referer?.split('://')[0] || 'https';
```
Comment (danielbeardsley) [inline frontend/pages/Parts/[handle].tsx:61]: Why are we using the `referer` here? Why does the protocol of the `referer` affect the way we render the page?

Comment (danielbeardsley) [reply frontend/pages/Parts/[handle].tsx:61]: Also, the `referer` isn't going to be in the cache key in the upstream cache, so our response can't really use it to affect the output.

frontend/components/product-list/ProductListView.tsx
danielbeardsley: Oh, that's typescript aparently. Hmm... doesn't seem necessary given that the type of `url` is already `string` and not `string|null`.

danielbeardsley:
Wow, that's a lot of code!

Good work @dhmacs, I'm glad we'll be using the new algolia hooks instead of sticking with our own, thanks for all this effort! Yes, I made a lot of comments, but there's only a small handful of important ones.

frontend/.env.development
```diff
@@ -1,4 +1,5 @@
-NEXT_PUBLIC_ALGOLIA_APP_ID=XQEP3AD9ZT
+# NEXT_PUBLIC_ALGOLIA_APP_ID=XQEP3AD9ZT
+NEXT_PUBLIC_ALGOLIA_APP_ID=9I2C2MSPT7
```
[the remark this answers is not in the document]
dhmacs: As mentioned in the PR description, there's a deploy block while we wait for the Algolia support ticket to go through

> ## Notes
>
>:warning: currently this branch is using an Algolia app from my personal account, because we're waiting for Algolia to enable the "Facet display" feature on iFixit account.

frontend/components/product-list/sections/FilterableProductsSection/CurrentRefinements.tsx
dhmacs: The reason I write the arrow function is that `onClick` passes `event` as an argument, and if for some reason the `refine` method implementation is changed to receive an argument this could potentially introduce an unexpected bug

frontend/components/product-list/ProductListView.tsx
dhmacs: You're right 👍

frontend/components/product-list/sections/FilterableProductsSection/FacetsDrawer.tsx
dhmacs: I'm doing this on purpose, since `onClick` passes `event` as an argument, and if for some reason the `refine` method implementation is changed to receive an argument this could potentially introduce an unexpected bug

frontend/components/product-list/sections/FilterableProductsSection/Pagination.tsx
dhmacs: It was throwing due to Algolia routing misconfiguration. I have fixed it, so I think it's safe now to drop the try catch 👍

frontend/components/product-list/sections/FilterableProductsSection/RangeInput.tsx
dhmacs: The behaviour of the Algolia `useRange` is tricky and it's optimized to limit refinement within the existing price range, in order to avoid showing empty results. 
In this specific case, if you input a number at the boundaries (or outside the range) it won't do anything, because when the lowest price option is `4` and you put `4` is the same thing as not adding the filter at all.

I think you're right though, my current implementation can feel buggy.  I've invested a little bit more time, and I think I've found a better trade off. I'll push the changes shortly, let me know what you think.

frontend/components/product-list/sections/FilterableProductsSection/RangeInput.tsx
dhmacs: This function is specific of this component, it basically decouples the input from the Algolia state, so that it doesn't feel laggy. I'll add the explanation in the code 👍

frontend/components/product-list/sections/FilterableProductsSection/RefinementList.tsx
dhmacs: I should have made a better job at documenting this. It's a UX optimization. `item.isRefined` is set to `true` after the response from the API has been received, and this can make the checkbox feel laggish. The purpose of this is to decouple the checkbox state from the Algolia context, so that it feels more responsive

frontend/components/product-list/sections/FilterableProductsSection/SearchInput.tsx
dhmacs: Do you have any specific concern here?

frontend/pages/Store/Tools/index.tsx
```diff
@@ -1,15 +1,11 @@
-import { Layout } from '@components/common';
```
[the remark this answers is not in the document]
dhmacs: I think it's the Github UI that's misleading.. If you check on this branch, the file is not there https://github.com/iFixit/react-commerce/tree/use-new-algolia-hooks/frontend/pages/Store

frontend/components/product-list/sections/FilterableProductsSection/SearchInput.tsx
dhmacs: I've refactored the code so hopefully is more readable. Basically it's decoupling input state from Algolia context state so the input feels more responsive.

frontend/components/product-list/sections/FilterableProductsSection/useFacets.tsx
dhmacs: I don't really have opinion about this, it depends on the max number of facets items. I've put this number because I think we were discussing with @sterlinghirsh that some facets might have a high number of items. Let me know what to set this to

frontend/pages/Parts/[handle].tsx
Comment (dhmacs) [reply frontend/pages/Parts/[handle].tsx:61]: To be honest I don't know, it's in the implementation [guide](https://www.algolia.com/doc/guides/building-search-ui/going-further/server-side-rendering/react-hooks/#with-nextjs) for server side rendering with Next.js.
What would you do? Just hard code `https` ?

dhmacs:
un_dev_block 🍏

frontend/.env.development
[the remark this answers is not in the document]
danielbeardsley: > As mentioned in the PR description, there's a deploy block while we wait for the Algolia support ticket to go through

Sorry I didn't see that :-(

Is this something that this pull should be blocked on? If we switch to the old account, does this pull break? Or is the experience jut not optimal?

frontend/components/product-list/sections/FilterableProductsSection/CurrentRefinements.tsx
danielbeardsley: :+1:

danielbeardsley:
deploy_block :+1:  on not merging this till we change the config back to using iFixit's algolia App. Merging as-is (depending on an algolia app that isn't the one we're pushing to / keeping up to date) isn't something we want to deploy.

frontend/components/product-list/sections/FilterableProductsSection/CurrentRefinements.tsx
```diff
@@ -0,0 +1,74 @@
...
+      <Collapse
+         in={currentRefinements.items.length > 0}
+         animateOpacity
+         data-testid="current-refinements"
+      >
+         <Flex pt="3" wrap="wrap" px={{ base: 6, sm: 0 }}>
+            {currentRefinements.items.map((item, index) => {
+               return (
+                  <React.Fragment key={item.label}>
+                     {item.refinements.map((refinement) => {
```
masonmcelvain, line 30: Maybe this is intentional, but I noticed that the `<CurrentRefinements>` [component provided by Algolia](https://instantsearchjs.netlify.app/stories/?path=/story/refinements-currentrefinements--default) lumps refinements from the same facet into a single tag, and here we put all refinements in their own tag

image

image

frontend/pages/Parts/[handle].tsx
Comment (danielbeardsley) [reply frontend/pages/Parts/[handle].tsx:61]: It looks like the `url` prop is used down the stack for the server-side rendering of the algolia history router (or something?). It appears as if we're just using it to get a "current url". I don't know precisely what that ends up being used for in algolia, but maybe it doesn't affect anything. I'd vote for yes, just choosing `https` always.

Given that we're doing the same bit of code for every page, I think we should perhaps move that to someplace common.

frontend/components/product-list/sections/FilterableProductsSection/SearchInput.tsx
```diff
@@ -86,3 +73,44 @@ export const SearchInput = forwardRef<SearchInputProps, 'div'>((props, ref) => {
       </InputGroup>
    );
 });
+
+type SearchBoxRenderState = ReturnType<typeof useSearchBox>;
+
+/**
+ * Maintains a local search query while keeping it in sync with the Algolia search box query.
```
danielbeardsley, line 80: Thanks for the comments, this whole thing is much more clear!

frontend/components/product-list/sections/FilterableProductsSection/SearchInput.tsx
danielbeardsley: Sorry, it was a lame comment. Your extracted code and the explanations are perfect. Good job solving this!

frontend/pages/Parts/[handle].tsx
Comment (dhmacs) [reply frontend/pages/Parts/[handle].tsx:61]: Algolia uses it because it needs to know the url server side in order to get the facets. The thing I don't know is if it makes any difference to have `http` or `https`, but I doubt it. It's weird though that their guide contains such a specific piece of code, it seems like it has been written that way for some reason.

> Given that we're doing the same bit of code for every page, I think we should perhaps move that to someplace common.

That's right 👍  I already have a plan for this, but I didn't want to complicate this PR any further. The gist is that we should have a `templates` folder that exports the various kind of pages that we have (for now, it's just the product list page). This way we can reuse the code across different routes that all render the same template.

frontend/components/product-list/sections/FilterableProductsSection/CurrentRefinements.tsx
dhmacs: I'm following our mockup, so it's expected to be that way. Thanks for pointing this out though, the Algolia approach it's interesting and we might want to iterate on this later to improve it 👍

dhmacs:
> deploy_block on not merging this till we change the config back to using iFixit's algolia App. Merging as-is (depending on an algolia app that isn't the one we're pushing to / keeping up to date) isn't something we want to deploy.

The feature we're using in this PR is called "Facet display" and it's used to decide which facets to show, so the current code wouldn't work without it, although it's possible to implement it differently if it turns out that we'll be stuck for some time.

@davidrans told me that it would probably take a couple of business days to enable the feature, so we can probably merge this as is, and redact the Algolia credentials when the feature is enabled.

danielbeardsley:
>  we can probably merge this as is, and redact the Algolia credentials when the feature is enabled

I do want to merge this code and not have it block other things. I think keeping it open is probably a worse state than us using an index that is not our own (I'm assuming you somehow *copied* our existing index or something?)

frontend/components/product-list/sections/FilterableProductsSection/RangeInput.tsx
danielbeardsley: The behavior is still a bit buggy and I feel like it's just trying to be too smart. I'd voter for just accepting the numbers they type in. Having numbers auto-change to something you didn't intend is surprising and a little annoying (and it accidentally hides the highest price item). Further searches may return different results and the auto-clamped values may then be irrelevant.

https://user-images.githubusercontent.com/26855/173158397-ccd20825-2b7c-48bb-8688-d6635fc43c8e.mp4

danielbeardsley: I think let's merge this and fix it in a follow-up.

dhmacs:
@danielbeardsley yes, I made a copy of the index (although right now is not a full copy, just a slice of 1000 records). I can replicate this on any account pretty easily, so if you want to open a quick account with iFixit credentials and use that, it's not a big deal for me to replicate the index there. Let me know 👍

frontend/components/product-list/sections/FilterableProductsSection/RangeInput.tsx
```diff
@@ -0,0 +1,219 @@
...
+   React.useEffect(() => {
+      refresh();
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+   }, []);
+
+   const range = useRange(props);
+   const appliedRangeMin = range.start[0];
+   const appliedRangeMax = range.start[1];
+
+   const refineMinValue = React.useCallback(
```
masonmcelvain, line 36: `refineMinValue` and `refineMaxValue` don't appear consistent with each other. I'm confused about a couple things:

- Why do we use `clamp(value, 0, maxAllowedValue)` here, but in refineMaxValue we do `clamp(value, undefined, maxAllowedValue)`? I would expect `clamp(value, minAllowedValue, Infinity)` there.
- Here, we use `clampedValue < appliedRangeMax` in isValidRange, but in refineMaxValue we use `clampedValue > minAllowedValue && clampedValue > appliedRangeMin` in isValidRange. I would expect only `clampedValue > minAllowedValue` there.
- the call to `range.refine(...)` is different here than in refineMaxValue

If any of these require changes, I'm also happy to have it happen in a followup pull.

masonmcelvain (approved):
CR ✌🏻 and bumping to cr req 2

erinemay:
Found some UI issues: 
@sterlinghirsh @dhmacs, a lot of these could/should be new issues. Since this PR has a pretty big scope, I'll defer to you to decide which should be their own issue and which should be fixed here.

### 1. Worksin filters not very usable at this time
   1. Text overflows to the right (including filter item count)
   2. `Show More` shows all worksin options (There are quite a few)
   3. Worksin facet text isn't very user friendly
   4. Selecting a worksin checkbox lower in the list refreshes the page and moves that facet to the top of the worksin list. (Try going to the [ipad page](https://react-commerce-git-use-new-algolia-hooks-ifixit.vercel.app/Parts/iPad) and selecting a worksin near the bottom of the expanded worksin list). Behavior is inconsistent with other filters.
<details><summary>screenshot of worksin</summary>

image

</details>

### 2. Inputting an unacceptable price range re-sets the filters
Steps:
     1. Select some filters (Select Item Type = `Batteries`)
     2. Enter a price range that doesn't return any results (25-26)
     3. Observe: The "No results found" page briefly flashes and then the filters are all reset

### 3. Item Type and Category are now even more similar
<details><summary>Category comparison</summary>

| Item Type | Category |
| ------------- | ------------- |
| image | image |

</details>

### 4. Now seeing multiple results for the same item
<details><summary>It appears that we're getting different results (one for each active variant) for the same item with only the price differentiating most of the cards</summary>

image

</details>

Is this expected? Will we display all results for one item in the product list here? Part only, kit, OLED version, OEM version, etc.
(Another example https://react-commerce-git-use-new-algolia-hooks-ifixit.vercel.app/Parts/iPad?filter%5Bfacet_tags.Item%20Type%5D%5B0%5D=Screens&filter%5Bdevice%5D%5B0%5D=iPad%208)


### 5. Search behavior is buggy
Behavior:
- After typing in a term, it briefly flashes loading without that term and then back to that term
- When removing any search terms, a term will show back up
- Sometimes it works and typing in a term doesn't result in flashing term and page reload

What actually happens each time appears inconsistent, but the bugginess is pretty consistent. (Confirmed present in Chrome and FF)

<details><summary>screen recording of search behavior</summary>

https://user-images.githubusercontent.com/72166715/173420774-1f499e52-293e-4c77-8a2f-7fed8a2e0375.mp4

</details>

### 6. Still seeing facet dividers after the last facet
Noted in https://github.com/iFixit/react-commerce/issues/239
Found at https://react-commerce-git-use-new-algolia-hooks-ifixit.cominor.com/Tools?filter%5Bcategory%5D%5B0%5D=Drivers%20and%20Wrenches&filter%5Bprice_range%5D%5B0%5D=%240%20-%20%2410
<details><summary>screenshot</summary>

image

</details>

dhmacs:
> Since this PR has a pretty big scope, I'll defer to you to decide which should be their own issue and which should be fixed here.

I think I can fix them today / tomorrow 👍 

> ### 1. Worksin filters not very usable at this time

I don't think this facet is relevant, it's there just for testing purposes. I'll hide it (this new implementation displays the facet that are set for display in the Algolia UI, so this is something we can change from the Algolia dashboard at any time)

> ### 2. Inputting an unacceptable price range re-sets the filters

Yeah the Algolia behaviour around the price filter is tricky. I didn't notice this one, I'll play with it a little bit more 👍 

> ### 3. Item Type and Category are now even more similar

Don't pay attention to the facet that we're showing. As mentioned above, they'll be selected from the Algolia dashboard, so they can be changed at any time.

> ### 4. Now seeing multiple results for the same item

This is expected. @sterlinghirsh is working on a new Algolia index without duplicates, so we'll be using that index and thus we won't have duplicates.

> ### 5. Search behavior is buggy

Mm this one is weird. I'll fix it

> ### 6. Still seeing facet dividers after the last facet

You're right, I need to account for when a facet is not shown because it doesn't have applicable options. I'll fix it 👍 


Thanks for the great feedback @erinemay !

sterlinghirsh:
We don't need to launch with the price range filter, so don't spend too much time on it. Also don't worry about hiding the worksin filter since we'll probably just make that a filter-only facet at index time.

frontend/components/product-list/sections/FilterableProductsSection/RangeInput.tsx
danielbeardsley: I think our code should not try to be "smart" about it. Let the user type whatever they want, no auto-fixing. If they type something nonsensical like "10 to 5" (instead of `5 to 10`) we can just ignore the price filter or show no results, both seem like reasonable behavior.

sterlinghirsh:
In fact let's disable the price range filter for now and come back to it if we have time before launch.

masonmcelvain:
@dhmacs I hope this isn't an inconvenience, but I am going to push two commits and merge this as is to unblock some work @danielbeardsley and I are doing. @sterlinghirsh is on board too. For the changes that you planned to make, please open a new PR.

masonmcelvain:
I think CI is failing because we are using the ALGOLIA_API_KEY environment variable for the personal app in the vercel preview, as opposed to the iFixit Algolia app. If I'm incorrect about this, we can make a followup pull to amend this mistake, but this configuration works for local dev.

PR #331, opened 2022-06-13
Search Facets: Hide irrelevant ones (algolia hooks version)

### LINKED ISSUE #299: Hide facets when corresponding Wiki Info is set

masonmcelvain:
Hide Facets that are already specified by the Device Parts page we are on.

Context: Assume we are on the iPhone 7 parts list page
The "Device Brand" facet and its "iPhone" option is useless and
confusing cause all products here are already compatible with iPhones.

Here we hide any facets that come from the wiki info of the current
device and its ancestors.

## QA
When you visit a refined page of parts or tools (like /Parts/Dell_Laptop) make sure that any facets that exist in the device's wiki info are excluded from the page, both on mobile and on desktop views.

Closes #299

### DIFF
--- a/frontend/components/product-list/ProductListView.tsx
+++ b/frontend/components/product-list/ProductListView.tsx
@@ -106,7 +106,9 @@ export function ProductListView({
                      {productList.children.length > 0 && (
                         <ProductListChildrenSection productList={productList} />
                      )}
-                     <FilterableProductsSection />
+                     <FilterableProductsSection
+                        wikiInfo={productList.wikiInfo}
+                     />
                      {productList.sections.map((section, index) => {
                         switch (section.type) {
                            case ProductListSectionType.Banner: {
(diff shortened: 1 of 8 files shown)


masonmcelvain:
blocking QA on the preview problem https://github.com/iFixit/react-commerce/pull/326#issuecomment-1154559687

danielbeardsley:
CR :+1: 

Noice! Thanks for updating this pull!

jordycosta:
QA 🟢 

Any facet of a products Wiki info is excluded from the facet list on preview ✔️ 

---

### Testing Procedure:
1. Navigate to a product Wiki API
> i.e. https://www.ifixit.com/api/2.0/wikis/CATEGORY/iPhone

2. Expand `🔽 info:`
3. Note the names of each element in the array
4. Ensure that none of these are shown on the facet list in the Vercel preview
> i.e. https://react-commerce-git-search-facets-hide-irrelevant-89a343-ifixit.vercel.app/Parts/iPhone

---

### Screen Capture(s):
wiki

**Before/After:**

beforeafter

PR #328, opened 2022-06-10
Collections: Add OEM partnership flare

masonmcelvain:
Adds a green OEM Partnership flare tag to product list items if the field exists in Algolia. Also updates the flare text for some badges.

## QA
Find products in product_en in [Algolia](https://www.algolia.com/apps/XQEP3AD9ZT/) that have the oem_partnership field. Also test when it is null.

Visit the vercel preview and view those products. Make sure the tag is present, and meets the conditions specified in the issue. The flare should come before 'Lifetime Warranty'.

(You can type 'teenage' into the search bar, and results come up with this flare)

Also make sure that these flares have the right text:
- 'Lifetime warranty' is now 'Lifetime Warranty'
- 'PRO' is now 'iFixit Pro'

CC @sterlinghirsh 

Closes #316
Closes #317

PR #332, opened 2022-06-14
Pin Node.js docker image to Alpine 3.14

dhmacs:
Apparently the last update to the Docker image of Node.js Alpine broke our dockerfile. I'd need to spend more time to understand exactly what's causing the issue with Alpine `3.16`, but my best guess is that it is related to the removal of `python2` (see [release notes](https://www.alpinelinux.org/posts/Alpine-3.16.0-released.html)).

This PR pin the node image to Alpine `3.14` which we were using before.

## QA

Verify that you can access the Govinor deploy [preview](https://debug-broken-docker-image.govinor.com/admin) for this branch.

## Notes

This PR should unblock #328

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -1,7 +1,7 @@
-FROM node:16-alpine
+FROM node:16-alpine3.14
 # Installing libvips for sharp compatibility
 RUN apk --update-cache add vips-dev \
- && rm -rf /var/cache/apk/*
+  && rm -rf /var/cache/apk/*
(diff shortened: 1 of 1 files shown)


k0rvusk0r4x:
I'm able to access the preview without seeing breaks or warnings. I can go through the various tabs, and push all the colorful buttons without being yelled at.

QA 😺

masonmcelvain (approved):
CR ✌🏻 Thanks!

PR #277, opened 2022-05-05
Filters: Sort capacity filter options

masonmcelvain:
This sorts hard drive capacities from largest to smallest in the filter
options dropdown list. Before, they were being sorted alphabetically.

Triplet programmed with @danielbeardsley and @sterlinghirsh 

## QA
Visit the preview link.
Click the capacity dropdown and confirm that capacity sizes are sorted from largest to smallest.

Closes https://github.com/iFixit/ifixit/issues/42631

PR #338, opened 2022-06-15
Facets: Sort `Capacity` from high to low

masonmcelvain:
This reintroduces Capacity facet sorting. It was accidentally erased
when we switched over to using Algolia hooks.

This is an adaptation of the original implementation in #277.

Closes https://github.com/iFixit/ifixit/issues/42631

### DIFF
--- a/frontend/components/product-list/sections/FilterableProductsSection/RefinementList.tsx
+++ b/frontend/components/product-list/sections/FilterableProductsSection/RefinementList.tsx
@@ -20,8 +20,7 @@ export function RefinementList(props: RefinementListProps) {
    const { items, refine, isShowingMore, toggleShowMore, canToggleShowMore } =
       useRefinementList({
          ...props,
-         sortBy:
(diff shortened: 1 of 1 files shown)


dhmacs (approved):
CR 👍 

To clarify, this was removed on purpose since "Facet display" in Algolia lets you pick the order in which you want to display facet option values.
Facet display is now enabled for the iFixit account, so we might want to use that instead of implementing custom code for sorting facet option values.

masonmcelvain:
deploy_block ✌🏻 on considering @dhmacs's suggestion

@sterlinghirsh @danielbeardsley

k0rvusk0r4x:
I'm going to Put the external block label on here so we don't start committing testing time unless we know we want this change rather than the above suggestion. Please remove the label if we decide we do want the change.

k0rvusk0r4x:
The "Capacity" facet is now sorted by actual capacity in descending order rather than alphabetically.

QA 😺

danielbeardsley:
un_deploy_block :+1:

PR #341, opened 2022-06-16
Cache: Bump it up

danielbeardsley:
Cloudfront doesn't respect stale-while-revalidate, so at 10s, nearly all
of our page requests will be misses which are quite expensive.

Let's bump this because product data in algolia changes relatively
infrequently.


dhmacs (approved):
TIL 
We're in 2022 and Cloudfront still does not support SWR header 😬 

CR 👍

k0rvusk0r4x:
- Went through the product list pages in the preview and messed with various facets.
- Not really more to check here under the circumstances.

QA 😺

PR #343, opened 2022-06-17
Strapi: Cache some rarely-changing results

danielbeardsley:
### Problem
We were making these three strapi requests on every next request.

### Do This
The data changes infrequently, so let's cache it for a bit and eliminate 3 requests for most page loads.

### DIFF
--- a/frontend/lib/cache.ts
+++ b/frontend/lib/cache.ts
@@ -1,0 +1,14 @@
+import LRU from 'lru-cache';
+const cacheStore = new LRU({ max: 100 });
+
+export function cache<T>(key: string, fetch: () => T, ttl: number): T {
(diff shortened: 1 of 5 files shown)


andyg0808 (approved):

andyg0808:
CR :goat: through 2ce8cc4

k0rvusk0r4x:
In the iFixit/ifixit repo, I would think to add a log so I could see each time the request got sent off. How would we do the equivalent here? Or is there a better way?

danielbeardsley:
> How would we do the equivalent here? Or is there a better way?

There already is a log:
image

First page load, followed by subsequent page load:
image

QA :+1:

PR #348, opened 2022-06-20
Webpack: Enable optimization on server-side bundles

andyg0808:
NextJS doesn't tree-shake server-side bundles by default:
https://github.com/vercel/next.js/issues/8956#issuecomment-557867732

We suspect that the large amount of code being loaded by our Lambdas is resulting in slow startup times.

This enables tree-shaking and other optimizations on the server-side code, in order to reduce bundle size. We're hoping this helps with startup times.

<details>
<summary>Before</summary>

image

</details>
<details>
<summary>After</summary>

image

</details>

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -82,18 +82,21 @@ const moduleExports = {
    }),
 };
 
-// Make sure adding Sentry options is the last code to run before exporting, to
-// ensure that your source maps include changes from all other Webpack plugins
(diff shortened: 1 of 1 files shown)


danielbeardsley:
CR :+1:

andyg0808:
QA :+1: 
See pictures above.

danielbeardsley (approved):
CR :+1:

PR #349, opened 2022-06-20
Webpack: install our own copy

andyg0808:
Several subdependencies have a peer dep on webpack. Next actually has webpack built *into* the package (in its `dist/` directory). It seems like serverless builds (at least on Amplify) are broken because one of these subdependencies (`@next/react-refresh-utils`) can't find webpack. It seems like the pull that added sentry caused `webpack` to not be hoisted to top level and thus not to be found by `@next/react-refresh-utils`.

This ensures all our dependencies use the same copy when they've got it as a peer dependency.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -100,6 +100,7 @@
       "jest-watch-typeahead": "0.6.1",
       "next-transpile-modules": "9.0.0",
       "prettier": "2.2.1",
-      "typescript": "4.3.2"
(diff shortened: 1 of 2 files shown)


andyg0808:
QA :+1: 
The Vercel preview doesn't break.

danielbeardsley:
CR :+1:

danielbeardsley (approved):
CR :+1:

PR #350, opened 2022-06-21
Images: add IfixitImage component

danielbeardsley:
We want fine grained control over when and how we use next/image and its image resizing pipeline. Given that each of our images are rendered in many sizes, we want to be able to choose the sizes to use and when.

This is a minimal starting point. This gives us one place where we can tune when to use which sizes and when to use next's image pipeline. For the time being, all images are served directly from their sources instead of being re-compressed on the fly.

CC @sterlinghirsh @andyg0808

### DIFF
--- a/frontend/components/ifixit-image/ifixit-image.tsx
+++ b/frontend/components/ifixit-image/ifixit-image.tsx
@@ -1,17 +1,19 @@
 import Image, { ImageProps, ImageLoader } from 'next/image';
 
 export function IfixitImage(props: ImageProps) {
+   let unoptimized = props.unoptimized;
+
    if (typeof props.src === 'string') {
       if (isGuideImage(props.src)) {
-         props.loader = passThroughLoader;
+         unoptimized = true;
       } else if (isCartImage(props.src)) {
-         props.loader = passThroughLoader;
+         unoptimized = true;
       } else if (isStrapiImage(props.src)) {
-         props.loader = passThroughLoader;
+         unoptimized = true;
       }
    }
 
-   return <Image {...props} />;
+   return <Image {...props} unoptimized={unoptimized} />;
 }
 
 function isGuideImage(src: string) {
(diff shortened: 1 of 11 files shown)


andyg0808 (approved):

andyg0808:
CR :goat: through 296bedf

frontend/components/ifixit-image/ifixit-image.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+   if (typeof props.src === 'string') {
+      if (isGuideImage(props.src)) {
+         unoptimized = true;
+      } else if (isCartImage(props.src)) {
+         unoptimized = true;
+      } else if (isStrapiImage(props.src)) {
+         unoptimized = true;
+      }
+   }
```
andyg0808, line 14: Might consider pulling the conditional below out into its own function, which allows setting `unoptimized` once:
```suggestion
   const unoptimized = props.unoptimized || !shouldOptimize(props.src);
```

Example `shouldOptimize`:
```ts
function shouldOptimize(src: any) {
   if (typeof src === 'string') {
      if (isGuideImage(src)) {
         return false;
      } else if (isCartImage(src)) {
         return false;
      } else if (isStrapiImage(src)) {
         return false;
      }
   }
   return true;
}
```

andyg0808 (approved):

andyg0808:
CR :goat: through 68505ab

frontend/components/ifixit-image/ifixit-image.tsx
danielbeardsley: In upcoming pulls we'll want to change both the loader *and* `unoptimized`. When that comes, we'll probably user a pattern like `const [unoptimized, loader] = getLoader(props)`

danielbeardsley:
Seems to be QA :+1: 

cdn and cominor.com images are used directly, while static images go through the pipeline.

image

andyg0808 (approved):

andyg0808:
CR :goat: through c8003cd

PR #353, opened 2022-06-21
Drop pages from current app

andyg0808:
This prevents `next/link` from trying to load them client-side, which can accidentally override the current app pages.


sterlinghirsh (approved):

djmetzle:
Err, that's a little unfortunate... the Vercel preview link is a 404 now, because we're dropping the index.
Maybe we can (separately?) update the preview link to point at one of the collection pages?

djmetzle (approved):
CR :gear: 

But see my comment, the bad preview link is a (minor) regression.

andyg0808:
QA :+1: 

The page does a real page load now.
image

PR #357, opened 2022-06-24
Make tools pages paths case insensitive

### LINKED ISSUE #327: Make strapi respond tool pages case-insensitively
Like /Tools/gripping vs /Tools/Gripping

The canonical should be what's set in Strapi (ideally /Tools/Gripping) and we should redirect to the canonical. We'll have to be sure to preserve any url params.

dhmacs:
closes #327 

## QA

1. Visit [Vercel preview](https://react-commerce-git-make-tools-pages-paths-case-in-01b5b9-ifixit.cominor.com/Tools)
2. Verify that tools categories paths are case insensitive (e.g.  `/Tools/gripping` vs `/Tools/Gripping`), that is the page is always redirected to the lowercase version

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -87,11 +87,11 @@ function withUniversalOptimization(plugins) {
    return {
       ...plugins,
       webpack(config, info) {
-         config = configureWebpack(config, info)
+         config = configureWebpack(config, info);
          config.optimization.minimize = true;
          return config;
-      }
-   }
+      },
+   };
 }
 
 // Make sure adding Sentry options is the last code to run before exporting, to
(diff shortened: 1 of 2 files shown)


sterlinghirsh:
I wrote up a short document examining the different approaches. https://docs.google.com/document/d/15bJZoTv3Sh7-ezn8cd6P-WzsnL5HQIk59o8n3NO3-Hc/edit Let's leave it as-is for now, and I'll discuss it with some people and come back to this next week.

sterlinghirsh:
I think we're going to leave it as-is for now. We may end up trying another approach later, but I don't want to get locked into lower-case urls.

PR #358, opened 2022-06-27
Make tagline an h2

### LINKED ISSUE #311: Make tagline an h2, and remove h2 from sub categories

sumukhipandey:
## Summary

Previously, the tagline was a `<Text></Text>` and the sub-categories, ie the child titles, were `<h2></h2>`. This PR makes the taglines `h2` and the child title [sub categories] `spans`

## CR/QA
WAP

Closes #311

### DIFF
--- a/frontend/components/product-list/sections/HeroSection.tsx
+++ b/frontend/components/product-list/sections/HeroSection.tsx
@@ -34,9 +34,9 @@ export function HeroSection({ productList }: HeroSectionProps) {
             {page > 1 ? ` - Page ${page}` : ''}
          </HeroTitle>
          {productList.tagline && productList.tagline.length > 0 && page === 1 && (
-            <Text fontWeight="bold" fontSize="xl" px={{ base: 6, sm: 0 }}>
+            <h2>
                {productList.tagline}
-            </Text>
+            </h2>
          )}
          {hasDescription && (
             <HeroDescription>{productList.description}</HeroDescription>


sumukhipandey:
dev_block 🔧

sumukhipandey:
un_dev_block 👍

frontend/components/product-list/sections/HeroSection.tsx
```diff
@@ -34,9 +34,9 @@ export function HeroSection({ productList }: HeroSectionProps) {
             {page > 1 ? ` - Page ${page}` : ''}
          </HeroTitle>
          {productList.tagline && productList.tagline.length > 0 && page === 1 && (
-            <Text fontWeight="bold" fontSize="xl" px={{ base: 6, sm: 0 }}>
+            <h2>
                {productList.tagline}
-            </Text>
+            </h2>
```
dhmacs, line 39: This changes the look & feel of the tagline

| Before  | With this PR |
| ------------- | ------------- |
| Screenshot 2022-06-28 at 07 43 37  | Screenshot 2022-06-28 at 07 44 03  |

dhmacs (changes requested):

frontend/components/product-list/sections/HeroSection.tsx
dhmacs: Hint: You can use `as` prop also for the Text component, to change its semantics

sumukhipandey: Resolved here cc056d290ff4ad93fba17296101b2bce33bc60f9

jordycosta:
QA 🟢 

Taglines are now `h2` instead of **bolded text** ☑️ 
child.titles are now `span` instead of `h2` ☑️ 

### Testing Procedure:
1. Navigate to the Vercel VR Preview
2. Check that taglines are now `h2` elements
3. Also check that the child.titles are now `span` elements

---

### Screen Captures:

h2
span

aburke07 (dismissed):
CR :alien:

dhmacs:
> Taglines are now `h2` instead of **bolded text**

@jordycosta I don't think it's expected that the tagline is not bold text. The change is only about semantic, so the design should be preserved

deploy_block 🚫

sumukhipandey:
> > Taglines are now `h2` instead of **bolded text**
> 
> @jordycosta I don't think it's expected that the tagline is not bold text. The change is only about semantic, so the design should be preserved
> 
> deploy_block 🚫

Thanks for pointing that out! Will restore the design changes. 
@dhmacs In doing this, will the text tag contain the `fontSize` info from before? ie will the change simple be 
`<Text as=h2 fontWeight="bold" fontSize="xl" px={{ base: 6, sm: 0 }}>` ? I was confused as to why we still need the fontSize info along with the h2?

dhmacs:
> In doing this, will the text tag contain the fontSize info from before? ie will the change simply be
<Text as=h2 fontWeight="bold" fontSize="xl" px={{ base: 6, sm: 0 }}> ?

@sumukhipandey That's right 👍 The tagline should look exactly the same. This change is only about semantics (for SEO purposes), it's not a design change.

sumukhipandey:
> > In doing this, will the text tag contain the fontSize info from before? ie will the change simply be
> > <Text as=h2 fontWeight="bold" fontSize="xl" px={{ base: 6, sm: 0 }}> ?
> 
> @sumukhipandey That's right 👍 The tagline should look exactly the same. This change is only about semantics (for SEO purposes), it's not a design change.

Thanks! Resolved 37561571c4f369bfd756f93059d1f38229cd5b2b
un_deploy_block 🦋

dhmacs (approved):
CR 💯

jordycosta:
QA 🟢 

h2

Design unchanged 👍

PR #382, opened 2022-07-04
Unique Username: Add a null check 

### LINKED ISSUE #372: Make sure the page doesn't break when unique_username is null

sumukhipandey:
## Summary:

Previously, we did not check if the `unique_username` was null and the page could break. This PR fixes this so that the null case for `unique_username` is handled.

## CR/QA:
Users created with https://www.cominor.com/dev/create_user don't have unique usernames by default. 
Visit this branch's [preview](https://react-commerce-git-unique-username-check-for-null-ifixit.cominor.com/Parts) while logged in as a user without a unique username and confirm log in/out and appearance is the same for a user with a unique username.

Closes #372 

CC: @sterlinghirsh @dhmacs

### DIFF
--- a/packages/auth-sdk/user.tsx
+++ b/packages/auth-sdk/user.tsx
@@ -5,7 +5,7 @@ import { useQuery } from 'react-query';
 type User = {
    id: number;
    username: string;
-   handle: string;
+   handle: string | null;
    thumbnail: string | null;
    is_pro: boolean;
    discountTier: string | null;
@@ -38,10 +38,6 @@ async function fetchAuthenticatedUser(apiOrigin: string): Promise<User | null> {
       invariant(
          typeof payload.username === 'string',
          'User username is not a string'
-      );
-      invariant(
-         typeof payload.unique_username === 'string',
-         'User handle is not a string'
       );
       let thumbnailUrl: string | null = null;
       if (
--- a/packages/auth-sdk/user.tsx
+++ b/packages/auth-sdk/user.tsx
@@ -43,6 +43,10 @@ async function fetchAuthenticatedUser(apiOrigin: string): Promise<User | null> {
          typeof payload.unique_username === 'string',
          'User handle is not a string'
       );
+      invariant(
+         payload.unique_username !== null,
+         'User handle is null'
+      );
       let thumbnailUrl: string | null = null;
       if (
          isRecord(payload.image) &&
--- a/packages/auth-sdk/user.tsx
+++ b/packages/auth-sdk/user.tsx
@@ -43,10 +43,6 @@ async function fetchAuthenticatedUser(apiOrigin: string): Promise<User | null> {
          typeof payload.unique_username === 'string',
          'User handle is not a string'
       );
-      invariant(
-         payload.unique_username !== null,
-         'User handle is null'
-      );
       let thumbnailUrl: string | null = null;
       if (
          isRecord(payload.image) &&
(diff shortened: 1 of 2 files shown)


packages/auth-sdk/user.tsx
```diff
@@ -43,6 +43,10 @@ async function fetchAuthenticatedUser(apiOrigin: string): Promise<User | null> {
          typeof payload.unique_username === 'string',
          'User handle is not a string'
       );
+      invariant(
+         payload.unique_username !== null,
+         'User handle is null'
+      );
```
dhmacs, line 49: The issue described by #372 is that if the user `unique_username` is not defined, then this hook does not return the user because of [this invariant](https://github.com/iFixit/react-commerce/blob/main/packages/auth-sdk/user.tsx#L42-L45).

In order for us to be able to support possibly null/undefined `unique_username` we need to get rid of the invariant, type the `unique_username` as nullable, and refactor the UI components that access `unique_username` unsafely, so that they take into account that the value might be null.

dhmacs (changes requested):

packages/auth-sdk/user.tsx
sumukhipandey: Thanks! I wasn't sure about this and it makes sense now.

sumukhipandey:
dev_block 🦋 on https://github.com/iFixit/react-commerce/pull/382#discussion_r912715031

packages/auth-sdk/user.tsx
sumukhipandey: Resolved via 19f4cc0085a3d47b323aacd4fd4cebfe183ac2d9, 16e184466f83edd6b9c09db6017b65258d2f95fd and dc101d2f422dc4106594c34ec45fffb03a2ea8e0

danielbeardsley (approved):
CR :+1: 

Thanks @sumukhipandey !

Kinda annoying that the code causing the breakage here was these `invariant()` calls. I'm really tempted to just not include them or at least make them *report* (to sentry) but not throw errors. Right now, a slight adjustment to our API could brick this app because of run-time type enforcement in the `invariant()` calls. Many times (i.e. null vs undefined) the rest of the code would likely handle it fine.

Thoughts @dhmacs @sterlinghirsh @masonmcelvain ?

masonmcelvain:
> make them report (to sentry) but not throw errors

In prod, this makes sense to me, a broken component seems better than a broken app. +1

danielbeardsley:
> we can break the api without (potentially) breaking the frontend

I think you mean "change" the API. I don't think changing a null to undefined or a blank string would count as breaking.

> I think there are pros and cons about dropping invariants:

Sorry, I don't mean that we should *drop* them, just that we should change them to *report* errors to sentry instead of throwing, but still provide the type protection they offer. I don't think it's all or nothing. i.e. we can describe our API with invariants or typescript types.

sumukhipandey:
un_dev_block 🦋

dhmacs (dismissed):
CR 👍

jordycosta:
QA 🟢 

unique not_unique

 Log in/out and appearance is the same for a user with a unique username

danielbeardsley:
un_dev_block :+1: 

Made the change.

CR :+1:

danielbeardsley:
QA :+1: 

Excludes the handle item in the menu when it's missing:

image
image

danielbeardsley (approved):
un_dev_block :+1: 

Made the change.

CR :+1:

PR #393, opened 2022-07-06
iFixit Origin: use relative urls if on same origin

danielbeardsley:
Effectively, if we are running the app on the same base domain as the
urls of the API, then we should use relative urls in for all front-end
API requests.

Backend requests still need to the full url because there is no "origin"
as they aren't in the context of a browser.

Determine if we should use relatively urls by comapring the base domain
of IFIXIT_ORIGIN and the base domain of the current url. We are using
base domain here so we allow using relative urls when on language
subdomains like es.ifixit.com (where IFIXIT_ORIGIN === www.ifixit.com).

Closes https://github.com/iFixit/ifixit/issues/43646

### DIFF
--- a/frontend/config/env.ts
+++ b/frontend/config/env.ts
@@ -8,10 +8,14 @@ export const ALGOLIA_API_KEY = checkEnv(
    'ALGOLIA_API_KEY'
 );
 
-export const IFIXIT_ORIGIN = checkEnv(
+const NEXT_PUBLIC_IFIXIT_ORIGIN = checkEnv(
    process.env.NEXT_PUBLIC_IFIXIT_ORIGIN,
    'NEXT_PUBLIC_IFIXIT_ORIGIN'
 );
+
+export const IFIXIT_ORIGIN = shouldUseRelativeOrigin(NEXT_PUBLIC_IFIXIT_ORIGIN) ?
+   "" :
+   NEXT_PUBLIC_IFIXIT_ORIGIN;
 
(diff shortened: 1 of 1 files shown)


masonmcelvain (approved):
CR ✌🏻 Straightforward thanks for taking this on!

frontend/config/env.ts
```diff
@@ -30,3 +34,18 @@ function checkEnv(env: string | null | undefined, envName: string): string {
...
 }
+
+function shouldUseRelativeOrigin(ifixitOrigin: string): boolean {
+   if (typeof window === 'undefined') {
+      return false;
+   }
+
+   function baseDomain(url: string): string|null {
+      const parsedURl = new URL(url);
+      const host = parsedURl.host.match(/[^.]+\.[^.]+$/);
```
sterlinghirsh, line 45:
```suggestion
      const parsedUrl = new URL(url);
      const host = parsedUrl.host.match(/[^.]+\.[^.]+$/);
```

sterlinghirsh:
CR :zap:

masonmcelvain:
CR ✌🏻

danielbeardsley:
QA :+1: 

I used `/etc/hosts` to access my local instance at `local.cominor.com` and it parsed the base domain just fine. Since these were running on separate ports, they weren't considered equal.

`current:cominor.com:3000 origin:cominor.com`

I then changed the code to ignore the port (by using `location.hostname`) and it started making relative requests to `/api/2.0/...`.

PR #394, opened 2022-07-06
Tools: Hide irrelevant facets

### LINKED ISSUE #375: Hide irrelevant facets on tool pages
These don't make much sense on tool pages since they're mostly for parts:

* Device
* Device Brand
* Device Category
* Device Type
* Item Type
* OS

This is almost every facet except for price range. Maybe that's okay if there aren't other facets we want to show for tools

sumukhipandey:
## Summary

Previously, we included facets, such as `Device Brand`, that were not needed on the `/Tools` page. This PR adds changes such that these facets are hidden on `/Tools`:
- Device
- Device Brand
- Device Category
- Device Type
- Item Type
- OS

## CR/QA:
- Visit the preview and add `/Tools`.
- On the tools homepage make sure that the above facets are hidden.
- Additionally, I also checked `/Parts` to make sure that page wasn't broken.

Closes #375

### DIFF
--- a/frontend/components/product-list/sections/FilterableProductsSection/useFacets.tsx
+++ b/frontend/components/product-list/sections/FilterableProductsSection/useFacets.tsx
@@ -36,8 +36,8 @@ export function useFilteredFacets(wikiInfo: WikiInfoEntry[], title?: string) {
       return usefulFacets;
    }, [facets, infoNames]);
 
-   if (title !== null && title === "Tools") {
-      return facets.slice(7,10);
+   if (title !== null && title === 'Tools') {
+      return facets.slice(7, 10);
    }
 
    return usefulFacets;
--- a/frontend/components/product-list/sections/FilterableProductsSection/useFacets.tsx
+++ b/frontend/components/product-list/sections/FilterableProductsSection/useFacets.tsx
@@ -22,7 +22,7 @@ export function useFacets() {
    // return attributesToRender;
 }
 
-export function useFilteredFacets(wikiInfo: WikiInfoEntry[]) {
+export function useFilteredFacets(wikiInfo: WikiInfoEntry[], title?: string) {
    const facets = useFacets();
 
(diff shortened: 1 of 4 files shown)


sumukhipandey:
dev_block 🦋

sumukhipandey:
un_dev_block 🦋 
However I want to address a couple thing:
- Sorry about putting everything in one commit. I think it would be easy if it was split into at least 2 commits:
   - Add `title?:` as an optional function parameter for `useFilteredFacets`
   - Add a `title` for the different types that call `useFilteredFacets` and where it makes sense.
- I wasn't sure if there was a more efficient way to do this. I tried to see where `facet_tags.Main Category: "Tools"`  was actually being used but I could not figure it out.

This solution seems to work but I would appreciate any suggestions for a better solution. Thanks!

frontend/components/product-list/sections/FilterableProductsSection/useFacets.tsx
```diff
@@ -36,5 +36,9 @@ export function useFilteredFacets(wikiInfo: WikiInfoEntry[]) {
       return usefulFacets;
    }, [facets, infoNames]);
 
+   if (title !== null && title === 'Tools') {
+      return facets.slice(7, 10);
```
sterlinghirsh, line 40: This won't work if we change the order of the facets. We'll want to specifically omit the facets listed on all tool pages. I think this is another place where we can use the new page type field.

sterlinghirsh:
CR :zap: dev_block :zap: Honestly you're probably best off starting with a new branch, since the main branch has changed a lot and there are so many conflicts.

sumukhipandey:
Close https://github.com/iFixit/react-commerce/pull/394#issuecomment-1175868168 with new branch [`tools--hide-facets`](https://github.com/iFixit/react-commerce/tree/tools--hide-facets)

PR #399, opened 2022-07-06
Facets: Hide refinement items whose count equals hits

### LINKED ISSUE #388: Regression: Hide facet options that filter no options
We want to hide facet options when the count is equal to the number of results. This used to work before we switched to Algolia's hooks.

image

masonmcelvain:
If a refinement item (aka a facet option) is applicable to the same
number of products as the number of products in the current view, then
that refinement item is essentially a no-op and we should hide it.

One exception: Refinement items that are currently selected should still
be displayed.

## QA
Click through some different product list views (tools and parts, mobile and desktop). Make sure that no facet options are ever displayed if they apply to the same number of products as are displayed in the current view, and no other facet options in that facet are selected. (This differs slightly from the original issue; see https://github.com/iFixit/react-commerce/pull/399#issuecomment-1179234914 and the conversation leading up to it).

CC @sterlinghirsh 

Closes #388


sterlinghirsh:
CR :zap: dev_block :zap: This doesn't seem to work right. I think there are two problems. First, all the items seem to be removed.

image

Second, when a facet becomes empty (i.e. no valid options), we should completely hide it. Make sure the logic to remove options happens before the logic for whether to render facet or the "Show More" button.

masonmcelvain:
un_dev_block ✌🏻 Thanks @sterlinghirsh, I think those should be addressed now

frontend/components/product-list/sections/FilterableProductsSection/useFilteredRefinementList.tsx
```diff
@@ -0,0 +1,20 @@
...
+   const { items, ...rest } = useRefinementList({ ...props });
+   const { results } = useHits();
+   const hitsCount = results?.nbHits ?? 0;
+
+   const filteredItems = items.reduce((acc: typeof items, item) => {
+      if (item.isRefined || hitsCount > item.count) {
+         acc.push(item);
+      }
+      return acc;
+   }, []);
```
sterlinghirsh, line 17: I think this should work equivalently.
```suggestion
   const filteredItems = items.filter((item) => item.isRefined || hitsCount > item.count);
```

danielbeardsley:
> For example, maybe the user wants to see all SSD's that are 1 GB and 1.5 GB.
> Say there are 2 1GB products (A and B) and 2 different 1.5 GB products (C and
> D). When the user selects 1GB, the facet option for 1.5GB will disappear
> because it also has a count of 2. The user wanted to see all 4 products (ABCD),
> but can only see A & B, which seems like a bug.

That's a really good point.

> Maybe this hiding behavior should only apply when there are no refinements selected?

Yes, I think that's appropriate. Though it will be weird when you click a facet and then *more* facets show up...   Maybe there's a count of *results without this filter applied* or something? i.e. something that still returns *4* in your example?

danielbeardsley:
I don't think such a count exists in my reading of the docs and the types returned from the hooks.

The example given in the issue seems like it should be dealt with by the "hide facets that match wiki info" logic...  Perhaps that has broken?

danielbeardsley:
Getting another invariant failure locally and it just bricks the app when one record has a device title that's null:
image

I'll make a reminder to make an issue to make invariants soft-fail.

masonmcelvain:
> Though it will be weird when you click a facet and then more facets show up

Yes I agree, this is strange UX.

> The example given in the issue seems like it should be dealt with by the "hide facets that match wiki info" logic... Perhaps that has broken?

Our wiki info filtering only filters out _entire facets_, not specific refinement list items (like Phone or iPhone).

https://github.com/iFixit/react-commerce/blob/01d849894cfb7476cb677203b154a9a8fc8b7797/frontend/components/product-list/sections/FilterableProductsSection/useFacets.tsx#L43-L51

"device" doesn't show up in the wiki info either, fwiw.

[Wiki info json](https://www.ifixit.com/api/2.0/wikis/CATEGORY/iPhone) for iphone:

```json
  "info": [
    {
      "name": "Device Brand",
      "value": "Apple",
      "inheritedFrom": null
    },
    {
      "name": "Device Type",
      "value": "iPhone",
      "inheritedFrom": null
    },
    {
      "name": "OS",
      "value": "iOS",
      "inheritedFrom": null
    },
    {
      "name": "Device Category",
      "value": "Phone",
      "inheritedFrom": "Phone"
    }
  ]
```

You might be on to something though. Maybe we can somehow exclude refinement list items if they show up in the wiki info.

dhmacs:
> Getting another invariant failure locally and it just bricks the app when one record has a device title that's null: image
> 
> I'll make a reminder to make an issue to make invariants soft-fail.

@danielbeardsley I can't find `iFixit Replacement Kits` on this Strapi branch, nor in the Strapi seed. Is this a product list that you've added? The problem here is that it's being wrongly classified as `device-parts` product list. This might happen for two reason:

1. Wrong type is set on Strapi
2. No `type` is set on Strapi, so it defaults to `device-parts`. I chose that a null `type` is considered `device-parts` because device parts are the vast majority of product lists and they weren't all set to `device-parts`. Now that I reason more about this, I think that a better behavior would be to check if the product list has a `device title` set, and only then default to `device-parts` type, otherwise it'll default to `marketing` type.

danielbeardsley:
From a high level, adding a record to strapi or changing a record to be inconsistent shouldn't *brick* the app. At **worst**, our app should pretend like the bad record doesn't exist.

masonmcelvain:
Proposed solution: Hide all facet options whose refinement counts match the total number of hits, but only if no refinements are selected within the facet.

masonmcelvain:
un_dev_block ✌🏻

rickisXP:
QA :penguin: but deploy_block :thought_balloon: on a clarification
- Item type refinement no longer populates parts aside from the selected device :+1:  
- I'm able to see facets whose refinement returns the same number of items (see img below). Is this supposed to be expected? :thought_balloon: 
<details><summary>Screenshots</summary>

image

image

</details>

sterlinghirsh:
CR :zap: QA :zap:

sterlinghirsh (approved):

PR #404, opened 2022-07-07
Facets: Custom sort order

### LINKED ISSUE #390: Sort and hide Facets
[`sort-and-hide-facets`](https://github.com/iFixit/react-commerce/compare/sort-and-hide-facets)

Algolia's Facet Display doesn't work for nested facets like facet_tags. We'll have to sort them in code.

On Parts pages:
* Item Type
  * Alphabetize
* Part or Kit
* Remaining facets sorted alphabetically
  * Values sorted by number of hits unless otherwise specified (i.e. Capacity)

On Tools pages:
* Tool Type
* Price Range

masonmcelvain:
Algolia's Facet Display doesn't work for nested facets like facet_tags. This adds code to customize the sort order.

It also adds code to alphabetize the "Item Type" refinement list items (aka facet options).

## QA
Make sure the facet accordions are sorted as spec'd on desktop and mobile. Check it across parts pages, tools pages, and other pages (ie marketing).

Also make sure the item type facet options are alphabetized.

CC @sterlinghirsh 

Closes #390

### DIFF
--- a/frontend/components/product-list/sections/FilterableProductsSection/useFacets.tsx
+++ b/frontend/components/product-list/sections/FilterableProductsSection/useFacets.tsx
@@ -8,14 +8,15 @@ export function useFacets() {
    });
 
    return [
-      'facet_tags.Item Type',
       'facet_tags.Capacity',
-      'device',
       'facet_tags.Device Brand',
       'facet_tags.Device Category',
       'facet_tags.Device Type',
+      'facet_tags.Item Type',
       'facet_tags.OS',
       'facet_tags.Part or Kit',
+      'facet_tags.Tool Category',
+      'device',
       'price_range',
       'worksin',
    ];
--- a/frontend/components/product-list/sections/FilterableProductsSection/useFacets.tsx
+++ b/frontend/components/product-list/sections/FilterableProductsSection/useFacets.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { useDynamicWidgets } from 'react-instantsearch-hooks-web';
 import { ProductList, ProductListType } from '@models/product-list';
+import { formatFacetName } from '@helpers/algolia-helpers';
 
 export function useFacets() {
    const { attributesToRender } = useDynamicWidgets({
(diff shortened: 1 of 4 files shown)


frontend/components/product-list/sections/FilterableProductsSection/useFacets.tsx
```diff
@@ -33,15 +35,65 @@ export function useFilteredFacets(productList: ProductList) {
...
+      if (ranking.get(a) && ranking.get(b)) {
+         return ranking.get(a)! - ranking.get(b)!;
+      } else if (ranking.get(a)) {
+         return -1;
+      } else if (ranking.get(b)) {
+         return 1;
+      } else {
+         return sortFacetsAlphabetically(a, b);
+      }
+   };
```
sterlinghirsh, line 93: This is a little simpler and allows negative rankings (where a facet could be pinned to the bottom). Plus if two facets have the same rank, they'll be sorted alphabetically. I didn't check this, so the aRank-bRank might be backwards.

```suggestion
   return (aName: string, bName: string): number => {
      const aRank = ranking.get(aName) || 0;
      const bBrank = ranking.get(bName) || 0;
      
      return aRank - bRank || sortFacetsAlphabetically(aName, bName);
   };
```

sterlinghirsh:
CR :zap: dev_block :zap: on a minor change. I think this looks good otherwise, though!

erinemay:
> Part or Kit

From https://github.com/iFixit/react-commerce/issues/390#issue-1294899243... @sterlinghirsh, is this ⬆️  specifying that the only two options for the `part or kit` dropdown are meant to be `Part` or `Kit`? I remember talking about it, but I don't know if it was intended here.

sterlinghirsh:
> > Part or Kit
> 
> From [#390 (comment)](https://github.com/iFixit/react-commerce/issues/390#issue-1294899243)... @sterlinghirsh, is this ⬆️ specifying that the only two options for the `part or kit` dropdown are meant to be `Part` or `Kit`? I remember talking about it, but I don't know if it was intended here.

The only ones we should see in the final version are "Part Only" or "Fix Kit". There are a few others that are currently mistakenly included in the index, but that's not a bug on the client.

frontend/components/product-list/sections/FilterableProductsSection/useFacets.tsx
danielbeardsley: Also, requires changing the mapping values above because higher number = closer to top.

frontend/components/product-list/sections/FilterableProductsSection/useFacets.tsx
```diff
@@ -34,15 +35,65 @@ export function useFilteredFacets(productList: ProductList) {
...
+   ['facet_tags.Item Type', 1],
+   ['facet_tags.Part or Kit', 2],
+]);
+
+const toolsFacetRanking = new Map([
+   ['facet_tags.Tool Category', 1],
+   ['price_range', 2],
+]);
+
+function sortFacets(productListType: ProductListType) {
```
danielbeardsley, line 68: I might name this `getFacetCompartor` instead as it currently doesn't sort.

danielbeardsley:
CR :+1:  But I agree with Sterling's suggestion.

danielbeardsley (approved):
un_dev_block :+1:

CR :+1: 

Made the suggested change.

danielbeardsley:
QA :+1: 

Seems to sort as expected in the mapping:

Tools:
image

Parts:
image

masonmcelvain:
Thanks for finishing this up! CR ✌🏻

PR #406, opened 2022-07-07
No Results Search: Add link to main search

### LINKED ISSUE #401: No-results search page: add link to main search
## Problem
When a user searches on a product list and gets no results, it's not always clear that their search is restricted to the current product list.

## Do This
Make the no-results component say something like `No matching products found in [page title]` instead of just "no results". Add a link that says "Search all of iFixit for [search term]". That should take the user to `https://www.ifixit.com/Search?query=[search term]`.

For example, `No matching products found in iPhone 7 Parts` `Search all of iFixit for "potato head"` => `https://www.ifixit.com/Search?query=potato%20head`

The link can be relative in production, but probably needs to be absolute in for it to work from vercel previews.

sterlinghirsh:
When a user searches but does't find any results, this changes the no-results page to clarify that they're only searching within their given page. There's now a link to the main iFixit search with their query pre-filled, too.

I also added a couple changes in the project setup to fix some issues I encountered while setting this up on windows.

Closes #401

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,6 @@
 {
    "printWidth": 80,
    "tabWidth": 3,
-   "singleQuote": true
+   "singleQuote": true,
+   "endOfLine": "auto"
 }
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ This command will start Strapi dev server and Next.js dev server:
 pnpm dev
 ```
 
-> :warning: Don't forget to create `.env.local` file with the required secrets (see `.env.local.example`)
+> :warning: Don't forget to create `backend/.env` and `frontend/.env.local` files with the required secrets (see `.env.example` and `.env.local.example`)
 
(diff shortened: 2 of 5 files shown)


dhmacs (approved):
I like how it turned out 👍 

CR ⚡

erinemay:
deploy_block 🦖 on a thought... 
Do we want the ifixit search to be `?doctype=product`

erinemay:
> deploy_block 🦖 on a thought... Do we want the ifixit search to be `?doctype=product`

Products are below the fold on some important mobile devices when the doctype isn't product:

<details><summary>iPhone 13</summary>

image

</details>

<details><summary>Samsung Galaxy S22</summary>

image

</details>

sterlinghirsh:
un_deploy_block :zap: I left it as a global search on purpose. The idea is that if someone searches for something that's not a product, we want to give them a quick way to do a global search.

PR #409, opened 2022-07-07
Tests: Parts page search test

batbattur:
## Summary 
Adds test for the Search on `/Parts` page. From https://github.com/iFixit/react-commerce/issues/124:

> Search
    - Search for a common word like "watch"
      - [ ] Assert that the remaining products all conform to the search specification above
    - Search for a nonexistent word like "asdfasdfasdf"
      - [ ] Assert that the "No results found" component is displayed

## QA Notes
Please re-run the E2E tests / cypress-run ci check and make sure it passes.

Connects: https://github.com/iFixit/react-commerce/issues/124 and https://github.com/iFixit/ifixit/issues/43590

### DIFF
--- a/frontend/test/cypress/integration/product-list/parts_page_search.spec.ts
+++ b/frontend/test/cypress/integration/product-list/parts_page_search.spec.ts
@@ -20,8 +20,8 @@ describe('parts page search', () => {
       user
          .findByTestId('list-view-products')
          .children('article')
-         .each((productCart) => {
-            user.wrap(productCart).contains('iphone', { matchCase: false });
+         .each((productListItem) => {
+            user.wrap(productListItem).contains('iphone', { matchCase: false });
          });
    });
 


frontend/test/cypress/integration/product-list/parts_page_search.spec.ts
```diff
@@ -0,0 +1,39 @@
+describe('parts page search', () => {
+   const user = cy;
+
+   beforeEach(() => {
+      user.intercept('/1/indexes/**').as('search');
+      user.visit('/Parts');
+   });
+
+   it('should search for products', () => {
```
mlahargou, line 9: I prefer to split tests like this into two because they are two different user flows. Thoughts?

```
should show results when the search term exists
should show no results when search term doesn't exist
```

mlahargou (approved):
CR 📱

frontend/test/cypress/integration/product-list/parts_page_search.spec.ts
batbattur: Yes, I agree!

batbattur:
dev_block 👍 on incoming changes

batbattur:
un_dev_block 🚛

frontend/test/cypress/integration/product-list/parts_page_search.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+         .findByTestId('collections-search-box')
+         .should('be.visible')
+         .type('asdasasdadasd');
+
+      user.wait('@search');
+
+      user.findByTestId('list-view-products').should('not.exist');
+      user.contains('No results found');
+      user.contains(
+         "Try adjusting your search or filter to find what you're looking for"
```
mlahargou, line 37: [minor] I wonder if we can pull this text into a constant. That way if we ever wanted to update the text, we wouldn't also have to update this test. Not a blocker.

mlahargou (approved):
CR 📱

batbattur:
dev_block 👍 looking into test failure

batbattur:
un_dev_block 👍

Just found that the search result is fuzzy search so when searching `camp`, it returns all results with camp and any products with part `cam`: https://fr.ifixit.com/Parts?q=camp. Therefore, modified the search text from `watch` to `iphone` to guarantee that all products in search results will have iphone text (we have enough iphone products/search results).

ardelato (approved):
CR 🦖

k0rvusk0r4x:
`E2E tests / cypress-run` run successfully.

QA 😺

PR #422, opened 2022-07-09
Tests: Add tests for products on /Parts

batbattur:
## Summary
Adds test for the Results view on /Parts page. From https://github.com/iFixit/react-commerce/issues/124:

> Results view
    - Body
        - [ ] Assert these are truncated to the desired length
    - Price
         - [ ] Ensure we display the price data the component is provided
    - View button
         - [ ] Ensure the button directs the user to the link that the component was provided 

## QA Notes
Please re-run the E2E tests / cypress-run ci check and make sure it passes.

Connects: https://github.com/iFixit/react-commerce/issues/124 and https://github.com/iFixit/ifixit/issues/43590

### DIFF
--- a/frontend/test/cypress/integration/product-list/parts_results_view.spec.ts
+++ b/frontend/test/cypress/integration/product-list/parts_results_view.spec.ts
@@ -18,26 +18,6 @@ describe('parts page results view', () => {
    });
 
    /**
-    * In ProductListItem, we set noOfLines prop to 3
-    * to truncate the product description text. This
-    * tests each product description to be less than 4
-    * lines.
-    */
-   it('product description is truncated', () => {
-      // Counting text lines https://stackoverflow.com/questions/783899
-      user.get('[data-testid="product-short-desc"]').each((descriptionDiv) => {
-         const divHeight = descriptionDiv.outerHeight() ?? 0;
-         const lineHeigth = parseInt(descriptionDiv.css('lineHeight'));
-         const noOfLines = divHeight / lineHeigth;
-
-         expect(noOfLines).to.be.below(
-            4,
-            'Product description text is more than 3 lines!'
-         );
-      });
-   });
-
-   /**
     * TODO: Unskip when the new product page is implemented
     *
     * We can link to the old product page by setting
--- a/frontend/test/cypress/integration/product-list/parts_results_view.spec.ts
+++ b/frontend/test/cypress/integration/product-list/parts_results_view.spec.ts
@@ -5,7 +5,7 @@ describe('parts page results view', () => {
       user.visit('/Parts');
    });
 
-   it('all products have price visible', () => {
+   it('all products have a visible price', () => {
       user
          .findByTestId('list-view-products')
          .children('article')
--- a/frontend/test/cypress/integration/product-list/parts_results_view.spec.ts
+++ b/frontend/test/cypress/integration/product-list/parts_results_view.spec.ts
@@ -1,0 +1,63 @@
+describe('parts page results view', () => {
+   const user = cy;
+
+   beforeEach(() => {
+      user.visit('/Parts');
+   });
+
+   it('all products have price visible', () => {
+      user
+         .findByTestId('list-view-products')
+         .children('article')
+         .each((productListItem) => {
+            user
+               .wrap(productListItem)
+               .findByTestId('product-price')
+               .should('be.visible');
+         });
+   });
+
+   /**
+    * In ProductListItem, we set noOfLines prop to 3
+    * to truncate the product description text. This
+    * tests each product description to be less than 4
+    * lines.
+    */
+   it('product description is truncated', () => {
+      // Counting text lines https://stackoverflow.com/questions/783899
+      user.get('[data-testid="product-short-desc"]').each((descriptionDiv) => {
+         const divHeight = descriptionDiv.outerHeight() ?? 0;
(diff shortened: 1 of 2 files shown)


frontend/test/cypress/integration/product-list/parts_results_view.spec.ts
```diff
@@ -0,0 +1,63 @@
+describe('parts page results view', () => {
+   const user = cy;
+
+   beforeEach(() => {
+      user.visit('/Parts');
+   });
+
+   it('all products have price visible', () => {
```
ardelato, line 8:
```suggestion
   it('all products have a visible price, () => {
```

frontend/test/cypress/integration/product-list/parts_results_view.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+
+   /**
+    * In ProductListItem, we set noOfLines prop to 3
+    * to truncate the product description text. This
+    * tests each product description to be less than 4
+    * lines.
+    */
+   it('product description is truncated', () => {
+      // Counting text lines https://stackoverflow.com/questions/783899
+      user.get('[data-testid="product-short-desc"]').each((descriptionDiv) => {
```
ardelato, line 28: How come we are using `get` instead of `findByTestId` in this line?

frontend/test/cypress/integration/product-list/parts_results_view.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+         const divHeight = descriptionDiv.outerHeight() ?? 0;
+         const lineHeigth = parseInt(descriptionDiv.css('lineHeight'));
+         const noOfLines = divHeight / lineHeigth;
```
ardelato, line 31: Are we retrieving the estimated number of lines by dividing the overall div size with the line height size?

ardelato (commented):
dev_block ☁️ I think the way we are counting the number of lines needs to be revisited

frontend/test/cypress/integration/product-list/parts_results_view.spec.ts
batbattur: It turns out `findByTestId` fails if there are multiple elements with `data-testid="..."` on the page. `findByTestId` seems to expect to find only one element. Therefore, I had to use `get` which works with getting multiple elements
fail

dhmacs: @batbattur I suggest not to rely too much on test ids.. I use test ids as the last resort generally.. I try to test based on accessible elements. This way tests also validates accessibility. 
While writing tests I've tried to follow [these principles](https://testing-library.com/docs/guiding-principles), especially point 2. The closest way to select an input like a user would do is to search using accessible names

batbattur: > It turns out findByTestId fails if there are multiple elements with data-testid="..." on the page. findByTestId seems to expect to find only one element. Therefore, I had to use get which works with getting multiple elements

I was wrong here. Just found that there is `findAllByTestId` in [cypress-testing-library.](https://github.com/testing-library/cypress-testing-library/blob/main/types/index.d.ts#L155)

<hr>

> @batbattur I suggest not to rely too much on test ids.. I use test ids as the last resort generally.. I try to test based on accessible elements. This way tests also validates accessibility.

@dhmacs Sounds good, and thank you for the advice! From this [article](https://betterprogramming.pub/why-you-should-avoid-testing-react-components-with-test-ids-ee50d20d37d2), I found that `getByRole`, `getByLabelText`, or `getByText` are recommended instead of using data-testid. 

In our case, if I want to get the product short description text element, how would you recommend? Maybe should I add `aria-label="product_short_desc"` instead of adding `data-testid="product_short_desc"` to the text?
short_description

dhmacs: @batbattur I'd start by analyzing what the test is trying to do. From what I understand here we're trying to test if the product description is truncated. Imo I wouldn't test this behaviour at all, as is not really critical for the functioning of the e-commerce.
Having said that, `aria-label` is something that is pronounced by the screen reader, so "product_short_desc" will not sound nice. In this case you can add a screen only span with "product description" content and and `id` and reference this `id` in the `aria-labelledby` attribute of the description paragraph. This way the description will have an accessible name. But I think using a test id is also a valid option here, as we probably don't need to give accessible names to every little thing.

batbattur: Dropped the test in https://github.com/iFixit/react-commerce/pull/422/commits/bc0835f7b30fcd28c34ed48fde336ce1e2936f01.

frontend/test/cypress/integration/product-list/parts_results_view.spec.ts
batbattur: We dropped this test case in https://github.com/iFixit/react-commerce/pull/422/commits/bc0835f7b30fcd28c34ed48fde336ce1e2936f01. I added a snapshot test for the whole component in https://github.com/iFixit/react-commerce/pull/438.

ardelato (approved):
CR 👍

jordycosta:
♻️ Re-running E2E tests / cypress-run

jordycosta:
QA 🟢 

e2e

PR #431, opened 2022-07-11
Filters: Forward when selecting item type

### LINKED ISSUE #420: Forward filters when selecting Item Type
## Problem
It's confusing when you select an item type and the page changes and loses its filters.

## Do This
Make it so that when selecting an item type, the other filters are maintained when switching pages. When on the item type page, make sure to show the item type filter along with the other filters. Removing the item type filter should switch you back to the previous page but maintain the filters.

masonmcelvain:
In brief, prevents page reloads when selecting an `Item Type` facet, and keeps the currently selected filters when an `Item Type` facet is selected/deselected.

We now manually create and parse the client's url based on the current Algolia search state. I followed these [docs](https://www.algolia.com/doc/guides/building-search-ui/going-further/routing-urls/js/#seo-friendly-urls), which walk you through the process. This lets us dynamically change the url of the `Item Type` facet links to include the currently selected filters. It also gives us the power to put the item type in the slug of the url, instead of adding it to the query parameters like a normal filter.

Because `Item Type` can now change without a full page reload, we have to update all the places that depended on `Item Type` being constant. This adds a new hook `useDevicePartsItemType()` that gives you the `Item Type` based on the current search state. 

## QA
Do a full smoke test to make sure nothing is broken compared to `main`. The new feature introduced in this pull is

> Prevents page reloads when selecting an Item Type facet, and keeps the currently selected filters when an Item Type facet is selected/deselected.

CC @sterlinghirsh 

Closes #420


masonmcelvain:
dev_block ✌🏻 there's still some navigation bugs I need to work out

frontend/pages/Parts/[deviceHandle]/[itemTypeHandle].tsx
```diff
@@ -48,16 +48,11 @@ export const getServerSideProps: GetServerSideProps<AppPageProps> = async (
          getGlobalSettings(),
```
Comment (masonmcelvain) [inline frontend/pages/Parts/[deviceHandle]/[itemTypeHandle].tsx:48]: [...deviceHandle] there is a different syntax for matching multiple path elements. we could collapse this file and frontend/pages/Parts/[deviceHandle]/index.tsx into the same file.

frontend/pages/Parts/[deviceHandle]/[itemTypeHandle].tsx
```diff
@@ -48,16 +48,11 @@ export const getServerSideProps: GetServerSideProps<AppPageProps> = async (
...
          getStoreList(),
          getStoreByCode('us'),
-         findProductList(
-            {
-               deviceTitle: {
-                  eq: deviceTitle,
-               },
+         findProductList({
```
Comment (masonmcelvain) [inline frontend/pages/Parts/[deviceHandle]/[itemTypeHandle].tsx:51]: Is the Algolia search state loaded by the time this code runs? If so, we could use the hook to get the item type, and pass it to findProductList()

frontend/components/product-list/sections/FilterableProductsSection/FacetFilter.tsx
```diff
@@ -1,61 +1,63 @@
...
    attribute: string;
    productList: ProductList;
 };
 
 export function FacetFilter({ attribute, productList }: FacetFilterProps) {
+   const { indexUiState } = useInstantSearch();
+   const queryString = React.useMemo(
+      () => uiStateToQueryString(indexUiState, ['facet_tags.Item Type']),
+      [indexUiState]
+   );
```
masonmcelvain, line 24: We could move this code to RefinementMenu, because it's only needed there

frontend/components/product-list/sections/FilterableProductsSection/FacetFilter.tsx
```diff
@@ -1,61 +1,63 @@
...
       (itemType: string) => {
-         return getProductListPath({
-            ...productList,
-            type: ProductListType.DeviceItemTypeParts,
-            itemType,
-         });
+         const path = getProductListPath(
+            {
+               ...productList,
+               type: ProductListType.DeviceParts,
```
masonmcelvain, line 31: We don't need to redefine the `type` field anymore, it's not changing anything. You can just pass productlist

frontend/components/product-list/sections/FilterableProductsSection/FacetFilter.tsx
```diff
@@ -1,61 +1,63 @@
...
+            />
+         );
+      case RefinementDisplayType.MultiSelect:
          return (
             <RefinementList
                attribute={attribute}
                showMore
                showMoreLimit={200}
             />
          );
```
masonmcelvain, line 61: default should throw

masonmcelvain:
un_dev_block ✌🏻 but qa block on investigating why the preview links are 500ing

masonmcelvain:
Found the [line](https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch-hooks-server/src/getServerState.tsx#L199) that throws, and the PR that introduced it https://github.com/algolia/react-instantsearch/pull/3515/files. Got released in version 6.28.0. We upgraded from 6.26.0 -> 6.30.1 in this PR, so maybe the preview didn't install the update for some reason?

image

masonmcelvain:
Oh also, I've been seeing this console warning from InstantSearch throughout most of this pull. It happens as soon as you select/deselect a facet, but it only ever warns once per page load.

image

Solutions tried:
- Not rendering this: https://github.com/iFixit/react-commerce/blob/0c0b82d7debc87af311a961dc79c2b263942e86f/frontend/components/product-list/sections/FeaturedProductListSection.tsx#L128-L134
- Adding an invisible `useRange` widget

dhmacs (commented):
I'm still trying to understand why this is throwing errors on Vercel, and not locally.. Meanwhile while reading the code I've left a few comments

frontend/components/product-list/ProductListBreadcrumb.tsx
```diff
@@ -28,10 +29,11 @@ export function ProductListBreadcrumb({
    const reverseAncestorList = React.useMemo(() => {
       return [...ancestors].reverse();
    }, [ancestors]);
+   const itemType = useDevicePartsItemType(productList);
 
    let currentItemTitle = productList.title;
-   if (productList.type === ProductListType.DeviceItemTypeParts) {
-      currentItemTitle = productList.itemType;
+   if (productList.type === ProductListType.DeviceParts && itemType) {
+      currentItemTitle = itemType;
```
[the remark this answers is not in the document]
masonmcelvain: Good catch! Fixed in https://github.com/iFixit/react-commerce/pull/431/commits/7e167c3336802db1478f6df77f8811280cae4ae2

frontend/components/product-list/sections/ProductListChildrenSection.tsx
```diff
@@ -134,6 +165,7 @@ const ChildLink = ({ child }: ChildLinkProps) => {
    return (
       <NextLink href={child.path} passHref>
          <chakra.a
+            href={child.path}
```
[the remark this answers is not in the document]
masonmcelvain: 👍🏻 https://github.com/iFixit/react-commerce/pull/431/commits/584c8d48766c1aac7ffc8f14e36743eb892f97b8

frontend/pages/Parts/[...deviceHandleItemType].tsx
```diff
@@ -30,8 +30,15 @@ export const getServerSideProps: GetServerSideProps<AppPageProps> = async (
...
-   const { deviceHandle } = context.params || {};
+   const { deviceHandleItemType } = context.params || {};
+   const [deviceHandle, itemTypeHandle, ...rest] = Array.isArray(
+      deviceHandleItemType
+   )
+      ? deviceHandleItemType
+      : [];
+
    invariant(typeof deviceHandle === 'string', 'device handle is required');
+   invariant(rest?.length === 0, 'invalid extra path components');
```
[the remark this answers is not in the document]
Comment (masonmcelvain) [reply frontend/pages/Parts/[...deviceHandleItemType].tsx]: Thanks for noticing this behavior! Fixed in https://github.com/iFixit/react-commerce/pull/431/commits/a14c37f2218625bfbb3a2a7365d50db06fbbac11

frontend/components/common/InstantSearchProvider.tsx
```diff
@@ -72,60 +82,123 @@ export function InstantSearchProvider({
...
+               path += `/${partsOrTools}`;
+               if (deviceHandle) {
+                  path += `/${deviceHandle}`;
+                  ignoreFilterKeys.push('facet_tags.Item Type');
+                  const raw: string | string[] | undefined =
+                     routeState.filter?.['facet_tags.Item Type'];
+                  console.log('raw', raw);
+                  const itemType = Array.isArray(raw) ? raw[0] : raw;
+                  if (itemType?.length) {
+                     const decodedItemType = decodeDeviceItemType(itemType);
```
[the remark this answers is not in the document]
masonmcelvain: https://github.com/iFixit/react-commerce/pull/431/commits/de0a32e3f9fe9cb21c0f96e125df83c37f27ae83

masonmcelvain:
Hmm I'm still seeing a problem where the current refinements aren't updated when the user navigates back to the device page from the item type page via the breadcrumbs:


https://user-images.githubusercontent.com/52104630/179866479-48cacde5-b175-47a1-9349-d441de2526bc.mp4

masonmcelvain:
https://github.com/iFixit/react-commerce/pull/431/commits/dafbbd7908139d50278e2d416825a763770a45a4 fixes the bug mentioned here https://github.com/iFixit/react-commerce/pull/431#issuecomment-1189655208

masonmcelvain:
The preview works!  🎉  Ready for review

erinemay:
Confirmed:
- Selecting an item type doesn't wipe out other selected filters
- Item type now shows up as a filter in the list of selected filters that can be cleared without losing other filters
- Item type is still in the url path as opposed to params (like /Parts/MacBook_Pro/Bluetooth_Board)
- Facet lists retain open/close state when entering or leaving an item type page
- /Tools remains unaffected

dev_block 🛑 
<details><summary>1. On mobile, if you select an item type field, you are no longer directed to that page automatically, and it's unclear that the type has been applied. You have to manually exit the filter fields.</summary>

First one is main. Second one is this branch.

https://user-images.githubusercontent.com/72166715/180330241-3f0c1bb0-1655-474c-8e25-cb0983814867.mp4

</details>


<details><summary>2. When entering an item type page, it's possible to get get into a state where you have facet boxes selected that are showing no results.</summary>

image

Repro:
1. Select several checkboxes in a category that will have at least one present when entering an item type page (or go [here](https://react-commerce-prod-orsaewj41-ifixit.vercel.app/Parts/Electronics?filter%5Bprice_range%5D%5B0%5D=%2475%20-%20%24100&filter%5Bprice_range%5D%5B1%5D=%24100%20-%20%24250&filter%5Bprice_range%5D%5B2%5D=%2450%20-%20%2475&filter%5Bprice_range%5D%5B3%5D=%2425%20-%20%2450) - I selected several of the price ranges.)
2. Enter an item type category with few results (use Accessories item type category for example above)
3. If you uncheck any of the 0 options, they disappear

</details>

(double checked these against the latest deploy [f0898f1](https://github.com/iFixit/react-commerce/pull/431/commits/f0898f1f1c3389c23d28bad431b80d7dda9350e4))

masonmcelvain:
> On mobile, if you select an item type field, you are no longer directed to that page automatically, and it's unclear that the type has been applied. You have to manually exit the filter fields.

Good catch! Maybe a better UX would be to navigate back to the facet overview when an Item Type is clicked, and maybe display that the refinement is selected just like any other refinement.

> When entering an item type page, it's possible to get get into a state where you have facet boxes selected that are showing no results.

Also a good catch! This seems like a less significant UX bug, maybe this is acceptable for now and we can address is later @sterlinghirsh?

erinemay:
> > When entering an item type page, it's possible to get get into a state where you have facet boxes selected that are showing no results.
> 
> Also a good catch! This seems like a less significant UX bug, maybe this is acceptable for now and we can address is later @sterlinghirsh?

I agree. This doesn't seem like a significant blocker.

erinemay:
> > On mobile, if you select an item type field, you are no longer directed to that page automatically, and it's unclear that the type has been applied. You have to manually exit the filter fields.
> 
> Good catch! Maybe a better UX would be to navigate back to the facet overview when an Item Type is clicked, and maybe display that the refinement is selected just like any other refinement.

That sounds like a reasonable UX, but I'm not sure if it matches with the current behavior/UX. Right now (on this branch), when you return to the facet overview after selecting an item type, the item type facet is hidden (as expected). I'm not sure of the implications of adding it back in with an indicator that it has a selected refinement. Will it be clickable, and if you enter it, how do you indicate which item type is selected without the checkboxes? I don't really have a better solution, though. Other than the main behavior of exiting refinements altogether.

masonmcelvain:
Yea, there is more nuance here than I thought.

> Other than the main behavior of exiting refinements altogether.

I think this is a simple solution that we can implement in this pull while we decide on the best UX, which we can do in a follow up issue if need be 👍🏻

frontend/helpers/product-list-helpers.ts
```diff
@@ -58,12 +55,11 @@ export function encodeDeviceTitle(deviceTitle: string): string {
...
    'type' | 'handle' | 'deviceTitle'
-> & {
-   itemType?: string;
-};
+>;
 
 export function getProductListPath(
-   productList: ProductListPathAttributes
+   productList: ProductListPathAttributes,
+   itemType?: string
```
masonmcelvain, line 62: If we don't actually pass itemType at any callsites, maybe we can drop this param

masonmcelvain: https://github.com/iFixit/react-commerce/pull/431/commits/198aa5a9e132b038ff6a0610908984d2361fac69

frontend/components/common/InstantSearchProvider.tsx
```diff
@@ -68,64 +77,138 @@ export function InstantSearchProvider({
       };
    }, [router]);
 
-   // We're using this to make `InstantSearch` unmount at every re-render, since as of version 6.28.0 it breaks when
-   // it re-renders. Re-rendering though should be relatively infrequent in this component, so this should be fine.
+   // We're using this to make `InstantSearch` unmount at every re-render, since
+   // as of version 6.28.0 it breaks when it re-renders. Re-rendering though
+   // should be relatively infrequent in this component, so this should be fine.
    const count = useCountRenders();
```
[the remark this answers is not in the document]
masonmcelvain: https://github.com/iFixit/react-commerce/pull/431/commits/e6d175f30e500061ac4344f26f85a446e456967e

masonmcelvain:
> > > On mobile, if you select an item type field, you are no longer directed to that page automatically, and it's unclear that the type has been applied. You have to manually exit the filter fields.
> > 
> > 
> > Good catch! Maybe a better UX would be to navigate back to the facet overview when an Item Type is clicked, and maybe display that the refinement is selected just like any other refinement.
> 
> That sounds like a reasonable UX, but I'm not sure if it matches with the current behavior/UX. Right now (on this branch), when you return to the facet overview after selecting an item type, the item type facet is hidden (as expected). I'm not sure of the implications of adding it back in with an indicator that it has a selected refinement. Will it be clickable, and if you enter it, how do you indicate which item type is selected without the checkboxes? I don't really have a better solution, though. Other than the main behavior of exiting refinements altogether.

After discussing with @sterlinghirsh, we decided to NOT hide single selects when they are selected (both on desktop and mobile).

Instead, they will continue to be displayed, and the selected option is highlighted. Selecting another option behaves as expected - the old option is removed from the search, and the search is refined according to the new selection.

We also chose to close the facet drawer on mobile when the user clicks a single select. I think this solves issue 1 from this comment: https://github.com/iFixit/react-commerce/pull/431#issuecomment-1192023207. We could change this in the future to navigate back to the facet overview instead, but I think either UX is fine.

un_dev_block ✌🏻

sterlinghirsh:
CR :zap:

sterlinghirsh (dismissed):

erinemay:
Looks good. The behavior is a lot more unified between mobile and desktop now. 👍 
Again confirmed: 

- Selecting an item type doesn't wipe out other selected filters
- Item type shows up as a filter in the list of selected filters that can be cleared without losing other filters
- Item type is in the url path as opposed to params (like /Parts/MacBook_Pro/Bluetooth_Board)
- Facet lists retain open/close state when entering or leaving an item type page
- /Tools remains unaffected

dev_block 🌵 I noticed that sometimes when there is an item type selected, the clear button is disabled. 
| Disabled | Enabled |
| -- | -- |
| [PC Laptop Adhesives](https://react-commerce-prod-6m7wyawev-ifixit.vercel.app/Parts/PC_Laptop/Adhesives) | [Mac Antennas](https://react-commerce-prod-6m7wyawev-ifixit.vercel.app/Parts/Mac/Antennas?filter%5Bfacet_tags.Device%20Type%5D%5B0%5D=MacBook%20Pro%2015%22&filter%5Bfacet_tags.Device%20Type%5D%5B1%5D=MacBook%20Pro%2013%22&filter%5Bfacet_tags.Device%20Type%5D%5B2%5D=Mac%20mini&filter%5Bprice_range%5D%5B0%5D=%2410%20-%20%2425&filter%5Bprice_range%5D%5B1%5D=%240%20-%20%2410) plus some other refinements | 
| Screenshot_20220722-142933_Chrome | Screenshot_20220722-143633_Chrome |

TBH, I couldn't get a clear pattern of reproduction steps. I saw the button both enabled and disabled while testing, but I couldn't find a clear pattern to enabling it. The example I found above where the clear button is enabled has other refinements selected, but just selecting other refinements doesn't always trigger it to be enabled.

masonmcelvain:
Thanks @erinemay! I was able to reproduce locally (and I couldn't find a pattern either!). I tried disabling the button a slightly different way. I can't reproduce the bug anymore, but I'd appreciate if you could give it a try when you have a chance.

un_dev_block ✌🏻

sterlinghirsh (approved):

sterlinghirsh:
CR :zap:

erinemay:
QA 👍 Looks good. I think I've exhausted clicking through it in every way I can conceive. 
The odd clear button behavior [here](https://github.com/iFixit/react-commerce/pull/431#issuecomment-1192957450) is fixed. And: 

- The transition is smooth and makes it look like you're on the same page
- Selecting or deselecting an item type doesn't wipe out other selected filters
- On desktop, item type shows up as a filter in the list of selected filters that can be cleared without losing other filters
- Item type is in the url path as opposed to params (like /Parts/MacBook_Pro/Bluetooth_Board)
- Facet lists retain open/close state when entering or leaving an item type page
- `Clear all` clears all filters and the item type selection
- /Tools remains unaffected

PR #434, opened 2022-07-13
Empty Tile: Get rid of it

### LINKED ISSUE #410: Empty tile in grid mode
empty

This was noticed during our hallway usability testing

---

### Reproduction Steps:
1. Navigate to [/Parts](https://d3lnoy3aplwg2k.cloudfront.net/Parts)
2. Select the grid view
3. Scroll down to the bottom of the page
4. Observe the blank entry

sumukhipandey:
## Summary:

Previously, when switched to grid view, an empty tile appeared if we scrolled all the way to the bottom. This PR gets rid of that tile.

## CR/QA:
- We had a `bg` color set to `gray.100` for the `<SimpleGrid`.
- While this was adding borders between each individual link box, if there was no link box displayed we could see an empty tile in the grid.
- I decided to get rid of that `bg` color and add borders to each individual `LinkBox`.
   - This took care of both things:
      - The empty tile
      - The border around the linkboxes.

<details><summary>Screenshot BEFORE Change</summary>
Screenshot 2022-07-12 184823
</details>

<details><summary>Screenshot AFTER Change</summary>
Screenshot 2022-07-12 184640
</details>

Closes #410 

CC: @jordycosta

### DIFF
--- a/frontend/components/product-list/sections/FilterableProductsSection/ProductGrid.tsx
+++ b/frontend/components/product-list/sections/FilterableProductsSection/ProductGrid.tsx
@@ -20,7 +20,6 @@ export function ProductGrid({ children }: ProductGridProps) {
    return (
       <SimpleGrid
          data-testid="grid-view-products"
-         bg="gray.100"
          borderBottomColor="gray.100"
          borderBottomWidth="1px"
          columns={{
@@ -46,7 +45,7 @@ export function ProductGridItem({ product }: ProductGridItemProps) {
       useProductSearchHitPricing(product);
 
    return (
-      <LinkBox as="article" display="block" w="full" role="group">
+      <LinkBox as="article" display="block" w="full" role="group" borderWidth={"thin"} borderColor={"gray.100"}>
          <ProductCard h="full">
             <ProductCardImage src={product.image_url} alt={product.title} />
             <ProductCardBadgeList>
--- a/frontend/components/product-list/sections/FilterableProductsSection/ProductGrid.tsx
+++ b/frontend/components/product-list/sections/FilterableProductsSection/ProductGrid.tsx
@@ -45,7 +45,7 @@ export function ProductGridItem({ product }: ProductGridItemProps) {
       useProductSearchHitPricing(product);
 
    return (
-      <LinkBox as="article" display="block" w="full" role="group" borderWidth={"thin"} borderColor={"gray.100"}>
+      <LinkBox as="article" display="block" w="full" role="group" borderWidth="thin" borderColor="gray.100">
          <ProductCard h="full">
             <ProductCardImage src={product.image_url} alt={product.title} />
             <ProductCardBadgeList>
--- a/frontend/components/product-list/sections/FilterableProductsSection/ProductGrid.tsx
+++ b/frontend/components/product-list/sections/FilterableProductsSection/ProductGrid.tsx
@@ -45,7 +45,18 @@ export function ProductGridItem({ product }: ProductGridItemProps) {
       useProductSearchHitPricing(product);
 
    return (
-      <LinkBox as="article" display="block" w="full" role="group" borderWidth="thin" borderColor="gray.100">
+      <LinkBox
+         as="article"
+         display="block"
+         w="full"
+         role="group"
+         borderTop="0"
+         borderRight="1px"
+         borderLeft="0"
+         borderBottom="1px"
+         borderRightColor="gray.100"
+         borderBottomColor="gray.100"
+      >
          <ProductCard h="full">
             <ProductCardImage src={product.image_url} alt={product.title} />
             <ProductCardBadgeList>
--- a/frontend/components/product-list/ProductListView.tsx
+++ b/frontend/components/product-list/ProductListView.tsx
@@ -46,7 +46,7 @@ export function ProductListView({
          <PageContentWrapper py="10">
             <VStack align="stretch" spacing="12">
                <Index indexName={indexName}>
-                  <Configure filters={filters} />
+                  <Configure filters={filters} hitsPerPage={18}/>
                   <MetaTags productList={productList} />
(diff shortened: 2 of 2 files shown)


sumukhipandey:
This seemed like a simple change to me but I wasn't sure if this issue needed changes in the logic to get rid of the tile. I'd be happy to do so if needed.

sterlinghirsh (dismissed):

sterlinghirsh:
CR :zap:

frontend/components/product-list/sections/FilterableProductsSection/ProductGrid.tsx
```diff
@@ -46,7 +45,7 @@ export function ProductGridItem({ product }: ProductGridItemProps) {
       useProductSearchHitPricing(product);
 
    return (
-      <LinkBox as="article" display="block" w="full" role="group">
+      <LinkBox as="article" display="block" w="full" role="group" borderWidth={"thin"} borderColor={"gray.100"}>
```
danielbeardsley, line 48:
```suggestion
      <LinkBox as="article" display="block" w="full" role="group" borderWidth="thin" borderColor="gray.100">
```

davidrans (dismissed):
CR :coffee:

k0rvusk0r4x:
On the one hand:
- The background is no longer there
- The border is fixed.

One the other hand, this doesn't really address the issue, which is that there is a space in the grid because 20 is not evenly divisible by 3. There should only be a space like that if there are no more pages of products in that category.

dev_block 😿

sumukhipandey:
> On the one hand:
> 
> * The background is no longer there
> * The border is fixed.
> 
> One the other hand, this doesn't really address the issue, which is that there is a space in the grid because 20 is not evenly divisible by 3. There should only be a space like that if there are no more pages of products in that category.
> 
> dev_block 😿

I tried looking into this initially. The columns are getting [set here](https://github.com/iFixit/react-commerce/blob/8ec0d5a04afa1ea8405c6567238ad8d431b6d884/frontend/components/product-list/sections/ProductListChildrenSection.tsx#L82) where each size `sm`, `md` and `lg` are window sizes I believe. If the value of  `lg` is changed to 4 then that makes it look the way it would be ideal. But will we always get 20 devices per page?  I wasn't sure about that. 
@sterlinghirsh @danielbeardsley @mason thoughts?

masonmcelvain:
The default hits per page is 20 in our product_group_en index, which is why we're getting 20.
image

We can override this in the query if we want by passing the `hitsPerPage` prop here

https://github.com/iFixit/react-commerce/blob/2ecbea829da988fb047c0ba96b497ebf217fcc05/frontend/components/product-list/ProductListView.tsx#L49

sumukhipandey:
Thanks @danielbeardsley @masonmcelvain. I made the change here 393709b91a6ee3593ebe5bbe9be84e56b2bc13e8
un_dev_block 🦋 
@k0rvusk0r4x now the products are displayed as you mentioned and the only time the empty tile/tiles appear are on the last page. The previous changes should take care of that.

frontend/components/product-list/sections/FilterableProductsSection/ProductGrid.tsx
```diff
@@ -20,7 +20,6 @@ export function ProductGrid({ children }: ProductGridProps) {
    return (
       <SimpleGrid
          data-testid="grid-view-products"
-         bg="gray.100"
```
masonmcelvain, line 23: Do we still want to remove these styles?

masonmcelvain (approved):
CR ✌🏻 but deploy_block ✌🏻 on a question

jordycosta:
QA 🟢 

Looks great! 👍 

tile

frontend/components/product-list/sections/FilterableProductsSection/ProductGrid.tsx
sumukhipandey: I believe these styles would still result in empty tiles on the last page if towards the end the # of products is < 18.
That's why I decided to leave the changes where I delete this. 
For example, if I add the `bg="gray.100"` we would end up with the last page looking like:
Screenshot 2022-07-13 142511

sumukhipandey: Unless there is a way to fix that as well, I think keeping the commit where `bg="gray.100"` is deleted would fix this last page issue.

dhmacs:
@jordycosta I don't think this should pass QA, as there's a regression in the grid cell border. We shouldn't have a double border:

image

dev_block 🎨

sumukhipandey:
> @jordycosta I don't think this should pass QA, as there's a regression in the grid cell border. We shouldn't have a double border:
> 
> image
> 
> dev_block 🎨

@dhmacs Thanks for pointing this out. I wasn't sure if we were okay with the double borders. I made some changes here ac3d655da6a28939d6320b7249afea5b746e7917. This fixes the issue you addressed above. 
I wasn't sure if there was a simpler/cleaner way to approach this so let me know if there is and I would be happy to do it!

jordycosta:
> @jordycosta I don't think this should pass QA, as there's a regression in the grid cell border. We shouldn't have a double border:

@dhmacs 
I assumed this was desired since it was in the spec above
> I decided to get rid of that bg color and add borders to each individual LinkBox

---

QA 🟢 

Border's are no longer present for each individual LinkBox 👍 

border
border-2

frontend/components/product-list/sections/FilterableProductsSection/ProductGrid.tsx
sumukhipandey: un_deploy_block 🦋

danielbeardsley:
CR :+1:

sumukhipandey:
un_deploy_block 🦋

danielbeardsley (approved):
CR :+1:

Issue #215, opened 2022-03-17
Hide some reviews if review count is low


erinemay:
### Current behavior:
Review stars and review count are always shown if the number of reviewers are > 0. This can lead to misleading, low ratings based on a small sample size.

### Do this:
Don't show the reviews in the product container if the average rating is < 4 stars unless the review count is > 10.

(following the same behavior as ifixit.com https://github.com/iFixit/ifixit/issues/27500)

PR #438, opened 2022-07-13
Tests: Update Jest setup and add snapshot test

batbattur:
## Summary of changes
1. This pull updates the existing `jest.config.js` and install jest `devDependencies` for jest tests.
2. Adds a snapshot test for `ProductListItem`.

## CR Notes
I'm not really familiar with babel config files therefore would like review on https://github.com/iFixit/react-commerce/commit/0e609146e4ededfec40bd12d72d7c945dcd3074f to make sure it doesn't break anything. 

## QA Notes
1. Please make sure the deployed app works properly.
2. It would awesome if you run the jest test locally by running `pnpm install:all`, `cd frontend` and `pnpm test` and make sure it works.

Closes: https://github.com/iFixit/react-commerce/issues/124 and https://github.com/iFixit/ifixit/issues/43590


deltuh-vee:
Frontend jest tests are running locally. Still have to check the app is still working.

frontend/test/jest/tests/ProductListItem.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+      ProductSearch.useProductSearchHitPricing = jest.fn().mockReturnValue({
+         price: 5,
+         compareAtPrice: 5,
+         isDiscounted: 5,
+         percentage: 5,
+      });
+
+      AppContext.useAppContext = jest
+         .fn()
+         .mockReturnValue({ ifixitOrigin: 'https://www.cominor.com' });
```
mlahargou, line 19: Do you think we will need to add this to multiple places as we add more tests? Perhaps it should go in some sort of global mock file.

frontend/test/jest/tests/ProductListItem.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+   });
+
+   afterEach(() => {
+      jest.clearAllMocks();
+   });
+
+   /**
+    * @jest-environment jsdom
```
mlahargou, line 27: Why do we need this?

frontend/test/jest/tests/ProductListItem.test.tsx
```diff
@@ -0,0 +1,38 @@
+// @ts-nocheck
```
mlahargou, line 1: Do we will need this?

frontend/test/jest/tests/ProductListItem.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+   afterEach(() => {
+      jest.clearAllMocks();
+   });
+
+   /**
+    * @jest-environment jsdom
+    */
+   it('renders ProductListItem', () => {
+      const { asFragment } = render(<ProductListItem product={mockProduct} />);
```
mlahargou, line 30: Nice job getting this working!

frontend/test/jest/tests/ProductListItem.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+   });
+
+   afterEach(() => {
+      jest.clearAllMocks();
+   });
+
+   /**
+    * @jest-environment jsdom
+    */
+   it('renders ProductListItem', () => {
```
mlahargou, line 29: Seems a bit odd to have the `describe` text and the `it` text be the same. Maybe `describe('ProductListItem test)` and `it('matches the ProductListItem snapshot')`?

mlahargou (commented):
CR 📱 

Nice job! A couple questions/comments deploy_block 📱

andyg0808:
We're no longer using Babel for our dev or production builds. I'd suggest that we should switch our Jest to using [`@swc/jest`](https://github.com/swc-project/jest), so that everything is compiled with the same engine. We might be able to move some of the SWC config out of the Webpack config and into a SWC config file, so that it can be shared between Jest and Webpack.

In general, I'd push for us to avoid duplicating settings between our tests and our actual builds, because any differences mean that the tests are testing slightly different code than is actually running in production.

frontend/test/jest/tests/ProductListItem.test.tsx
batbattur: Moved to a global mock file in https://github.com/iFixit/react-commerce/pull/438/commits/8c37f35950332ac60c52cd675fbd297ff026d78d.

frontend/test/jest/tests/ProductListItem.test.tsx
batbattur: It's from here https://jestjs.io/docs/25.x/configuration#testenvironment-string but looks like we don't need it anymore because the default is `jsdom`. Removed it in https://github.com/iFixit/react-commerce/pull/438/commits/8c37f35950332ac60c52cd675fbd297ff026d78d.

frontend/test/jest/tests/ProductListItem.test.tsx
batbattur: Yeah, unfortunately, we need this to bypass type checking inside the test, otherwise, we get typechecking errors when setting up the mocks functions and objects like these: 
image
image

andyg0808: I'd recommend that instead we add [`@ts-ignore` comments](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-6.html#suppress-errors-in-ts-files-using--ts-ignore-comments) above the troublesome lines, in order to enable us to have type checking on the rest of the code in this file. That is potentially helpful in quickly finding places that need to be changed when refactoring.

batbattur: Sounds good! Added `@ts-ignore` and removed `@ts-ignore` comments in https://github.com/iFixit/react-commerce/pull/438/commits/90165dc4d20cc331a3b6ca50f0470f118b61ec35.

batbattur:
un_deploy_block 🚛

mlahargou (approved):
CR 📱

frontend/test/jest/tests/ProductListItem.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+      );
+
+      const reviewStars = screen.getByTestId('reviewStars');
+      (expect(reviewStars) as any).toBeInTheDocument();
+
+      (expect(shortDescription) as any).toBeInTheDocument();
+      (expect(asFragment()) as any).toMatchSnapshot();
+   });
+
+   it('renders without the review stars and matches the snapshot', () => {
```
mlahargou, line 37: Do you think it's necessary to check all permutations?

```
rating < 4 && rating_count < 10 {
   shouldn't show stars
}
rating < 4 && rating_count > 10 {
   shouldn't show stars
}
raing > 4 && rating_count < 10 {
   shouldn't show starts
}
rating > 4 && rating_count > 10 {
   should show stars
}
```
Also, is it necessary to do a snapshot for this test?

mlahargou (dismissed):
CR 📱 Nice job on this!

frontend/test/jest/tests/ProductListItem.test.tsx
batbattur: > Do you think it's necessary to check all permutations?

We definitely should! I think you caught a bug on master, [here](https://github.com/iFixit/react-commerce/blob/main/frontend/components/product-list/sections/FilterableProductsSection/ProductList.tsx#L145).
https://github.com/iFixit/react-commerce/blob/fb76f57c0e1e414db4cdf36da83e12879fd213c7/frontend/components/product-list/sections/FilterableProductsSection/ProductList.tsx#L145

We display the stars if the rating is greater than or equal to 4, **OR** when the rating count is greater than 10. 
This means we would end up displaying 2 stars if the review count is greater than 10. 

I'll clarify the expected behavior for displaying the stars and then make a follow-up pr to add all permutations. 

> Also, is it necessary to do a snapshot for this test? Meaning can we just check that the element is in the document instead of a whole snapshot test.

I don't think it's necessary to do a snapshot for this test. Removed the snapshot in https://github.com/iFixit/react-commerce/pull/438/commits/dc1403a39a423727539f9c5e798d5271210e6f06.

batbattur: >We display the stars if the rating is greater than or equal to 4, OR when the rating count is greater than 10.
This means we would end up displaying 2 stars if the review count is greater than 10.

Oh, nevermind. I think it is the expected behavior. We will show any review stars if the review count is greater than 10.
Found it from here https://github.com/iFixit/react-commerce/issues/215:
 
> Don't show the reviews in the product container if the average rating is < 4 stars unless the review count is > 10.

batbattur:
un_deploy_block 🚛

mlahargou (approved):
CR 📱

mlahargou:
> Don't show the reviews in the product container if the average rating is < 4 stars unless the review count is > 10.

I still think we could benefit from testing this logic:

- rating < 4; count < 10
  - don't show
- rating < 4; count > 10
  - show
- rating > 4; count < 10
  - show
- rating > 4; count > 10
  - show

frontend/test/jest/tests/ProductListItem.test.tsx
batbattur: Addressed in https://github.com/iFixit/react-commerce/pull/438/commits/63d6578415daf3f5e739a5a434d308aa1b41d8a1.

batbattur:
> > Don't show the reviews in the product container if the average rating is < 4 stars unless the review count is > 10.
> 
> I still think we could benefit from testing this logic:


Added another test to cover for all cases for review stars in https://github.com/iFixit/react-commerce/pull/438/commits/63d6578415daf3f5e739a5a434d308aa1b41d8a1.

frontend/test/jest/tests/ProductListItem.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+      const shortDescription = screen.getByText(
+         'Replace a dead or malfunctioning model EB-BG96aasd5ABE battery in a Samsung Galaxy S9 Plus smartphone.'
+      );
+
+      (expect(shortDescription) as any).toBeInTheDocument();
+      (expect(asFragment()) as any).toMatchSnapshot();
+   });
+
+   it('renders without the review stars', () => {
+      // We don't render the stars if the rating <= 4 or rating_count < 10
```
mlahargou, line 35: [minor]

I think this should be:
```suggestion
      // We don't render the stars if the rating <= 4 AND rating_count < 10
```

mlahargou (dismissed):
CR 📱 With one comment update

batbattur:
@deltuh-vee this is ready for another round of QA 🙇

deltuh-vee:
QA 🎬 
* Deployed app still works
* jest test passes locally

mlahargou (approved):
CR 📱

PR #457, opened 2022-07-19
Performance profiling: Add a bunch of logging

danielbeardsley:
It's not perfect and it's a bit verbose, but we have log info on most of our backend requests now.

Future Work
====
* Set a random string per request and prefix all log entries with that so log entries from one request can be differentiated
* Hunt around and find any other async requests (might be a few more strapi requests)
* ~~Fix bug: parallel requests cause multiple timers to start with the same name~~
* Log a request-start line
* ~~Add way to silence or reduce log verbosity (env var, more sophisticated logging library?)~~ (disabled logging in prod)


masonmcelvain (approved):
CR ✌🏻 through dc569e93d767465d6afb12fd73e8be28595db9df, looks great!

frontend/models/product-list/index.ts
```diff
@@ -47,23 +48,30 @@ export async function findProductList(
    filters: ProductListFiltersInput,
    options: ProductListOptions = {}
 ): Promise<ProductList | null> {
+   console.time('getProductList');
    const result = await strapi.getProductList({
       filters,
    });
+   console.timeEnd('getProductList');
```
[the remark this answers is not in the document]
danielbeardsley: This has been replaced with `logAsync()` I started with the commit from @federicobadini and slowly replaced bits of it / added to it.

frontend/pages/Parts/index.tsx
```diff
@@ -10,6 +10,7 @@ import {
    ProductListViewProps,
 } from '@components/product-list';
 import { ALGOLIA_DEFAULT_INDEX_NAME } from '@config/constants';
+import { asyncTimings } from '@helpers/performance-helpers';
```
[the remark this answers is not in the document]
danielbeardsley: Removed!

packages/helpers/index.tsx
```diff
@@ -31,3 +31,13 @@ export function invariant(
 export function isError(x: any): x is Error {
    return x instanceof Error;
 }
+
+export function logAsync<T>(
```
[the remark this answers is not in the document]
danielbeardsley: That other file has been deleted.

frontend/models/product-list/index.ts
```diff
@@ -432,3 +434,9 @@ function createPublicAlgoliaKey(appId: string, apiKey: string): string {
    });
    return publicKey;
 }
+
+function createIFixitAPIClient() {
```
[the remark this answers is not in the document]
danielbeardsley: Good question! Because this function is called from `getServerSideProps()` which is outside of a react context and thus can't use hooks.

danielbeardsley:
Pushed a bunch of changes. I think this is ready to go now.

packages/auth-sdk/user.tsx
```diff
@@ -29,8 +31,11 @@ export function useAuthenticatedUser() {
    return query;
 }
 
-async function fetchAuthenticatedUser(apiOrigin: string): Promise<User | null> {
-   const response = await fetch(`${apiOrigin}/api/2.0/user`, {
+async function fetchAuthenticatedUser(
+   apiClient: IFixitAPIClient,
+   apiOrigin: string
```
sterlinghirsh, line 36: Looks like apiOrigin isn't used. Any reason why we pass the client and origin in here instead of getting the client from inside fetchAuthenticatedUser and getting the origin inside the client?

danielbeardsley: Just forgot to remove a now-unused argument.

packages/ifixit-api-client/index.tsx
```diff
@@ -44,12 +45,14 @@ export class IFixitAPIClient {
...
    async fetch(endpoint: string, init?: RequestInit) {
-      const response = await fetch(
-         `${this.origin}/api/${this.version}/${endpoint}`,
-         {
-            credentials: 'include',
-            ...init,
-         }
+      const url = `${this.origin}/api/${this.version}/${endpoint}`;
+      const response = await logAsync(
+         `${init?.method || 'GET'}:${truncate(url, 70)}`,
```
sterlinghirsh, line 50: We can save some space by skipping origin/api/version
```suggestion
         `API ${init?.method || 'GET'}:${truncate(endpoint, 70)}`,
```

sterlinghirsh:
CR :zap: dev_block :zap: on some minor changes. also there are test failures

sterlinghirsh:
Ideally rather than a random string, if we could identify requests by their start time, that would be nice. Not sure how easy it is to get a time since the beginning of the request, but even a sequential timer id or request id, like `1: GET:wikis/CATEGORY/iPhone_7 125ms`.

danielbeardsley:
> Ideally rather than a random string, if we could identify requests by their start time, that would be nice.

Yes, both of those options would be better. Practically, it seems like it'd be annoying to pass an id around to every place that does logging, or an instance of the logger that prefixes it. This isn't PHP, there are multiple requests being handled per process.

I'm going to leave that to future pulls.

packages/ifixit-api-client/index.tsx
danielbeardsley: Yeah, that is better. I was interested in including the hostname... but the endpoint is the crucial one.

danielbeardsley:
So I've run the cypress tests locally and they all pass. I'm running out of things to try here... I'm not sure why the tests fail.

The *only* error we see is "Failed to fetch" on the client side in an unhandled promise rejection. This sounds to me like one of the api requests made from the client fails and we *used* to ignore the error and in this branch we *don't*?

frontend/lib/cache.ts
```diff
@@ -1,14 +1,18 @@
...
+   ttl: number
+): Promise<T> {
+   const fromCache = cacheStore.get(key) as Promise<T> | undefined;
    if (fromCache !== undefined) {
-      console.log(`Got '${key}' from cache`);
       return fromCache;
    }
-   const result = fetch();
-   console.log(`Storing '${key}' in cache`);
+   const result = logAsync(`Cache miss for '${key}'`, () => fetch());
```
sterlinghirsh, line 15: I think this could be a bare `fetch`. Also can we use another name since fetch is overloaded?

frontend/lib/ifixit-api/devices.ts
```diff
@@ -12,14 +13,17 @@ export async function fetchDeviceWiki(
          deviceHandle.length > 0,
          'deviceHandle cannot be a blank string'
       );
-      const response = await fetch(
-         `${IFIXIT_ORIGIN}/api/2.0/cart/part_collections/devices/${deviceHandle}`,
+      const response = await client.get(
+         `cart/part_collections/devices/${deviceHandle}`,
```
sterlinghirsh, line 17: Let's put the headers in the client

frontend/lib/cache.ts
danielbeardsley: Fixed!

frontend/lib/ifixit-api/devices.ts
danielbeardsley: Fixed by removing the headers.

danielbeardsley:
un_dev_block :+1: 

Fixed the problems!  Turns out `promise.finally()` didn't work exactly like I thought.

danielbeardsley:
QA :+1: 

image
image

This highlights something weird though...  the API endpoint in question is really fast (20ms) but from vercel it seems to take 800ms!

danielbeardsley:
Disappointing that the floor for network requests seem to be ~300ms and sometimes much higher even though the underlying requests are fast:
image
image

I'm curling these pages from us-east FYI.

To be clear... that ifixit API is fast. Here it is from log.ifixit.com (to remove network overhead):
image

danielbeardsley:
Well crap... scrap all that, this deploy is using cominor.com:
image

sterlinghirsh:
CR :zap:

sterlinghirsh (approved):

PR #459, opened 2022-07-19
Non EN subdomains: Redirect to the English domain

jarstelfox:
cr_req 2

Closes: https://github.com/ifixit/ifixit/issues/43658


frontend/next.config.js
```diff
@@ -47,13 +47,43 @@ const moduleExports = {
...
+         return redirects;
+      }
+
+      return redirects.concat([
+         {
+            source: '/Parts',
+            has: [
+               {
+                 type: 'host',
+                 value: '^(?!www).*.*.*'
```
jarstelfox, line 69: This is the Magic™


See: regexr.com/6q3nv

frontend/pages/Parts/index.tsx
```diff
@@ -28,6 +28,12 @@ export const getServerSideProps: GetServerSideProps<AppPageProps> = async (
       'public, s-maxage=600, stale-while-revalidate=1200'
    );
 
+   const isEN = !!context.req.headers.host?.match(/www\.ifixit\.com/g);
+
+   if (!isEN) {
+      context.req.headers['X-Robots-Tag'] = 'noindex, nofollow, nosnippet, noarchive, noimageindex';
```
djmetzle, line 34: ???
```suggestion
      context.resp.headers['X-Robots-Tag'] = 'noindex, nofollow, nosnippet, noarchive, noimageindex';
```

frontend/next.config.js
```diff
@@ -47,13 +47,43 @@ const moduleExports = {
...
+      const redirects =[
          {
             source: '/Store/Guide/:guideid',
             destination: `${process.env.NEXT_PUBLIC_IFIXIT_ORIGIN}/Guide/_/:guideid`,
             permanent: true,
          },
       ];
+
+      // Do not hide /Parts and /Tools in dev
+      if (process.env.NODE_ENV == "development") {
```
Michelle5102, line 59:
```suggestion
      if (process.env.NODE_ENV === "development") {
```

Michelle5102 (changes requested):
CR :octocat: but dev_block :octocat: on using `===` instead of `==`

frontend/next.config.js
batbattur, line 59: I believe the cypress tests are failing because we are running the production build when we are running cypress tests: https://github.com/iFixit/react-commerce/blob/c06691b50e0830eaab40a12938a2e8eb55de819c/.github/workflows/main.yml#L29-L30

For a quick/short-term fix, I think we could run tests on the development build to get cypress working, But ideally I think we want to use the production build for tests?
```diff
- build: pnpm run build 
- start: pnpm start 
+ start: pnpm dev
```

frontend/pages/Parts/index.tsx
```diff
@@ -28,6 +28,15 @@ export const getServerSideProps: GetServerSideProps<AppPageProps> = async (
       'public, s-maxage=600, stale-while-revalidate=1200'
    );
 
+   const isEN = !!context.req.headers.host?.match(/www\.ifixit\.com/g);
+
+   if (!isEN) {
```
sctice-ifixit, line 33: Shouldn't we be redirecting if this is the case? How does condition get executed?

frontend/next.config.js
sctice-ifixit: It's unclear to me from [the docs](https://nextjs.org/docs/api-reference/next.config.js/redirects#header-cookie-and-query-matching) what kind of regex `value` is (if it is a regex). Is the `.` a literal `.` here or a regex metacharacter? In either case, this doesn't seem right to me. If it's a literal `.`, then we're expecting 4 domain parts, but wouldn't we just have 3, like `de.ifixit.com`? If it's a metacharacter, then this would match basically any string that doesn't start with `www`, wouldn't it?

Here's an [updated regexr](https://regexr.com/6q9h7) with what I'm guessing is the intended regex, but I'm really not sure.

sctice-ifixit (commented):
dev_block :hand: on what I think are a couple issues.

frontend/next.config.js
```diff
@@ -47,13 +47,42 @@ const moduleExports = {
...
          {
             source: '/Store/Guide/:guideid',
             destination: `${process.env.NEXT_PUBLIC_IFIXIT_ORIGIN}/Guide/_/:guideid`,
             permanent: true,
          },
       ];
+
+      if (process.env.FORCE_SUBDOMAIN_REDIRECTS !== "true") {
+         return redirects;
+      }
```
dhmacs, line 60: Not a big deal, but I like the flow to test the "positive" condition instead of having a negation (reason being negative conditions are generally harder to grasp)..
I find the code a little bit easier to follow this way, but feel free to ignore this suggestion

```suggestion
      if (process.env.FORCE_SUBDOMAIN_REDIRECTS === "true") {
         redirects.concat(...);
      }
      return redirects;
```

jarstelfox:
Putting my knowledge on here:

It looks like next.js middleware would make the redirects in the configs much easier to work with.

We decided to do it serverside, which mostly works. The problem is not with next.js, but cloudfront and vercel. The host header is being eaten (I believe by cloudfront) so we are missing.

We definitely want to solve this problem in the long run, but we found for deploy we can just not serve from couldfront for subdomains and rely on the app code redirecting us.

frontend/pages/Tools/index.tsx
```diff
@@ -28,6 +29,12 @@ export const getServerSideProps: GetServerSideProps<AppPageProps> = async (
       'public, s-maxage=600, stale-while-revalidate=1200'
    );
 
+   const redirects = getSubDomainRedirect(context.req, '/Parts');
```
andyg0808, line 32:
```suggestion
   const redirects = getSubDomainRedirect(context.req, '/Tools');
```

frontend/pages/Tools/[handle].tsx
```diff
@@ -29,6 +30,12 @@ export const getServerSideProps: GetServerSideProps<AppPageProps> = async (
       'public, s-maxage=600, stale-while-revalidate=1200'
    );
 
+   const redirects = getSubDomainRedirect(context.req, '/Parts');
```
Comment (andyg0808) [inline frontend/pages/Tools/[handle].tsx:33]:
```suggestion
   const redirects = getSubDomainRedirect(context.req, '/Tools');
```

frontend/helpers/redirect-helper.ts
```diff
@@ -0,0 +1,18 @@
+import { IncomingMessage } from "http";
+
+export function getSubDomainRedirect(request: IncomingMessage, redirectPath: string) {
+   const regex = new RegExp(process.env.SUBDOMAIN_REDIRECT_UNLESS_REGEX || '^');
+   const host = request.headers.host;
+   const shouldRedirect =  host ? !host.match(regex) : false;
+
+   if (shouldRedirect) {
+      const url = new URL(process.env.NEXT_PUBLIC_IFIXIT_ORIGIN || "");
+      url.pathname = redirectPath;
```
andyg0808, line 10: It might be cute to use https://nodejs.org/dist/latest-v15.x/docs/api/http.html#http_message_url, instead of `redirectPath`, because it looks like that should basically always do exactly what we want: have just the path part of the URL which was used to get here in the first place.

Free bonus, it lets us get rid of `redirectPath`, which simplifies all the calls to this function.
```suggestion
      url.pathname = request.url;
```

frontend/pages/Parts/[deviceHandle]/[itemTypeHandle].tsx
```diff
@@ -0,0 +1,129 @@
+import {
```
Comment (andyg0808) [inline frontend/pages/Parts/[deviceHandle]/[itemTypeHandle].tsx:1]: I'm doubtful that this whole file should be getting added; did it slip in by accident?

andyg0808:
CR :dog: through 50eb959
dev_block :+1: on comments.

jarstelfox:
So we determined:

* The host header is a blocker
* Next v12 would allow a middleware to make this easy
* The new CloudFront distribution allows the subdomains to fallback to the PHP which behaves how we want

---

I think we should regroup and determine how to best approach this problem.

My gut says let's upgrade to v12!

sterlinghirsh:
I'm in favor of next 12! Hopefully the upgrade path is easy.

andyg0808:
I think we'll need a host-header-equivalent no matter what; without that, there's nothing for react-commerce code to hook to in order to determine when to redirect.

Actually, an alternative might be to host NextJS ourselves, so our routing doesn't need to care about the `Host` header. Sounds like we need to have a discussion.

sumukhipandey:
I created an [issue ](https://github.com/iFixit/react-commerce/issues/511)to upgrade to next 12

andyg0808:
And an issue to forward the header: https://github.com/iFixit/ifixit/issues/44023

jarstelfox:
I think we will want to redo this work once we solve https://github.com/iFixit/ifixit/issues/44023

I am going to close this for now.

Issue #430, opened 2022-07-11
Support collection pages that match an iFixit device but no strapi page


sterlinghirsh:
## Problem
We may have some auto-generated links to product lists that don't exist in strapi. These aren't common, but some /Store/id/x pages can redirect. These may not have parts for sale, but we shouldn't just 404.

## Do This
When visiting a collection page like /Parts/SomeDevice, if a page for SomeDevice doesn't exist in strapi but does exist in the api, still show the page. Use the parent device from the api for the parent breadcrumb and show no children. The title should be `[Device Title] Parts`, and there should be no description text. We should also no-index the page, but we can continue to show the Parts/Guides/Answers links in the top right.

PR #463, opened 2022-07-20
Use iFixit API to Create ProductList Pages if Data Missing In Strapi

danielcliu-ifixit:
## Overview
We currently just 404 on pages where the product list data isn't in Strapi, even if a wiki is on iFixit. This calls to the iFixit API and tries to fill in the parts pages if we are missing some data. This will make the breadcrumbs hierarchy on these pages no longer directly tied to the productList in Strapi, but it shouldn't matter too much because we will just source the missing Strapi info from iFixit anyway.  It also replaces the root page from the wiki hierarchy with the `All Parts` productList. 

## CR
The main thing I'm not super happy with is the `const` that consolidate the Strapi and iFixit data together, but I couldn't think of a way to make it better. Also is `wikiid` the best way to identify the Root Page.

## QA
So for your dev Strapi `Parts/Asus_Zenfone_2` should be a vaild page. However `Parts/Asus_Phone` shouldn't be valid for your dev server on master (Note: it is on production see https://es.ifixit.com/Parts/Asus_Phone ) but on this branch it should get  info from iFixit and not be a 404 page. Note that the breadcrumbs on es.ifixit.com for this page and your dev page on this branch should be different because one gets parent data from Strapi and the other from iFixit wikis. Also make sure that most parent breadcrumb should always be `All Parts`.

Also make sure that performance isn't hurt significantly by this pull on Strapi pages, and that performance is acceptable on the pages that have to fetch iFixit wikis.

Closes: #430


frontend/models/product-list/index.ts
```diff
@@ -50,66 +50,81 @@ export async function findProductList(
...
 
-   const deviceWiki = productList.deviceTitle
-      ? await fetchDeviceWiki(productList.deviceTitle)
+   const handle = productList?.handle ?? '';
+   const parents =
+      productList?.parent ??
+      (deviceWiki?.ancestors
+         ? convertAncestorsToStrapiFormat(deviceWiki.ancestors)
+         : null);
+   const type = productList?.type ?? deviceWiki?.namespace;
```
sterlinghirsh, line 70: I think this type wants to be one of the productList type enums. So this should probably fall back to a "device parts" type.

frontend/models/product-list/index.ts
```diff
@@ -50,66 +50,81 @@ export async function findProductList(
...
+   const title =
+      productList?.title ??
+      (deviceWiki?.title ? deviceWiki?.title + ' Parts' : '');
+   const description =
+      productList?.description ?? deviceWiki?.description ?? '';
+   const productListImageAttributes = productList?.image?.data?.attributes;
+   const image = productListImageAttributes
```
sterlinghirsh, line 77: IIRC we aren't showing the main image at all anymore (only child images) so we can probably lose this.

frontend/models/product-list/index.ts
```diff
@@ -227,6 +242,31 @@ type ApiProductList = NonNullable<
...
+      ancestor['type'] = 'all_parts';
+      ancestor['title'] = 'All';
+      ancestor['handle'] = 'Parts';
+   }
+   return {
+      data: {
+         attributes: {
+            type: ancestor.type,
+            title: ancestor.title + ' Parts',
+            handle: ancestor.handle ?? '',
```
sterlinghirsh, line 262: What is this handle?

frontend/models/product-list/index.ts
```diff
@@ -227,6 +242,31 @@ type ApiProductList = NonNullable<
...
 >;
 
+function convertAncestorsToStrapiFormat(
+   ancestors: any
+): ApiProductList['parent'] | null {
+   const ancestor = ancestors.shift();
+   if (ancestor == null) {
+      return null;
+   }
+   if (ancestor.wikiid === 11000) {
```
sterlinghirsh, line 252: Any reason to check wikiid rather than title === 'Root'?

frontend/lib/ifixit-api/devices.ts
```diff
@@ -16,7 +16,7 @@ export async function fetchDeviceWiki(
          }
       );
       const payload = await response.json();
-      return payload;
+      return response.ok ? payload : null;
```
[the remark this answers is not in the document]
sterlinghirsh: This is already changed in another pull so don't worry about it.

frontend/models/product-list/index.ts
danielcliu-ifixit: i thought that `wikiid` would be the most stable identifier but I asked about this in my CR section.

frontend/models/product-list/index.ts
danielcliu-ifixit: the ancestors but for most of them that information isn't available to be created from the ifixit info. But its a required attribute by ts rn

frontend/models/product-list/index.ts
danielcliu-ifixit: we only use type for `getProductListType()` and that has a default in its switch so im just going to ditch the const and have the conditional as the parameter.

danielcliu-ifixit:
un_dev_block :v:

frontend/models/product-list/index.ts
```diff
@@ -50,66 +50,77 @@ export async function findProductList(
...
 
    const algoliaApiKey = createPublicAlgoliaKey(
       ALGOLIA_APP_ID,
       ALGOLIA_API_KEY
    );
-   const baseProductListType = getProductListType(productList.type);
+
+   const baseProductListType = getProductListType(
+      productList?.type ?? deviceWiki?.namespace
+   );
```
sterlinghirsh, line 83:
```suggestion
   const baseProductListType = getProductListType(productList?.type);
```

sterlinghirsh (dismissed):

sterlinghirsh:
CR :zap:

sterlinghirsh:
CR :zap:

sterlinghirsh (dismissed):

danielcliu-ifixit:
Just changed some formatting to what the formatter wanted (we should commit those automatically) and also a merge from main so CR should carry over cause nothing should have changed really.

k0rvusk0r4x:
- The pages no longer 404.
- Most parent breadcrumbs are "All Parts."
- Performance does not appear to be hurt.

QA 😺

sterlinghirsh (approved):

sterlinghirsh:
CR :zap:

Issue #412, opened 2022-07-07
Mini-Cart: Empty Cart emoji isn't rendering on Windows + Chrome Desktop


erinemay:
![image](https://user-images.githubusercontent.com/72166715/177872981-0b99d88f-99d9-4c31-98ed-c3274661f1ba.png)

Confirmed not working on Windows 10 and lower on Chrome (102,103, and 104-beta). ❌ 

Confirmed working on:
- iOS + Safari ✔️ 
- Android + Chrome ✔️ 
- MacOS + Safari ✔️ 
- MacOS + Chrome ✔️ 
- Windows + FF ✔️ 

(All tested via Browserstack)

Also, are we sure the smiling/crying emoji is what we want here?
![image](https://user-images.githubusercontent.com/72166715/177873112-e7f831b0-2452-4ad3-8e39-82f928c23eb0.png)

## Do This
Make it a shopping cart icon

dhmacs:
> Also, are we sure the smiling/crying emoji is what we want here? ![image](https://user-images.githubusercontent.com/72166715/177873112-e7f831b0-2452-4ad3-8e39-82f928c23eb0.png)

This was a temporary solution. @mmarcon91 will think about this

danielbeardsley:
How about change it to just a slightly larger shopping cart icon?

mmarcon91:
I created a new empty state with font-awesome icon : https://www.figma.com/file/sfj6r7Bn6eREMvuqo0f7Du/iFixit---Components?node-id=284%3A17687

PR #467, opened 2022-07-21
Mini-Cart: Fix Empty Cart Icon

danielcliu-ifixit:
## Overview

The Empty Cart icon was just a crying emoji, and so we wanted to replace that with the design [here](https://www.figma.com/file/sfj6r7Bn6eREMvuqo0f7Du/iFixit---Components?node-id=284%3A17687). We also wanted to make sure that the cart icon works on on Windows 10 and lower on Chrome (102,103, and 104-beta).

This adds `Font Awesome` dependencies to use the specified design. 

## QA
Does the install work fine? You will have to add the `FONT_AWESOME_NPM_TOKEN` to the proper `.env` file for this install to work.

Click on the cart icon on your app and make sure you don't have any items added. If it doesn't the previously mentioned design should now show up.

Closes: #412


frontend/.npmrc
```diff
@@ -0,0 +1,2 @@
+@fortawesome:registry=https://npm.fontawesome.com/
+//npm.fontawesome.com/:_authToken=7FADC5BA-2678-4A03-95C9-A4B3F422FCFF
```
masonmcelvain, line 2: I see in https://github.com/iFixit/ifixit/commit/e7cc5a4262380c8c350385b868859269cb3b6409 we also have this line:

```
//npm.fontawesome.com/:always-auth=true
```

Do we want that here too? Not really sure what this authentication is for, but thought I'd point it out

masonmcelvain (approved):
CR ✌🏻 but deploy_block ✌🏻 on a question

batbattur:
dev_block 🚒 on resolving conflicts for `pnpm-lock.yaml`

danielcliu-ifixit:
Uh I was really hoping that the failures would go away once i force pushed :-1: 

deploy_block :v:  on investigating the `.env` files

.github/workflows/main.yml
```diff
@@ -16,7 +18,9 @@ jobs:
          - name: Install PNPM
            run: corepack enable && corepack prepare pnpm@6.32.11 --activate
          - name: Install workspaces
-           run: pnpm install
+           env:
+              FONT_AWESOME_NPM_TOKEN: ${{ secrets.FONT_AWESOME_NPM_TOKEN }}
+           run: echo "$FONT_AWESOME_NPM_TOKEN" && pnpm install
```
sterlinghirsh, line 23: nope

sterlinghirsh:
CR :zap:

sterlinghirsh (approved):

sterlinghirsh:
CR :zap:

sterlinghirsh:
QA :zap: 
image

package.json
```diff
@@ -7,7 +7,7 @@
       "node": ">= 14.19"
    },
    "scripts": {
-      "install:all": "pnpm install && npm run install:backend",
+      "install:all": "export $(cat .env) && pnpm install && npm run install:backend",
```
masonmcelvain, line 10: Why are we `cat`ing and exporting here?

.env.local.example
```diff
@@ -0,0 +1 @@
+FONT_AWESOME_NPM_TOKEN=
```
masonmcelvain, line 1: Maybe we could move this to `frontend/.env.local.example`, since it seems to only be needed for the frontend? Adding some instructions on where to find this env var in the README would be nice too

masonmcelvain (commented):
Had some trouble `pnpm install`ing, and I think I just need the new env var tho

image

.npmrc
```diff
@@ -1 +1,5 @@
-shamefully-hoist=true
\ No newline at end of file
+shamefully-hoist=true
+always-auth=false
+@fortawesome:registry=https://npm.fontawesome.com/
+//npm.fontawesome.com/:_authToken=${FONT_AWESOME_NPM_TOKEN}
+//npm.fontawesome.com/:always-auth=true
```
masonmcelvain, line 5: Why do we have this here and in `frontend/.npmrc`?

Issue #300, opened 2022-05-18
Add analytics


dhmacs:
We need to setup Analytics so that we're not losing insights when we launch the new product list page. It seems that currently iFixit is on a migration from Google Analytics to [Matomo](https://matomo.org/).

What's the current setup? Should we only use Matomo? What do we want to track?

cc @sterlinghirsh @ifixitmaxb

sterlinghirsh:
Let's try to get both GA and Matomo

PR #474, opened 2022-07-22
Analytics: Add Matomo script

masonmcelvain:
Tracking code from: https://minerva.ifixit.com/index.php?module=CoreAdminHome&action=trackingCodeGenerator

## QA

Click around on dev and prod previews, and make sure the event show up in [matomo.ubreakit.com](http://matomo.ubreakit.com/) or [minerva.ifixit.com](https://minerva.ifixit.com). Thanks!

_Note:_ You should disable all ad blockers when testing this, as they might prevent the Matomo script from running.

CC @sterlinghirsh

Connects #300

### DIFF
--- a/frontend/components/analytics/Matomo.tsx
+++ b/frontend/components/analytics/Matomo.tsx
@@ -4,21 +4,24 @@ import Head from 'next/head';
 export function Matomo() {
    return MATOMO_URL ? (
       <Head>
-         <script key="matomo-analytics">
-            {`
-            var _paq = window._paq = window._paq || [];
-            /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
-            _paq.push(['trackPageView']);
-            _paq.push(['enableLinkTracking']);
-            (function() {
-               var u='${MATOMO_URL}';
-               _paq.push(['setTrackerUrl', u+'/js/tracker.php']);
-               _paq.push(['setSiteId', '1']);
-               var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
-               g.async=true; g.src=u+'/js/tracker.php'; s.parentNode.insertBefore(g,s);
-            })();
-            `}
-         </script>
+         <script
+            key="matomo-analytics"
+            dangerouslySetInnerHTML={{
+               __html: `
+                  var _paq = window._paq = window._paq || [];
+                  /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
+                  _paq.push(['trackPageView']);
+                  _paq.push(['enableLinkTracking']);
+                  (function() {
+                     var u='${MATOMO_URL}';
+                     _paq.push(['setTrackerUrl', u+'/js/tracker.php']);
+                     _paq.push(['setSiteId', '1']);
+                     var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
+                     g.async=true; g.src=u+'/js/tracker.php'; s.parentNode.insertBefore(g,s);
+                  })();
+               `,
+            }}
+         />
       </Head>
    ) : null;
 }
--- a/frontend/components/analytics/Matomo.tsx
+++ b/frontend/components/analytics/Matomo.tsx
@@ -16,10 +16,10 @@ export function Matomo() {
             _paq.push(['enableLinkTracking']);
             (function() {
                var u='${MATOMO_URL}';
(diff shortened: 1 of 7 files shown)


masonmcelvain:
dev_block ✌🏻

masonmcelvain:
un_dev_block ✌🏻 Not sure how to test these previews, because they seem to be trying to talk to minerva.ifixit.com. @jarstelfox maybe you have some insight here?

image

sterlinghirsh:
CR :zap: FYI I also found this package https://github.com/SocialGouv/matomo-next. Not sure if it's necessary or not, but we'll have to validate that in QA. The preview should be talking to matomo.ubreakit.com right?

masonmcelvain:
Cool package, thanks for finding! Happy to install it if we think it's simpler, looks like it's only 37 kb.

> The preview should be talking to matomo.ubreakit.com right?

It depends which env vars the previews are using. If they use the Prod env vars, they'll try to talk to minerva, not ubreakit.

masonmcelvain:
> Not sure how to test these previews

I also forgot that you need to disable ad blockers when testing analytics, I'll add that to QA steps

deltuh-vee:
dev_block ✊ 
Both previews are reporting to minerva.ifixit.com, but I didn't see any events show up. I took the POST request to the tracker API and modified it to send to matomo.ubreakit.com so that I could find the event a little more easily, but still wasn't able to find the event.

jarstelfox:
> Not sure how to test these previews, because they seem to be trying to talk to minerva.ifixit.com.

I found this on https://github.com/iFixit/react-commerce/pull/459

It looks like the E2E Tests / vercel are using the production `.env`. I made a small change for the `NODE_ENV` so that tests were their own environment

djmetzle (approved):
Nice! CR :footprints:

jarstelfox:
QA 🌵 

I ran a prod build locally and received the prod URL:

image

Running the dev build produces `https://matomo.ubreakit.com`

The vercel preview shows the correct URL:

image

frontend/components/analytics/Matomo.tsx
```diff
@@ -0,0 +1,27 @@
...
+                  var _paq = window._paq = window._paq || [];
+                  /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
+                  _paq.push(['trackPageView']);
+                  _paq.push(['enableLinkTracking']);
+                  (function() {
+                     var u='${MATOMO_URL}';
+                     _paq.push(['setTrackerUrl', u+'/js/tracker.php']);
+                     _paq.push(['setSiteId', '1']);
+                     var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
+                     g.async=true; g.src=u+'/js/tracker.php'; s.parentNode.insertBefore(g,s);
```
djmetzle, line 20: Quick check:

In production, we lightly obfuscate the tracker path (to work around basic adblockers):
https://github.com/iFixit/ifixit/blob/95a280d5a046342b69196ae04f6c082643ef3200/iFixit/Service/Matomo/MatomoPageComponent.php#L72

djmetzle: @jarstelfox i think we made `minerva.php` an alias, and `tracker.js` should still work. But do we want to keep config parity there? :thinking:

masonmcelvain:
@jarstelfox That QA makes sense to me! We want to see `minerva` when using prod env vars, and `matomo` when we use dev env vars, and it looks like you saw that.

frontend/components/analytics/Matomo.tsx
jarstelfox: Yea that would be nice to add, I can re-qa if needed

masonmcelvain: Can do. Will `matomo.ubreakit.com/minerva.php` work? Or should this obfuscation only be for `minerva.ifixit.com`?

djmetzle: Both work, right?
https://matomo.ubreakit.com/minerva.php

masonmcelvain: Didn't even think to follow the link myself 😅 thanks!

masonmcelvain:
un_deploy_block ✌🏻

jarstelfox:
QA 🌵 

image

jarstelfox (approved):
CR 🌵

PR #475, opened 2022-07-22
Googlebot: Disallow crawling

sumukhipandey:
Connects https://github.com/iFixit/ifixit/issues/43878

qa_req 0

CC: @sterlinghirsh

### DIFF
--- a/backend/public/robots.txt
+++ b/backend/public/robots.txt
@@ -1,3 +1,8 @@
 # To prevent search engines from seeing the site altogether, uncomment the next two lines:
 # User-Agent: *
-# Disallow: /
+Disallow: /Parts*?*filter
+Disallow: /Parts*?*q=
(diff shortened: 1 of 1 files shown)


sterlinghirsh (approved):

sterlinghirsh:
CR :zap: dev_block :zap: This isn't wrong per-se, but we need to change this in ifixit/ifixit since robots.txt will be served from there. Also I'm not sure, but we might need the `user-agent: *` line too.

sumukhipandey:
> CR ⚡ dev_block ⚡ This isn't wrong per-se, but we need to change this in ifixit/ifixit since robots.txt will be served from there. Also I'm not sure, but we might need the `user-agent: *` line too.

Done here https://github.com/iFixit/ifixit/pull/43939

PR #476, opened 2022-07-23
Facets: Convert most to single select

### LINKED ISSUE #374: Make most facets single-select instead of checkboxes
It's rare that we want to select products that match multiple values of the same facet. When someone selects a facet value, hide the list of other values.

masonmcelvain:
Makes all displayed facets single selects, except for Price Range and Capacity.

## QA

Make sure faceting still functions properly, on desktop and mobile. Make sure the proper links are still rendered for each Item Type facet option. Thanks!

@sterlinghirsh @sumukhipandey 

Closes #374


masonmcelvain:
dev_block ✌🏻 Just remembered we'll have to change the item type hook to not care which page we're on now. Before, it returned undefined if we weren't on the device parts page

masonmcelvain:
un_dev_block ✌🏻

masonmcelvain:
That's probably a legit failure, though we might just need to update the test with the new behavior

jordycosta:
dev_block 🟥 

Getting an undefined pathway when navigating Item Type facet options

> https://react-commerce-prod-git-facets-convert-some-to-si-f0b6fb-ifixit.vercel.app/Parts/undefined/Accessories

frontend/components/product-list/sections/FilterableProductsSection/RefinementSingleSelect.tsx
```diff
@@ -68,66 +64,78 @@ export function RefinementMenu({
...
+            attribute,
+            type: 'disjunctive',
+            value: item.value,
+            label: item.label,
+         })
+      );
+      // The url created by InstantSearch doesn't have the correct item type slug.
+      const path = url.pathname.split('/').filter((part) => part !== '');
+      const itemTypeHandle = encodeDeviceItemType(item.value);
+      const href = `${url.origin}/${path[0]}/${path[1]}/${itemTypeHandle}${url.search}`;
```
sterlinghirsh, line 119: We got rid of the origin in another pull since relative urls will work fine here and it prevents a mismatch between server and client rendering.
```suggestion
      const href = `/${path[0]}/${path[1]}/${itemTypeHandle}${url.search}`;
```

frontend/helpers/product-list-helpers.ts
```diff
@@ -102,21 +102,14 @@ export function getProductListTitle(
...
-               return RefinementDisplayType.SingleSelect;
-            default:
-               return RefinementDisplayType.MultiSelect;
-         }
-      }
-      default: {
-         return RefinementDisplayType.MultiSelect;
-      }
-   }
+const refinementDisplayTypeMap: Partial<Record<string, RefinementDisplayType>> =
```
sterlinghirsh, line 105: Why is this partial?

masonmcelvain: No good reason. I think it was an artifact of me playing around with the shape of this data structure.

masonmcelvain: https://github.com/iFixit/react-commerce/pull/476/commits/7bf04ec8a13081c784bd5078ab709fff6b4b39c8

jordycosta:
QA 🟢 

All facet options are single select aside from Capacity and Price Range ✔️ 



---

deploy_block 🟨 

**Some questions:**

I noticed that deselecting a facet in Item Types removes the bold property in the quantity shown to the right, but keeps the filter on for the user to clear it. Is this expected?

Should the number of facets selected be shown for single-select categories? _(since it will always be 1)_

single_select

masonmcelvain:
> I noticed that deselecting a facet in Item Types removes the bold property in the quantity shown to the right, but keeps the filter on for the user to clear it. Is this expected?

That's a bug, nice catch!

> Should the number of facets selected be shown for single-select categories? (since it will always be 1)

Yea, it definitely seems a little unnecessary. On mobile, however, there is no indicator that the facet has a selection without that (1) bubble. I think we'll leave as-is in this pull, and we can come up with a better design in another issue.

masonmcelvain:
dev_block ✌🏻 un_deploy_block ✌🏻

masonmcelvain:
un_dev_block ✌🏻 Clicking a selected facet option now deselects it visually _and_ removes it from the search.

jordycosta:
QA 🟢 


All facet options are single select aside from Capacity and Price Range ✔️ 

All **Item Types** facet options show proper links 👍 
> i.e. `?filter[facet_tags.Item Type][0]=Case Components`

Clicking a selected facet option now deselects it visually and removes it from the search. ✅

masonmcelvain:
@jordycosta I made an issue about the (1) bubble on single selects: https://github.com/iFixit/react-commerce/issues/522

sterlinghirsh (approved):

sterlinghirsh:
CR :zap:

PR #478, opened 2022-07-25
Facets: Fix item type urls

masonmcelvain:
I noticed that the item type facet urls were linking to the wrong place, typically the current window.location.href.

image

## Fix

The item type urls were previously all the same as window.location.href,
indicating that something wasn't working right with InstantSearch's
`createURL()`. A quick solution is to just add the correct item type
slug ourselves.

## QA

Make sure the item type urls always link to the right place now.

## Background

Bug introduced in #431, thought I could have sworn they were linking to the right place in that pull.

CC @sterlinghirsh @erinemay

### DIFF
--- a/frontend/components/product-list/sections/FilterableProductsSection/RefinementMenu.tsx
+++ b/frontend/components/product-list/sections/FilterableProductsSection/RefinementMenu.tsx
@@ -12,6 +12,7 @@ import { useSortBy } from './useSortBy';
 import { useFilteredRefinementList } from './useFilteredRefinementList';
 import { ProductList } from '@models/product-list';
 import { useDevicePartsItemType } from './useDevicePartsItemType';
+import { encodeDeviceItemType } from '@helpers/product-list-helpers';
 
 export type RefinementMenuProps = UseRefinementListProps & {
    productList: ProductList;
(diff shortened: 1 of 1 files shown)


sumukhipandey:
CR 🦋 Thanks for fixing this.

erinemay:
QA 🤦 
Hovering down the entire list of item type facets now shows the correct URL for each item type. 
image

Checked on a couple collection pages.
Confirmed actual links and pages linked to remain correct.

frontend/components/product-list/sections/FilterableProductsSection/RefinementMenu.tsx
```diff
@@ -90,12 +91,18 @@ const MenuItem = React.memo(function RefinementListItem({
...
+         attribute,
+         type: 'disjunctive',
+         value,
+         label,
+      })
+   );
+   // The url created by InstantSearch doesn't have the correct item type slug.
+   const path = url.pathname.split('/').filter((part) => part !== '');
+   const itemTypeHandle = encodeDeviceItemType(value);
+   const href = `${url.origin}/${path[0]}/${path[1]}/${itemTypeHandle}${url.search}`;
```
sterlinghirsh, line 105: We only want to build it this way if "value" is the itemType, which means this assumes that the only RefinementMenu is the ItemType one. Might be okay for now, but I thought I saw another pull changing that too.

sterlinghirsh (approved):

frontend/components/product-list/sections/FilterableProductsSection/RefinementMenu.tsx
masonmcelvain: Yea, we'll handle that when we start reusing this component for different facet types

PR #479, opened 2022-07-25
Fix broken fast refresh

dhmacs:
#348 broke [Fast Refresh](https://nextjs.org/docs/basic-features/fast-refresh), so here I'm proposing to drop `withUniversalOptimization` to fix it. This [comment](https://github.com/vercel/next.js/issues/8956#issuecomment-557871585) from Next.js core member says:

```
Minification is disabled by default for serverless bundles as it significantly increases 
build times and doesn't give any significant benefits when deploying serverless bundles.
```

@andyg0808 do you have better insights about this? 

## Changelog

1. Fix fast refresh
2. Update readme to reflect new dependencies install requirements

## QA

1. Clone the process locally
3. Install dependencies
4. Run dev server
5. Try to change a component and verify that you can see changes applied in the browser almost instantly (less than 1 sec)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,5 @@ tmp
 **/.rush/temp
 
 tsconfig.tsbuildinfo
+
+.env.local
--- a/README.md
+++ b/README.md
@@ -40,6 +40,8 @@ Here's one way you can get all the right versions installed and setup:
 
 This command will install both backend and frontend dependencies:
 
+Copy `.env.local.example` to `.env.local` and fill in the values.
+
 ```sh
 pnpm install:all
 ```
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -82,21 +82,9 @@ const moduleExports = {
    }),
 };
 
-function withUniversalOptimization(plugins) {
-   const configureWebpack = plugins.webpack || ((config, info) => config);
-   return {
-      ...plugins,
-      webpack(config, info) {
-         config = configureWebpack(config, info);
-         config.optimization.minimize = true;
-         return config;
-      },
-   };
-}
-
 // Make sure adding Sentry options is the last code to run before exporting, to
 // ensure that your source maps include changes from all other Webpack plugins
 module.exports = withSentryConfig(
-   withUniversalOptimization(withBundleAnalyzer(withTM(moduleExports))),
+   withBundleAnalyzer(withTM(moduleExports)),
    SENTRY_AUTH_TOKEN ? sentryWebpackPluginOptions : undefined
 );
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
       "node": ">= 16.14.2"
    },
    "scripts": {
-      "install:all": "export $(cat .env) && pnpm install && npm run install:backend",
+      "install:all": "export $(cat .env.local) && pnpm install && npm run install:backend",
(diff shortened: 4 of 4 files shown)


andyg0808:
CR :dog: through 5feefe
I'm fine with this, because I'm guessing it's holding up development, and it shouldn't currently be essential to performance. But I'd also like to know why it broke fast refresh eventually, because without it, the server-side bundle includes, like, _all_ the icons from the combined icon set we're using (it's like multiple megabytes of JS), and I'd rather drop the ones we don't need.

andyg0808 (approved):

deltuh-vee:
dev_block ✊ 
I'm unable to finish running `pnpm install:all`
CleanShot 2022-07-25 at 11 12 40

federicobadini:
> CR 🐶 through [5feefed](https://github.com/iFixit/react-commerce/commit/5feefed41d19583753bd747a1912daa8ba41b809) I'm fine with this, because I'm guessing it's holding up development, and it shouldn't currently be essential to performance. But I'd also like to know why it broke fast refresh eventually, because without it, the server-side bundle includes, like, _all_ the icons from the combined icon set we're using (it's like multiple megabytes of JS), and I'd rather drop the ones we don't need.

It was "breaking" fast refresh because it was applying minification even during development on every file save.

However Next v.12 has radically changed the way in which it deals with dependencies during compilation: 
the `target` option itself has been deprecated and a new [dependency tracing util](https://github.com/vercel/nft) has been introduced.

I'm confident that just upgrading to v.12 will resolve the problem that inspired #348 without any webpack tuning on our side.

Here some reference:

- [Target option deprecation](https://nextjs.org/docs/upgrading#target-option-deprecated)
- [New dependency tracing approach](https://nextjs.org/docs/advanced-features/output-file-tracing)

sterlinghirsh:
CR :zap: secondary un_dev_block :zap: @deltuh-vee I think you ran into a problem with missing an auth token for NPM. @danielcliu-ifixit just dealt with it so he should be able to help if you can't figure it out.

deltuh-vee:
>I'm confident that just upgrading to v.12 will resolve the problem that inspired https://github.com/iFixit/react-commerce/pull/348 without any webpack tuning on our side.

Do we still need this pull?

sterlinghirsh:
I think until we update to next 12 then yes

deltuh-vee:
QA 🎬 
Changes to components are instantly applied.

danielbeardsley:
If this was making development slow, why didn't we disable optimization in development?

> Minification is disabled by default for serverless bundles as it significantly increases 
build times and doesn't give any significant benefits when deploying serverless bundles.

I'm sure this core member was talking about *minification* (i.e. variable renaming, whitespace removal, ...), but tree-shaking is done too and we can easy measure its affect on bundle size.

andyg0808:
I believe I saw a _massive_ effect on bundle size: I think it went from ~2MB to a few hundred KB. I don't find the team member's argument convincing (I saw that post before I made the original change), but I also can't prove that tree shaking actually makes a meaningful difference on the server.

danielbeardsley:
> but I also can't prove that tree shaking actually makes a meaningful difference on the server.

We can prove that the bundle size decreases, but proving a reduction in cold-start-time is harder and definitely not a given.

Issue #489, opened 2022-07-25
Mini cart checkout button not working


erinemay:
When trying to check out from the mini cart on ifixit.com/Parts, the user is hit with an alert in the UI and unable to proceed.
View cart button works as expected.

Whether you are logged in or out, you get the same alert in the UI, but different errors:
![image](https://user-images.githubusercontent.com/72166715/180893052-dcabc35c-58fb-4872-a179-ca3c50389cab.png)

### Logged in
![image](https://user-images.githubusercontent.com/72166715/180893136-13867b90-1244-4414-9228-c595820c8f20.png)

### Logged out
![image](https://user-images.githubusercontent.com/72166715/180893084-5f494bf4-a5cb-4544-8cfe-5af684de266d.png)

### Repro steps

1. Log in, or stay logged out
2. Add something to your cart
3. Navigate to https://www.ifixit.com/Parts
4. Open the mini cart and click the check out button

CC @sterlinghirsh

dhmacs:
@erinemay @sterlinghirsh I think it's just a matter of setting the right storefront api access token in the production Strapi instance. Right now it's using `ifixit-test` token. I don't have access to production stores, so you have to do it

<img width="886" alt="image" src="https://user-images.githubusercontent.com/4640135/180980831-5feb1c9e-659a-4093-bf96-ffa760260505.png">

masonmcelvain:
I just saved the ifixit-us storefront access token to strapi.ifixit.com. The EU doesn't have the storefront api set up on their shopify store, but I think that's not a problem for now since they're not using this frontend.

masonmcelvain:
I'm not sure how environment variables work in vercel, but it seems like `NEXT_PUBLIC_IFIXIT_ORIGIN` might be undefined

![image](https://user-images.githubusercontent.com/52104630/181021088-b8b004a0-726d-48b4-9f1d-b97582d6fa8e.png)

masonmcelvain:
I also see that we're talking to cominor and govinor in `.env.production`, thought I'm not sure if Vercel is using that.

https://github.com/iFixit/react-commerce/blob/b86e929d4d1e9835899baa2f4c835027a92271b0/frontend/.env.production#L2-L3

I made a pull #493, if we want to talk to ifixit.com instead. Again, not sure if Vercel actually uses `.env.production` or if it has it's own concept of env vars.

dhmacs:
[...]

PR #493, opened 2022-07-26
Talk to ifixit.com in production

masonmcelvain:
We don't want to talk to cominor or govinor in production.

Not sure if Vercel actually uses .env.production or if it has it's own concept of env vars. If it doesn't use this fail, maybe we shouldn't merge this pull.

Connects #489


dhmacs:
This will make all Vercel previews use `ifixit.com` and `strapi.ifixit.com` by default, and I'm not sure that's what we want.

I think we should stick to set these env variable in Vercel dashboard, and only for the `production` environment.

deploy_block 🔢

masonmcelvain:
Great, let's do that instead. Thanks @dhmacs!

PR #496, opened 2022-07-26
Revert  #393: "iFixit Origin: use relative urls if on same origin"

masonmcelvain:
This is causing issues in production because we are unable to build
absolute urls with the ifixit origin. In production, we shouldn't have the CORS problem because we'll be on www.ifixit.com.

See https://github.com/iFixit/react-commerce/issues/489#issuecomment-1195598782

## QA

Smoke test that all basic features are working. We will probably see CORS errors in the previews as detailed in https://github.com/iFixit/ifixit/issues/43646

CC @danielbeardsley @dhmacs 

Connects #489

### DIFF
--- a/frontend/config/env.ts
+++ b/frontend/config/env.ts
@@ -8,14 +8,10 @@ export const ALGOLIA_API_KEY = checkEnv(
    'ALGOLIA_API_KEY'
 );
 
-const NEXT_PUBLIC_IFIXIT_ORIGIN = checkEnv(
+export const IFIXIT_ORIGIN = checkEnv(
    process.env.NEXT_PUBLIC_IFIXIT_ORIGIN,
    'NEXT_PUBLIC_IFIXIT_ORIGIN'
 );
-
-export const IFIXIT_ORIGIN = shouldUseRelativeOrigin(NEXT_PUBLIC_IFIXIT_ORIGIN)
-   ? ''
-   : NEXT_PUBLIC_IFIXIT_ORIGIN;
 
 export const STRAPI_ORIGIN = checkEnv(
    process.env.NEXT_PUBLIC_STRAPI_ORIGIN,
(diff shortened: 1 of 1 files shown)


davidrans (approved):
CR :coffee:

erinemay:
deploy_block 🌵 on some things. 

1. Losing our valuable prod preview
Is there a way around this fix? It looks like this sets the prod deployment to talk to development strapi making the prod and development links to basically identical pages. I really don't want to lose our ability to preview code changes with production data. 

2. Logout behavior off - just a development thing?
Logout behavior is a little off, but it's hard to tell if this is expected based on these changes. When I log out in production (ifixit.com/Parts), it leaves me on the same page. When I log out from https://react-commerce-5dsquxr17-ifixit.vercel.app/Parts, it redirects met to the base URL without /Parts (https://react-commerce-5dsquxr17-ifixit.vercel.app), which 404s.


Otherwise confirmed the following works:
- Checkout from both [prod](https://react-commerce-prod-kzz0rs6w3-ifixit.vercel.app/) and [development](https://react-commerce-5dsquxr17-ifixit.vercel.app/) mini carts both successfully create checkout on ifixit-test (store.cominor)
- [Prod](https://react-commerce-prod-kzz0rs6w3-ifixit.vercel.app/) and [development](https://react-commerce-5dsquxr17-ifixit.vercel.app/) previews both sync up with Cominor login state
- Navigating up and down the hierarchy is the same
- Item Type collection pages still work as expected
- Guide/Answers buttons link to appropriate pages

masonmcelvain:
Let's get the checkout link working for now, and I'll open another pull to use relative urls in some places, and always us absolute urls in others so that the prod preview works again.

un_deploy_block ✌🏻

erinemay:
@masonmcelvain That fixed the checkout button on the mini cart 🥳 . Do you want an issue to add back some relative URLs?

masonmcelvain:
Sure! I'm working on a fix right now

PR #497, opened 2022-07-26
getProductListPath: Use /Shop for marketing links

masonmcelvain:
This change is for links we generate on the frontend, but I don't think we actually generate any marketing links anywhere. I made this change for correctness sake, but I don't think we can QA because we don't currently generate marketing links.

qa_req 0

### DIFF
--- a/frontend/helpers/product-list-helpers.ts
+++ b/frontend/helpers/product-list-helpers.ts
@@ -79,7 +79,7 @@ export function getProductListPath(
          return `/Tools/${productList.handle}`;
       }
       case ProductListType.Marketing: {
-         return `/Store/${productList.handle}`;
+         return `/Shop/${productList.handle}`;
       }
       default: {
          throw new Error(`unknown product list type: ${productList.type}`);


sumukhipandey (commented):
CR 🦋

rickisXP:
What would be an example of a marketing link?

erinemay:
> What would be an example of a marketing link?

Marketing links are any collections that are set to type=marketing.
You can access strapi for this branch at https://fix-marketing-links.govinor.com/admin/auth/login with the standard email strapi@ifixit.com and password Password1 login credentials. 

(In general, you can get a branch's instance of strapi by replacing "main" in main.govinor.com with the branch name)

I don't think there are any collections actually set as marketing in Strapi, so you might have to set one. 

I just set this collection to type=marketing.
image

Now, `https://react-commerce-1r3obknuk-ifixit.vercel.app/Shop/acer-cb3-111-chromebook` works and the /Parts/ version 404's (`https://react-commerce-1r3obknuk-ifixit.vercel.app/Parts/acer-cb3-111-chromebook`)

@masonmcelvain, is there an expected change here, or is confirming that ⬆️ this all works what we want?

masonmcelvain:
Sorry I never circled back to add a description! Now that I think about it, I'm not sure there's anything to QA here. This change is for links we generate on the frontend, but I don't think we actually generate any marketing links anywhere. I made this change for correctness sake, I just noticed it was wrong. I think I'll qa_req 0 and we can move on

sumukhipandey (approved):
CR 🦋

PR #503, opened 2022-07-26
parts and tools:  Redirect to Parts and Tools

sumukhipandey:
Previously, `www.ifixit.com/parts` and `www.ifixit.cam/Parts` would load different pages. That shouldn't be the case and it was happening due to the lower case `parts`. This PR redirects `/parts` to `/Parts`. I noticed the same thing for `/tools` so I added a redirect for that as well [ `/tools` -> `/Tools`].

Closes https://github.com/ifixit/ifixit/issues/43952

CC: @sterlinghirsh

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -71,6 +71,12 @@ const moduleExports = {
             destination: `${process.env.NEXT_PUBLIC_IFIXIT_ORIGIN}/sitemap/marketing.xml`,
             permanent: true,
          },
+         {
+            source: '/parts*',
+            destination: '/Parts*',
+            permanent: true,
+         },
+
       ];
    },
    images: {
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -72,13 +72,13 @@ const moduleExports = {
             permanent: true,
          },
          {
-            source: '/parts*',
-            destination: '/Parts*',
+            source: '/parts/:path*',
+            destination: '/Parts/:path*',
             permanent: true,
          },
          {
-            source: '/tools*',
-            destination: '/Tools*',
+            source: '/tools/:path*',
+            destination: '/Tools/:path*',
             permanent: true,
          },
       ];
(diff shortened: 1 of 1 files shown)


andyg0808:
CR :dog: through 213112b

andyg0808:
Actually, I suspect this isn't going to help, unfortunately, because `/parts` and `/tools` don't get sent to react-commerce at all. I think we need to do the redirect on the PHP side, since that's what's going to serve these requests.

We might be able to do them via Walrus Map; that might be the best way.
See the Overlord docs at https://overlord.ifixit.com/Wiki/WalrusMap.

andyg0808:
dev_block :+1: on possibly doing this in ifixit/ifixit.

andyg0808:
Hm. Might make sense to keep this even if we need to add the redirect in the Walrus Map as well, because this'll apply in dev or if we ever pass all paths to NextJS.

sterlinghirsh:
CR :zap: un_dev_block :zap: @andyg0808 Like you said, at the very least it will prevent an annoying 404 when manually typing /parts or /tools.

sterlinghirsh (approved):

sterlinghirsh:
The test says it has a redirect loop.

masonmcelvain:
When I visit https://react-commerce.vercel.app/parts, the url is immediately changed to https://react-commerce.vercel.app/Parts, which is probably the cause of the infinite redirects. After a quick search I couldn't find docs on case sensitivity in Next.js routing, so I don't know the mechanism that causes this. Given the behavior though, I think we might not need to make this change. dev_block ✌🏻

sumukhipandey:
dev_block 🦋 on waiting to upgrade to next 12
https://github.com/iFixit/react-commerce/issues/511

sumukhipandey:
This issue is now resolved [iFixit/ifixit/pull/44095 and iFixit/ifixit/pull/44125] so I am closing this PR.

PR #505, opened 2022-07-27
InstantSearch: Render facets with URL from environment on the server

andyg0808:
We were having problems where the HTML page returned from the Vercel SSR included facet links pointing at the Vercel subdomain where it's actually hosted, instead of at `www.ifixit.com`. This switches things around to use the host from `NEXT_PUBLIC_IFIXIT_ORIGIN` when rendering on the server, which means that we're in full control of where the code thinks its origin is.

I didn't handle the case where `NEXT_PUBLIC_IFIXIT_ORIGIN` has a path on it; it doesn't seem currently relevant, and it's kinda a change from the previous behavior (which breaks up the `location` and uses it to generate the new URL; the path is handled separately).

## QA Notes
Check the page source on https://react-commerce-prod-git-instant-search-use-correc-eb6415-ifixit.vercel.app/Parts/Mac (use the `View Source` function in your browser; the inspector won't show what you need), search for `Adhesive` in the source, and verify that the `href` on the `a` tag you should find makes sense. If you're not sure, you should be able to try the same thing on https://react-commerce-prod.vercel.app/Parts/Mac and see how the output is different.

The thing you're searching for should look a bit like this:
image

I'm a little unsure about why the branch preview version seems to have `www.cominor.com` configured as the server URL for `react-commerce-prod`, but given that it also doesn't seem to have all the products I'd have expected, I'm guessing it's not using a real prod config, despite the name.

Fixes ifixit/ifixit#43961

### DIFF
--- a/frontend/components/common/InstantSearchProvider.tsx
+++ b/frontend/components/common/InstantSearchProvider.tsx
@@ -129,7 +129,7 @@ export function InstantSearchProvider({
             return window.location;
          },
          createURL({ qsModule, routeState, location }) {
-            const baseUrl = location.origin;
+            const baseUrl = getBaseOrigin(location)
             const pathParts = location.pathname
                .split('/')
                .filter((part) => part !== '');
@@ -260,4 +260,15 @@ function RefreshSearchResults({
       }
    }, [apiKey, prevApiKey]);
    return null;
+}
+
+function getBaseOrigin(location: Location): string {
+   if (typeof window === 'undefined' && process.env.NEXT_PUBLIC_IFIXIT_ORIGIN) {
+      // On the server, use the IFIXIT_ORIGIN url
+      // This ensures that the SSR produces the correct links on Vercel
+      // (where the Host header doesn't match the page URL.)
+      const publicOrigin = new URL(process.env.NEXT_PUBLIC_IFIXIT_ORIGIN)
+      return publicOrigin.origin
+   }
+   return location.origin
 }
--- a/frontend/components/product-list/sections/FilterableProductsSection/RefinementMenu.tsx
+++ b/frontend/components/product-list/sections/FilterableProductsSection/RefinementMenu.tsx
@@ -102,7 +102,7 @@ const MenuItem = React.memo(function RefinementListItem({
    // The url created by InstantSearch doesn't have the correct item type slug.
    const path = url.pathname.split('/').filter((part) => part !== '');
    const itemTypeHandle = encodeDeviceItemType(value);
-   const href = `${url.origin}/${path[0]}/${path[1]}/${itemTypeHandle}${url.search}`;
+   const href = `/${path[0]}/${path[1]}/${itemTypeHandle}${url.search}`;
    return (
       <HStack
          key={label}


frontend/components/common/InstantSearchProvider.tsx
```diff
@@ -129,7 +129,7 @@ export function InstantSearchProvider({
             return window.location;
          },
          createURL({ qsModule, routeState, location }) {
-            const baseUrl = location.origin;
```
sterlinghirsh, line 132: Can we omit baseUrl and let these be relative urls?

sterlinghirsh:
CR :zap: dev_block :zap: relative urls might be a simpler approach that also would solve the mismatch between server-side and client-side rendering on other domains like dev domains.

frontend/components/common/InstantSearchProvider.tsx
masonmcelvain: If we choose to make the url returned by this method relative, also consider that this might cause the urls created by calls to `createURL` elsewhere in the app (links below) might also become relative. Maybe that's fine, just wanted to point it out.

https://github.com/iFixit/react-commerce/blob/206ff8be780f671407ccf096500e21823d3be6b4/frontend/components/product-list/sections/FilterableProductsSection/RefinementMenu.tsx#L93-L101

https://github.com/iFixit/react-commerce/blob/959d07417c8016352bd507cb0e7b510028fb6a00/frontend/components/product-list/sections/FilterableProductsSection/Pagination.tsx#L33

andyg0808: I like this idea, but it looks trickier to implement than I'd like.

Looking at the code for those two places, they both rely on `URL` to parse apart the URL, in order to modify it. `URL` [does not support](https://github.com/whatwg/url/issues/531) relative URLs, so we would have to find an alternative technique. If I could be sure `createURL` would never return a search string, I could just make the assumption that we only ever get a path back from it (we're already ignoring fragments). But that seems like a risk longer-term, because it violates the implied interface of a method named `createURL`: the assumption is that we get a URL when calling it.

Further, the argument about the name `createURL` also implies that we should always return a complete URL from it.

Given all that, I think I'd like to keep `createURL` returning absolute URLs. What I wouldn't mind doing is dropping the `origin` part of the URL from the `RefinementMenu` use, which should result in relative URLs being rendered for the case of interest.

andyg0808: Talked to @sterlinghirsh, and decided on [9cdd74d](https://github.com/iFixit/react-commerce/pull/505/commits/9cdd74d680439be8b2392fde7fb259f4aaf59445).

andyg0808:
un_dev_block :+1: Using relative URLs now.

jordycosta:
QA 🟢

The `href` on the `a` tag makes sense 👍 

---

### Testing Procedure:
1. Prepend `view-source:` to [any](https://react-commerce-prod-git-instant-search-use-correc-eb6415-ifixit.vercel.app/Parts/Mac) collections page
2. Search the source for `Adhesives`
3. Observe that the `href` is correct

---

### Screen Capture:
a-class

sterlinghirsh:
CR :zap:

sterlinghirsh (approved):

PR #507, opened 2022-07-27
Add env variable for Algolia index name

dhmacs:
This PR will let us use different Algolia index for different environments

## Changelog

- Refactor algolia index name from being a const to be an env variable
- Override production `NEXT_PUBLIC_ALGOLIA_PRODUCT_INDEX_NAME` env variable to  point to `product_group_en`

## QA

1. Open Vercel preview
2. Visit parts product list
3. Open dev tools network tab
4. Trigger an Algolia search (e.g. by clicking on a filter)
5. Inspect the request and verify that it's using `dev_product_group_en` index
    image

### DIFF
--- a/frontend/.env.production
+++ b/frontend/.env.production
@@ -9,3 +9,6 @@ NEXT_PUBLIC_MATOMO_URL=https://minerva.ifixit.com
 NEXT_PUBLIC_GA_KEY=UA-30506-1
 NEXT_PUBLIC_GA_DEBUG=false
 NEXT_PUBLIC_GA_URL=https://www.google-analytics.com/analytics.js
+# This is being set to dev_product_group_en because we want all preview environments to use this by default.
+# We'll add an override in the Vercel dashboard to use the prod_product_group_en for the production environment.
+NEXT_PUBLIC_ALGOLIA_PRODUCT_INDEX_NAME=dev_product_group_en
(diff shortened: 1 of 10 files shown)


masonmcelvain (approved):
CR ✌🏻

frontend/.env.production
```diff
@@ -9,3 +9,6 @@ NEXT_PUBLIC_MATOMO_URL=https://minerva.ifixit.com
 NEXT_PUBLIC_GA_KEY=UA-30506-1
 NEXT_PUBLIC_GA_DEBUG=false
 NEXT_PUBLIC_GA_URL=https://www.google-analytics.com/analytics.js
+# This is being set to dev_product_group_en because we want all preview environments to use this by default.
+# We'll add an override in the Vercel dashboard to use the prod_product_group_en for the production environment.
+NEXT_PUBLIC_ALGOLIA_PRODUCT_INDEX_NAME=dev_product_group_en
```
danielbeardsley, line 14: The docs are unclear about when env vars need to be prefixed with `NEXT_PUBLIC`.

> This loads [environment variables] into the Node.js environment automatically allowing you to use them in [Next.js data fetching methods](https://nextjs.org/docs/basic-features/data-fetching/overview) and [API routes](https://nextjs.org/docs/api-routes/introduction).

Then later about vars with the prefix: 

> This loads [env vars with a `NEXT_PUBLIC_` prefix] into the Node.js environment automatically, allowing you to use it anywhere in your code.

I *think* they mean the substitution is only done in certain files or maybe methods? It's not clear cause there isn't a clear line between client vs server code.

https://nextjs.org/docs/basic-features/environment-variables#exposing-environment-variables-to-the-browser

dhmacs: @danielbeardsley I'm not sure I understand what is the question here. When you add `NEXT_PUBLIC` the environment variable is also inlined at build time in the browser bundle (beside being available server side as well, thanks to `process.env`). Effectively this make the env variable available in both environments.

PR #517, opened 2022-07-27
Routing: Don't 404 on item type visits

### LINKED ISSUE #516: Direct entry to item type product lists 404ing
Enter directly https://www.ifixit.com/Parts/iPhone_11/Adhesives and get a 404 with the following error:
image

Enter via the parent page (https://www.ifixit.com/Parts/iPhone_11) by selecting "Adhesives" in the UI, the page is displayed.

masonmcelvain:
Fixes a bug where a visit to /Parts/Device/Item_Type would 404. The item type slug is optional, but any extra path components after that should 404.

This bug was unintentionally introduced in https://github.com/iFixit/react-commerce/commit/c32b62fe8859fa04f68b4f6e855b0938120544fd (and I missed it in CR!).

Admittedly it was bad idea on my part to store the itemTypeHandle in an unused variable, and it got "cleaned up" in that commit. Now, we'll include it in the `...` and only 404 if there is > 1 extra path component.

## QA
Visit /Parts/iPhone_11/Adhesives and make sure it doesn't 404.

Closes #516

### DIFF
--- a/frontend/pages/Parts/[...deviceHandleItemType].tsx
+++ b/frontend/pages/Parts/[...deviceHandleItemType].tsx
@@ -31,12 +31,14 @@ export const getServerSideProps: GetServerSideProps<AppPageProps> = async (
    );
 
    const { deviceHandleItemType } = context.params || {};
-   const [deviceHandle, ...rest] = Array.isArray(deviceHandleItemType)
+   const [deviceHandle, ...itemTypeAndRest] = Array.isArray(
+      deviceHandleItemType
+   )
       ? deviceHandleItemType
       : [];
 
    invariant(typeof deviceHandle === 'string', 'device handle is required');
-   if (rest?.length > 0) {
+   if (itemTypeAndRest?.length > 1) {
       return {
          notFound: true,
       };


frontend/pages/Parts/[...deviceHandleItemType].tsx
```diff
@@ -31,12 +31,14 @@ export const getServerSideProps: GetServerSideProps<AppPageProps> = async (
    );
 
    const { deviceHandleItemType } = context.params || {};
-   const [deviceHandle, ...rest] = Array.isArray(deviceHandleItemType)
+   const [deviceHandle, ...itemTypeAndRest] = Array.isArray(
+      deviceHandleItemType
+   )
```
Comment (andyg0808) [inline frontend/pages/Parts/[...deviceHandleItemType].tsx:36]: This seems odd. `isArray` returns a boolean, not an array, if I'm reading [the docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray) right. So I'm not sure how unpacking that works.

andyg0808 (changes requested):

andyg0808:
dev_block :+1: on what I suspect is a bug.

frontend/pages/Parts/[...deviceHandleItemType].tsx
Comment (dhmacs) [reply frontend/pages/Parts/[...deviceHandleItemType].tsx:36]: @andyg0808 this condition is part of a ternary expression, so this instruction is actually returning an array 👍

Comment (andyg0808) [reply frontend/pages/Parts/[...deviceHandleItemType].tsx:36]: Oh, got it. Thanks.

andyg0808 (approved):

andyg0808:
CR :dog: through ca1e87b
un_dev_block :+1:

djmetzle (approved):
Thanks @masonmcelvain 
LGTM CR :link:

deltuh-vee:
QA 🎬 
Direct entry to Item Type pages no longer 404s

PR #520, opened 2022-07-28
Hide device children on unfiltered item type pages with no hits

### LINKED ISSUE #462: Hide children on device item type collections with no hits
Google will end up crawling a lot of unnecessary pages if we expose the entire tree for every item type, so when someone gets to a dead end, stop giving them options to keep going down the tree.

This might be a little tricky if we load the page with more filters than just device and item type. We don't want to hide if the user searches for something weird, for example.

masonmcelvain:
Hides the device children links on item type pages (like /Parts/iPhone/Batteries) when there are no product results and there are no other refinements applied. This prevents crawlers from continuing to descend down the device tree, when we know they will keep hitting empty pages.

## QA
- Find a device that has children device links on `main`. You might have to go into [strapi admin](https://hide-children-on-no-hits-item-type.govinor.com/admin) for this branch and set the `parent` field for some devices in order to see them.
- Visit an item type page for this device that has no results.
  - One way to do this: Visit the device parts page, like /Parts/iPhone. Type gibberish as the last component, like /Parts/iPhone/asdf.
- On main, the children links should still show on the empty item type page. On this feature, they shouldn't appear.
- In any other conditions, the children device links should display as usual.

CC @sterlinghirsh 

Closes #462

### DIFF
--- a/frontend/components/common/InstantSearchProvider.tsx
+++ b/frontend/components/common/InstantSearchProvider.tsx
@@ -129,7 +129,7 @@ export function InstantSearchProvider({
             return window.location;
          },
          createURL({ qsModule, routeState, location }) {
-            const baseUrl = getBaseOrigin(location)
+            const baseUrl = getBaseOrigin(location);
             const pathParts = location.pathname
                .split('/')
                .filter((part) => part !== '');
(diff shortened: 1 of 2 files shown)


masonmcelvain:
deploy_block ✌🏻 Do we want to disable the refinement search bar when there are no results to prevent the UX bug described in https://github.com/iFixit/react-commerce/pull/520/commits/9312648223c934c1f024e44a1faa2ef4bbebba35?

masonmcelvain:
un_deploy_block ✌🏻 Nevermind, that bug isn't actually happening in practice for me.

deltuh-vee:
QA 🎬 
Queries for non existent items with an Item Type filter set will remove the child device links.

deltuh-vee:
QA 🎬 
Child device links disappear when a search query for a non existent item is performed with an Item Type filter set.

sterlinghirsh (approved):

sterlinghirsh:
CR :zap:

Issue #525, opened 2022-07-28
Can We Parallelize Calls to Strapi and iFixit API 


danielcliu-ifixit:
We sometimes make sequential calls to Strapi and iFixit API, when ideally we wouldn't have to wait for one to return if we need to ask for both.

PR #544, opened 2022-07-29
Have findProductList() Make Async Calls in Parallel.

danielcliu-ifixit:
## Overview
Before, we had a call to strapi and a conditional call to iFixit API afterwards. This makes it so the two requests are done at the same time so we can cut out the overhead of the faster request. This should make the two calls take less time overall as one is not conditional on the other and they are practically always both called anyway.

## CR
Pretty simple

## QA
Do product lists still load? Are they any faster? I've found a slight improvement when running locally (~60 ms speed up) but couldn't find that same performance gain in lighthouse for whatever reason.

Closes: #525

### DIFF
--- a/frontend/lib/ifixit-api/devices.ts
+++ b/frontend/lib/ifixit-api/devices.ts
@@ -7,6 +7,9 @@ export async function fetchDeviceWiki(
 ): Promise<DeviceWiki | null> {
    const deviceHandle = getDeviceHandle(deviceTitle);
    try {
+      if(!deviceHandle) {
+         throw new Error(`deviceHandle cannot be a blank string!`);
+      }
       const response = await fetch(
          `${IFIXIT_ORIGIN}/api/2.0/cart/part_collections/devices/${deviceHandle}`,
          {
--- a/frontend/lib/ifixit-api/devices.ts
+++ b/frontend/lib/ifixit-api/devices.ts
@@ -1,4 +1,5 @@
 import { IFIXIT_ORIGIN } from '@config/env';
+import { invariant } from '@ifixit/helpers';
 
 export type DeviceWiki = Record<string, any>;
 
@@ -7,9 +8,7 @@ export async function fetchDeviceWiki(
 ): Promise<DeviceWiki | null> {
    const deviceHandle = getDeviceHandle(deviceTitle);
    try {
-      if(!deviceHandle) {
-         throw new Error(`deviceHandle cannot be a blank string!`);
-      }
+      invariant(deviceHandle.length > 0, "deviceHandle cannot be a blank string");
       const response = await fetch(
          `${IFIXIT_ORIGIN}/api/2.0/cart/part_collections/devices/${deviceHandle}`,
          {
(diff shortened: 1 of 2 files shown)


frontend/lib/ifixit-api/devices.ts
```diff
@@ -7,6 +7,9 @@ export async function fetchDeviceWiki(
 ): Promise<DeviceWiki | null> {
    const deviceHandle = getDeviceHandle(deviceTitle);
    try {
+      if(!deviceHandle) {
+         throw new Error(`deviceHandle cannot be a blank string!`);
+      }
```
dhmacs, line 12:
```suggestion
     invariant(deviceHandle.length > 0, "deviceHandle cannot be a blank string");
```

masonmcelvain: Have you seen cases where this is called with a blank string @danielcliu-ifixit?

frontend/models/product-list/index.ts
```diff
@@ -47,16 +47,17 @@ export type {
 export async function findProductList(
    filters: ProductListFiltersInput
 ): Promise<ProductList | null> {
-   const result = await strapi.getProductList({
-      filters,
-   });
+   const deviceTitle = filters.deviceTitle?.eq ?? '';
```
masonmcelvain, line 50: This seems like a change in behavior. We never want to use `productList.deviceTitle` anymore?

frontend/models/product-list/index.ts
```diff
@@ -47,16 +47,17 @@ export type {
...
 
+   const [result, deviceWiki] =
+      await Promise.all([
+         strapi.getProductList({
+            filters,
+         }),
+         fetchDeviceWiki(deviceTitle)
+      ]);
```
masonmcelvain, line 58: I like how succinct this is. I would personally avoid calling `fetchDeviceWiki(deviceTitle)` if `deviceTitle === ''`, but the result is the same so I'm fine with it either way.

```suggestion
   const [result, deviceWiki] =
      await Promise.all([
         strapi.getProductList({
            filters,
         }),
         deviceTitle ? fetchDeviceWiki(deviceTitle) : null,
      ]);
```

masonmcelvain (changes requested):
CR ✌🏻 with a question about `productList.deviceTitle`

k0rvusk0r4x:
- I'm seeing an improvement in speed-based metrics (time-to xxx) and in the Speed Index specifically. 
- The page load is working without issue.

QA 😺

frontend/models/product-list/index.ts
danielcliu-ifixit: Apparently the strapi api is case-sensitive so I'm not worried about the two being different. I don't think they'll be a behavior change

frontend/models/product-list/index.ts
danielcliu-ifixit: I like handling it in the function so if anyone uses the function in future they don't have to account for their input being `== null`

masonmcelvain (approved):
CR ✌🏻

deltuh-vee:
Carry Over QA 🎬

Issue #103, opened 2021-10-13
Add OpenGraph tags


dhmacs:
Although they don't offer any benefit to SEO ranking, OpenGraph tags are still recommended as people may share content on social media.

## To do

- [ ] Add of tags
    ```tsx
    <Head>
      <meta 
        property="og:title" 
        content="Social Title for Cool Page" 
      />
      <meta 
        property="og:description" 
        content="And a social description for our cool page" 
      />
      <meta
        property="og:image"
        content="https://example.com/images/cool-page.jpg"
      />
    </Head>
    ```

PR #545, opened 2022-07-29
Collections: Add meta title to strapi

### LINKED ISSUE #539: Collections: Meta Title field in Strapi
I want the ability to write a custom meta title in Strapi for any given page. I assume it will be difficult or impossible to allow for this on Item Type pages currently, so for now, I expect this would only exist on the unfiltered Parts page. 

Here is a use case in case it matters:
The DualShock 4 is technically what we call the Collection, and it is the correct name, but people 10x more commonly search "ps4 controller parts" - I'd like to write a custom meta title to optimize for both DualShock 4 and PS4

[This is a related Issue](https://github.com/iFixit/ifixit/issues/39157) where we did this for Product Pages in Akeneo in March this year.

masonmcelvain:
- Adds a new field, `metaTitle` to Strapi. It has a 130 character limit based on the spec in this issue: https://github.com/iFixit/ifixit/issues/39157
- Uses the `metaTitle` field in the `<title>` of the `<head>`, and nowhere else on the page.

## QA
Pages with a value in `metaTitle` should have the value in the `<title>` tag, otherwise it should default to the normal device title. You can edit the `metaTitle` in Strapi here: https://collections--add-meta-title-to-strapi.govinor.com/admin

CC @ifixitmaxb

Closes #539


masonmcelvain:
This implementation can cause the browser tab to have a different title than the h1 title of the page. Not sure if that's important or not.

image

erinemay:
QA 🦖 
- New metaTitle field is present in Strapi
- No issues or errors when saving that value
- The value entered into that field is now set as the `<title>` in `<head>`
image
- Adding a metaTitle value for one collection doesn't affect the title for other collections



deploy_block 🧷 
1. Further discussion on the change in browser tab titles. It's a little weird to see a different title on the page than on the browser tab. I imagine if they're related and sensibly used, this shouldn't be an issue. It looks like this behavior is present for the product page meta titles. 
2. What's this property? It's also getting set as the metaTitle text
image

masonmcelvain:
It looks like it's an OpenGraph tag, they were added in this issue: https://github.com/iFixit/react-commerce/issues/103

> Although they don't offer any benefit to SEO ranking, OpenGraph tags are still recommended as people may share content on social media.

I don't see any harm in changing the title of this tag too, I think it just gets used in social media previews. If @ifixitmaxb plans to use the colloquial version of product names in `metaTitle`, then maybe it's a good thing.

ifixitmaxb:
I think it's good/fine for the OG tag title to follow the meta title - probably ideal honestly.

sterlinghirsh (approved):

sterlinghirsh:
Deploy this carefully since live strapi won't be updated automatically. We might have to revert if it breaks.

sterlinghirsh:
CR :zap:

masonmcelvain:
un_deploy_block ✌🏻 Doesn't sound like the mismatch between the h1 page title and the browser tab title will cause an issue assuming we use reasonable `metaTitle` values.

PR #527, opened 2022-07-28
Add item type to canonical

sterlinghirsh:
Closes #519

Issue #526, opened 2022-07-28
Default results appear when switching single-select facets


jordycosta:
### Overview:

There is a brief flash of the no-filtered results when switching from one single-select facet to another

![search](https://user-images.githubusercontent.com/95656772/181639990-0d789021-d3fc-4edd-adb6-d5290e0f2147.gif)

https://react-commerce-ifixit.vercel.app/Parts
https://www.ifixit.com/Parts

sumukhipandey:
This is probably happening [here ](https://github.com/iFixit/react-commerce/blob/dc15c19fbd5d0ed3a08ef263cd4acdbf56eb2653/frontend/components/product-list/sections/FilterableProductsSection/RefinementSingleSelect.tsx#L99). I think the quick clearing of refinements causes the default page to appear briefly.

PR #546, opened 2022-07-29
Fix SSR for item type product lists

### LINKED ISSUE #531: Fix SSR for item type product lists
In #527 we fixed canonical for item type, but we didn't fix the bug issue entirely, since the item type product list is still rendered client side (e.g. we render the parent device title).

## Possible solution

Anything that depends on `useDevicePartsItemType` should be using `productList.deviceItemType` instead

cc @sterlinghirsh

masonmcelvain:
Modifies the `useDevicePartsItemType()` hook to use the item type from the slug on the server.

## Background

Algolia's `useCurrentRefinements()` hook doesn't include the 'Item Type' filter on page load, even when we visit an item type page like /Parts/iPhone/Batteries. This is unexpected because we explicitly add the 'Item Type' to the current UIState in the InstantSearchProvider. This approach is a workaround for this unexpected behavior.

Uses the fix in https://github.com/iFixit/react-commerce/pull/527 everywhere else.

## QA

A smoke test of the app. Notably, when you visit item type pages directly, you shouldn't see a flash of unfiltered results before the page finishes loading and you shouldn't see the "Parts" version of the title flash.

CC @sterlinghirsh @dhmacs 

Closes #531

### DIFF
--- a/frontend/components/product-list/MetaTags.tsx
+++ b/frontend/components/product-list/MetaTags.tsx
@@ -34,7 +34,7 @@ export function MetaTags({ productList }: MetaTagsProps) {
       title += ` - Page ${page}`;
    }
    title += ' | iFixit';
-   const itemTypeHandle = `/${encodeDeviceItemType(itemType ?? '')}`;
+   const itemTypeHandle = itemType ? `/${encodeDeviceItemType(itemType)}` : '';
    const canonicalUrl = `${appContext.ifixitOrigin}${
       productList.path
    }${itemTypeHandle}${page > 1 ? `?${PRODUCT_LIST_PAGE_PARAM}=${page}` : ''}`;
--- a/frontend/components/product-list/MetaTags.tsx
+++ b/frontend/components/product-list/MetaTags.tsx
@@ -28,18 +28,13 @@ export function MetaTags({ productList }: MetaTagsProps) {
       refinementAttributes.length === 1 &&
       refinementAttributes[0] === 'facet_tags.Item Type';
    const isFiltered = currentRefinements.items.length > 0 && !isItemTypeFilter;
-   // Use the original device item type on the server.
-   const algoliaDeviceItemType = useDevicePartsItemType(productList);
-   const itemType =
-      typeof window === 'undefined' && productList.deviceItemType
-         ? encodeDeviceItemType(productList.deviceItemType)
-         : encodeDeviceItemType(algoliaDeviceItemType ?? '');
+   const itemType = useDevicePartsItemType(productList);
    let title = getProductListTitle(productList, itemType);
    if (!isFiltered && page > 1) {
       title += ` - Page ${page}`;
    }
    title += ' | iFixit';
-   const itemTypeHandle = itemType ? `/${itemType}` : '';
+   const itemTypeHandle = `/${encodeDeviceItemType(itemType ?? '')}`;
    const canonicalUrl = `${appContext.ifixitOrigin}${
       productList.path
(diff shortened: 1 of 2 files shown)


masonmcelvain:
> Anything that depends on useDevicePartsItemType should be using productList.deviceItemType instead

@dhmacs Is this what you had in mind? Wasn't sure if you meant on the server only, or in all cases.

frontend/components/product-list/MetaTags.tsx
```diff
@@ -28,18 +28,13 @@ export function MetaTags({ productList }: MetaTagsProps) {
...
-         ? encodeDeviceItemType(productList.deviceItemType)
-         : encodeDeviceItemType(algoliaDeviceItemType ?? '');
+   const itemType = useDevicePartsItemType(productList);
    let title = getProductListTitle(productList, itemType);
    if (!isFiltered && page > 1) {
       title += ` - Page ${page}`;
    }
    title += ' | iFixit';
-   const itemTypeHandle = itemType ? `/${itemType}` : '';
+   const itemTypeHandle = `/${encodeDeviceItemType(itemType ?? '')}`;
```
sterlinghirsh, line 37: This might have subtly changed behavior to add a trailing slash when itemType is empty. When itemType was empty before, itemTypeHandle would be `''`, but now it's `'/'`.

sterlinghirsh, line 37: I think this will work but not sure if TS will approve this.

```suggestion
   const itemTypeHandle = itemType ? `/${encodeDeviceItemType(itemType)}` : '';
```

sterlinghirsh:
CR :zap: dev_block :zap: on a minor change

masonmcelvain:
un_dev_block ✌🏻

sterlinghirsh (approved):

sterlinghirsh:
CR :zap:

sterlinghirsh:
I tried to QA this, but I don't know what should be different from main. I didn't see a flash of wrong content or anything, since I think those types of bugs have already been fixed elsewhere. Am I missing something?

masonmcelvain:
> I tried to QA this, but I don't know what should be different from main. I didn't see a flash of wrong content or anything, since I think those types of bugs have already been fixed elsewhere. Am I missing something?

I'm still seeing the wrong title flash on `main` (also see the wrong breadcrumbs, applied facets, etc.). I think this change should fix that bug, and anywhere else where the item type is used in the UI.

https://user-images.githubusercontent.com/52104630/182042890-d7c30664-795e-4d40-bad1-37e3e1513f90.mp4

erinemay:
@jordycosta was also still seeing the flash of incorrect title https://github.com/iFixit/react-commerce/issues/526 as well. He could probably confirm that it's working here.

jordycosta:
The flash of incorrect title(s) and breadcrumb(s) is now resolved 👍  

However, I'm still seeing a brief flash of default/non-filtered results when switching from one facet to another. I noticed that this doesn't occur if you have already used one of the single-select facets, so it seems that if the content has loaded prior, it won't show the non-filtered results.

firefox_Pfx5bKQaZJ

Looking at this screen capture, we can see that it doesn't display any non-filtered results for single-select facets that I have already select _(e.g. Batteries -> Cables)_; but when I select a facet that I haven't loaded prior _(e.g. Cameras)_, it will display the non-filtered results briefly.

masonmcelvain:
Let's deal with #526 in a separate PR, as it's a little more complex.

> The flash of incorrect title(s) and breadcrumb(s) is now resolved

Thanks @jordycosta! I think that's QA for #531. un_dev_block ✌🏻

sterlinghirsh:
CR :zap:

PR #550, opened 2022-07-29
Strapi: Add custom meta title field

masonmcelvain:
https://github.com/iFixit/react-commerce/pull/545 failed because we started querying for this field before strapi got redeployed. Let's deploy Strapi first, then modify our queries in another pull.

## QA

Make sure `metaTitle` shows up in https://add-custom-meta-title-on-strapi-ONLY.govinor.com/admin

## On Deploy

Redeploy Strapi

CC @sterlinghirsh 

Connects #539

PR #551, opened 2022-07-29
Use `metaTitle` from Strapi

### LINKED ISSUE #539: Collections: Meta Title field in Strapi

masonmcelvain:
Uses the `metaTitle` field in the `<title>` of the `<head>`, and nowhere else on the page. This field was added to Strapi in https://github.com/iFixit/react-commerce/pull/550.

## QA
Pages with a value in `metaTitle` should have the value in the `<title>` tag, otherwise it should default to the normal device title. You can edit the `metaTitle` in Strapi here: https://collections--add-meta-title-to-strapi.govinor.com/admin

Closes #539

### DIFF
--- a/frontend/components/product-list/MetaTags.tsx
+++ b/frontend/components/product-list/MetaTags.tsx
@@ -30,7 +30,8 @@ export function MetaTags({ productList }: MetaTagsProps) {
       refinementAttributes[0] === 'facet_tags.Item Type';
    const isFiltered = currentRefinements.items.length > 0 && !isItemTypeFilter;
    const itemType = useDevicePartsItemType(productList);
-   let title = getProductListTitle(productList, itemType);
+   let title =
+      productList.metaTitle || getProductListTitle(productList, itemType);
    if (!isFiltered && page > 1) {
       title += ` - Page ${page}`;
    }
--- a/frontend/lib/strapi-sdk/generated/sdk.ts
+++ b/frontend/lib/strapi-sdk/generated/sdk.ts
@@ -1512,6 +1512,7 @@ export type GetProductListQuery = {
             tagline?: Maybe<string>;
             description: string;
             metaDescription?: Maybe<string>;
+            metaTitle?: Maybe<string>;
             filters?: Maybe<string>;
             childrenHeading?: Maybe<string>;
             image?: Maybe<{
@@ -2669,6 +2670,7 @@ export const GetProductListDocument = `
         tagline
         description
         metaDescription
+        metaTitle
         filters
         image {
           data {
--- a/frontend/lib/strapi-sdk/operations/getProductList.graphql
+++ b/frontend/lib/strapi-sdk/operations/getProductList.graphql
@@ -14,6 +14,7 @@ query getProductList($filters: ProductListFiltersInput) {
             tagline
             description
             metaDescription
+            metaTitle
             filters
             image {
                data {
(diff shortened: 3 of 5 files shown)


hcutrone (approved):
CR :burrito: 
Makes sense to me, but gonna cr_req 2 since I'm not familiar with Strapi :+1:

davidrans (dismissed):
CR :coffee:

sterlinghirsh:
CR :zap:

sterlinghirsh (approved):

erinemay:
QA 🍓 
- metaTitle value is now set as the `<title>` in `<head>`
- metaTitle value is [set as the OG:title tag](https://github.com/iFixit/react-commerce/pull/545#issuecomment-1199847866)
- Parent and children `<title>` in `<head>` remain unchanged
- The metaTitle doesn't appear anywhere else on the page (only present in tab title)
- No new console errors with usage

deploy_block 🌵 @masonmcelvain, I'm seeing a caching bug, but I'm not sure how caching here relates to production caching behavior. 

After changing the metaTitle twice for [iPhone 13 Pro Parts in Strapi](https://use-meta-title.govinor.com/admin/content-manager/collectionType/api::product-list.product-list/103?plugins[i18n][locale]=en), I'm still seeing the first one persist when I go to the page, even in incognito. Only after using `?emptyCache`, do I see the new change.

sterlinghirsh:
I don't think emptyCache does what you want on nextjs, but we should be able to skip some caching at least with ?_vercel_no_cache=1

davidrans (approved):
CR :coffee:

masonmcelvain:
Good catch on the caching @erinemay. Given that the cache is only fresh for 10 minutes and served if stale for up to 20 minutes (see #562), I think the behavior you're seeing is probably ok. Part of the point of caching is so next.js doesn't have to talk to Strapi, so if any field is updated in Strapi, next.js won't know about it until the cache becomes stale. Maybe we could set something up where Strapi can tell next.js to invalidate the cache when a field changes (idk how easy or difficult that would be to implement), but something like that should be a separate issue.

un_deploy_block ✌🏻

dhmacs:
> Good catch on the caching @erinemay. Given that the cache is only fresh for 10 minutes and served if stale for up to 20 minutes (see #562), I think the behavior you're seeing is probably ok. Part of the point of caching is so next.js doesn't have to talk to Strapi, so if any field is updated in Strapi, next.js won't know about it until the cache becomes stale. Maybe we could set something up where Strapi can tell next.js to invalidate the cache when a field changes (idk how easy or difficult that would be to implement), but something like that should be a separate issue.

The quickest way to improve this is to revert #341. Since we're using Vercel (which supports Stale While Revalidate) there's no point in keeping `s-maxage` to 10 minutes 

cc @masonmcelvain

PR #554, opened 2022-07-31
Facets: Drop `device` facet

### LINKED ISSUE #536: UX: Device category shows unrelated device filters
### Overview:

When on a device part page, such as https://www.ifixit.com/Parts/iPhone_13, there are filters in the Device category that are for different devices (i.e. iPad, iPod, etc.)

A user would expect to see filters that are only related to the current device page, such as iPhone 13, iPhone 13 mini, iPhone 13 Pro, and iPhone 13 Pro Max

Device_facets

### LINKED ISSUE #553: Remove Devices from Facet list
Remove Devices from Facet list. We still want the data to be searchable, so do not remove it from Algolia.

@sterlinghirsh and I have talked about this at length. We don't yet have a way to make this more usable. In the meantime, it's a much bigger hinderance than help. 

Perfect example on [Sony DualShock 4](https://www.ifixit.com/Parts/DualShock_4) . All but one of these are irrelevant and confusing: 

image

masonmcelvain:
We think there are too many device options too be helpful as a facet. We still want device to be searchable, so we'll keep indexing it in Algolia.

CC @jeffsnyder @sterlinghirsh 

Closes #553 
Closes #536

### DIFF
--- a/frontend/components/product-list/sections/FilterableProductsSection/useFacets.tsx
+++ b/frontend/components/product-list/sections/FilterableProductsSection/useFacets.tsx
@@ -43,7 +43,7 @@ export function useFilteredFacets(productList: ProductList) {
       const usefulFacets = facets
          .slice()
          .filter((facet) => {
-            return !infoNames.has(facet);
+            return facet !== 'device' && !infoNames.has(facet);
          })
          .sort(sortBy);
       return usefulFacets;


jordycosta:
QA 🟢 

The device facet is removed, devices are still searchable 👍 

---

### Testing Procedure:
1. Navigate to [remove-device-facet-ifixit.vercel.app/Parts](https://react-commerce-prod-git-remove-device-facet-ifixit.vercel.app/Parts)
2. Ensure that the **Device** filter is no longer present
3. Click on a device model _(iPhone Parts)_
4. Check that each model is searchable

---

### Screen Capture:

dvRIAlfD51

sterlinghirsh (approved):

Issue #541, opened 2022-07-29
Collections: lower case URLs just get marked no-index instead of canonicalised


ifixitmaxb:
Pages with lower case URL suffix, like, https://www.ifixit.com/Parts/iphone, https://www.ifixit.com/Parts/ipad, are just no-indexed. I think we had planned to canonicalise them? 

I don't know where we landed on desired case sensitivity but I don't think this was the desired behavior. 

https://docs.google.com/document/d/15bJZoTv3Sh7-ezn8cd6P-WzsnL5HQIk59o8n3NO3-Hc/edit

@sterlinghirsh

PR #555, opened 2022-07-31
Strapi: Upgrade so we can later have case insensitive urls

masonmcelvain:
## What's new

- Bumps Strapi and all it's plugins from 4.1.8 -> 4.3.0. This is so we can use a new `eqi` filter for case insensitive matching on product list handles.
  - There is 1 [breaking change](https://docs.strapi.io/developer-docs/latest/update-migration-guides/migration-guides.html) in this upgrade, but I don't think it applies to us. See 6e260ab1e81468f07a38a9e496bf8b060a04ceb1. In strapi.ifixit.com, we have 49 media assets, and all of them have MIME type of `image/png`, ie none are null.

## QA

Full smoke test. Make sure nothing is broken.

## On deploy

Redeploy Strapi.

Followup pull: #556

CC @sterlinghirsh

Connects #541

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -11,12 +11,12 @@
    },
    "dependencies": {
       "@graphql-tools/utils": "8.6.7",
-      "@strapi/plugin-graphql": "4.1.8",
-      "@strapi/plugin-i18n": "4.1.8",
+      "@strapi/plugin-graphql": "4.1.10",
+      "@strapi/plugin-i18n": "4.1.10",
       "@strapi/plugin-sentry": "^4.1.12",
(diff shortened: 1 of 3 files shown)


dhmacs:
deploy_block 🚫 

In [this branch](https://github.com/iFixit/react-commerce/tree/upgrade-strapi-version) I already have a working version of the upgrade, including refactoring the backend folder to use Strapi

davidrans:
dev_block :hand: on deciding if we should wait for Massimo's PR

sterlinghirsh:
Closing in favor of #557

PR #564, opened 2022-08-02
Font load optimization

### LINKED ISSUE #559: Font performance: appears to be hurting LCP and TBT scores
I ran a couple SIte Performance checks:
- https://www.webpagetest.org/result/220801_BiDcWE_C7G/
- https://www.webpagetest.org/result/220801_BiDc52_C8N/

Here's a video showing that the LCP on mobile is being held up waiting for a font: https://www.webpagetest.org/video/view.php?tests=220801_BiDc52_C8N-r:1-c:0

https://www.webpagetest.org/vitals.php?test=220801_BiDc52_C8N&run=1
image

CC @federicobadini

federicobadini:
closes #559 

This PR moves font links (and other link/meta) from `next/head` to `_document.tsx`.
This enables [nextjs font optimization](https://nextjs.org/docs/basic-features/font-optimization).

## QA

1. Visit Vercel preview
2. Verify that the font is not only linked, but that css `@font-face` instruction have been embedded in the page at build time


davidrans (approved):
The docs for `pages/_document` are here: https://nextjs.org/docs/advanced-features/custom-document

Thanks for making this change @federicobadini!

CR :coffee:

k0rvusk0r4x:
Font is linked and css is there: 
image

QA 😺

frontend/pages/_document.tsx
```diff
@@ -0,0 +1,32 @@
+import { IFIXIT_ORIGIN } from '@config/env';
+import Document, { Html, Head, Main, NextScript } from 'next/document';
+
+class MyDocument extends Document {
+   render() {
+      return (
+         <Html>
+            <Head>
+               <link
+                  href="https://fonts.googleapis.com/css2?family=Archivo+Black&family=Lato:wght@400;700&display=swap"
```
jarstelfox, line 10: Google dislikes any second paint above the fold. It's likely setting font-display: optional would fix it (https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display)
Optional will force the first load of "slow" users to not use our font(s) and thus fallback to the browser defaults.
It's possible to style the browser defaults close to the actual font, which is what we do on the main site. It's a pain to style and may not be worth it right now.

```suggestion
                  href="https://fonts.googleapis.com/css2?family=Archivo+Black&family=Lato:wght@400;700&display=optional"
```

davidrans: I think what we have here in this pull is an improvement, so let's kick this suggestion to a new issue

Issue #571, opened 2022-08-03
Redirect with Correctly Capitalized Device Title in URL


danielcliu-ifixit:
## Overview
Right now, we are handling all incorrectly capitalized queries for devices, but its good for SEO if we only have one url per device and not allow `iBook`, `ibook`, `iBoOk` etc. We should do this by always redirecting to the correctly capitalized version of the url by grabbing the correct device title data.

PR #556, opened 2022-07-31
Routing: Redirect lowercase urls

masonmcelvain:
- Uses the new [`eqi` filter](https://docs.strapi.io/developer-docs/latest/developer-resources/database-apis-reference/rest/filtering-locale-publication.html#filtering) ([introduced](https://github.com/strapi/strapi/releases) in release 4.3.0) to lookup product lists in a case insensitive manner.

## QA

Lower case urls should get redirected (with a 308 permanent redirect) to the canonical page.

CC @sterlinghirsh 

Closes #541

PR #573, opened 2022-08-03
Redirect Device Pages If Capitalization Isn't Right

danielcliu-ifixit:
## Overview
This got super easy after #556 
We want to now 1.) handle iFixit only devices (ghost devices) redirect to the correct capitalization 2.) prioritize iFixit's capitalization over Strapi's. This will make it easier to link from iFixit to the store so they don't have to check strapi for the official url, etc.

## QA
While testing watch out for permanent redirecting (308s and the like). You may have to clear your browsers cache (or in the code change the `[...deviceHandleItemType].tsx` redirect calls permanent attribute to false)
You should test an device not in strapi (`/Phone` is what I use) and make sure that even if you mess around with the capitalization it reroutes back to the correct URL. Also make sure that iFixit data is prioritized over strapi's, so maybe change some local strapi data to be miscapitalized and then make sure the iFixit data is used instead.

Closes: #571

### DIFF
--- a/frontend/models/product-list/index.ts
+++ b/frontend/models/product-list/index.ts
@@ -56,13 +56,13 @@ export async function findProductList(
       }),
       fetchDeviceWiki(filterDeviceTitle),
    ]);
-   const productList = result.productLists?.data?.[0]?.attributes;
 
+   const productList = result.productLists?.data?.[0]?.attributes;
    if (productList == null && deviceWiki == null) {
       return null;
    }
 
-   const deviceTitle = productList?.deviceTitle ?? filterDeviceTitle;
+   const deviceTitle = deviceWiki?.deviceTitle ?? productList?.deviceTitle;
    const handle = productList?.handle ?? '';
    const parents =
       productList?.parent ??
--- a/frontend/models/product-list/index.ts
+++ b/frontend/models/product-list/index.ts
@@ -65,7 +65,7 @@ export async function findProductList(
       return null;
    }
 
-   const deviceTitle = deviceWiki?.deviceTitle ?? productList?.deviceTitle;
+   const deviceTitle = deviceWiki?.deviceTitle ?? productList?.deviceTitle ?? null;
    const handle = productList?.handle ?? '';
    const parents =
       productList?.parent ??


frontend/models/product-list/index.ts
```diff
@@ -61,12 +61,11 @@ export async function findProductList(
    ]);
 
    const productList = result.productLists?.data?.[0]?.attributes;
-
    if (productList == null && deviceWiki == null) {
       return null;
    }
 
-   const deviceTitle = productList?.deviceTitle ?? filterDeviceTitle;
+   const deviceTitle = deviceWiki?.deviceTitle ?? productList?.deviceTitle;
```
davidrans, line 68: Does this need to be a string?
```suggestion
   const deviceTitle = deviceWiki?.deviceTitle ?? productList?.deviceTitle ?? '';
```

What was `filterDeviceTitle`?

masonmcelvain: `filterDeviceTitle` is taken directly from the url. So for `/Parts/ibook` it would be `ibook`.

masonmcelvain: We should probably default to string or null here: https://github.com/iFixit/react-commerce/blob/4b59f9c916a424b5e38b49f9a93781e4364ccf02/frontend/models/product-list/types.ts#L50

sterlinghirsh, line 68:
```suggestion
   const deviceTitle = deviceWiki?.deviceTitle ?? productList?.deviceTitle ?? null;
```

sterlinghirsh: Good call. I made it null.

sterlinghirsh:
CR :zap:

sterlinghirsh (approved):

masonmcelvain (approved):
CR ✌🏻

frontend/models/product-list/index.ts
danielcliu-ifixit: This is cleaner so I'm not complaining but I think the previous version still worked

masonmcelvain (approved):
CR ✌🏻

rickisXP:
QA :writing_hand: but deploy_block :thought_balloon: on a potential issue
- Wonky capitalization gets casted to the correct capitalization :heavy_check_mark: 
- Changing Strapi `deviceTitle` to be miscapitalized will use iFixit data instead :heavy_check_mark: 
- Navigating to [`/pArTs`](https://react-commerce-prod-git-redirect-if-strapi-device-59a5ab-ifixit.vercel.app/pArTs) or [`/tOOls`](https://react-commerce-prod-git-redirect-if-strapi-device-59a5ab-ifixit.vercel.app/tOOls) will 404 :warning:

sterlinghirsh:
un_deploy_block :zap: that's fine, since the parts/tools segment is outside the scope of this pr

danielcliu-ifixit:
@rickisXP I should have clarified the scope of the change, it doesn't fix capitalization for `/Tools` or `/Parts` i think that change is coming with next js is updated. 
un_deploy_block :v:

PR #575, opened 2022-08-03
Revert #341: Cache for only 10 sec, not 10 min

masonmcelvain:
We were setting the max age of the cache to 10 min when we were on Cloudfront because Cloudfront doesn't respect the SWR header.

Now that we're on Vercel, we can drop max age back to 10 seconds and set a 10 minute SWR.

https://vercel.com/docs/concepts/edge-network/caching#serverless-functions---lambdas

## QA

Not sure that we need it here?

## Background

Motivation: https://github.com/iFixit/react-commerce/pull/551#issuecomment-1203982043
#341

CC @danielbeardsley @dhmacs 
qa_req 0

### DIFF
--- a/frontend/pages/Parts/[...deviceHandleItemType].tsx
+++ b/frontend/pages/Parts/[...deviceHandleItemType].tsx
@@ -30,7 +30,7 @@ export const getServerSideProps: GetServerSideProps<AppPageProps> = async (
 ) => {
    context.res.setHeader(
       'Cache-Control',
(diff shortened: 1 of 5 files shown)


sterlinghirsh:
CR :zap:

sterlinghirsh (approved):

sterlinghirsh:
I think this is fine. I'm a little nervous that it will increase load on strapi, but hopefully not too much.

PR #578, opened 2022-08-03
Filter bubble: Fix for single select

### LINKED ISSUE #522: Facet Design: Don't show (1) bubble on single selects
## Problem

Many of our facets are now single selects as of #476. On desktop, it looks a little funny to have an expanded single select that has a (1) bubble on it, when the selected facet option is already highlighted.

However, collapsed facets on desktop and the facet overview on mobile currently depend on the (1) bubble to indicate that there is an option selected within that facet.

Desktop view:
image

Mobile facet overview:
image

Mobile facet panel:
image

## Do this

For active filter feedback, let's try this:

`<Box w="2" h="2" borderRadius="full" bg="brand.500"></Box>`

Screenshot 2022-08-02 at 08 07 00

CC @mmarcon91

sumukhipandey:
Previously, we displayed the `1` bubble when we had a facet select for the [single select facet categories](https://github.com/iFixit/react-commerce/pull/476). We don't want that anymore because in case of single select that number will always be a 1. Instead we want something of this sort:
Screenshot 2022-08-03 112849

## QA
- We only want this for single select facet categories [ie everything except for `Capacity` and `Price Range`].
   -  `Capacity` and `Price Range` should still display the count as before when the options under these categories are selected [where count = # of options selected]
- Visit the preview
- Make sure the blue bubble [refer screenshot] is visible when we select an option for single select categories on desktop and mobile
- Price Range and Capacity should still display the number

Closes #522 

CC: @masonmcelvain

### DIFF
--- a/frontend/components/product-list/sections/FilterableProductsSection/FacetsAccordion.tsx
+++ b/frontend/components/product-list/sections/FilterableProductsSection/FacetsAccordion.tsx
@@ -19,6 +19,8 @@ import { FacetFilter } from './FacetFilter';
 import { useCountRefinements } from './useCountRefinements';
 import { MAX_VALUES_PER_FACET, useFilteredFacets } from './useFacets';
 import { useFilteredRefinementList } from './useFilteredRefinementList';
+import { RefinementDisplayType } from '@models/product-list/types';
+import { getRefinementDisplayType } from '@helpers/product-list-helpers';
 
 type FacetsAccordianProps = {
    productList: ProductList;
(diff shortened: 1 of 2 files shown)


jordycosta:
QA 🟢 

Single-select facets now show a bubble instead of a count indicator👍 
Multi-select facets still use a counter ✔️ 

---

### Testing Procedure:
1. Head to https://react-commerce-prod-git-filter-bubble-change-for-1ef660-ifixit.vercel.app/Parts
2. Test clicking on multiple single and multi-select facets
3. Ensure that bubbles are shown for _single-select_ and numbers are shown for _multi-select_
4. Also check that clearing filters _(with X and the clear all button)_ will remove the indicator(s)

> Repeat steps 1-4 on mobile

---

### Screen Capture:

firefox_UhegOn1LZ9

masonmcelvain (approved):
CR ✌🏻

Issue #577, opened 2022-08-03
Fix Images


sterlinghirsh:
https://ifixit.slack.com/archives/C01K32B4FPX/p1659539928557239

https://www.ifixit.com/Parts/MacBook_Pro

PR #580, opened 2022-08-03
Fix Images and Part Collections API calls

danielcliu-ifixit:
## Overview

We introduced logs in #457, and this makes sure that the places that use `iFixtAPIClient` now can handle the response's format. Before our failures were basically nulling out and thus making it so no information gotten from the api was being used in the pages. This makes sure that we expect the appropriate data formats so we don't default to nulls.

## CR

Uh I think this should work but make sure that nothing I'm doing is funky. You know, the usual.

## QA
Make sure the normal pages work and load the pictures for the children again. Also make sure that pages that aren't in strapi are loading correctly from the iFixit API and stuff. You can use `Asus_Phone` and `Phone` for this. 
You also need to test authenticating a user in react-commerce.

Closes: #577

### DIFF
--- a/frontend/lib/ifixit-api/devices.ts
+++ b/frontend/lib/ifixit-api/devices.ts
@@ -13,14 +13,7 @@ export async function fetchDeviceWiki(
          deviceHandle.length > 0,
          'deviceHandle cannot be a blank string'
       );
-      const response = await client.get(
-         `cart/part_collections/devices/${deviceHandle}`
-      );
-      if (!response.ok) {
-         return null;
-      }
-      const payload = await response.json();
-      return response.ok ? payload : null;
+      return await client.get(`cart/part_collections/devices/${deviceHandle}`);
    } catch (error: any) {
       return null;
    }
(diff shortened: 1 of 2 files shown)


sterlinghirsh (approved):

sterlinghirsh:
CR :zap: so the problem was that we already return the resulting object from parsing the json from client.get() but we were then checking that object's `.ok` value which wouldn't exist? Weird one. Hopefully we can figure out a test to cover this in the future.

jordycosta:
Non-strapi loading:
- [ ] /Asus_Phone
- [ ] /Phone

sterlinghirsh:
CR :zap:

sterlinghirsh (approved):

masonmcelvain:
@jordycosta was that QA? Or are you still working on it?

jordycosta:
> @jordycosta was that QA? Or are you still working on it?

Double-checking on the user authentication part

masonmcelvain (approved):
CR ✌🏻

danielbeardsley:
CR :+1: 

Nice work!

PR #581, opened 2022-08-03
Sentry: Add `fetch()` override

### LINKED ISSUE #455: Improve error handling around broken api calls
[`improve-api-error-handling`](https://github.com/iFixit/react-commerce/compare/improve-api-error-handling)

## Problem
We had an issue where the product page mvp branch was completely crashing because the reviews api was behind the cominor whitelist. The request would error out, and a non-descriptive error page would show. The logs and sentry report had very limited information:

```
[GET] /Products/iphone-se-2020-replacement-battery
18:57:04:71
Function Status:
None
Edge Status:
500
Duration:
2686.06 ms
Init Duration:
472.93 ms
Memory Used:
133 MB
ID:
sfo1::q9cqk-1658195828263-f76fcbab59b7
User Agent:
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/103.0.0.0 Safari/537.36
info  - Loaded env from /var/task/frontend/.env.production
Storing 'globalSettings' in cache
Storing 'storeList' in cache
Storing 'store-us' in cache
2022-07-19T01:57:08.207Z	02817147-a27e-4933-ba44-886d68e47181	ERROR	Error: Internal Error
    at re (/var/task/frontend/.next/server/pages/Products/[handle].js:1:18227)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async Q (/var/task/frontend/.next/server/pages/Products/[handle].js:1:17063)
    at async Je (/var/task/frontend/.next/server/pages/Products/[handle].js:1:39785)
    at async Object.renderToHTML (/var/task/node_modules/.pnpm/next@11.1.3_bc463dac3dbdf444efc002b2c9d7ec7a/node_modules/next/dist/server/render.js:428:24)
    at async doRender (/var/task/node_modules/.pnpm/next@11.1.3_bc463dac3dbdf444efc002b2c9d7ec7a/node_modules/next/dist/server/next-server.js:1149:38)
    at async /var/task/node_modules/.pnpm/next@11.1.3_bc463dac3dbdf444efc002b2c9d7ec7a/node_modules/next/dist/server/next-server.js:1241:28
    at async /var/task/node_modules/.pnpm/next@11.1.3_bc463dac3dbdf444efc002b2c9d7ec7a/node_modules/next/dist/server/response-cache.js:64:36 {
  page: '/Products/[handle]'
}
RequestId: 02817147-a27e-4933-ba44-886d68e47181 Error: Runtime exited with error: exit status 1
Runtime.ExitError
```

## Do This
When there's an error with a remote api, make sure the error log and sentry report include info like the requested url, any parameters, and any http headers if possible.

For a request like this (list of reviews), an error shouldn't crash the whole page, so we should be able to catch and report the error, then move on with page rendering and show no reviews.

masonmcelvain:
This adds an override for `fetch()` that adds [Sentry context](https://docs.sentry.io/platforms/javascript/guides/nextjs/enriching-events/context/) and Vercel error logging (via `console.error`).
All usages of `fetch()` are replaced with this override.

## QA

Make sure we get a context section when fetch requests fail.

CC @sterlinghirsh 

Closes #455


packages/sentry/index.tsx
```diff
@@ -0,0 +1,17 @@
+import * as Sentry from '@sentry/nextjs';
+
+export const sentryFetch: typeof fetch = async (input, init) => {
```
sterlinghirsh, line 3: Minor point but the fetch params are officially called "resource" and "options". https://developer.mozilla.org/en-US/docs/Web/API/fetch Though options has a type of "RequestInit" according to TypeScript so who knows.

packages/sentry/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+      input,
+      credentials: init?.credentials,
+      headers: init?.headers,
+      method: init?.method,
+      parameters: init?.body,
+   };
+   Sentry.configureScope((scope) => scope.setContext('request', context));
+   return fetch(input, init).catch((error) => {
+      console.error(error);
+      console.error(context);
```
sterlinghirsh, line 14: I haven't tried this. Not sure if the output looks any better or worse but maybe.
```suggestion
      console.error(error, context);
```

packages/sentry/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+      throw error;
```
sterlinghirsh, line 15: Does using Promise.catch and then throwing here cause any difference in error handling around fetch?

sterlinghirsh:
CR :zap: dev_block :zap: on a few minor points / questions but this looks like a great approach to helping us track down these bugs.

packages/sentry/index.tsx
masonmcelvain: Oh weird, typescript says they're different than mdn does
https://github.com/microsoft/TypeScript/blob/d6d26430c7a64d3a740a7705db6ad63a8bd1c59c/lib/lib.dom.d.ts#L18664

packages/sentry/index.tsx
masonmcelvain: I don't think so

https://www.typescriptlang.org/play?#code/AoJw9gtglgzgpgOhHAVnAxgFwBQHI4CGMmcIABHAOaW4CUC6Bm6AFttnLWQLwB8ZAbwBQZUWXRgAdjDAAbRKXAg8AMyghi4pq1wAaCrQDcIsZhbgA7hWMBfembiT2yGFz6CToidLmJZYSjwHSTIAI390AGs9MhcjITsGbTYON35hMXEpGXkERTBlXHhvABMtZhYYzlt4oVBIWEQXOQA3ODxCYlIKajokivZOHnTPLJ9c-MK1DUxynX1q0bNLawT7FkdnOFdhj0zvHL8AoI2Q8LAomLia-tZBtL2xA988kCU8Yqkyxgqq+LtDEA

packages/sentry/index.tsx
masonmcelvain: https://github.com/iFixit/react-commerce/pull/581/commits/7ff5c5564c74db2fbfd0c1b0114f097f45861b64

masonmcelvain:
un_dev_block ✌🏻

packages/sentry/index.tsx
danielbeardsley: Does this produce an "error" for 400s? Cause our API produces 400s on the norm:
https://ifixit.slack.com/archives/C90126AN4/p1659398584649409

masonmcelvain: I wouldn't think so, but I don't know how to test it easily. If that's something we want, then we could add a `.then()` block and check `response.ok`

> The Promise returned from fetch() won't reject on HTTP error status even if the response is an HTTP 404 or 500. Instead, it will resolve normally (with ok status set to false), and it will only reject on network failure or if anything prevented the request from completing.

https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#differences_from_jquery

danielbeardsley: Great! Let's make sure that doesn't include *cancellation* i.e. some browsers cancel requests when navigating away or when pages are in the background.

sterlinghirsh (dismissed):

sterlinghirsh:
CR :zap:

deltuh-vee:
dev_block ✊ 
I tried blocking the fetch requests from the Network tab in devtools, but it looks like the errors are getting caught and handled by React query(so Sentry isn't receiving any errors).

packages/sentry/index.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export const sentryFetch: typeof fetch = async (resource, options) => {
+   const context = {
+      resource,
+      credentials: options?.credentials,
+      headers: options?.headers,
+      method: options?.method,
+      parameters: options?.body,
+   };
+   Sentry.configureScope((scope) => scope.setContext('request', context));
```
danielbeardsley, line 11: I think this is not quite right. `configureScope()` sets the scope *globally* for all future sentry reports. So every request will be flipping this "request" context around when its placed. The *failure* may happen *after* another request has been initiated and thus will be reported with the wrong context. Given that we are catching the errors here, I vote for at least only adding the scopes in the `catch()` to reduce the chance of reporting with the wrong context.

danielbeardsley:
dev_block :+1:  on the above comment.

packages/sentry/index.tsx
masonmcelvain: Ah you're right, I think we'd rather use [`withScope`](https://docs.sentry.io/platforms/javascript/guides/nextjs/enriching-events/context/#clearing-context) here so that we drop the `request` context after the `fetch()`.

danielbeardsley: `withScope()` is only for synchronous actions, so I don't think it applies for the promise-returning `fetch()`

packages/sentry/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+            Sentry.captureException(new Error(msg), (scope) => {
+               scope.setContext('request', context);
+               return scope;
+            });
+            console.error(msg, context);
+         }
+         return response;
+      })
+      .catch((error) => {
+         // We don't want to hear about network errors in Sentry
```
danielbeardsley, line 25: I think you're probably right, but this code doesn't prevent them from getting to sentry. They may still be unhandled promise rejections, but I suppose we should handle those at each callsite... Carry on :-).

packages/sentry/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+      _resource: resource,
+      headers: options?.headers,
+      method: options?.method,
+      // Parse to pretty print GraphQL queries
+      body: options?.body ? JSON.parse(String(options?.body)) : undefined,
+   };
+   return fetch(resource, options)
+      .then((response) => {
+         if (response.status >= 400 && response.status !== 401) {
+            const msg = `\`fetch()\` HTTP error: ${response.status} ${response.statusText}`;
```
danielbeardsley, line 15:
```suggestion
            const msg = `fetch() HTTP error: ${response.status} ${response.statusText}`;
```

Did the backticks do something special in the error message in sentry? Otherwise, I think they are superfluous.

masonmcelvain: Nope they don't get rendered as anything special. Happy to drop them

masonmcelvain:
un_dev_block ✌🏻 I think I got around to all the comments, thanks for the feedback/QA!

danielbeardsley:
CR :+1:

k0rvusk0r4x:
Errors caused when there is a problem with fetch requests now has context.

QA 😺

danielbeardsley (approved):
CR :+1:

PR #552, opened 2022-07-29
SEO: Setup no-index exemption logic

masonmcelvain:
This sets up logic to exempt certain device pages and item type pages from being no-indexed if the number of results in that collection is 1.

MacBook_Pro is included here as an example, more data will be filled in on the next commit.

## QA

Visit some of the urls in the CSV below. If they have at least 1 result, they should have a `<link rel="canonical" href={canonicalUrl} />` in the `<head>` of the html. For context, normal pages (that aren't exempted here) will have this canonical when there are 2 or more results.

[1-product-item-type-collections_exempt.xlsx](https://github.com/iFixit/react-commerce/files/9223728/1-product-item-type-collections_exempt.xlsx)

Closes #543

PR #585, opened 2022-08-04
Delete `parseNoIndexExemptDeviceUrls.ts`

masonmcelvain:
It was used as a one time script in #552, and now we don't need it anymore.

CC @dhmacs


masonmcelvain:
QA ✌🏻 quick smoke test, the preview isn't broken

dhmacs (approved):
CR 🔥

PR #591, opened 2022-08-05
Algolia: Drop unused hook (make much faster!)

danielbeardsley:
OMG, this was causing our server-side algolia query (and react dom
rendering) to happen twice. This was effectively doubling our response
times.

See this line in algolia's getServerState():
https://github.com/algolia/react-instantsearch/blob/affcbb58ee5c63345cc9424ad8ab566f3a01c9f9/packages/react-instantsearch-hooks-server/src/getServerState.tsx#L48-L49

Connect https://github.com/ifixit/ifixit/issues/43498

### DIFF
--- a/frontend/components/product-list/sections/FilterableProductsSection/useFacets.tsx
+++ b/frontend/components/product-list/sections/FilterableProductsSection/useFacets.tsx
@@ -1,15 +1,10 @@
 import * as React from 'react';
-import { useDynamicWidgets } from 'react-instantsearch-hooks-web';
 import { ProductList, ProductListType } from '@models/product-list';
 import { formatFacetName } from '@helpers/algolia-helpers';
 
 export const MAX_VALUES_PER_FACET = 200;
 
 export function useFacets() {
-   const { attributesToRender } = useDynamicWidgets({
-      maxValuesPerFacet: MAX_VALUES_PER_FACET,
-   });
-
    return [
       'facet_tags.Capacity',
       'facet_tags.Device Brand',
@@ -23,7 +18,6 @@ export function useFacets() {
       'price_range',
       'worksin',
    ];
-   // return attributesToRender;
 }
 
 export function useFilteredFacets(productList: ProductList) {


dhmacs (approved):
CR 👍 

Good job at finding this 👏 

Right now we're not even taking advantage of it, so it makes sense to drop it. If later on we want to avoid hard-coding facets in our codebase, I think we can still use this without performance penalty, as their [docs](https://www.algolia.com/doc/api-reference/widgets/dynamic-facets/react-hooks/) says:

> All matching widgets mount after the first network request completes. To avoid a second network request, facets are set to ['*'] and maxValuesPerFacet is set to 20 by default.

So I think it's just a matter of configuring `maxValuesPerFacet` correctly to avoid a second query.

federicobadini:
I had seen this while I was profiling performance. However at the time, the second iteration was much faster than the first. 

From what I saw it was due to the fact that when we hit algolia for the first time we spend a lot in negotiating the connection. The second iteration is benefitting from having a connection already open.

So yeah, if we don't need that dynamic facet anymore good catch, let's get rid of that 🎉.
However from what I saw I do not expect that we will see times halved.

Screenshot 2022-08-05 at 09 13 56

Screenshot 2022-08-05 at 09 14 18

danielbeardsley:
> However from what I saw I do not expect that we will see times halved.

Neat tool, I'm really curious what tool you're using to peek on requests made by the nextjs backend?

That said, the second network request was the small part of the overall cost.

### Before
Here's a CPU profile for a page *befire* this change. The sections are as follows:

1. Code: user request code
2. Blank: waiting on strapi and iFixit network
3. Code: first algolia evaluation from getServerState()
4. Blank: waiting on first algolia query
5. Code: second evaluation of algolia react
6. Blank: waiting on second algolia query
7. Code: Next.js generating html from props 

image


### After
Cpu capture *after* this change. Note: horizontal scales are not equivalent

1. Code: user request code
2. Blank: waiting on strapi and iFixit network
3. Code: algolia evaluation from getServerState()
4. Blank: waiting on first algolia query
5. Code: Next.js generating html from props 

image

k0rvusk0r4x:
-  Collections pages continue to function.
- Facets work as expected.

QA 😺

danielbeardsley:
deploy_block :+1:  till the end of the day so we can see the change in analytics.

sterlinghirsh:
CR :zap:

danielbeardsley:
un_deploy_block :+1: It's been enough time.

federicobadini:
> Neat tool, I'm really curious what tool you're using to peek on requests made by the nextjs backend?

Oh nothing fancy: I dug into the algolia implementation, extracted the request with all its parameters and replied it in Postman.

PR #592, opened 2022-08-05
Add Sentry Logging to Frontend Performance

danielcliu-ifixit:
## Overview
We want to add sentry logging for the frontend performance, so this pr introduces changes to do so at the same rate as `ifixit/fixit`. 

## CR
The `pnpm-lock.yaml` changes seemed necessary for `ts` standards, but I'm not sure how to determine issues of changing tons of packages or if I should do this a better way to reduce the code change.

## QA
You'll want to change the value of `tracesSampleRate` in `frontend/sentry.client.config.ts` to `1` to make sure 100% of logging goes to sentry. Then run `pnpm dev` and check if `react-commerce-dev` on sentry is getting performance logs.

Closes: https://github.com/iFixit/ifixit/issues/44070

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -41,6 +41,7 @@
       "@ifixit/shopify-storefront-client": "workspace:*",
       "@ifixit/ui": "workspace:*",
       "@sentry/nextjs": "^7.0.0",
+      "@sentry/tracing": "^7.8.1",
       "@xstate/fsm": "1.6.1",
       "@xstate/react": "1.5.1",
       "algoliasearch": "4.13.1",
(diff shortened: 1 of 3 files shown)


sterlinghirsh (approved):

sterlinghirsh:
CR :zap:

frontend/package.json
```diff
@@ -41,6 +41,7 @@
       "@ifixit/shopify-storefront-client": "workspace:*",
       "@ifixit/ui": "workspace:*",
       "@sentry/nextjs": "^7.0.0",
+      "@sentry/tracing": "^7.8.1",
```
davidrans, line 44: Should we upgrade `@sentry/nextjs` so they're on the same version?

davidrans: Never mind, you did do that :+1:

deltuh-vee:
QA 🎬 
Sentry now logs frontend performance for react-commerce.

PR #593, opened 2022-08-05
Export KIP Banner from WIP Branch

hcutrone:
## Background
Our updated footer does not include our KPIs, but we still want to show them on the homepage. We have a KPI banner in a WIP branch, so let's grab that and put it in its own package to be used in IFixit/ifixit.

This is the first pull for [IFixit/ifixit: #42248](https://github.com/iFixit/ifixit/issues/42248)

## Summary of Changes
1. Made a new package called 'homepage-kpis'
2. Copied over the `StatsSection` component from WIP branch ([add-store-home-page](https://github.com/iFixit/react-commerce/compare/main...add-store-home-page#diff-73027adcba3536a56ebd1902baa816a9b6906230de97856338c2f2f37de0654e))
3. Copied `PageContentWrapper` into package
4. Ran `pnpm install`

### DIFF
--- a/packages/homepage-kpis/PageContentWrapper.tsx
+++ b/packages/homepage-kpis/PageContentWrapper.tsx
@@ -1,0 +1,21 @@
+import { Box, BoxProps } from '@chakra-ui/react';
+import * as React from 'react';
+
+export type PageContentWrapperProps = React.PropsWithChildren<{
+   className?: string;
+}>;
+
+export function PageContentWrapper(props: BoxProps) {
+   return (
+      <Box
+         w={{ base: 'full', lg: '960px', xl: '1100px' }}
+         mx="auto"
+         px={{
+            base: 0,
+            sm: 6,
+            lg: 0,
+         }}
+         {...props}
+      />
+   );
+}
--- a/packages/homepage-kpis/index.tsx
+++ b/packages/homepage-kpis/index.tsx
@@ -1,0 +1,81 @@
+import React from 'react';
+import {
+  Box,
+  SimpleGrid,
+  Stat,
+  StatHelpText,
+  StatNumber,
+} from '@chakra-ui/react';
+import { PageContentWrapper } from './PageContentWrapper';
+
+interface Stat {
+   value: string;
+   label: string;
+}
+
+export interface StatsSectionProps {
+  data: SectionData;
+}
+
+export interface SectionData {
+   type: 'Stats';
+   id: string;
+   stats: Stat[];
+}
+
+export function StatsSection({ data: { stats } }: StatsSectionProps) {
+  return (
+     <Box
+        as="section"
+        position="relative"
+        w="full"
+        bg="blue.50"
+        borderWidth={1}
+        borderColor="blue.100"
+     >
+        <PageContentWrapper>
+           <SimpleGrid
+              columns={{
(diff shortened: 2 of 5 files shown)


packages/homepage-kpis/PageContentWrapper.tsx
```diff
@@ -0,0 +1,21 @@
...
+         mx="auto"
+         px={{
+            base: 0,
+            sm: 6,
+            lg: 0,
+         }}
+         {...props}
+      />
+   );
+}
```
dhmacs, line 21: There is already a `PageContentWrapper` exported by the `ui` package

```suggestion
```

dhmacs (changes requested):
Thanks for taking care of this @hcutrone 👍  

I think I would put the KPI component under the `ui` package rather than creating a specific package for this. What do you think?

dev_block 🔢

packages/homepage-kpis/PageContentWrapper.tsx
hcutrone: I think dropping it in the UI package is a great idea, especially if it already has the `PageContentWrapper`. I'll move it over now

hcutrone:
un_dev_block :burrito: 
KPI banner is now in the `ui` package

packages/ui/package.json
```diff
@@ -6,7 +6,8 @@
    "license": "MIT",
    "dependencies": {
       "@ifixit/app": "workspace:*",
-      "@ifixit/cart-sdk": "workspace:*"
+      "@ifixit/cart-sdk": "workspace:*",
+      "@ifixit/commerce-frontend": "workspace:*"
```
dhmacs, line 10: What's the reason for adding commerce-frontend as a dep of the ui package?

dhmacs (changes requested):
dev_block 🔢  on a question

packages/ui/package.json
hcutrone: Gonna pull @andyg0808 in for this one. I wasn't sure if I needed to include it, but I remembered him saying I might need it (he helped me set this up). Do you not think it is needed? Easy change to remove it.

andyg0808: If we didn't need to depend on it, I'm totally fine with not adding it as a dependency. I'd love to avoid it if we can. I think I thought we were gonna need files from it.

hcutrone: In that case, I'm gonna go ahead and remove it. Thanks y'all!

hcutrone:
un_dev_block :burrito:
Removed unnecessary dependency :+1:

jarstelfox (approved):
CR 🌵

packages/ui/package.json
dhmacs: `@ifixit/commerce-frontend` is the package that contains the next.js app, it would be weird for a UI package to depend on the app package. If ui package needs some code that currently is in Next.js, then the best thing to do is to extract that piece of code into a separate package. Otherwise we'd start having circular dependencies between `@ifixit/commerce-frontend` and ui package.

hcutrone: Ooh, ok. That makes sense. I don't see why we would need any Next.js code, so I think we should be good. Thanks @dhmacs !

PR #594, opened 2022-08-05
Fix google analytics

### LINKED ISSUE #576: Fix Google Analytics
[`fix-google-analytics`](https://github.com/iFixit/react-commerce/compare/fix-google-analytics)

https://ifixit.slack.com/archives/C01K32B4FPX/p1659483805779779

Google Tag Assistant says it isn't set up right.

Implemented in #485 . QA verified that the js was in the page, but didn't verify that data was sent properly.

masonmcelvain:
- Fixes a bug where page views weren't being tracked properly.
- Follows the Next.js [reference implementation](https://nextjs.org/docs/messages/next-script-for-ga#using-analyticsjs) of Google Analytics. Note that this puts the script in the `<body>`, and it was in the `<head>` before. This doesn't seem to make a difference functionally. I couldn't find a way to use the Next's `<Script>` component in the `<head>`.
- Sends a GA pageview when the Next router indicates that the route has changed.

## QA
- Install the Legacy Google Tag Manager extension on Chrome: https://chrome.google.com/webstore/detail/tag-assistant-legacy-by-g/kejbdjndbnbjgmefkgdddjlbokphdefk?hl=en
- Run react-commerce locally, and set `NEXT_PUBLIC_GA_URL=https://www.google-analytics.com/analytics.js` in `.env.development`.
- Click "Record" in the extension, and load any page of the app.
- When the page is finished loading, click "Stop Recording" in the extension. It should tell you the page was tracked properly. It will warn you that the script is in the `<body>` instead of the `<head>`, but as long as tracking works properly, I don't think this matters too much.

This is what it should look like:
image

CC @sterlinghirsh @davidrans 

Closes #576

### DIFF
--- a/frontend/components/analytics/GoogleAnalytics.tsx
+++ b/frontend/components/analytics/GoogleAnalytics.tsx
@@ -1,7 +1,25 @@
 import { GA_URL, GA_KEY, GA_DEBUG } from '@config/env';
+import * as React from 'react';
 import Script from 'next/script';
+import { useRouter } from 'next/router';
 
 export function GoogleAnalytics() {
+   const router = useRouter();
+   React.useEffect(() => {
+      const handleRouteChange = (url: string) => {
+         if (typeof ga !== 'undefined') {
+            ga('ifixit.set', 'page', url);
+            ga('ifixit.send', 'pageview');
+         }
+      };
+      router.events.on('routeChangeComplete', handleRouteChange);
+      router.events.on('hashChangeComplete', handleRouteChange);
+      return () => {
+         router.events.off('routeChangeComplete', handleRouteChange);
+         router.events.off('hashChangeComplete', handleRouteChange);
+      };
+   }, [router?.events]);
+
    return GA_URL ? (
       <>
          <Script id="google-analytics" strategy="afterInteractive">
@@ -24,6 +42,7 @@ export function GoogleAnalytics() {
                // Enable Remarketing and Advertising Reporting Features in GA
                ga('ifixit.require', 'displayfeatures');
 
+               ga('ifixit.set', 'page', window.location.pathname);
                ga('ifixit.send', 'pageview');
             `}
          </Script>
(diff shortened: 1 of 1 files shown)


davidrans (approved):
Nice work!
CR :coffee:

frontend/components/analytics/GoogleAnalytics.tsx
```diff
@@ -1,35 +1,54 @@
...
+import * as React from 'react';
+import Script from 'next/script';
+import { useRouter } from 'next/router';
+
+declare const ga: (command: string, hitType: string, url?: string) => void;
 
 export function GoogleAnalytics() {
+   const router = useRouter();
+   React.useEffect(() => {
+      const handleRouteChange = (url: string) => {
```
davidrans, line 11: Do we need to do this same thing for Matomo?

frontend/components/analytics/GoogleAnalytics.tsx
```diff
@@ -1,35 +1,54 @@
...
+               ga('ifixit.require', 'ec');
+
+               // Enable Remarketing and Advertising Reporting Features in GA
+               ga('ifixit.require', 'displayfeatures');
+
+               ga('ifixit.set', 'page', window.location.pathname);
+               ga('ifixit.send', 'pageview');
+            `}
+         </Script>
+         <Script src={`${GA_URL}`} strategy="afterInteractive" />
```
masonmcelvain, line 51:
```suggestion
         <Script src={GA_URL} strategy="afterInteractive" />
```

frontend/components/analytics/GoogleAnalytics.tsx
masonmcelvain: I'll check if navigation is working, didn't think of that.

davidrans (approved):
CR :coffee:

frontend/components/analytics/GoogleAnalytics.tsx
masonmcelvain: It doesn't seem like Matomo knows about route changes. I made an issue: https://github.com/iFixit/react-commerce/issues/595

deltuh-vee:
QA 🎬 
Page views are now tracked with Google Analytics.

PR #596, opened 2022-08-05
Preserve product list description newlines

### LINKED ISSUE #450: Change line break style in ProductList Description
Let's update the linebreak style on ProductList. 

[Strapi](https://strapi.ifixit.com/admin/content-manager/collectionType/api::product-list.product-list/17?plugins[i18n][locale]=en) 
image

[Live test](https://es.ifixit.com/Parts/Android_Phone)
image

We have an 8px margin on `<p>` tags in descriptions on Product Pages. Let's follow that standard. (Er, I don't know if it's a relative margin that changes with font size. Someone who can read code better than I can can figure out the actual standard and suggest an implementation. 🤷‍♂️ )
image

masonmcelvain:
Converts newlines to `<br />` tags before rendering product list descriptions from markdown to html.
Leading and trailing newlines are stripped.

## QA
- Visit the preview link, and choose a product list page.
- Find that product list in strapi admin: https://productlist-description-preserve-newlines.govinor.com/admin
- Edit the product list description to have newlines.
- The newlines in the UI should be one-to-one with the number of newlines in Strapi. (Leading and trailing newlines will be stripped).

CC @jeffsnyder

Closes #450

### DIFF
--- a/frontend/components/product-list/sections/HeroSection.tsx
+++ b/frontend/components/product-list/sections/HeroSection.tsx
@@ -137,7 +137,8 @@ type DescriptionRichTextProps = Omit<BoxProps, 'children'> & {
 const DescriptionRichText = forwardRef<DescriptionRichTextProps, 'div'>(
    ({ children, ...other }, ref) => {
       const html = React.useMemo(() => {
-         return snarkdown(children);
+         const preserveNewlines = children.trim().replace(/\n/g, '<br />');
+         return snarkdown(preserveNewlines);
       }, [children]);
 
       return (


sterlinghirsh:
CR :zap: dev_block :zap: This feels like a weird solution to me. Maybe it's the best we can do, but I wonder if snarkdown should be producing `<p>` tags instead of just linebreaks for this kind of thing. Then we'd get to add a margin with css (which looks like what we had before) instead of messing with the html directly and doubling the linebreaks.

sterlinghirsh:
Also it looks like snarkdown already adds `<br>` tags there. Is \n gets replaced here, how will the default <br> get inserted?

masonmcelvain:
I found an issue from 2019 about snarkdown producing plain text instead of `<p>` tags. Snarkdown doesn't have any config options, so the best we could do is add a wrapper like someone else suggested:
```js
const snarkdownEnhanced = (md: string) => {
    const htmls = md
        .split(/(?:\r?\n){2,}/)
        .map(l =>
            [' ', '\t', '#', '-', '*'].some(ch => l.startsWith(ch))
                ? snarkdown(l)
                : `<p>${snarkdown(l)}</p>`,
        )

    return htmls.join('\n\n')
}
```

> Also it looks like snarkdown already adds <br> tags there. Is \n gets replaced here, how will the default
get inserted?

By default, two newlines create a single `<br />` in snarkdown.

sterlinghirsh:
un_dev_block :zap: QA :zap: This is a little weird but I think it's fine for now. I might want to try the enhanced option later if we don't end up liking the styling here. Personally I prefer a smaller break between paragraphs. I saw that there was an even more complicated hack in there to support <p> even better. Maybe someday. It's convenient to be able to parse markdown on the client, though with server components, we might be able to do just fine with a heavier markdown parser.

sterlinghirsh (approved):

PR #600, opened 2022-08-08
Sentry: Add resolved url tag

masonmcelvain:
Sometimes it's hard to tell which page a user was on when an error occurred, so adding it in a tag will be useful for debugging.

The resolved url is the url that the user requested (like /Parts/iPhone). This is different than `context.req.url`, which can be a link like /_next/data..., which is not the helpful for debugging.

## QA
Throw an error on any page while running local dev. It should get reported to react-commerce-dev in sentry, and have a `resolved_url` tag.

### DIFF
--- a/frontend/pages/Tools/index.tsx
+++ b/frontend/pages/Tools/index.tsx
@@ -10,18 +10,18 @@ import {
    ProductListViewProps,
 } from '@components/product-list';
 import { ALGOLIA_PRODUCT_INDEX_NAME } from '@config/env';
-import { setSentryPageContext } from '@ifixit/sentry';
 import { getGlobalSettings } from '@models/global-settings';
 import { findProductList } from '@models/product-list';
 import { getStoreByCode, getStoreList } from '@models/store';
 import { urlFromContext } from '@ifixit/helpers/nextjs';
 import { GetServerSideProps } from 'next';
 import { getServerState } from 'react-instantsearch-hooks-server';
+import { serverSidePropsWrapper } from '@helpers/next-helpers';
 
 type PageProps = WithLayoutProps<ProductListViewProps>;
 type AppPageProps = WithProvidersProps<PageProps>;
 
-export const getServerSideProps: GetServerSideProps<AppPageProps> = async (
+const getServerSidePropsInternal: GetServerSideProps<AppPageProps> = async (
    context
 ) => {
    context.res.setHeader(
@@ -89,6 +89,10 @@ export const getServerSideProps: GetServerSideProps<AppPageProps> = async (
       props: pageProps,
    };
 };
+
+export const getServerSideProps = serverSidePropsWrapper(
+   getServerSidePropsInternal
+);
 
 const ProductListPage: NextPageWithLayout<PageProps> = (pageProps) => {
    return <ProductListView {...pageProps} />;
--- a/frontend/pages/Tools/index.tsx
+++ b/frontend/pages/Tools/index.tsx
@@ -14,6 +14,7 @@ import { setSentryPageContext } from '@ifixit/sentry';
 import { getGlobalSettings } from '@models/global-settings';
 import { findProductList } from '@models/product-list';
 import { getStoreByCode, getStoreList } from '@models/store';
+import { urlFromContext } from '@ifixit/helpers/nextjs';
 import { GetServerSideProps } from 'next';
 import { getServerState } from 'react-instantsearch-hooks-server';
 
@@ -23,10 +24,6 @@ type AppPageProps = WithProvidersProps<PageProps>;
 export const getServerSideProps: GetServerSideProps<AppPageProps> = async (
    context
 ) => {
-   const protocol = context.req.headers.referer?.split('://')[0] || 'https';
-   const url = `${protocol}://${context.req.headers.host}${context.resolvedUrl}`;
-   setSentryPageContext({ url });
-
    context.res.setHeader(
       'Cache-Control',
       'public, s-maxage=10, stale-while-revalidate=600'
@@ -57,7 +54,7 @@ export const getServerSideProps: GetServerSideProps<AppPageProps> = async (
    const appProps: AppProvidersProps = {
       algolia: {
          indexName,
-         url,
+         url: urlFromContext(context),
          apiKey: productList.algolia.apiKey,
       },
    };
--- a/frontend/pages/Tools/index.tsx
+++ b/frontend/pages/Tools/index.tsx
@@ -10,6 +10,7 @@ import {
    ProductListViewProps,
 } from '@components/product-list';
 import { ALGOLIA_PRODUCT_INDEX_NAME } from '@config/env';
+import { setSentryPageContext } from '@ifixit/sentry';
 import { getGlobalSettings } from '@models/global-settings';
 import { findProductList } from '@models/product-list';
 import { getStoreByCode, getStoreList } from '@models/store';
@@ -22,6 +23,10 @@ type AppPageProps = WithProvidersProps<PageProps>;
 export const getServerSideProps: GetServerSideProps<AppPageProps> = async (
    context
 ) => {
+   const protocol = context.req.headers.referer?.split('://')[0] || 'https';
+   const url = `${protocol}://${context.req.headers.host}${context.resolvedUrl}`;
+   setSentryPageContext({ url });
+
    context.res.setHeader(
       'Cache-Control',
       'public, s-maxage=10, stale-while-revalidate=600'
@@ -47,8 +52,6 @@ export const getServerSideProps: GetServerSideProps<AppPageProps> = async (
 
    const title = `iFixit | ${productList.title}`;
 
-   const protocol = context.req.headers.referer?.split('://')[0] || 'https';
-   const url = `${protocol}://${context.req.headers.host}${context.resolvedUrl}`;
    const indexName = ALGOLIA_PRODUCT_INDEX_NAME;
 
    const appProps: AppProvidersProps = {
--- a/packages/helpers/nextjs.tsx
+++ b/packages/helpers/nextjs.tsx
@@ -1,0 +1,6 @@
+import { GetServerSidePropsContext } from 'next';
+
+export function urlFromContext(context: GetServerSidePropsContext): string {
+   const protocol = context.req.headers.referer?.split('://')[0] || 'https';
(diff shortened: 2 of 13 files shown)


deltuh-vee:
QA 🎬 
Resolved URL tag is now present in Sentry.

frontend/pages/Tools/index.tsx
```diff
@@ -22,6 +23,10 @@ type AppPageProps = WithProvidersProps<PageProps>;
 export const getServerSideProps: GetServerSideProps<AppPageProps> = async (
    context
 ) => {
+   const protocol = context.req.headers.referer?.split('://')[0] || 'https';
+   const url = `${protocol}://${context.req.headers.host}${context.resolvedUrl}`;
+   setSentryPageContext({ url });
```
sterlinghirsh, line 28: What do you think about changing setSentryPageContext to accept a GetServerSideProps context object so it can calculate the url and protocol internally without having to copy/paste that. I imagine there will be something like this on every page. Not sure if there's a way to globally add this, but that could be good too. I also think for something like resolvedurl, we might want that to be a sentry tag rather than context, since tags are searchable but context isn't.

sterlinghirsh: Oh I see you used setTag down in the actual implementation so nevermind on that part.

packages/sentry/index.tsx
```diff
@@ -27,3 +27,11 @@ export const sentryFetch: typeof fetch = async (resource, options) => {
          throw error;
       });
 };
+
+type PageContextProps = {
+   url: string;
+};
+
+export const setSentryPageContext = ({ url }: PageContextProps) => {
+   Sentry.setTag('resolved_url', url);
```
danielbeardsley, line 36: I think this will be inconsistent and flaky. We're in the context of a JS server that does a bunch of async work. This JS server has only one global context yet it may be serving several requests in parallel. I think we can kill two birds though. I'm gonna try something.

frontend/pages/Tools/index.tsx
masonmcelvain: > What do you think about changing setSentryPageContext to accept a GetServerSideProps context object so it can calculate the url and protocol internally

Yea that seems reasonable, would definitely simplify calling it as the number of pages grows.

packages/sentry/index.tsx
masonmcelvain: > I think this will be inconsistent and flaky. We're in the context of a JS server that does a bunch of async work. This JS server has only one global context yet it may be serving several requests in parallel.

I'm not sure I understand. What exactly are you worried about?

Feel free to push to this branch if you come up with something 👍🏻

frontend/pages/Tools/index.tsx
jarstelfox: I agree. FYI, Next v12 has a [middleware concept](https://nextjs.org/docs/advanced-features/middleware) and I suspect we might want to create custom hooks in the long run.

packages/sentry/index.tsx
danielbeardsley: Wow... looks like they already auto-inject code to wrap `getServerSideProps()` (given that pull request). But it seems (based on experimentation) that they don't capture the relevant url? I wonder if there's a mechanism (via config) to augment or add to the code that auto-wraps `getServerSideProps()`.

packages/helpers/nextjs.tsx
```diff
@@ -0,0 +1,6 @@
+import { GetServerSidePropsContext } from 'next';
+
+export function urlFromContext(context: GetServerSidePropsContext): string {
```
masonmcelvain, line 3: FYI I didn't include this in `@ifixit/helpers/nextjs` because doing so would cause a cyclic dependency between `@ifixit/sentry` and `@ifixit/helpers`.

packages/helpers/nextjs.tsx
```diff
@@ -0,0 +1,6 @@
...
+   const protocol = context.req.headers.referer?.split('://')[0] || 'https';
+   return `${protocol}://${context.req.headers.host}${context.resolvedUrl}`;
+}
```
jarstelfox, line 6: You could use the actual URL object. I know you may toString it later, but the class may be nice down the road

```suggestion
export function urlFromContext(context: GetServerSidePropsContext): URL {
   const protocol = context.req.headers.referer?.split('://')[0] || 'https';
   return new URL(`${protocol}://${context.req.headers.host}${context.resolvedUrl}`);
}
```

packages/sentry/index.tsx
```diff
@@ -27,3 +29,8 @@ export const sentryFetch: typeof fetch = async (resource, options) => {
          throw error;
       });
 };
+
+export const setSentryPageContext = (context: GetServerSidePropsContext) => {
+   Sentry.configureScope((scope) => scope.clear());
```
jarstelfox, line 34: This is a little worrisome since the mutation-based `configureScope` will blow other data away:

```js
Sentry.withScope(scope => {
	scope.SetTag('ShopifyLoaded', 'true');
	loadPage();
	
	// ShopifyLoaded was just unset
});

function loadPage() {
	setSentryPageContext();
}
```

masonmcelvain:
Hmm the url is showing up as a sentry tag, but when I add arbitrary tags, they aren't showing up. dev_block ✌🏻

packages/sentry/index.tsx
masonmcelvain: https://github.com/iFixit/react-commerce/pull/600/commits/16b6857716d04c697dc36b085f54b3af0ecc9a2f

packages/helpers/nextjs.tsx
masonmcelvain: I agree, but we use string urls throughout most of this app unfortunately, so this would be a more siginificant refactor if we were to do it right. I'd rather leave it to another pull.

packages/sentry/index.tsx
masonmcelvain: > But it seems (based on experimentation) that they don't capture the relevant url?

That's what I had seen, but now that I'm testing this branch, I see the correct `url` tag.

image


> I wonder if there's a mechanism (via config) to augment or add to the code that auto-wraps getServerSideProps().

Maybe via the middleware concept that @jarstelfox was talking about? https://github.com/iFixit/react-commerce/pull/600#discussion_r941570980

jarstelfox: The part that is setting the url is: https://docs.sentry.io/platforms/javascript/guides/nextjs/configuration/integrations/default/#httpcontext-previously-useragent

If you look at the code it's just grabbing it from event.request:
https://github.com/getsentry/sentry-javascript/blob/ac2dc121226e0b1aa1fc199e5b7549862fc99f5f/packages/browser/src/integrations/httpcontext.ts#L31

which is set here: https://github.com/getsentry/sentry-javascript/blob/ac2dc121226e0b1aa1fc199e5b7549862fc99f5f/packages/nextjs/src/utils/withSentry.ts#L44

tl;dr: I *think* it should just work, I don't know why we are having issues.

masonmcelvain: Great thanks for finding all that! [Here's an example](https://sentry.io/share/issue/7b644c341c5f4359a5712f3025733c2b/) of a Sentry event from `getServerSideProps` that didn't get the `url` tag. But many other events from the server are. I can't explain what's causing the discrepancy, I agree with you that it seems like it should work.

sterlinghirsh: I see that it's setting the url, but it isn't setting resolvedUrl. Is there something in this code that makes you think we should be seeing resolvedUrl already?

danielbeardsley (approved):
CR :+1: 

Man, crazy that sentry/next does the auto-wrapping of the `getServerSideProps()` function. Bummer that it doesn't seem to do the thing all the time.

packages/sentry/index.tsx
jarstelfox: Okay, I did some skimming and cross-referenced the example you linked above.

It looks like sentry hooks into the server and wraps a ton of logic. Most notably [Next's Request Handler Factory](https://github.com/getsentry/sentry-javascript/blob/ac2dc121226e0b1aa1fc199e5b7549862fc99f5f/packages/nextjs/src/utils/instrumentServer.ts#L147-L149) which is what [calls down to setting the url](https://github.com/getsentry/sentry-javascript/blob/ac2dc121226e0b1aa1fc199e5b7549862fc99f5f/packages/nextjs/src/utils/instrumentServer.ts#L247).

I still believe it should be working at the time of the error(even though it clearly is not).

Sentry is passing that data down to itself when a request handler is made. i.e. `afterRouted`

Maybe someone else can spy what's missing, but I think we should provide our own way of getting the data until we can find a reliable way for sentry to hook the url

sterlinghirsh: Just to make sure we're all on the same page, this should only affect page loads where it uses next-link to load json. In those cases, we've seen url set as `/_next/...`. `resolvedUrl` will be the url of the page that we were really trying to link to. In the code I've read, I see sentry adding the url field, but I haven't seen them add the resolvedUrl field, so I wouldn't expect it to just work without some changes like in this pull.

deltuh-vee:
Carry over QA 🎬

packages/sentry/index.tsx
lforst: Hi, Sentry JS SDK maintainer here. Just stumbled onto this thread.

We're currently working on wrapping Next.js data fetching methods (including `getServerSideProps`) to improve instrumentation. Currently, the efforts mainly go into properly setting parameterized transaction names (transactions are part of the Sentry performance monitoring product and parameterization means that transactions won't show up as `/user/42/post/1337` but rather as `/user/[userId]/post/[postId]`, which is generally more useful).

We might explore adding the parameterized route + parameters to errors in some way (or at least provide a way for SDK user's to figure out what parameterized route they're in - maybe we can contribute this to the Next.js repo upstream).

We would love for you to try out the changes as soon as they land!

lobsterkatie:
> Man, crazy that sentry/next does the auto-wrapping of the `getServerSideProps()` function. Bummer that it doesn't seem to do the thing all the time.

Weighing in as the other half of the team working on this project. So glad y'all are excited for this! It's still very much a WIP, so expect it to get better over the next few weeks.

Also, feel free to follow along on https://github.com/getsentry/sentry-javascript/issues/5505. It's mostly just some notes I wrote up as I was planning the project, but if there are thoughts about it you want to share, that's a great place.

Issue #444, opened 2022-07-14
"Back to shopping" button on mini cart is interactive when invisible


erinemay:
![image](https://user-images.githubusercontent.com/72166715/179037323-2360cfe4-e4f9-4aee-bdef-85a076ff8257.png)

Not sure if it's meant to be gone entirely or visible in this case. It still works when clicked on. Just invisible.

dhmacs:
it might be there just to make it easy to animate. It shouldn't be visible, and definitely not clickable 👍

sterlinghirsh:
Should this be closed or is there something to do here?

dhmacs:
@sterlinghirsh the invisible button should be interactive

sterlinghirsh:
OK I was confused by "It shouldn't be visible, and definitely not clickable"

PR #603, opened 2022-08-10
Make Sure "Back to Shopping" Button isn't on Page When Not Visible

danielcliu-ifixit:
## Overview

Previously, we would not have the "Back to Shopping" button and associated text elements visible but they would still be interactable. This has been fixed by making that section now collapse instead of not being rendered.

## QA
Are the aforementioned buttons and text not in the drawer anymore when they shouldn't be (having items in the cart)?

I've changed the `scalefade` section to a `collapse` section, which you can see how these animate differently [here](https://chakra-ui.com/docs/components/transitions/usage#scalefade-transition) and [here](https://chakra-ui.com/docs/components/transitions/usage#scalefade-transition) respectively. This is seen when your cart is clear and the "Back to Shopping" button is valid again. If this isn't acceptable design lmk, but I feel like it is ok.

Closes: #444

### DIFF
--- a/packages/ui/cart/drawer/CartDrawer.tsx
+++ b/packages/ui/cart/drawer/CartDrawer.tsx
@@ -163,7 +163,8 @@ export function CartDrawer() {
                            </AnimatePresence>
                         </Box>
                      </ScaleFade>
-                     <ScaleFade
+                     <Collapse
+                        animateOpacity
                         in={cart.data != null && cart.data.totalNumItems === 0}
                      >
                         <VStack spacing="5" p="5">
@@ -179,7 +180,7 @@ export function CartDrawer() {
                               Back to shopping
                            </Button>
                         </VStack>
-                     </ScaleFade>
+                     </Collapse>
                   </DrawerBody>
 
                   <Collapse


dhmacs:
Any idea why `cominor.com` preview links aren't working?

jordycosta:
QA 🟢 

The elements that display an empty cart are no longer present/interactive on the mini-cart with items selected 👍 

---

### Testing Procedure:
1. Navigate to the [Vercel preview](https://react-commerce-prod-git-collapse-back-to-shopping-610b67-ifixit.vercel.app/Tools)
2. Add any item to your cart
3. Ensuring you're signed in, click on the cart icon
4. Check that no empty cart elements are shown/interactive

---

### Screen Capture:

chrome_EGQAHRC6Jo

danielcliu-ifixit:
> Any idea why `cominor.com` preview links aren't working?

what do you mean?

masonmcelvain (approved):
CR ✌🏻

dhmacs:
@danielcliu-ifixit basically we replace `.vercel.app` with `.cominor.com` in the preview link otherwise some api like cart won't work.. But yesterday when I tried it wasn't working. 
Anyway imo the collapse animation looks a little bit off in this context, but it's a minor thing that we can fix later.

PR #607, opened 2022-08-10
Community Page: Import from IFixit/ifixit

hcutrone:
The transition to NextJS continues! Now that we have a route dedicated to `/Community`, let's add the (basic) page. This pull does not include support for the header and footer, since those will take some more work and will be done in another pull. 

Note: the only functional difference from `master` is the "Join the Community" button. On `master`, this displays a popup modal that the user can login/sign up through. We don't have an implementation of this in Next yet, so we are just redirecting to `/login` instead.

## CR
This is a pretty big pull. I did my best to break things up into small commits. The first commit is a straight copy/paste from IFixit/ifixit, then the following commits are my changes.

## QA
Make sure everything looks and behaves as `master` does. The only (current) exception is with the login/reset/register modals.

---

qa_req 2


hcutrone:
dev_block :burrito: found a bug:
In mobile view, the option cards are not clickable. Will work on a fix tomorrow

jordycosta:
The background color varies from our `main` [/community](https://www.ifixit.com/Community)

background

jordycosta:
Changed this to `qa_req 2` for more testing

hcutrone:
> The background color varies from our main [/community](https://www.ifixit.com/Community)

I feel like this is ok. The Next pages have a global style that sets the background to that blue/gray color, and both /Parts and /Tools have this background. Wiling to put in the effort to change it if needed, though.

>  In mobile view, the option cards are not clickable. Will work on a fix tomorrow

It's tomorrow! Pushed a fix.

un_dev_block :burrito:

jordycosta:
QA 🟢 

Nice work, this is neat!
Option cards are clickable on mobile 👍 
Appearance and functionality remain the same as [`main`](https://ifixit.com/Community)

---

### Testing Procedure:
1. Navigate to https://react-commerce-prod-git-community-add-page-ifixit.vercel.app/
2. Ensure that the appearance of the page layout is the same as https://ifixit.com/Community
3. Test for full-page functionality on desktop & mobile

deltuh-vee:
QA 🎬 
Looks and works the same as on `main`.

frontend/components/community/activity.tsx
```diff
@@ -92,25 +99,31 @@ export default function ActivityDisplay({ data }: { data: Activity[] }) {
...
+               md: `${space[10]} 0 ${space[7]}`,
             }}
          >
             Latest Community Activity
          </Heading>
          {data ? (
-            <SimpleGrid columns={{ base: 1, sm: 2, md: 3 }} spacing={7}>
-               {data.map((activity, i) => (
-                  <ActivityCard key={i} data={activity} />
+            <SimpleGrid columns={{ base: 1, sm: 2, md: 3 }} spacing={5}>
```
klmork, line 123: Was it intentional to change the 7 to a 5?  Everything else seemed to map 1:1 so just wanted to check :)

frontend/components/community/navigation.tsx
```diff
@@ -100,14 +101,17 @@ export default function NavigationDisplay({
                key={index}
                minWidth="100px"
                onClick={() => router.push(url)}
+               color={`${cp.color.gray[500]}`}
+               fontWeight={`${cp.fontWeight.bold}`}
```
klmork: You mentioned in the commit message:
> I also added some more styling because I was unable to get the theme that we used in the IFixit/ifixit version to work correctly here.

Sorry, I'm still trying to figure out how everything is connected.  Can you send a link to the theme in iFixit/ifixit you're referring to - is that a chakra thing?  And then follow up question is would getting that theme to work correctly here be something to add to the eventual TODO list?  Thanks and sorry for all the questions!

frontend/components/community/options.tsx
```diff
@@ -62,10 +63,10 @@ function OptionCard({ option }: { option: Option }) {
          align="center"
          overflow="hidden"
          minHeight="100%"
-         borderRadius="var(--border-radius-lg)"
+         borderRadius={`${borderRadius.lg}`}
          direction={{ base: 'row', md: 'column' }}
-         boxShadow={{ base: 'var(--shadow-2)', md: 'var(--shadow-3)' }}
-         padding={{ base: 'var(--space-4)', md: '0 0 var(--space-6)' }}
+         boxShadow={{ base: `${shadow[2]}`, md: `${shadow[3]}` }}
+         padding={{ base: `${space[4]}`, md: `0 0 ${space[5]}` }}
```
klmork, line 86: Same question as above - I'm assuming the change from 6 to 5 was intentional, but just wanted to check since the rest mapped.

frontend/components/community/options.tsx
```diff
@@ -88,27 +89,32 @@ function OptionCard({ option }: { option: Option }) {
...
-            backgroundColor="var(--color-blue)"
-            boxShadow="var(--shadow-2)"
+            backgroundColor={`${color.blue['ifixit']}`}
+            boxShadow={`${shadow[2]}`}
             objectFit="scale-down"
             margin={{
-               base: '0',
-               md: 'calc(-1 * var(--space-7)) auto var(--space-5)',
+               base: `${space[4]} ${space[4]} 0px`,
+               md: `calc(-1 * ${space[7]}) auto ${space[2]}`,
```
klmork, line 116: same question as above

frontend/components/login/register.tsx
```diff
@@ -27,10 +28,20 @@ export function RegisterHeader({ goToLogin }: { goToLogin: () => void }) {
...
+               fontSize={`${fontSize.md}`}
+               color={`${color.gray6}`}
+            >
                Been here before?
             </Text>
-            <Button variant="link" fontSize="14px" color="var(--color-blue)" onClick={goToLogin}>
+            <Button
+               variant="link"
+               fontSize={`${fontSize.md}`}
+               color={`${color.blue}`}
```
klmork: In `options.tsx` you had replaced `--color-blue` with `backgroundColor={'${color.blue['ifixit']}'}` 

Are these just different shades of the same blue?

klmork:
CR 🌸 Just had some questions since I'm not super familiar with core-primitives etc.  

Nice job on breaking up commit messages and everything - very smooth to read through 😸

klmork (commented):
CR 🌸 Just had some questions since I'm not super familiar with core-primitives etc.  

Nice job on breaking up commit messages and everything - very smooth to read through 😸

frontend/components/community/navigation.tsx
hcutrone: Good question! Yes, the theme is a chakra thing. To use chakra components, you have to wrap all of them in the `ChakraProvider` component. You can pass this component an object containing default styling for components. The theme I was trying to import is in `nextjs-community/chakraTheme`.

This should probably be an eventual ToDo item, but it seemed like I had to mess with the current themes they already had set in react-commerce, so I avoided it. Will definitely make things a lot cleaner and easier to expand upon if we can get that working. Thanks for bringing that up, I'll add it to the epic (or add it to this pull, if people think I should).

frontend/components/login/register.tsx
hcutrone: Yes, they are slightly different! Here's the JSON for the core-primitives, for future reference: https://github.com/iFixit/core-primitives/blob/master/index.json

frontend/components/community/activity.tsx
hcutrone: Double checked, and yup! There were a few places that I had to adjust the styling slightly to get it to match up. 

Should have made that more clear in commits, my bad! Thanks for checking

frontend/components/community/options.tsx
hcutrone: You know what, that one was intentional, but when I went to double check I think it looks better with `space[6]`. Good catch!

jordycosta:
QA 🟢 

Carry-over from https://github.com/iFixit/react-commerce/pull/607#issuecomment-1212436478

hcutrone:
QA :burrito: carryover from https://github.com/iFixit/react-commerce/pull/607#issuecomment-1212506562

frontend/components/login/register.tsx
klmork: Oh thanks!

klmork (approved):
CR 🌸 but feel free to bump up the cr req if you want someone more familiar to take a look :)

frontend/components/community/navigation.tsx
klmork: Thanks for taking the time to explain 😄 

Gotcha, that seems fair.

Issue #583, opened 2022-08-04
Images don't always seem to exist


andyg0808:
We found that the 4 TB 3.5" Hard Drive on [this page](https://www.ifixit.com/Parts/iMac_G5_20%22_Model_A1145)
![image](https://user-images.githubusercontent.com/1283490/182738434-c0722be8-b1a3-4159-a811-e5ea48c8c2f6.png)

uses a URL that errors:
https://cart-products.cdn.ifixit.com/cart-products/pDOcJteXb1BFWN6U.size400?width=384
![image](https://user-images.githubusercontent.com/1283490/182738494-acc12b13-e357-4b36-a1d1-1d792fcc31e8.png)

Other images on that page use similar URLs, so it's not clear what's wrong.

Also, making the URL
https://cart-products.cdn.ifixit.com/cart-products/pDOcJteXb1BFWN6U?size=400&width=384
does load; no idea if that's relevant.

![hard drive](https://cart-products.cdn.ifixit.com/cart-products/pDOcJteXb1BFWN6U?size=400&width=384)

danielbeardsley:
I was worrying if this would come up. Effectively, we don't upscale images, so if a source image is smaller than a particular size, we don't generate it. But, react-commerce only gets urls of the original size (`cart-products/{guid}`) and doesn't know if an individual image exists or not.

I guess this means *some* of our product images are relatively low res and thus don't have the larger sizes. I'm a little curious where this is happening, can you provide a link?

andyg0808:
Yup: https://www.ifixit.com/Parts/iMac_G5_20%22_Model_A1145
Sorry, missed including that in the issue. Added now.

danielbeardsley:
Turns out this was an image in the search results. I think there's a bug cause it's using an image (400px) that's way too big for the natural size of the div (160px)

Here's the resultant html:
```html
[...]

PR #608, opened 2022-08-11
Scale iFixit Images from iFixit API Better

danielcliu-ifixit:
## Overview

We had some images failing to load because we didn't have said images at the dimensions we were requesting. It turns out a reason for this possibly happening was that we would scale up from a given width to the next js preset sizes ( so 180px to 256px for example) and then when we would look up our image dimensions from our api we would scale again off of the next js size (so from 256px to 400px possibly). This is obviously wasteful and unwanted so lets fix this by making sure we only scale off of our expected iFixit dimensions when we can.

## QA 

This should fix images in `ProductList` that were having issues (ie https://www.ifixit.com/Parts/iMac_G5_20%22_Model_A1145). It also affects images in `ProductListSetSection` but I'm not sure how to replicate that exactly, and also if it works on the `ProductList` I'm not sure if it matters to check every use. Any other images should not be affected by this change.

Connects: #583


jordycosta:
dev_block 🟥 

Still seeing image issues on this preview

image

danielcliu-ifixit:
OK so for QA there are some changes. This PR wont fix all the images possible on the gridview product list. In fact this PR doesn't guarantee any images are loaded it only makes it so more reasonably sized images are loaded. You can expect the following though

For grid view:
Screens max-width is under 629px a 250px image should be used.
Screens max-width is under 767px a 400px image should be used. 
Screens max-width is under 895px a 250px image should be used. 
Screens max-width is under 1000px a 400px image should be used. 
Else a 250px image is used.

For list view:
A 250 px image is used

danielcliu-ifixit:
I'm not sure why differences are showing up on the merge it looks the same to me as main...

frontend/components/ifixit-image/ifixit-image.tsx
```diff
@@ -66,12 +66,28 @@ function isStrapiImage(src: string) {
...
+      // If next wants to see the 2x image here then this accounts for that in our calculations of resolution
+      // see next's image.js getWidths() for more details
+      if (width / realWidth > 2) {
+         realWidth *= 2;
+      }
+
+      const sizeName = getImageSize(realWidth, sizeMap, defaultSize);
       // We don't use the ?width param server-side, but it gets rid of a nextjs warning
-      return baseSrc.concat('.', sizeName, `?width=${width}`);
+      return baseSrc.concat('.', sizeName, `?width=${realWidth}`);
```
jarstelfox, line 90: I think it may be a good idea to:

1. Cast to a URL (new URL(string))
2. Add a few jest tests around this function

The issue seems to indicate we were generating non-urls:

`GUID.size400?width=384`

Given the last line `baseSrc.concat('.',...` I'd bet I could break this just by passing a src with a query param:

`const src = "www.ifixit.com?withSomeParam=1"` would like end up close to `www.ifixit.com?withSomeParam=1?width=100` which is not a valid url

danielcliu-ifixit: This issue isn't that we are generating non-urls, its that our databases of pictures aren't complete. Next js basically expects us to have our pictures available in various sizes, and our pictures are not standardized to having any specific sizes available as far as I'm aware. 

> Given the last line baseSrc.concat('.',... I'd bet I could break this just by passing a src with a query param:

I think the picture sources are coming out of algolia, not provided by the user

k0rvusk0r4x:
The images selected are as described [here](https://github.com/iFixit/react-commerce/pull/608#issuecomment-1215596473).

QA 😺

frontend/components/common/ProductCard.tsx
```diff
@@ -50,7 +50,7 @@ export const ProductCardImage = ({ src, alt }: ProductCardImageProps) => {
    return (
       <AspectRatio ratio={1} flexGrow={0} flexShrink={0} position="relative">
          <IfixitImage
-            sizes="30vw"
+            sizes="(max-width: 629px) 250px, (max-width: 767px) 400px, (max-width: 895px) 250px, (max-width: 1000px) 400px, 250px"
```
andyg0808, line 53: This looks to me like it will use `250px` at over-`1000px` screen sizes. Is that correct?

andyg0808, line 53: This looks to me like it will use `250px` at over-`1000px` screen sizes. Is that intended?

Also, I might consider doing these the other way around: start with the mobile size, then do `min-width` for all the larger sizes. Kinda mobile-first style. Not important, so if you don't wanna, don't worry about it.

frontend/components/ifixit-image/ifixit-image.tsx
```diff
@@ -66,12 +66,28 @@ function isStrapiImage(src: string) {
...
+      if (typeof baseWidth === 'string') {
+         realWidth = parseInt(baseWidth);
+      } else if (baseWidth) {
+         realWidth = baseWidth;
+      }
+      // If next wants to see the 2x image here then this accounts for that in our calculations of resolution
+      // see next's image.js getWidths() for more details
+      if (width / realWidth > 2) {
+         realWidth *= 2;
+      }
```
andyg0808, line 86: Consider avoiding the continued mutation of `realWidth` by making this a new constant:
```suggestion
      const scaledWidth = (width / realWidth > 2) ? realWidth * 2 : realWidth;
```

Also, I'm wondering if this needs to be
```js
width / realWidth >= 2
```
because if we expect `width` to be 2x `realWidth`, I think the `>` version won't actually get correct answers.

frontend/pages/Community/index.tsx
```diff
@@ -75,4 +75,4 @@ export async function getStaticProps() {
          patrolEnabled,
       },
    };
-}
+}
```
andyg0808, line 78: This looks unintentional.

andyg0808 (approved):

andyg0808:
CR :dog: through c2b0d60

frontend/components/common/ProductCard.tsx
danielcliu-ifixit: yes. as far as I can tell on larger displays the grid that contains the products basically caps out in width so the pictures do too. But these various viewport sizes not scaling with the picture size really explains why our responsive image sizing isn't great rn

andyg0808:
deploy_block :+1: on discussion.

frontend/components/common/ProductCard.tsx
andyg0808: I mean that, if I'm reading these numbers right, we have a 400px image available, but we only use it for screens between 630px and 767px or between 896px and 1000px.

danielcliu-ifixit: yep thats the intended sizing. feel free to verify for yourself if youd like i was doing my best to handcode this so i could be wrong at what viewport size should be what picture size

frontend/components/ifixit-image/ifixit-image.tsx
danielcliu-ifixit: So it needs to be `> 2` because if its  `= 2` next will just handle it in the backend. This case is basically supposed to handle if we for some reason  have an image size between `width * 2` and `realwidth * 2`, then we should scale to the minimal one.  However, after syncing the sizes in 880f96a8d9a051e5aaf4de131165fae9b49cc0c1 im not sure how likely that case is to run into (it happened more when we could scale up to our cart image size then `next` would scale _that_ up to its next size). So i actually need to think over if this change is necessary at all

andyg0808:
CR :dog: through 0688ebd

andyg0808 (approved):

k0rvusk0r4x:
Still looking good.

QA 😺

andyg0808:
un_deploy_block :+1: 
I don't think there's a reason this needs to stay blocked.

frontend/components/ifixit-image/ifixit-image.tsx
```diff
@@ -8,13 +8,13 @@ type SizeMap = Array<SizeMapEntry>;
...
-   { name: 'mini', width: 56 },
-   { name: 'thumbnail', width: 96 },
-   { name: '140x105', width: 140 },
-   { name: '200x150', width: 200 },
-   { name: 'standard', width: 300 },
-   { name: '440x330', width: 440 },
-   { name: 'medium', width: 592 },
+   { name: 'mini', width: 41 },
+   { name: 'thumbnail', width: 70 },
+   { name: '140x105', width: 110 },
```
danielbeardsley, line 13: I think I'm missing something here. Why did we change the sizes of these images? I feel like this is supposed to be our knowledge of what size these images are. Changing them to not reflect reality seems like a bad idea.

PR #611, opened 2022-08-13
Improve previews seed

### LINKED ISSUE #498: Seed data from the Strapi production instance
https://github.com/iFixit/react-commerce/pull/496 was done to fix the production mini cart checkout, but in the process, we lost relative URLs that made it possible to have a prod preview. This preview was immensely helpful for testing. 

> ~~I'll open another pull to use relative urls in some places, and always us absolute urls in others so that the prod preview works again.~~

We don't want to go back to be using production strapi in a development preview, but we want production data to allow for easier and more confident testing. 

> The problem of seeing the same set of data is solved by improving how we seed the strapi preview instances in Govinor. I'm planning to make them able to seed data from the Strapi production instance, so they'll be virtually the same.

https://github.com/iFixit/react-commerce/issues/498#issuecomment-1195954150

CC: @masonmcelvain @dhmacs

dhmacs:
closes #498 

## Changelog

- 🆕 Strapi plugin: iFixit addons. This new plugin is entirely written in typescript and replaces the old seed plugin. I gave the plugin a generic name because the intent is to use it as a place where we can put our custom made Strapi tools.
This initial implementation makes the plugin able to seed data using `strapi.ifixit.com` data.
- 🆙 Minor Strapi upgrade to v4.3.3 to fix some minor issues

👁️  **NOTE**: There are some changes to the Docker file that are necessary to build the typescript plugin. These changes have been tested to work with Govinor docker deployment, so I expect them to work in production, but it's worth to double check.

> ℹ️ Most of the stuff has been auto generated by Strapi using the `strapi generate` command. The interesting bits are in the folder `backend/src/plugins/addons/server/services/seed`

## QA

1. Visit the Vercel Preview
2. Verify that the preview is showing production-like data
3. Clone the project and install dependencies (make sure to follow install instructions from the readme)
4. Run `pnpm dev` and log into the local Strapi instance at http://localhost:1337/admin
    > ⚠️ If you already have the project cloned on your computer, chances are that the Strapi db is already initialized. If that's the case then reset it by deleting the file at `backend/.tmp/data.db`
6. Verify that the instance has a copy of production data


deltuh-vee:
dev_block ✊ 
I wasn't able to start the dev server.
CleanShot 2022-08-15 at 11 55 45

dhmacs:
@deltuh-vee I think you need to create the env file:

Screenshot 2022-08-16 at 09 36 06

un_dev_block 🏎️

deltuh-vee:
dev_block ✊ 
There's no preview?
CleanShot 2022-08-16 at 09 37 06

dhmacs:
un_dev_block 🏎️ 

> There's no preview? 

Something went wrong with the Vercel/Github integration. I've just redeployed 👍

deltuh-vee:
@dhmacs react-commerce-prod still looks stuck.

dhmacs:
@deltuh-vee it's the Vercel github integration that is not working properly right now, the Vercel preview is fine. Here's the link to the prod one

https://react-commerce-prod-git-improve-previews-seed-ifixit.vercel.app/Parts

deltuh-vee:
QA 🎬 
* Preview is now showing production data
* Localdev now has a copy of the production data

backend/src/plugins/addons/server/services/seed/index.ts
```diff
@@ -0,0 +1,142 @@
...
+            break;
+         }
+      }
+
+      if (repos.length === 0 || !shouldSeed) {
+         strapi.log.info('🌱 Nothing to seed. Skipping..');
+         return { contentTypes: {}, media: { count: 0 } };
+      }
+
+      await mediaRepo.dangerouslyReset();
```
masonmcelvain, line 62: Just so I understand correctly: seeding only occurs for previews and local dev? i.e. this code won't be run on strapi.ifixit.com?

backend/src/plugins/addons/server/services/seed/content-type-repository.ts
```diff
@@ -0,0 +1,295 @@
...
+         isComponent(value) ||
+         isDynamicZone(value) ||
+         isMedia(value)
+      );
+   });
+}
+
+function isDynamicZone(value: any) {
+   return Array.isArray(value);
+}
```
masonmcelvain, line 225: Above, I see we type `dynamicZone` as `Record<string, any>[]`. Should this verify that each element is a component, or is that unnecessary?

```suggestion
function isDynamicZone(value: any) {
   return Array.isArray(value) && value.every((element) => isComponent(element));
}
```

masonmcelvain (approved):
CR ✌🏻 Well done @dhmacs! deploy_block ✌🏻 on confirming this won't wipe out data on strapi.ifixit.com. It's hard for me to tell from the code whether that's possible or not. Feel free to merge if there are measures to prevent that.

backend/src/plugins/addons/server/services/seed/index.ts
dhmacs: Good question. Seeding occurs during the bootstrap process. Here's the  the relevant code:

https://github.com/iFixit/react-commerce/blob/improve-previews-seed/backend/src/plugins/addons/server/bootstrap.ts#L4-L5

As you can see the seeding starts only when `SEED_DB` environment variable is set to `true`. So we have to make sure that we're not setting that to `true` in production. The old plugin had the same behaviour, so I assume the production environment is not setting `SEED_DB`, but it's worth double checking.
Additionally, seeding will only occur if at least one content type is empty (which shouldn't be case, although we don't want to rely on this, since in the future we might add new content types.
Another thing to notice is that seeding against production is probably an idempotent operation, because you're effectively inserting the same data (but I won't put this to test because it's not a guarantee).

As a side note, do we have a backup process in place?

backend/src/plugins/addons/server/services/seed/content-type-repository.ts
dhmacs: We potentially could assert that, but I'd have to test the implications. I think this plugin could use several improvements, especially in types (Strapi plugin types are not that great). 
I'd like to merge this as is, so that we can ship this and evaluate improvements in isolation. What do you think?

backend/src/plugins/addons/server/services/seed/index.ts
masonmcelvain: @danielbeardsley @sterlinghirsh We're discussing this in our standup.

masonmcelvain: @dhmacs You probably can't access this [file](https://github.com/iFixit/server-templates/blob/master/terraform/strapi/prod/main.tf#L64-L77), but we are not setting the `SEED_DB` environment variable in production.

> As a side note, do we have a backup process in place?

@danielbeardsley confirmed that we take snapshots of our Amazon RDS machines daily, so we have backups.

masonmcelvain:
@sterlinghirsh is gonna take a look at this before we merge it.

backend/src/plugins/addons/server/services/seed/index.ts
masonmcelvain: I think we all feel comfortable that the content is safe 👍🏻

masonmcelvain:
un_deploy_block ✌🏻 I'll merge this and deploy it

masonmcelvain:
Successfully redeployed 👍🏻

PR #612, opened 2022-08-16
@ifixit/ui: Add `build` and `clean` scripts

andyg0808:
Our main monorepo requires these for all dependencies of our main package. Now that we're wanting to start using the `@ifixit/ui` package from that main package, we need them.

qa_req 0

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -4,6 +4,10 @@
    "main": "./index.tsx",
    "types": "./index.tsx",
    "license": "MIT",
+   "scripts": {
+      "build": "",
+      "clean": ""
+   },
    "dependencies": {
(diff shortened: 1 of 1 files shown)


masonmcelvain (approved):
CR ✌🏻

masonmcelvain:
deploy_block ✌🏻 the preview is 500ing

masonmcelvain:
Ah, it's because there's a `/` in the branch name, which is used as the govinor subdomain. So `https://ifixit/ui--add-build-and-clean-scripts.govinor.com` is not a valid url. I think this is fine to merge as is because the code is ok. @andyg0808 

un_deploy_block ✌🏻

PR #615, opened 2022-08-16
Mobile Filtering: Improve UX

### LINKED ISSUE #579: Improve UX: Mobile filtering
### Overview:

When testing https://github.com/iFixit/react-commerce/pull/578, I noticed that there could be some improvement on the way we navigate out of the facet list when clicking on a single-select facet option.

Currently, when a single-select facet option is selected, the user is navigated out of the filtering completely. I feel that a better option for this would be to navigate the user back instead of exiting out entirely, allowing them to get more filtering done if desired.

back

---

improve_ux

sumukhipandey:
In https://github.com/iFixit/react-commerce/commit/e905db9176b532d461d469711a0e3e0f64e49dbb, we made it so that the on mobile view, when we select a facet in the single select facet category, we close the drawer and get back to the product page. The issue with that is if the user needs to filter further, they have to repeat certain steps.

Instead, lets keep the behavior so that when the user selects a facet in the single select category, we bring the user back to the facet list instead of closing the drawer.

## QA
- On mobile/tablet view, make sure that when a single select facet is selected, it brings you back to the list of facets instead of exiting out.

**NOTE: https://github.com/iFixit/react-commerce/issues/579#issuecomment-1209078103 hasn't been implemented in this PR and if this scrolling problem still persists, then we will open a new issue for it.**

Closes #579 

CC: @jordycosta @masonmcelvain @sterlinghirsh

### DIFF
--- a/frontend/components/product-list/sections/FilterableProductsSection/FacetsDrawer.tsx
+++ b/frontend/components/product-list/sections/FilterableProductsSection/FacetsDrawer.tsx
@@ -147,7 +147,6 @@ export function FacetsDrawer({
                            isOpen={facet === currentFacet}
                            productList={productList}
                            onClose={() => {
-                              onClose();
                               setCurrentFacet(null);
                            }}
                         />


masonmcelvain (approved):
CR ✌🏻 Seems like it will do the thing, good work @sumukhipandey!

> Also scroll so that the result count / filters button is at the top of the page after closing the filters menu.

deploy_block ✌🏻 Is this already happening, and that's why there's no code change for it?

rickisXP:
QA :file_cabinet:  
- Filter drawer no longer disappears after a filter is selected :heavy_check_mark: 
- Filter drawer returns to all filters once a child filter has been selected :heavy_check_mark: 
- deploy_block :scroll: on confirming if this pull will address https://github.com/iFixit/react-commerce/issues/579#issuecomment-1209078103

sumukhipandey:
@masonmcelvain @rickisXP Sterling and I discussed this and since we `useLockBodyScroll`, we have to figure out a way to set the position of the page to the desired location when we exit out of the facet filter drawer. We can define this behavior in the `onClose `for the `x` on the top right and/or `onSelect` for `Apply`. We decided we can figure that out as a separate issue and I made one right now https://github.com/iFixit/react-commerce/issues/618.

sumukhipandey:
un_deploy_block 🦋 on #618

PR #617, opened 2022-08-16
Have Single Select Refinement UI Change Look Smoother

### LINKED ISSUE #526: Default results appear when switching single-select facets

danielcliu-ifixit:
## Overview

When we used to swap single select refinements we would sometimes have the no-refinements selection blip on the screen for a half second (see the issue #526 for an example). This pr should solve that problem by updating the full Index UI State at once (including the refinements) and not clearing out the refinements and then setting them which was why we would have a slight blip of wrong results.

## CR
Is there a more efficient way to organize the data for the IndexUIState than what I have?

## QA
Make sure that pages don't have the blip of wrong results when switching single select refinements (note that even on master this blip disappears on switching if youve already switched between the two refinements on a page load. it must be caching the refinement results in the browser). Make sure that refinements are working as intended. Also make sure that algolia search is still working in general/no glaring issues were introduced by the package upgrade (i looked through the patch notes an nothing seemed hugely worrying but this is safest). 

Closes #526

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -62,8 +62,8 @@
       "react": "17.0.2",
       "react-dom": "17.0.2",
       "react-icons": "4.2.0",
-      "react-instantsearch-hooks-server": "6.26.0",
-      "react-instantsearch-hooks-web": "6.26.0",
+      "react-instantsearch-hooks-server": "6.31.1",
+      "react-instantsearch-hooks-web": "6.31.1",
       "react-query": "3.34.19",
       "sharp": "0.30.7",
       "snarkdown": "2.0.0"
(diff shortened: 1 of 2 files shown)


sterlinghirsh:
I know a few weeks ago, we weren't able to update to the new InstantSearch Hooks package because of a bug. I don't remember off hand what that bug was, so let's definitely make sure we figure out what the problem was and whether or not it's fixed before we proceed with this.

rickisXP:
Looks like the preview is 500'ing 
image

danielcliu-ifixit:
deploy_block :v: on getting the instantsearch library figured out

masonmcelvain:
Closing since we merged #627 instead

PR #619, opened 2022-08-17
Add product list template

### LINKED ISSUE #572: Code Duplication across pages: Reduce
### Problem
We now have [five pages](https://github.com/iFixit/react-commerce/tree/main/frontend/pages). They are 100-140 lines long and most of them are *nearly* identical. five copies of nearly identical code can be hard to maintain.

### Do This
Reduce duplication. Not sure if extracting re-used bits is better, or making a more generic *Page* component is better.

Here's a diff of two of the files:
image

dhmacs:
closes #572 

To reduce code duplication for routes I'm proposing to introduce a `templates` folder. We can add a "template" for each kind of page that we have (e.g. product list, product, community, etc).
The advantages of this approach are:

- **Reduced code duplication**: the logic to handle similar routes is handled in a single place
- **Code colocation**: We can locate components that are tightly coupled with the template under that template folder (doing these with `pages` folder was not possible because these files would also become routes). Note that in an effort to keep this PR short I'm not moving `/frontend/components/product-list` components under `/frontend/templates/product-list`. I'll do this in a separate PR

## QA

1. Open Vercel preview
2. Verify that `/Parts` renders correctly
3. Verify that device part and device item type parts render correctly 
4. Verify that devices or item types that don't exist return a 404 page
5. Verify that devices handles typed in the wrong case are redirected to the canonical page
6. Verify that `/Tools` renders correctly
7. Verify that tool categories render correctly
8. Verify that tool categories that don't exist return a 404 page
9. Verify that tool categories handles typed in the wrong case are redirected to the canonical page

cc @danielbeardsley @sterlinghirsh @masonmcelvain


frontend/pages/Parts/[...deviceHandleItemType].tsx
```diff
@@ -1,145 +1,11 @@
...
-      const canonicalDeviceTitle = encodeDeviceTitle(productList.deviceTitle);
-      return {
-         redirect: {
-            permanent: true,
-            destination: `/Parts/${canonicalDeviceTitle}${slug}`,
-         },
-      };
-   }
-
-   const title = `iFixit | ${productList.title}`;
```
Comment (masonmcelvain) [inline frontend/pages/Parts/[...deviceHandleItemType].tsx:90]: We now prefer `iFixit` for the title as opposed to `iFixit | Productlist Title`? I didn't see it mentioned anywhere, so just wanted to check.

masonmcelvain (approved):
CR ✌🏻 This is great, thanks @dhmacs!

frontend/pages/Parts/[...deviceHandleItemType].tsx
Comment (dhmacs) [reply frontend/pages/Parts/[...deviceHandleItemType].tsx:90]: This is just a default value, which is overridden [here](https://github.com/iFixit/react-commerce/blob/main/frontend/components/product-list/MetaTags.tsx#L68).
I expect we'll do similar overrides for every template, so that value is only there to ensure that we have a default value.
You can check the preview to see that the behaviour is unchanged. 
I did this change to make it explicit that this is a default value, and to simplify the generation of layout props.

k0rvusk0r4x:
- `/Parts` renders correctly.
- Device part and device item type parts render correctly.
- Devices or item types that don't exist 404.
- Devices with the wrong casing get redirected to the correct page.
- `/Tools` renders correctly.
- Tool categories render correctly.
- Tool categories that don't exist 404.
- Tools with the wrong casing get redirected to the correct page.

QA 😺

frontend/templates/product-list/index.tsx
```diff
@@ -0,0 +1,279 @@
...
+      );
+
+      const indexName = ALGOLIA_PRODUCT_INDEX_NAME;
+      let layoutProps: LayoutProps;
+      let productList: ProductList | null;
+      let shouldRedirectToCanonical = false;
+      let canonicalPath: string | null = null;
+
+      switch (productListType) {
+         case ProductListType.AllParts: {
```
danielbeardsley, line 69: I appreciate the de-duplication! It seems like moving the differentiating code into a 120-line switch statement is a different kind of anti-pattern though.

sterlinghirsh: @danielbeardsley If you have an improvement in mind, feel free to suggest it. I appreciate short switches too, but this is a big improvement over what we had before.

danielbeardsley: Yeah, I would probably have left the bits that differ across pages in their respective `pages/*` files. That said, maybe this would have resulted in more overhead, I don't know. I commented cause it seems like this swung the pendulum too far the other direction.

dhmacs: I thought about that too, putting reusable pieces in a separate folder works too (e.g. `lib`). Maybe it's also possible to transform `getProductListServerSideProps` into a higher order function, so you can move the customization back into the `/pages` folder, but I didn't want to spend too much time thinking about that. I'm sure that as we move forward and accumulate use cases we'll have a better idea on how to improve this.

PR #624, opened 2022-08-17
Product List Template: DRY up layout props calls

danielbeardsley:
This function takes no args and we don't need it till we return the
final result, so let's call it early and await on the result before
returning.


masonmcelvain (approved):
CR ✌🏻

k0rvusk0r4x:
Collections pages continue to function as they do on master.

QA 😺

PR #625, opened 2022-08-17
strapi: Use long git ref for deploy

masonmcelvain:
We ran into a problem where different versions of `git` had different length short refs (old version had 7 chars, new version had 8 chars).

Let's always use the long ref so we don't have that problem in the future.

## On deploy

We should be able to [redeploy](https://github.com/ifixit/server-templates/wiki/strapi) strapi

CC @danielbeardsley

Connects https://github.com/iFixit/server-templates/pull/3512

qa_req 0

### DIFF
--- a/backend/build-docker-and-push
+++ b/backend/build-docker-and-push
@@ -2,8 +2,8 @@
 
 set -eu
 
-GIT_SHORT_REF=$(git rev-parse --short HEAD)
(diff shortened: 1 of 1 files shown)


danielcliu-ifixit:
CR :v:

danielbeardsley (approved):
CR :+1:

PR #626, opened 2022-08-17
Lato: replace with system fonts

davidrans:
This changes the default font from Lato -> system fonts.

### DIFF
--- a/packages/ui/theme/foundations/fonts.ts
+++ b/packages/ui/theme/foundations/fonts.ts
@@ -2,7 +2,7 @@ import type { ThemeOverride } from '@chakra-ui/react';
 import primitives from '@core-ds/primitives';
 
 export const fonts: ThemeOverride['fonts'] = {
-   body: primitives.fontFamily.lato,
-   heading: primitives.fontFamily.lato,
+   body: primitives.fontFamily.sansSystem,
+   heading: primitives.fontFamily.sansSystem,
    mono: primitives.fontFamily.monoSystem,
 };


davidrans:
deploy_block :hand: on deciding if/when to do this

dhmacs:
@davidrans @federicobadini is already working on this and will post his comments here soon

davidrans:
Cool, I just made this change so I could poke around and see what it looked like and figured I might as well open a pull. Feel free to close this favor of @federicobadini's

federicobadini:
@davidrans I have just a couple of observations that I think it's worth clarifying before proceeding:

- I think we should remove the lato+ArchivoBlack font imports from `_document.tsx`
- We should remove all the fontFamily rules referring to Lato or Archivo Black from code (I can find at least 3 right now)

Finally I see no `package.json` updates so I think there has been no updates to the core primitives library. 
This means that the fonts already defined there have been approved to substitute Lato? Otherwise we have to wait merging this until the right set of fonts is defined and shipped in core primitives.

davidrans:
This fixes the LCP issues that webpagetest.org was complaining about on mobile:

| | webpagetest.org output | 
| --------- | ------ |
| Before  | image |
| After | image |

davidrans:
> * I think we should remove the lato+ArchivoBlack font imports from `_document.tsx`
> * We should remove all the fontFamily rules referring to Lato or Archivo Black from code

Makes sense to me. Are you working on that in a separate pull @federicobadini or should I make those changes here?

sterlinghirsh:
CR :zap: dev_block :zap: on some more info from @federicobadini

PR #627, opened 2022-08-18
Have Single Select Refinement UI Change Look Somewhat Smoother 

danielcliu-ifixit:
## Overview

When we used to swap single select refinements we would sometimes have the no-refinements selection blip on the screen for a half second (see the issue #526 for an example). This pr helps make it smoother while we try to figure out whats wrong with vercel and #617. It does this by adding in the new refinement before removing the old refinement.

## CR
This isn't a bad pattern for react right?

## QA
Make sure that pages have less of the blip of wrong results when switching single select refinements (note that even on master this blip disappears on switching if youve already switched between the two refinements on a page load. it must be caching the refinement results in the browser). Make sure that refinements are working as intended. 

Connects #526

### DIFF
--- a/frontend/templates/product-list/sections/FilterableProductsSection/RefinementSingleSelect.tsx
+++ b/frontend/templates/product-list/sections/FilterableProductsSection/RefinementSingleSelect.tsx
@@ -32,16 +32,19 @@ export function RefinementSingleSelect({
          sortBy: useSortBy(otherProps),
       });
 
-   const onClickSingleSelect = (newSelected: RefinementListItem) => {
-      refine(newSelected.value);
+   const onClickSingleSelect = React.useCallback(
+      (newSelected: RefinementListItem) => {
+         refine(newSelected.value);
 
-      const oldSelected: RefinementListItem | undefined = items.find(
-         (item) => item.isRefined
-      );
-      if (oldSelected && oldSelected !== newSelected) {
-         refine(oldSelected.value);
-      }
-   };
+         const oldSelected: RefinementListItem | undefined = items.find(
+            (item) => item.isRefined
+         );
+         if (oldSelected && oldSelected !== newSelected) {
+            refine(oldSelected.value);
+         }
+      },
+      [items, refine]
+   );
    return (
       <Box>
          <VStack align="stretch" spacing="1" role="listbox">
--- a/frontend/templates/product-list/sections/FilterableProductsSection/RefinementSingleSelect.tsx
+++ b/frontend/templates/product-list/sections/FilterableProductsSection/RefinementSingleSelect.tsx
@@ -14,7 +14,6 @@ import { useSortBy } from './useSortBy';
 import { useFilteredRefinementList } from './useFilteredRefinementList';
 import { ProductList, ProductListType } from '@models/product-list';
 import { encodeDeviceItemType } from '@helpers/product-list-helpers';
-import { useDecoupledState } from '@ifixit/ui';
 
 type RefinementSingleSelectProps = UseRefinementListProps & {
    productList: ProductList;
@@ -96,7 +95,6 @@ const SingleSelectItem = React.memo(function SingleSelectItem({
    onClose,
    onClick,
 }: SingleSelectItemProps) {
-   const [isRefined, setIsRefined] = useDecoupledState(item.isRefined);
    const { createURL } = useCurrentRefinements();
    const shouldBeLink =
       attribute === 'facet_tags.Item Type' &&
@@ -109,7 +107,6 @@ const SingleSelectItem = React.memo(function SingleSelectItem({
          as={shouldBeLink ? 'a' : 'button'}
          onClick={(event) => {
             event.preventDefault();
-            setIsRefined((current) => !current);
             onClick && onClick(item);
             onClose?.();
          }}
@@ -149,8 +146,8 @@ const SingleSelectItem = React.memo(function SingleSelectItem({
       <HStack
          key={item.label}
          justify="space-between"
-         color={isRefined ? 'brand.500' : 'inherit'}
-         fontWeight={isRefined ? 'bold' : 'inherit'}
+         color={item.isRefined ? 'brand.500' : 'inherit'}
+         fontWeight={item.isRefined ? 'bold' : 'inherit'}
       >
          {RefinementTitle}
          <Text size="sm" fontFamily="sans-serif" color={'gray.500'}>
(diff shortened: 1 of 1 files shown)


jordycosta:
QA 🟢 

No longer seeing non-filtered search results when switching from one single-select facet option to another 👍 

---

### Testing Procedure:
1. Go to [`react-commerce-prod`](https://react-commerce-prod-git-filter-refinement-for-sin-ad0090-ifixit.vercel.app/)
2. Switch between some single-select facets
3. Ensure that there is no brief flash of non-filtered results

---

### Screen Capture:

<details>
<summary> Toggle View </summary> <br>

results
</details>

frontend/templates/product-list/sections/FilterableProductsSection/RefinementSingleSelect.tsx
```diff
@@ -29,6 +32,19 @@ export function RefinementSingleSelect({
          ...otherProps,
          sortBy: useSortBy(otherProps),
       });
+   const onClickSingleSelect = (newSelected: RefinementListItem) => {
+      const oldSelected: RefinementListItem | undefined = items.find(
+         (item) => item.isRefined
+      );
+
+      newSelected.isRefined = !newSelected.isRefined;
```
dhmacs, line 40: seems like this is trying to directly mutate Algolia library internal state?

frontend/templates/product-list/sections/FilterableProductsSection/RefinementSingleSelect.tsx
```diff
@@ -29,6 +32,19 @@ export function RefinementSingleSelect({
...
+      refine(newSelected.value);
+
+      if (oldSelected && oldSelected !== newSelected) {
+         oldSelected.isRefined = !oldSelected?.isRefined;
```
dhmacs, line 44: Same comment as above

jordycosta:
QA 🟢 

Followed testing procedure in https://github.com/iFixit/react-commerce/pull/627#issuecomment-1219043384

frontend/templates/product-list/sections/FilterableProductsSection/RefinementSingleSelect.tsx
```diff
@@ -29,6 +30,16 @@ export function RefinementSingleSelect({
...
+   const onClickSingleSelect = (newSelected: RefinementListItem) => {
+      refine(newSelected.value);
+
+      const oldSelected: RefinementListItem | undefined = items.find(
+         (item) => item.isRefined
+      );
+      if (oldSelected && oldSelected !== newSelected) {
+         refine(oldSelected.value);
+      }
+   };
```
masonmcelvain, line 42: Not sure how much of an optimization this is, but doesn't seem like a bad thing to do here.

```suggestion
   const onClickSingleSelect = React.useCallback((newSelected: RefinementListItem) => {
      refine(newSelected.value);

      const oldSelected: RefinementListItem | undefined = items.find(
         (item) => item.isRefined
      );
      if (oldSelected && oldSelected !== newSelected) {
         refine(oldSelected.value);
      }
   }, [items, refine]);
```

frontend/templates/product-list/sections/FilterableProductsSection/RefinementSingleSelect.tsx
```diff
@@ -69,21 +80,17 @@ type SingleSelectItemProps = {
...
 
 const SingleSelectItem = React.memo(function SingleSelectItem({
    item,
    attribute,
    productListType,
-   refine,
    onClose,
+   onClick,
 }: SingleSelectItemProps) {
-   const [isRefined, setIsRefined] = useDecoupledState(item.isRefined);
```
masonmcelvain, line 83: I realize we can no longer use the decoupled state because the `SingleSelectItem`s can now change the `isRefined` value of their sibling components.

The consequence of dropping the decoupled state is that there is a delay between when the user clicks the button and when the button is highlighted. This might cause some users to click the button again before it highlights, which would effectively unrefine that selection. I think this is a bad UX, but I can't think of a way to circumvent it. Maybe when we figure out how to upgrade `react-instantsearch` we can use different hooks that will let each `SingleSelectIem` control it's own decoupled `isRefined` state again, while also preventing the flash of intermediate search results.

masonmcelvain (approved):
CR ✌🏻 Nice solution @danielcliu-ifixit! deploy_block ✌🏻 on a couple things to think about, but I'm fine merging this as is.

danielcliu-ifixit:
un_deploy_block :v: on comments being addressed.

frontend/templates/product-list/sections/FilterableProductsSection/RefinementSingleSelect.tsx
```diff
@@ -29,6 +31,17 @@ export function RefinementSingleSelect({
...
+   const onClickSingleSelect = React.useCallback(((newSelected: RefinementListItem) => {
+      refine(newSelected.value);
+
+      const oldSelected: RefinementListItem | undefined = items.find(
+         (item) => item.isRefined
+      );
+      if (oldSelected && oldSelected !== newSelected) {
+         refine(oldSelected.value);
+      }
+   }), [items]);
```
masonmcelvain, line 44: Shouldn't `refine` be a part of the dependency array here? We don't want to use a stale version of the `refine` callback.

frontend/templates/product-list/sections/FilterableProductsSection/RefinementSingleSelect.tsx
```diff
@@ -137,8 +146,8 @@ const SingleSelectItem = React.memo(function SingleSelectItem({
       <HStack
          key={item.label}
          justify="space-between"
-         color={isRefined ? 'brand.500' : 'inherit'}
-         fontWeight={isRefined ? 'bold' : 'inherit'}
+         color={item.isRefined ? 'brand.500' : 'inherit'}
+         fontWeight={item.isRefined ? 'bold' : 'inherit'}
```
masonmcelvain, line 150: Since we're not using `isRefined` anymore, does the color actually change before Algolia has finished refining the search? If not, I'd say we just drop `const [, setIsRefined] = useDecoupledState(item.isRefined);` altogether.

masonmcelvain (dismissed):
CR ✌🏻 Thanks for considering my suggestions! dev_block ✌🏻 on a couple more comments

frontend/templates/product-list/sections/FilterableProductsSection/RefinementSingleSelect.tsx
danielcliu-ifixit: how would a function be stale?

danielcliu-ifixit: oh cause its a prop i see

danielcliu-ifixit: i see we craft it out of some props it looks like so it could change potentially

danielcliu-ifixit:
un_dev_block :v: comments addressed now fr

masonmcelvain (approved):
CR ✌🏻 Thanks @danielcliu-ifixit!

rickisXP:
QA :zap: 
- Confirmed flash of unfiltered results no longer occurs :heavy_check_mark: 

followed https://github.com/iFixit/react-commerce/pull/627#issuecomment-1219043384

PR #628, opened 2022-08-18
Add layouts folder

### LINKED ISSUE #621: Create layouts folder
Move layouts components into a separate folder where we can include also the data fetching code

dhmacs:
closes #621 

This PR moves the default layout into a separate `layouts` folder. Right now we have only one layout, but in the future we might have pages that will require different layouts.
The main reason for this change is that layout components as a concept are more closely related to templates than to normal components, since they also have data fetching requirements associated with them (e.g. `getLayoutSeverSideProps`) that we'd like to reuse across templates. For this reason I think it makes sense to promote them as first class citizens in the project structure.


masonmcelvain (approved):
CR ✌🏻

k0rvusk0r4x:
Collections pages continue to look and function as before.

QA 😺

PR #606, opened 2022-08-10
Upgrade to Next.js 12

federicobadini:
The project has been updated to Next.js 12 and is now using the [SWC compiler](https://nextjs.org/docs/advanced-features/compiler).
Jest tests however are still run via `babel-jest` / `next/babel`.

This is due to the fact that SWC strictly adhere to the ESModules specification so, according to how a module is declared, we may have difficulties in mocking up functions.   Here is an example:

Screenshot 2022-08-11 at 00 33 09

 Function 1 is exported via named export, but since it is imported with a namespace notation, it ends up being not modifiable. So any mocking will fail.

This is tracked in several issues: 
https://github.com/swc-project/swc/issues/5205
https://github.com/swc-project/swc/issues/3843#issuecomment-1061382158
https://github.com/facebook/jest/issues/9430

There seems to be a [plugin](https://github.com/swc-project/swc/discussions/5151) specifically designed to solve this issue but currently overriding .swcrc in Next [is not admitted](https://github.com/vercel/next.js/discussions/30413#discussioncomment-1543513)

For those reasons I decided to - at least temporarily - continue using Babel for the Jest transform

cc @ardelato 

## QA
1. Visit the Vercel preview
2. Verify that there are no inconsistencies between the old Next.js 11 version and this version

PR #633, opened 2022-08-18
Packages: Update scripts/dependencies for monorepo compatibility

jyee27:
Summary:
- Add empty build/clean scripts for packages used in the monorepo
- Add explicit peer/dev dependencies to `ui` package

Rush expects a `build` and `clean` script for every project when running `make` in the monorepo. We were also running into issues with dependencies not loading when using the `@ifixit/ui` package, so this makes them explicit.

CC @andyg0808 @hcutrone


packages/app/package.json
```diff
@@ -10,5 +10,9 @@
       "@types/react-dom": "17.0.11",
       "@ifixit/tsconfig": "workspace:*",
       "typescript": "4.5.3"
+   },
+   "scripts": {
+      "build": "",
+      "clean": ""
```
masonmcelvain, line 16: If we forget to add these scripts to new packages in the future, we probably wouldn't realize it until we tried to update react-commerce in the monorepo. I wonder if there's a simple way we can require that these are specified. But maybe that's more work than it's worth.

packages/ui/package.json
```diff
@@ -14,12 +14,24 @@
    },
    "peerDependencies": {
       "@chakra-ui/react": ">= 1.8.3 < 2",
-      "@core-ds/primitives": ">2.4"
+      "@core-ds/primitives": ">2.4",
+      "@emotion/react": ">=11.0.0 <12.0.0",
+      "framer-motion": ">=3.0.0 <4.0.0 || >=4.0.0 <5.0.0 || >=5.0.0 <6.0.0 || >=6.0.0 <7.0.0",
+      "@emotion/styled": ">=11.0.0 <12.0.0",
+      "react-dom": ">=16.8.6 <17.0.0 || >=17.0.0 <18.0.0",
+      "@babel/core": ">=7.0.0 <8.0.0"
```
masonmcelvain, line 22: I don't understand why several of these are peer dependencies _and_ dev dependencies. Could you explain it or point me to somewhere I can learn about it?

masonmcelvain (approved):
CR ✌🏻 Nice work! deploy_block ✌🏻 because I wasn't clear on peer v. dev deps, and I just wanna make sure that was intentional/best practice.

packages/ui/package.json
andyg0808: Yup: Basically, peer dependencies say, "At runtime, let me have access to the copy of this thing which my parent is using" ("parent" here being the thing which has a dependency on this package). This is useful for libraries like React, where we want to all share a copy, because all the React-using pieces of the system need to interact with the same React instance.

However, this creates a problem when developing the package. During development, there's no parent package (because we're working on the package all on its own). So there's nobody to provide that peer dependency. Thus, it's typical to _also_ include the peer dependency as a dev dependency, so that while developing we can use the dev dependency copy. When installed from NPM, dev dependencies won't be installed, so those modules should have no effect at runtime.

That's the basic logic. All this is complicated by the monorepo environment, where a package can be simultaneously used as a dependency _and_ be being developed. From what we've seen, Rush seems to ignore the peer dependencies of packages within the monorepo (that is, where both the parent and the package live in the monorepo). This is troublesome, and we're still figuring out what to do about it.

masonmcelvain:
un_deploy_block ✌🏻 Thanks for the great explanation @andyg0808!

jyee27:
> @jyee27, this is dev only, right? Can you or someone running into those dependency issues confirm that this fixes it?

We did add peer dependencies in `packages/ui`, so this might change versions of packages outside of dev.
We're using these changes on https://github.com/iFixit/ifixit/tree/homepage--use-new-kpi-banner. This fixes the issues on that branch.

erinemay:
> > @jyee27, this is dev only, right? Can you or someone running into those dependency issues confirm that this fixes it?
> 
> We did add peer dependencies in `packages/ui`, so this might change versions of packages outside of dev. We're using these changes on https://github.com/iFixit/ifixit/tree/homepage--use-new-kpi-banner. This fixes the issues on that branch.

How might I go about applying these changes in a way that I could see them within the context of iFixit/ifixit? Do we have something like an embed mode for this repo? 

We can do a smoke test of the preview to make sure nothing breaks the collection pages here, but I'm not sure how to confirm that the dependency issues are cleared up.

andyg0808:
@erinemay, I'd probably say, don't worry too hard about the iFixit/ifixit context. We've had a  hard time getting the issues to happen consistently, so I'd figure as long as this works in react-commerce, we'll trust the development efforts on the iFixit/ifixit uses to expose any issues.

erinemay:
Ok, I did a smoke test of the collection pages. They're working as expected except for one issue:
dev_block 🧷 
Opening the mobile hamburger menu modal is causing an error
image

Repro: Go to the [prod preview](https://react-commerce-prod-git-packages-add-build-clean-scripts-ifixit.vercel.app/Parts), set screen size to mobile, try to open the hamburger menu. Try on the [main branch prod preview](https://react-commerce-prod.vercel.app/Parts) - able to open menu without errors.

Otherwise, the collection pages continue to work as expected:
- Header and footer 
     - Links to other pages
     - User profile menu
     - Mini cart (modifying cart and checking out)
     - Newsletter signup is recording email to `ifixit_guide.newsletter`
- Toolbar
     - Guides/Answers links
     - Breadcrumbs and breadcrumb toolbar on desktop and mobile (/Parts and /Tools)
- Pagination
- Product cards
     - Title, description, flair, price, discounts
     - List view/tile view
- Navigating up and down the heirarchy
- Back button
- Search
- Filters
- Item Type collections

(Note, some mini cart functionality isn't working in the prod preview, but this is expected and same as main branch. Mini cart functionality confirmed on cominor with development preview)

erinemay:
Interesting... That modal is also the only thing I found problematic in upgrading to Next 12 https://github.com/iFixit/react-commerce/pull/606

jyee27:
un_dev_block :coffee: The mobile menu should be fixed. We're making additional changes on https://github.com/iFixit/ifixit/tree/homepage--use-new-kpi-banner to match the changes here.

k0rvusk0r4x:
- The console errors with the mobile menu are gone.
- Otherwise, smoke tests look good.

QA 😺

masonmcelvain (approved):
CR ✌🏻

PR #643, opened 2022-08-22
Add storefront sdk lib

dhmacs:
This PR adds a Typescript sdk to have typed queries for the Shopify Storefront GraphQL api (similar to what we do for Strapi GraphQL api)

Nothing changed in term of functionality.

PR #646, opened 2022-08-23
Report failing SDK queries to Sentry

masonmcelvain:
Adds Sentry reporting for failed GraphQL queries in our Strapi and Shopify Storefront SDKs.

## QA

It's difficult to mock graphql errors in practice, this is better suited for a test case. Since we'll see these errors in Sentry, we can decide if we more more/less metadata as events start rolling in.

qa_req 0

## Background

Our Strapi and Storefront SDKs use GraphQL to talk to their respective APIs. If the queries failed with errors in the response, we were logging them but not reporting them to Sentry.

https://github.com/iFixit/react-commerce/pull/643#discussion_r952791331

### DIFF
--- a/frontend/lib/shopify-storefront-sdk/index.ts
+++ b/frontend/lib/shopify-storefront-sdk/index.ts
@@ -1,4 +1,5 @@
 import { getSdk, Requester } from './generated/sdk';
+import * as Sentry from '@sentry/nextjs';
 export * from './generated/sdk';
 
 export type ShopCredentials = {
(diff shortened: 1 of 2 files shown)


masonmcelvain:
deploy_block on #643 merging first

masonmcelvain:
> Should we add a test then?

Ideally we'll add tests for all these SDK's, but that's better for a separate issue/PR.

masonmcelvain:
un_deploy_block ✌🏻

danielcliu-ifixit (approved):

PR #648, opened 2022-08-24
Sentry: Report when device handle is `data:image`

masonmcelvain:
We're seeing fetch's sent to our device wiki api that have a device of `data:image`.
https://sentry.io/share/issue/2fb1c3733997425ea19313d477f3d28e/

This is likely a bug in our frontend code, because we shouldn't be making requests like that. This will give us more "context" (pun intended) about what's causing the problem by creating a Sentry event with all the context info.

## CR

I chose to not change any behavior here until we get more context on what's going on, so this just does reporting for now.

## QA

Try to reproduce the original 404 and see if a Sentry event gets reported. If you can't, then we'll skip QA on this and hopefully get some more debug info when it's merged.

Connects #631

### DIFF
--- a/frontend/templates/product-list/index.tsx
+++ b/frontend/templates/product-list/index.tsx
@@ -25,6 +25,7 @@ import {
 import { GetServerSideProps, GetServerSidePropsContext } from 'next';
 import { ParsedUrlQuery } from 'querystring';
 import { getServerState } from 'react-instantsearch-hooks-server';
+import * as Sentry from '@sentry/nextjs';
 import { ProductListView } from './ProductListView';
 
 export type ProductListTemplateProps = WithProvidersProps<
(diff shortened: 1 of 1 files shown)


danielcliu-ifixit:
CR :v:

dhmacs:
@masonmcelvain did you check on Google Analytics (or matomo) if we have visits to url like `/Parts/data:` ?

masonmcelvain:
> @masonmcelvain did you check on Google Analytics (or matomo) if we have visits to url like `/Parts/data:` ?

Just checked and it doesn't appear so.

However, my concern is that the code in `frontend/templates/product-list/index.tsx` is running for urls like `/_next/data:image/png...`. If that were the case, then maybe that's why we're seeing `data:image` as the `deviceHandle` in the Sentry event.

I'm not really sure how Next uses these urls, so it's hard for me to know if this is possible. Looking at the [`context` object](https://nextjs.org/docs/api-reference/data-fetching/get-server-side-props#context-parameter) docs, there's nothing implying that the `params` could ever be wrong, but the `resolvedUrl` field implies that the `context.req.url` is sometimes prefixed with `/_next/data`. That's why I'd like to see these `context`s in Sentry, so I can have a better idea of what these fields are in practice.

masonmcelvain:
@deltuh-vee just noticed this Sentry issue hasn't happened for 10 days. I vote we close this until we start seeing these issues again, we've got more frequent errors to deal with.

Issue #635, opened 2022-08-18
Item Type: Facet selection not moving to the top


jordycosta:
### Overview:

I noticed that selected facets in the **Item Type** filter option don't move up like other single-select options
This can be closed if it is intentional, but I wanted to bring it up as an issue just in case this isn't

![firefox_G0UpvBIeDl](https://user-images.githubusercontent.com/95656772/185481672-3afc5317-311a-4d07-b992-11d58293f0bd.gif)

sterlinghirsh:
It's intentional

sterlinghirsh:
Wait looking again I think there is a bug here since Item Type stays but Device Brand floats. We should at least be consistent. I think leaving them in place makes sense.

PR #649, opened 2022-08-25
Don't Resort Facets When One is Selected

danielcliu-ifixit:
## Overview

We were previously using to the default sorting of algolia for facets, which would sort by `isRefined`, then `count:desc` and `name:asc`. This would rearrange our facets when you would select them, and wasn't consistent with our custom sorting for specific facets. This makes sure we no longer sort by `isRefined` by default.

## QA

Make sure that when clicking on facets that aren't `price_range`, `Capacity`, or `Item Type`, they no longer resort and still work as expected

Closes: #635

### DIFF
--- a/frontend/templates/product-list/sections/FilterableProductsSection/useSortBy.ts
+++ b/frontend/templates/product-list/sections/FilterableProductsSection/useSortBy.ts
@@ -63,6 +63,6 @@ export function useSortBy(
       case 'facet_tags.Item Type':
          return sortAlphabetically;
       default:
-         return props.sortBy;
+         return ["count:desc", "name:asc"];
    }
 }


frontend/templates/product-list/sections/FilterableProductsSection/useSortBy.ts
```diff
@@ -63,6 +63,6 @@ export function useSortBy(
       case 'facet_tags.Item Type':
          return sortAlphabetically;
       default:
-         return props.sortBy;
+         return ["count:desc", "name:asc"];
```
masonmcelvain, line 66: I think this approach makes sense. According to the [docs](https://www.algolia.com/doc/api-reference/widgets/refinement-list/react-hooks/?client=Function+%28TypeScript%29#hook-params-sortby), it looks like we can also update the default ordering when we [configure](https://www.algolia.com/doc/api-reference/api-parameters/renderingContent/?client=php) the index in PHP, but the intent is a little more explicit this way.

masonmcelvain (approved):
CR ✌🏻 Nice!

jordycosta:
QA 🟢 

Single-select facets no longer resort and still work as expected 👍 

---

### Testing Procedure:
1. Navigate to [/Parts](https://react-commerce-prod-git-facets-do-not-raise-isref-b76bde-ifixit.vercel.app/Parts) or [/Tools](https://react-commerce-prod-git-facets-do-not-raise-isref-b76bde-ifixit.vercel.app/Tools)
2. Click on a single-select facet shown in [the issue](https://github.com/iFixit/react-commerce/issues/635)
3. Observe that the facet no longer jumps to the top of the list

---

### Screen Capture:

screenshot

frontend/templates/product-list/sections/FilterableProductsSection/useSortBy.ts
danielcliu-ifixit: I like us defining it explicitly here, considering the rest of the sorts are also defined in this file

PR #651, opened 2022-08-25
Handle deviceTitles with Multiple Spaces In Them

danielcliu-ifixit:
## Overview

We have some sentry errors like [this](https://sentry.io/organizations/ifixit/issues/3504489498/events/053ed493b7db4427868e074dec9ddc18/?project=6469069) where the page was failing because we had multiple spaces in the `deviceTitle`/`wiki_title`. This change makes sure when we encode/decode `deviceTitle` we maintain a character per space.

## QA

Make sure that you can click on `Samsung Galaxy A8+ (2018)` from [here](http://localhost:3000/Parts/Samsung_Galaxy_A) and it now works instead of 404ing. Also make sure other pages aren't broken i guess.

Connects: #632

### DIFF
--- a/frontend/helpers/product-list-helpers.ts
+++ b/frontend/helpers/product-list-helpers.ts
@@ -42,14 +42,14 @@ export function encodeDeviceItemType(itemType: string): string {
  * Convert URL slug to product list device title
  */
 export function decodeDeviceTitle(handle: string): string {
-   return handle.replace(/_+/g, ' ');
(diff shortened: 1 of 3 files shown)


deltuh-vee:
QA 🎬 
Parts pages for devices with multiple spaces in their titles are now accessible and don't 404.

hcutrone (approved):
CR :burrito: 
Nice fix!

PR #654, opened 2022-08-25
Encode Canonical URL

danielcliu-ifixit:
## Overview

We've had some errors like [this](https://sentry.io/organizations/ifixit/issues/3504489498/events/d9719aa880c2447abada45d57cdcf3fe/?project=6469069) where there seems to be some sort of link that is not encoded correctly. This bug could potentially come from our canonical url not being encoded, so this PR just makes sure that we encode the url correctly.

## CR
is there any harm in url encoding these urls?

## QA
Make sure that the links still work I guess. 
Check on David's comment: https://github.com/iFixit/react-commerce/pull/654#issuecomment-1235663872

Connects #632


danielbeardsley:
Thanks for looking into this. I think the sentry error immediately highlights the problem which you probably saw:
image

But after looking around on that particular page, I found no problems.

All the times the url in question was in an html attribute, it was encoded correctly:

```
<link rel="canonical" href="https://www.ifixit.com/Parts/MacBook_Pro_15&quot;_Core_2_Duo_Models_A1226_and_A1260"/>
```
image

i.e. the double quote inside the attribute value is correctly encoded: `&quot;`

And when it was in json, it was also encoded correctly (for json):
image

So, I don't think we're doing anything wrong here. I *think* this is a bad browser (or bot) that's just hunting for strings that look like `https://` and visiting them without decoding the urls based on context. I don't want to risk messing up our canonical urls and I don't think this change will fix the problem cause the urls in the app already look correct. It's possible there's still a bug somewhere (wrongly encoding a url in some link) but I'd want us to demonstrate a broken-ness before we try to fix it.

dev_block :+1:

danielcliu-ifixit:
> I think this is a bad browser (or bot)

This was my first guess as well

However is it possible that the canonical url is not being decoded and thats why this error is being thrown though?

danielbeardsley:
> However is it possible that the canonical url is not being decoded and thats why this error is being thrown though?

Feel free to add some `console.log()` to the code to check the generated canonical if you're unsure. `encodeUri()` is a weird function but it will *url-encode* (i.e. with `%`) characters that are not allowed in a url, so it wouldn't every produce `&quot;`. The generated HTML is correct (a double quote needs to be html encoded in attribute values).

Generally, needing to use `encodeUrl()` would be indicative of a bug (i.e. not building a proper url by encoding pieces of it).

---

We should change something, but it should be that we *don't* report 404s.

Right now, when I visit:
https://www.ifixit.com/Parts/MacBook_Pro_is_nonsense

I see the [same error in sentry](https://sentry.io/organizations/ifixit/issues/3504489498/events/e49a9029757f4388bfcda382cccf63d6/?project=6469069):
image

frontend/templates/product-list/MetaTags.tsx
```diff
@@ -37,9 +37,11 @@ export function MetaTags({ productList }: MetaTagsProps) {
    }
    title += ' | iFixit';
    const itemTypeHandle = itemType ? `/${encodeDeviceItemType(itemType)}` : '';
-   const canonicalUrl = `${appContext.ifixitOrigin}${
-      productList.path
-   }${itemTypeHandle}${page > 1 ? `?${PRODUCT_LIST_PAGE_PARAM}=${page}` : ''}`;
+   const canonicalUrl = encodeURI(
+      `${appContext.ifixitOrigin}${productList.path}${itemTypeHandle}${
+         page > 1 ? `?${PRODUCT_LIST_PAGE_PARAM}=${page}` : ''
+      }`
```
sterlinghirsh, line 43: I think I'd rather encodeURIComponent on just the ItemTypeHandle (probably inside encodeDeviceItemType) rather than encoding the whole url. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI There's a subtle difference between the two functions and in this case it might not matter (at least most of the time) but I don't really trust encoding the whole url.

frontend/templates/product-list/ProductListDeviceNavigation.tsx
```diff
@@ -42,12 +42,12 @@ export function ProductListDeviceNavigation({
       >
          <SecondaryNavbarItem isCurrent>Parts</SecondaryNavbarItem>
          <SecondaryNavbarItem>
-            <NextLink href={guideUrl} passHref>
+            <NextLink href={encodeURI(guideUrl)} passHref>
```
sterlinghirsh, line 45: Same thing here. Instead of encoding the whole url, let's encode deviceHandle on line 22. In fact, we have a function encodeDeviceTitle that we should probably be using.

sterlinghirsh:
@danielbeardsley I see your point, though I think I'd still rather urlencode the segment than let the htmlencoding do it.

danielcliu-ifixit:
un_dev_block :v:

davidrans:
I think we need to double-check a couple things:
  - how many canonicals does this end up changing?
  - do the canonical URLs still match the URLs in the sitemap?

frontend/templates/product-list/MetaTags.tsx
```diff
@@ -36,16 +37,21 @@ export function MetaTags({ productList }: MetaTagsProps) {
...
    title += ' | iFixit';
-   const itemTypeHandle = itemType ? `/${encodeDeviceItemType(itemType)}` : '';
-   const canonicalUrl = `${appContext.ifixitOrigin}${
-      productList.path
-   }${itemTypeHandle}${page > 1 ? `?${PRODUCT_LIST_PAGE_PARAM}=${page}` : ''}`;
+   const itemTypeHandle = itemType
+      ? `/${encodeDeviceItemTypeSpaces(itemType)}`
+      : '';
+   const deviceAndItemTypePath = encodeDevicePath(
+      `${productList.path}${itemTypeHandle}`
```
danielbeardsley, line 44: I think this is a pretty odd bit of code and seeing it would make me question a bunch of surrounding code. If we know we have pieces of a url that need to be url-encoded, we should encode them *before* building a url. i.e.

`path = encodeUriComponent(piece1) + "/" + encodeUriComponent(piece2);`

k0rvusk0r4x:
The links appear to still work.

QA 😺

frontend/templates/product-list/MetaTags.tsx
sterlinghirsh: I agree. We already had encodeDeviceItemType and encodeDeviceTitle, so why make a new function that splits them apart and then encodes them?

sterlinghirsh:
CR :zap: dev_block :zap: I think we can have one function for encoding the item type and maybe another for encoding the device title. Not sure if they practically have any difference but whatever. But I don't think I want to have one that splits a path and then encodes each element and rejoins it. Just seems roundabout and like asking for trouble. Or maybe I'm just not seeing what the reasoning is here?

frontend/templates/product-list/MetaTags.tsx
danielcliu-ifixit: the `encodeDeviceItemType` and `encodeDeviceTitle` were not actually url encoding they were simply replacing ' ' with '_'. And theyre being used all over the place to basically just do that, and thats used in things like `const noIndexExemptions` which i wasn't sure if we wanted to change its readability to making all the keys url encoded, especially when it seems unnecessary. Which is why theres a commit to change the names so they are more accurate to what theyre doing.

As for why i join them and then encode, its because the path contains the entire base route so like `/Parts/MacBook_Air_13"` would be route. So we have to split it anyway, because encoding that will ruin the `/`s and break the route. Which is why originally i was using `encodeURI()`. And if we have to split them anyway and do the same exact thing to both the `Path` and `Item Type` I don't see a disadvantage to condensing that to one call.

sterlinghirsh: I think if there are places we're using encodeDevice* in a way that doesn't work with urlencoding, we should take another look and maybe remove that. I'm not sure why we would want to encode it partway. For example, in the noIndexExemptions, we should store the decoded values so we don't have to encode before looking it up. Then we can only encode immediately before creating a url, and we can encode the device and item type separately.

danielcliu-ifixit:
un_dev_block :v:

danielbeardsley (dismissed):
CR :+1: 

@danielcliu-ifixit and I worked through this until we understood everything and then made a few changes (including another function rename :smile: )

erinemay:
Links still work/urls still update appropriately:
- Choose a Model
- Item type lists
- Adding search facets
- Parts/Guides (top level and nested devices)
- Breadcrumb links
- All pagination buttons
- Product links (prod goes to ifixit product page, non-prod goes to new react-commerce pages)
- Search
- View cart
- Modify cart contents via mini cart

dev_block 👟 
Checkout is different on this branch.
When I try to check out from the mini cart on this [branch preview](https://react-commerce-git-encode-canonical-uri-ifixit.vercel.app/Parts), I'm redirected to https://www.ifixit.com/Store and no checkout is initiated. When I try to check out from the mini cart on the [master preview](https://react-commerce.vercel.app/Parts), it starts checkout on ifixit-test as expected.

Note: Mini cart checkout on the prod preview continues to not work - same on master.

erinemay:
Ok tried again with a fresh main merge and it's still not working for me. 

With @danielcliu-ifixit, confirmed that the error I'm seeing can also be triggered on the main branch. 
To do so, add https://www.cominor.com/Store/Android/Galaxy-Note20-Ultra-Battery-Adhesive/IF463-001?o=1 to your cart and try to check out from the mini cart on https://react-commerce.vercel.app/Parts

🛑 But I'm still seeing the error on this branch only for other products:

Repro steps:

Can use any of the following user states:
- Logged in user without unique username
- Logged in user with unique username
- Logged out 

Product: Mako kit - https://www.cominor.com/Store/Tools/Mako-Driver-Kit--64-Precision-Bits/IF145-299?o=4

Add only the Mako to your cart and try checking out from the mini cart on main and this branch

Checking out on the main branch (https://react-commerce.vercel.app/Parts): Starts ifixit-test (store.cominor) checkout as expected ✔️ 

Checking out on this branch ([direct to latest build](https://react-commerce-riyvliqrc-ifixit.vercel.app/Parts), or the [static branch url](https://react-commerce-git-encode-canonical-uri-ifixit.vercel.app/Parts)): 
I get the "Checkout unavailable. Please contact..." message in the UI and the following errors in the console:
image

danielbeardsley:
Good job being thorough @erinemay !!

Given that you were able to capture the error: "Variant is invalid" I'm 100% certain it's not related to this branch. This branch doesn't affect which variant you put in your cart nor does it affect the checkout process. It mostly deals with urls. I'm fairly certain there's no functional change that users would notice with this branch, just that some canonicals will be url-encoded in a more compatible way.

I'm betting it has to do with the particular products that were chosen to checkout with... Though the behavior being different across branches is more concerning.

danielcliu-ifixit:
ok @erinemay i looked into that bug and its because our `shopifyClient.shopDomain` is set to `store.ifixit.com` for the branches not on master for whatever reason. Im not sure where that is set, but I am sure that wasn't changed as part of this pull.

un_dev_block :v:

danielbeardsley:
> `shopifyClient.shopDomain` is set to store.ifixit.com for the branches not on master for whatever reason

Looks like this is set to `store.shopify.storefrontDomain` from the `us` store in strapi at https://{branch}.govinor.com

Where that comes from I don't know, but this branch didn't change that value.

danielcliu-ifixit:
Also note, the bug that Erin references is visible on all other pull branches rn, as well as replicate-able on all branches locally including `main`.

dhmacs:
> shopifyClient.shopDomain is set to store.ifixit.com for the branches not on master for whatever reason

This is due to the fact that govinor instances are seeded using production data (from strapi.ifixit.com). So for now you need to set storefront domain and storefront access token whenever you create a new branch instance (take those value from the main.govinor.com).
We're thinking how we could improve the seed, but right now the only way is to go and set those value for every new branch.

cc @danielcliu-ifixit @danielbeardsley

frontend/templates/product-list/index.tsx
```diff
@@ -5,9 +5,10 @@ import {
 } from '@components/common';
 import { ALGOLIA_PRODUCT_INDEX_NAME } from '@config/env';
 import {
-   decodeDeviceItemType,
-   decodeDeviceTitle,
-   encodeDeviceTitle,
+   destylizeDeviceItemType,
+   destylizeDeviceTitle as destylizeDeviceTitle,
```
sterlinghirsh, line 9: ?

sterlinghirsh:
CR :zap: I'm still not sure I agree that we need to keep these encode and stylize separate long-term since it looks like there's only one place where we're using stylize without encode, but I'm more interested in merging this and moving on than sending it around again. We can always clean it up later if we want.

sterlinghirsh (approved):

k0rvusk0r4x:
- Links working nicely.
- Oddities above explained.

QA 😺

frontend/templates/product-list/index.tsx
danielcliu-ifixit: must have been some auto import bug

PR #658, opened 2022-08-28
Log findProductList Failures With More Context

danielcliu-ifixit:
## Overview

We've had some errors like [this](https://sentry.io/organizations/ifixit/issues/3504489498/events/d9719aa880c2447abada45d57cdcf3fe/?project=6469069) where there seems to be some sort of link that is not encoded correctly. To investigate this more lets try to get the request/response context with the error for sentry.

## QA
Check if bad devices (so like `/Parts/badItemHere`) reports to sentry with a `Extra Info` json.

Connects #632

### DIFF
--- a/frontend/helpers/next-helpers.ts
+++ b/frontend/helpers/next-helpers.ts
@@ -1,11 +1,20 @@
 import { GetServerSideProps } from 'next';
 import { logAsync } from '@ifixit/helpers';
 import { setSentryPageContext } from '@ifixit/sentry';
+import * as Sentry from '@sentry/nextjs';
 
 export function serverSidePropsWrapper<T>(
    getServerSidePropsInternal: GetServerSideProps<T>
 ): GetServerSideProps<T> {
    return async (context) => {
+      Sentry.setContext("Extra Info", {
+         headers: context?.req.headers,
+         url: context?.req.url,
+         method: context?.req.method,
+         locale: context?.locale,
+         ...context?.params,
+         ...context?.query,
+       });
       return logAsync('getServerSideProps', () =>
          getServerSidePropsInternal(context)
       ).catch((err) => {


deltuh-vee:
QA 🎬 
Context object is now included in Sentry reports

CleanShot 2022-08-29 at 09 33 46

danielcliu-ifixit:
dev_block :v: rethinking if we want entire context sent

danielcliu-ifixit:
un_dev_block :v: I think its good to review again

deltuh-vee:
dev_block ✊ 
I'm not sure if these are the values that we would expect from visiting something like `/Parts/noitem`.
CleanShot 2022-08-29 at 15 16 02

danielcliu-ifixit:
@deltuh-vee hmm those are from the requests I made while testing this. Weird

danielcliu-ifixit:
@deltuh-vee yours looks like it made it [here](https://sentry.io/organizations/ifixit/issues/3541752609/events/f1538d1bf7364ef0bc5cf56743f8b50d/?project=6475315)?

deltuh-vee:
Carry over QA 🎬 
un_dev_block 👍

frontend/lib/ifixit-api/devices.ts
```diff
@@ -16,6 +20,15 @@ export async function fetchDeviceWiki(
       );
       return await client.get(`cart/part_collections/devices/${deviceHandle}`);
    } catch (error: any) {
+      Sentry.withScope((scope) => {
+         scope.setExtra('headers', context?.req.headers);
+         scope.setExtra('url', context?.req.url);
+         scope.setExtra('method', context?.req.method);
+         scope.setExtra('device', deviceTitle);
+         scope.setExtra('local', context?.locale);
+         scope.setExtra('query', context?.query);
```
jarstelfox, line 29: I'd think about moving these to a context. It looks like `setExtra` is deprecated: https://docs.sentry.io/platforms/javascript/enriching-events/context/#additional-data

danielcliu-ifixit:
Ok new solution is much cleaner, in terms of code and also producing redundant errors in sentry.

sterlinghirsh:
CR :zap: though is there a reason not to put this in the existing setSentryPageContext function?

sterlinghirsh:
deploy_block :zap: on a question

sterlinghirsh (approved):

danielcliu-ifixit:
@sterlinghirsh originally thats what I tried, but it seems like 404 is not an error in the system so it doesn't actually call that function normally. I can move the function but didn't feel like that was necessary and I wasn't sure if the context is needed when it doesn't error out

sterlinghirsh:
OK that makes sense. We can try it like this for now and see if it adds the context we want, and then we can change it later if we want.

sterlinghirsh:
un_deploy_block :zap:

deltuh-vee:
Carry over QA 🎬

PR #659, opened 2022-08-29
Add Ability to Ignore Cache (Maybe)

### LINKED ISSUE #548: Make something like disableCacheGets
When updating pages in strapi, it would be nice if there was an option to bust the cache on nextjs. Maybe this is something that can be triggered on save, but it would also be nice to be able to force a cache miss.

danielcliu-ifixit:
## Overview

This pull pertains to Vercel and `react-commerce`, and doesn't directly handle any Cloudfront related caching. Hopefully they respect `cache-control` headers.

Allow the `_vercel_no_cache=1` to disable use of the [vercel cache ](https://vercel.com/docs/concepts/edge-network/caching#cacheable-responses)for that request. Also makes it so that response has `cache-control` [headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) that should indicate to the browser to no cache. Also clears the `react-commerce` cache. 

## QA 
its pretty easy to check if vercel is ignoring its caches. You can go the `react-commerce-prod` [instance](https://vercel.com/ifixit/react-commerce-prod/8FoSXqTSmrgn43XKwDF8nb3xne3m/functions) of this branch's functions, verify that going to a page like `/Parts/iPhone` has cache misses like `Cache miss for 'storeList': 52ms` on the first call, but subsequent should not miss. Then if you add `_vercel_no_cache=1` as a parameter then it will always miss the caches. 
You should make sure that the `_vercel_no_cache` parameter works as intended, stays in the url and doesn't break. NOTE: I **now** have this set so it does need be set to `1` to work, any other string shouldnt work.

Also make sure that the prod's headers for `cache-control` are changed to `'no-store, no-cache, must-revalidate'` when you have a `_vercel_no_cache=1`.

I'm not sure how to test the cache on the `react-commerce` side honestly. I'm not sure how to create a scenario where the cache is wrong. Strapi seemed to update the data fine when messed with. 

Closes #548 until further notice

### DIFF
--- a/frontend/components/common/InstantSearchProvider.tsx
+++ b/frontend/components/common/InstantSearchProvider.tsx
@@ -161,24 +161,21 @@ export function InstantSearchProvider({
                // Item Type is the slug on device pages, not in the query.
                delete filterCopy['facet_tags.Item Type'];
             }
+            const customQueryParams = new URLSearchParams(location.search);
             const queryString = qsModule.stringify(
-               { ...routeState, filter: filterCopy },
+               {
+                  ...routeState,
+                  filter: filterCopy,
+                  _vercel_no_cache:
+                     customQueryParams.get('_vercel_no_cache') || undefined,
+               },
                {
                   addQueryPrefix: true,
                   arrayFormat: 'indices',
                }
             );
 
-            let customQueries = '';
-            const customQueryParams = new URLSearchParams(location.search);
-            if (customQueryParams.has('_vercel_no_cache')) {
-               customQueries =
-                  (queryString ? '&' : '?') +
-                  '_vercel_no_cache=' +
-                  customQueryParams.get('_vercel_no_cache');
-            }
-
-            return `${baseUrl}${path}${queryString}${customQueries}`;
+            return `${baseUrl}${path}${queryString}`;
          },
          parseURL({ qsModule, location }) {
             const pathParts = location.pathname
--- a/frontend/components/common/InstantSearchProvider.tsx
+++ b/frontend/components/common/InstantSearchProvider.tsx
@@ -169,7 +169,13 @@ export function InstantSearchProvider({
                }
             );
 
-            return `${baseUrl}${path}${queryString}`;
+            let customQueries = '';
+            const customQueryParams = new URLSearchParams(location.search);
+            if (customQueryParams.has('disableCacheGets')) {
+               customQueries = (queryString ? '&' : '?') + 'disableCacheGets';
+            }
+
+            return `${baseUrl}${path}${queryString}${customQueries}`;
          },
          parseURL({ qsModule, location }) {
             const pathParts = location.pathname
--- a/frontend/components/common/InstantSearchProvider.tsx
+++ b/frontend/components/common/InstantSearchProvider.tsx
@@ -171,8 +171,11 @@ export function InstantSearchProvider({
 
             let customQueries = '';
             const customQueryParams = new URLSearchParams(location.search);
-            if (customQueryParams.has('disableCacheGets')) {
-               customQueries = (queryString ? '&' : '?') + 'disableCacheGets';
+            if (customQueryParams.has('_vercel_no_cache')) {
+               customQueries =
+                  (queryString ? '&' : '?') +
+                  '_vercel_no_cache=' +
+                  customQueryParams.get('_vercel_no_cache');
             }
 
             return `${baseUrl}${path}${queryString}${customQueries}`;
--- a/frontend/templates/product-list/index.tsx
+++ b/frontend/templates/product-list/index.tsx
@@ -55,7 +55,7 @@ export const getProductListServerSideProps = ({
    productListType,
 }: GetProductListServerSidePropsOptions): GetServerSideProps<ProductListTemplateProps> => {
    return async (context) => {
-      if ('disableCacheGets' in context.query) {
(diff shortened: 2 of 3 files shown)


frontend/templates/product-list/index.tsx
```diff
@@ -55,10 +56,18 @@ export const getProductListServerSideProps = ({
...
 }: GetProductListServerSidePropsOptions): GetServerSideProps<ProductListTemplateProps> => {
    return async (context) => {
-      context.res.setHeader(
-         'Cache-Control',
-         'public, s-maxage=10, stale-while-revalidate=600'
-      );
+      if (context.query._vercel_no_cache) {
+         context.res.setHeader(
+            'Cache-Control',
+            'no-store, no-cache, must-revalidate'
```
deltuh-vee, line 62: Why is it not enough to just have `no-store`? Doesn't setting `no-store` contradict `no-cache` and `must-revalidate`?

danielcliu-ifixit: It's not contradictory, its redundant. And its like this because from what people said on SO some browsers don't accept all of them so this should cover all our bases. In addition, because this is basically just for in house use, I don't have a problem with some slight header redundancy.

frontend/components/common/InstantSearchProvider.tsx
```diff
@@ -169,7 +169,16 @@ export function InstantSearchProvider({
...
+            let customQueries = '';
+            const customQueryParams = new URLSearchParams(location.search);
+            if (customQueryParams.has('_vercel_no_cache')) {
+               customQueries =
+                  (queryString ? '&' : '?') +
+                  '_vercel_no_cache=' +
+                  customQueryParams.get('_vercel_no_cache');
+            }
+
+            return `${baseUrl}${path}${queryString}${customQueries}`;
```
andyg0808, line 181: Is there a reason that this isn't integrated into the `qsModule.stringify` call above? Something like this might work:
```js
            const queryParams = new URLSearchParams(location.search);
            const cacheParamName = '_vercel_no_cache'
            const cacheParam = queryParams.get(cacheParamName) || undefined;

            const queryString = qsModule.stringify(
               { ...routeState, filter: filterCopy, [cacheParamName]: cacheParam },
               {
                  addQueryPrefix: true,
                  arrayFormat: 'indices',
               }
            );
```

[These seem to be the `stringify` docs](https://github.com/ljharb/qs#stringifying); found via [Algolia's docs](https://www.algolia.com/doc/guides/building-search-ui/going-further/routing-urls/react/#createurlsearchstate).

andyg0808 (changes requested):

andyg0808:
This looks pretty good to me, but I'd try to avoid the manual string munging for the query parameters.

dev_block :+1: but CR otherwise.

danielcliu-ifixit:
> So the query parameter shouldn't disappear when clicking on a link to somewhere else on the react-commerce page(e.g. iPhone Parts -> iPhone 13 Parts)?

It doesn't do that on ifixit so im ok with it not following here either

frontend/components/common/InstantSearchProvider.tsx
masonmcelvain, line 181: What you have seems like it works. However, I'd rather pass everything through `qsModule` just to be safe, so we don't get mixed up with `&` and `?` and url encoding and such.

```ts
            const params = new URLSearchParams(location.search);
            const vercelCache = params.has('_vercel_no_cache')
               ? { _vercel_no_cache: params.get('_vercel_no_cache') }
               : {};

            const queryString = qsModule.stringify(
               { ...routeState, filter: filterCopy, ...vercelCache },
               {
                  addQueryPrefix: true,
                  arrayFormat: 'indices',
               }
            );

            return `${baseUrl}${path}${queryString}`;
```

masonmcelvain (approved):
Looks like a reasonable approach! I'm gonna defer to @sterlinghirsh for CR since he requested the feature, just to make sure this is what he had in mind.

frontend/components/common/InstantSearchProvider.tsx
masonmcelvain: > Properties that are set to undefined will be omitted entirely

I like this better than my suggestion!

masonmcelvain: I prefer @andyg0808's solution actually.

masonmcelvain:
> doesn't directly handle any Cloudfront related caching. Hopefully they respect cache-control headers.

I think these are the cloudfront [caching](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Expiration.html) docs. I remember having a discussion about CloudFront not respecting the `must-revalidate` header, but I could be wrong about that.

andyg0808:
I'm out for the rest of the week; feel free to dismiss my review once it's addressed and let @masonmcelvain re-review.

danielcliu-ifixit:
@masonmcelvain the docs say that CloudFront and browsers respect the `no-cache`, `no-store` headers.

un_dev_block :v:

deltuh-vee:
QA 🎬 
* Adding `_vercel_no_cache=1` as a parameter causes the cache to always miss.
* The Vercel response headers for cache-control are `no-store, no-cache, must-revalidate, stale-if-error=0` when `_vercel_no_cache=1` is set.

masonmcelvain (approved):
CR ✌🏻

PR #664, opened 2022-08-30
Mobile Productlist: Scroll on facet drawer close

### LINKED ISSUE #618: Fix scroll on facet selection
This was noticed here https://github.com/iFixit/react-commerce/issues/579#issuecomment-1209078103
On mobile/tablet view, when the user selects facets and exits out of the the facet list, the page scrolls back to the top. Instead, we want to bring the page to the top of the section where we start seeing the filters followed by the products, ie here instead of the top of the page:

- Some ideas:
   - We can have the page scroll to the desired position when we exit of the facet drawer [here](https://github.com/iFixit/react-commerce/blob/90218f6642b6683073a38de3ec1134285d33b071/frontend/components/product-list/sections/FilterableProductsSection/FacetsDrawer.tsx#L124)
       - If we go with this then we should probably do the same thing for the `Apply` button.
@sterlinghirsh

masonmcelvain:
When the facet drawer is closed on mobile, the view now scrolls so that the toolbar is at the top of the viewport.

## QA
- Preview /Parts on mobile.
- Click on "Filters" button.
- After closing the dialog, the view should have the number of results at the top of the screen. See the issue for a screenshot of the desired behavior.
- When the page first renders, nothing should be different (ie, there is no scroll effect).

CC @sumukhipandey

Closes #618

### DIFF
--- a/frontend/templates/product-list/sections/FilterableProductsSection/FacetsDrawer.tsx
+++ b/frontend/templates/product-list/sections/FilterableProductsSection/FacetsDrawer.tsx
@@ -146,9 +146,7 @@ export function FacetsDrawer({
                            attribute={facet}
                            isOpen={facet === currentFacet}
                            productList={productList}
-                           onClose={() => {
-                              setCurrentFacet(null);
-                           }}
+                           onClose={() => setCurrentFacet(null)}
                         />
                      );
                   })}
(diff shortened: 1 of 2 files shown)


jordycosta:
QA 🟢 

After closing the filter list on mobile; the view is moved to the start of the results, and not the top of the page. 👍 

---

### Testing Procedure:
1. Navigate to [`react-commerce-prod`](https://react-commerce-prod-git-scroll-on-facet-drawer-close-ifixit.vercel.app/) on mobile
2. Click on **Filters**
3. Close the drawer or apply a filter
4. Ensure that the view is moved to the start of the results

---

### Screen Capture:

img_4645

hcutrone (commented):
CR :burrito:

hcutrone (approved):

PR #665, opened 2022-08-30
Actions: Auto-commit formatting diff

masonmcelvain:
Much like we do for updating the psalm baseline, this automatically commits formatting changes produced by prettier every time a dev pushes to their branch.

The intent is to prevent formatting diffs sneaking through to `main`, which causes diffs on unrelated feature branches.

## QA

- Make a commit to this branch that would cause a diff if you ran `pnpm format`.
- Push the commit to this branch.
- The action should push a new commit that applies the diff.

Closes #616

Issue #616, opened 2022-08-16
Linting: Add a GitHub Action


masonmcelvain:
## Problem

We have a linter that runs as a post-commit hook or as an npm script `pnpm run format`. Sometimes, code sneaks into `main` that hasn't been formatted, and it causes unrelated code changes on feature branches if developers commit the formatting changes.

## Do this

Add a GitHub action that runs our linter. It could fail when there is a linting diff (forcing the dev to format and commit). Alternatively, it could apply the formatting changes in an automated commit, like we do for psalm baseline fixes in iFixit/ifixit.

PR #671, opened 2022-08-31
GH Actions: Skip `format` workflow for `main`

masonmcelvain:
The "Fail in production" step [failed](https://github.com/iFixit/react-commerce/runs/8120146230?check_suite_focus=true) even when there was no diff for some reason.

Let's use a more idiomatic pattern to skip workflows for the main branch.
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-branches-and-tags

## QA
Same as #665

Connects #616

### DIFF
--- a/frontend/sentry.client.config.ts
+++ b/frontend/sentry.client.config.ts
@@ -21,7 +21,5 @@ Sentry.init({
          'next.runtime': 'client',
       },
    },
-   ignoreErrors: [
-      'TypeError: NetworkError when attempting to fetch resource.'
-   ],
+   ignoreErrors: ['TypeError: NetworkError when attempting to fetch resource.'],
 });
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,6 +1,9 @@
 name: Format
 
-on: [push]
+on:
+   push:
+      branches-ignore:
+         - 'main'
 
 jobs:
    prettier:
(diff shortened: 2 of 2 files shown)


masonmcelvain:
QA ✌🏻 Still works!

frontend/sentry.client.config.ts
```diff
@@ -21,7 +21,5 @@ Sentry.init({
          'next.runtime': 'client',
       },
    },
-   ignoreErrors: [
-      'TypeError: NetworkError when attempting to fetch resource.'
-   ],
+   ignoreErrors: ['TypeError: NetworkError when attempting to fetch resource.'],
```
ardelato, line 24: Was this linter action supposed to be triggered?

ardelato: Basically, I see that the action just runs prettier and if there were changes it commits those changes. However, since your pull did not modify `frontend/sentry.client.config.ts` I am assuming this file was pushed to `main` without it being formatted first and is now being formatted with the GH linter?

ardelato (approved):
CR 👍 with question

frontend/sentry.client.config.ts
masonmcelvain: Correct, this file was pushed to main without being formatted. The action formatted it on this branch and pushed it in a new commit.

Going forward, we shouldn't encounter this situation because feature branches will be formatted by the action before being merged into main.

Issue #533, opened 2022-07-28
Results shown outside price range selected


jordycosta:
### Overview:

When selecting one of the price range facet options, there are some results that are displayed to users that are out of bounds from what they chose. These results are included because the product has either a Fix Kit or a Part Only option that is included in that price range.

![price_range](https://user-images.githubusercontent.com/95656772/181648006-5654db60-b88a-4bef-9b8d-f6e236d3aee3.jpg)

![part-only](https://user-images.githubusercontent.com/95656772/181648247-3a332569-8d8f-4539-a32b-08d06213beb5.jpg) ![fix-kit](https://user-images.githubusercontent.com/95656772/181648254-096fb737-e8f2-45a1-adfa-1fa554e7bb5b.jpg)

---

masonmcelvain:
Related: https://github.com/iFixit/ifixit/pull/43862

sterlinghirsh:
@mmarcon91 Can we experiment with putting a price range on the product card again?

dhmacs:
@sterlinghirsh Just to double check, is it ok that we're using `price_float` instead of for example `group_min_price` to display the product price?

mmarcon91:
> @mmarcon91 Can we experiment with putting a price range on the product card again?

I do some research to understand how to manage different prices. My concern is when we have to show discounts, it would be a bit confusing to have two ranges

sterlinghirsh:
>  is it ok that we're using price_float instead of for example group_min_price to display the product price?

I think it's okay, though it's the price of the default variant, which isn't necessarily the lowest. It's confusing when it shows one price on search but another price after clicking the link.

mmarcon91:
@sterlinghirsh I did a bit of research to understand how other sites handle this case and actually their approach is as you suggested to have min and max price.
I think it is the most effective solution.

I took the opportunity to create a new component of the product card to keep as a reference. prices have been resized to avoid taking up too much space.
https://www.figma.com/file/sfj6r7Bn6eREMvuqo0f7Du/iFixit---Components?node-id=308%3A17651

Here is a preview of the collection with the updated product cards.
https://www.figma.com/file/jRG4EyQXRDMGfyLqut08KP/iFixit---Working-files?node-id=5677%3A63004

federicobadini:
@sterlinghirsh can I go on implementing the price range visible the mockup above ☝️ ?

sterlinghirsh:
Sure let's give it a shot

PR #673, opened 2022-08-31
ProductPrice component

### LINKED ISSUE #662: Align price appearance to mockup
The price field in the product section should be updated to match the one in the mockup:

federicobadini:
close #662 

This PR introduce a new ProductPrice component.

The product page heading will use the red color scheme in horizontal direction as per mockup:

Screenshot 2022-08-30 at 17 54 13

The component can however be configured to cover all the layouts visible in the mockup:

Screenshot 2022-08-31 at 23 02 17

So, it may already cover most - if not all - the use cases that are being discussed [here](https://github.com/iFixit/react-commerce/issues/533). 

## QA
1. Open Vercel preview
2. Go to [here](https://react-commerce-git-enrich-price-component-ifixit.vercel.app/Products/macbook-air-13-late-2010-2017-replacement-battery) to see a discounted product
3. Go to [here](https://react-commerce-git-enrich-price-component-ifixit.vercel.app/Products/iphone-11-screen) to see a product without discount

### DIFF
--- a/frontend/components/common/ProductPrice.tsx
+++ b/frontend/components/common/ProductPrice.tsx
@@ -1,0 +1,79 @@
+import { Box, Text, ThemeTypings, useTheme } from '@chakra-ui/react';
+import { IconBadge } from '@components/ui';
+import { faRectanglePro } from '@fortawesome/pro-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+
+export type ProductPriceProps = {
+   formattedPrice: string;
+   formattedCompareAtPrice?: string | null;
+   isDiscounted: boolean;
+   discountLabel?: string;
+   showDiscountLabel?: boolean;
+   showProBadge?: boolean;
+   size?: 'large' | 'medium' | 'small';
+   colorScheme?: ThemeTypings['colorSchemes'];
+   direction?: 'row' | 'column' | 'column-reverse';
+};
+
+export function ProductPrice({
+   formattedPrice,
+   formattedCompareAtPrice,
+   isDiscounted,
+   discountLabel,
+   showDiscountLabel = true,
+   showProBadge = false,
+   size = 'large',
+   colorScheme = 'red',
+   direction = 'row',
+}: ProductPriceProps) {
+   const theme = useTheme();
+   const priceFontSize =
+      size === 'large' ? 'xl' : size === 'medium' ? 'md' : 'sm';
+   const compareAtPriceFontSize = size === 'large' ? 'md' : 'sm';
+
+   return (
+      <Box
+         display="flex"
+         flexDir={direction}
+         alignSelf="flex-start"
+         alignItems={direction === 'row' ? 'center' : 'flex-end'}
+      >
+         <Text
+            mr={direction === 'row' ? 1 : 0}
+            fontSize={priceFontSize}
+            fontWeight="semibold"
+            color={isDiscounted ? theme.colors[colorScheme][600] : 'gray.900'}
+         >
+            {showProBadge && direction !== 'row' && (
+               <FontAwesomeIcon
+                  icon={faRectanglePro}
+                  size="1x"
+                  color={theme.colors[colorScheme][500]}
+                  style={{ marginRight: '6px' }}
+               />
+            )}
+            {formattedPrice}
+         </Text>
+         {isDiscounted && (
+            <>
+               <Text
+                  mr={direction === 'row' ? '10px' : 0}
+                  fontSize={compareAtPriceFontSize}
+                  color="gray.500"
+                  textDecor="line-through"
+               >
+                  {formattedCompareAtPrice}
+               </Text>
+               {isDiscounted && showDiscountLabel && direction === 'row' && (
+                  <IconBadge
+                     icon={showProBadge ? faRectanglePro : undefined}
+                     colorScheme={colorScheme}
+                  >
+                     {discountLabel}
+                  </IconBadge>
+               )}
+            </>
+         )}
+      </Box>
+   );
+}
--- a/frontend/models/product.ts
+++ b/frontend/models/product.ts
@@ -22,11 +22,12 @@ export async function findProduct(shop: ShopCredentials, handle: string) {
    const variants = response.product.variants.nodes.map((variant) => {
       const isDiscounted =
          variant.compareAtPriceV2 != null &&
-         variant.compareAtPriceV2.amount > variant.priceV2.amount;
+         parseFloat(variant.compareAtPriceV2.amount) >
+            parseFloat(variant.priceV2.amount);
       const discountPercentage = isDiscounted
          ? computeDiscountPercentage(
-              variant.priceV2.amount * 100,
-              variant.compareAtPriceV2!.amount * 100
+              parseFloat(variant.priceV2.amount) * 100,
+              parseFloat(variant.compareAtPriceV2!.amount) * 100
            )
          : 0;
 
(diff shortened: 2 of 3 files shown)


frontend/templates/product/sections/ProductSection/index.tsx
```diff
@@ -106,8 +107,12 @@ export function ProductSection({
...
                <ProductTitle mb="2.5">{product.title}</ProductTitle>
                <ProductPrice
-                  price={selectedVariant.formattedPrice}
-                  compareAt={selectedVariant.formattedCompareAtPrice}
+                  formattedPrice={selectedVariant.formattedPrice}
+                  formattedCompareAtPrice={
+                     selectedVariant.formattedCompareAtPrice
+                  }
+                  isDiscounted={selectedVariant.isDiscounted}
+                  discountLabel={`${selectedVariant.discountPercentage}% OFF`}
```
masonmcelvain, line 115: I'm curious why you chose to have the caller pass a formatted discount label, as opposed to having them pass the `discountPercentage` and formatting the discount label internally? It seems like a simpler API to just pass the `discountPercentage` to me.

frontend/components/common/ProductPrice.tsx
```diff
@@ -0,0 +1,79 @@
...
+export type ProductPriceProps = {
+   formattedPrice: string;
+   formattedCompareAtPrice?: string | null;
+   isDiscounted: boolean;
+   discountLabel?: string;
+   showDiscountLabel?: boolean;
+   showProBadge?: boolean;
+   size?: 'large' | 'medium' | 'small';
+   colorScheme?: ThemeTypings['colorSchemes'];
+   direction?: 'row' | 'column' | 'column-reverse';
```
masonmcelvain, line 15: Minor: I'm not sure if you're intentionally limiting the `flexDir` types here, but this pattern also works if you weren't.

```suggestion
   direction?: BoxProps['flexDir'];
```

deltuh-vee:
dev_block ✊ 
I'm having trouble interacting with `/Parts`(searching for products, clearing out filters, etc.) and I'm encountering these errors:
CleanShot 2022-08-31 at 14 46 27

masonmcelvain (approved):
CR ✌🏻 Very nice! I had some questions, but I'm happy to merge this as is.

federicobadini:
> dev_block ✊ I'm having trouble interacting with `/Parts`(searching for products, clearing out filters, etc.) and I'm encountering these errors: CleanShot 2022-08-31 at 14 46 27

I'm not experiencing any of those errors. I've just opened the preview on two different devices and I see nothing strange on `/Parts`. Also because all these changes should impact only the product page.
@masonmcelvain have you experienced any of these errors?

masonmcelvain:
I'm not seeing those errors either. I'm using Firefox on Ubuntu 22.04.

@deltuh-vee I think the collection pages should be ok given that both @federicobadini and I aren't seeing those errors, it might be specific to your system for some reason (or maybe it was a transient outage of Algolia or something?). I'd say we should carry on with the QA of the product page.

un_dev_block ✌🏻

frontend/templates/product/sections/ProductSection/index.tsx
federicobadini: It was all due to [this open discussion](https://github.com/iFixit/react-commerce/issues/533)

The design proposal for prices is still not definitive. But a few scenarios make embedding logic in the component itself complicates. For instance we will probably need to show `min`/`max` prices:

Screenshot 2022-09-01 at 00 10 20

This leads to the fact that we will potentially need to compute the discount percentage differently, according to the context in which we are rendering the price component.
If we are showing price ranges there is also the possibility that the label needs to be a bit richer to be explicative. Imagine the case in which not all variants have the same discount

Having labels allows us to compute the values outside the component and to avoid having a lot of complex conditions inside. 

If the API will be clearer in the future and we see the possibility to embed logic inside the component I will do for sure 🚀 . But if the price in the product card ends up diverging too much we may also need to separate the two.

masonmcelvain: Thanks for the explanation, makes complete sense!

rickisXP:
I'm unable to see the discount. After playing with Akeneo and setting the `retail_price` attribute (and running `Exec/console shopify:save-products --store=test --productcodes=188113` and `Admin/Exec/exportAllProductsToiFixit.php --run --skus=1881132`)

@masonmcelvain Do you know how I can get a discount to show?

image

masonmcelvain:
It looks like we are using the US store's product data for the previews.

https://github.com/iFixit/react-commerce/blob/1662e3537d6a8b2a2fd29792662c745be0983bf7/frontend/layouts/default/getLayoutServerSideProps.ts#L9

So you'll need to find a real product that is discounted to QA this pull unfortunately.

I opened a separate issue #675 to discuss using `ifixit-test` product data in the `react-commerce` preview.

deltuh-vee:
@rickisXP Try https://www.ifixit.com/Store/Tools/Fix-the-World-Transfer-Sticker/IF179-071?o=1

deltuh-vee:
dev_block ✊ 
CleanShot 2022-09-01 at 10 56 26
This item isn't being shown as being On Sale even though it is.

frontend/models/product.ts
```diff
@@ -19,8 +20,20 @@ export async function findProduct(shop: ShopCredentials, handle: string) {
       return null;
    }
    const variants = response.product.variants.nodes.map((variant) => {
+      const isDiscounted =
+         variant.compareAtPriceV2 != null &&
+         variant.compareAtPriceV2.amount > variant.priceV2.amount;
```
masonmcelvain, line 25: I think these amounts are serialized as strings (e.g. "29.99") so we probably need to cast them as numbers before treating them as such.

federicobadini: Good catch!

masonmcelvain:
I think this can be un_dev_block ✌🏻 now

k0rvusk0r4x:
- In the mockup, the text for the percentage off is the same color as the sale price while the actual is quite a bit darker.
- In the mockup, the spacing between the sale price and slashed-out regular price is the same as the spacing between the slashed-out regular price and the percentage off. In the actual, the space is a bit less than half.

image

dev_block 😺

federicobadini:
> * In the mockup, the text for the percentage off is the same color as the sale price while the actual is quite a bit darker.

Actually the two colors are different in the mockup too: percentage off is a `red.700` while the actual price is `red.600`. That is to keep an high contrast ratio. Otherwise it will be much lower given the light red background.
This is the contrast score with the values that are present in the mockup:

Screenshot 2022-09-13 at 08 22 44

This is what the score would be using the same red.600 that is used for the actual price:

Screenshot 2022-09-13 at 08 12 18

> * In the mockup, the spacing between the sale price and slashed-out regular price is the same as the spacing between the slashed-out regular price and the percentage off. In the actual, the space is a bit less than half.

The mockup shows a 6px spacing between the actual price and the slashed-out price and a 12px spacing between the slashed-out price and the percentage off. The only difference from there is that I used a 10px spacing instead of 12px because in the actual implementation 12px seemed a bit too much. 

Screenshot 2022-09-13 at 08 37 23
Screenshot 2022-09-13 at 08 37 48

masonmcelvain:
un_dev_block ✌🏻 for another QA

k0rvusk0r4x:
Explanation [here](https://github.com/iFixit/react-commerce/pull/673#issuecomment-1244970150) shows the issue was interpreting the mockup.

QA 😺

PR #677, opened 2022-09-01
Add `STORE_CODE` env var

### LINKED ISSUE #675: Use `ifixit-test` product data in `react-commerce` preview
## Problem

We use `ifixit-us` product data in our `react-commerce` PR previews. QA has no way to edit product data for testing different frontend features.

## Do this

Pass a new environment variable instead of hardcoding `"us"` here. (Or consider using/refactoring the existing `CODEGEN_STOREFRONT_SHOP_NAME` variable).

https://github.com/iFixit/react-commerce/blob/1662e3537d6a8b2a2fd29792662c745be0983bf7/frontend/layouts/default/getLayoutServerSideProps.ts#L9

masonmcelvain:
Allows us to configure which storefront api we talk to. This is useful for toggling between real product data from the US store, and test data from the Test store.

I added a new "Test" store to main.govinor.com and strapi.ifixit.com. I also set `STORE_CODE=test` in `react-commerce`'s Vercel dasbhoard, since the previews use `.env.production`, which uses the `us` store code.

image

## QA

Make sure `react-commerce` uses ifixit-test data product data (and reviews data).

I don't think the product page is turned on for `react-commerce-prod`? If so, we'll just have to trust that it uses ifixit-us product data for now.

Closes #675


frontend/.env.test
```diff
@@ -1,4 +1,5 @@
 NEXT_PUBLIC_ALGOLIA_APP_ID=XQEP3AD9ZT
 NEXT_PUBLIC_IFIXIT_ORIGIN=https://www.cominor.com
 NEXT_PUBLIC_STRAPI_ORIGIN=https://strapi.cominor.com
-NEXT_PUBLIC_SHOPIFY_STOREFRONT_VERSION=2021-07
\ No newline at end of file
+NEXT_PUBLIC_SHOPIFY_STOREFRONT_VERSION=2021-07
```
danielcliu-ifixit, line 4: Not this issue but should this be `2022-07` now like it is in `.env.production` and `.env.development`.

danielcliu-ifixit:
CR :v:

danielcliu-ifixit (approved):

frontend/.env.test
masonmcelvain: Let's try it!

```suggestion
NEXT_PUBLIC_SHOPIFY_STOREFRONT_VERSION=2022-07
```

deltuh-vee:
@masonmcelvain I see that react-commerce is using the review data from ifixit-test, but not price data. Do we want to include that as well?

masonmcelvain:
@deltuh-vee Yes we want to use ifixit-test price data. However, it looks like [Strapi](https://use-ifixit-test-in-dev-preview.govinor.com/admin) for this branch was missing the `test` store.

image

I just added the test store to this branch's Strapi, let's give it another try.

davidrans (approved):
CR :coffee:

deltuh-vee:
dev_block ✊ 
Price is still doesn't match up with the one on cominor/ifixit-test for the same variant.
CleanShot 2022-09-02 at 12 58 00

masonmcelvain:
I used the wrong name for the environment variable in `react-commerce` 🤦🏻

I corrected it:

image

un_dev_block ✌🏻

deltuh-vee:
QA 🎬 
`react-commerce` now uses ifixit-test data product data (and reviews data).

dhmacs (commented):
dev_block 🔢 

After implementing the seed plugin, we're seeding data from production and that's why we also get the production access token. An easy solution to see product data from `ifixit-test` would have been to just set`ifixit-test` storefront token on the Strapi govinor instance (or localhost for dev) for the `us` store, and that's how I was doing it.

I think you have come up with a good solution though, because we also need to localize product reviews, and we need the correct store code for that, so it makes sense to me to create a test store as you're suggesting 👍 
I've just commented with a minor suggestion to make it clear that we'd use env variable only as a default value

masonmcelvain:
> set ifixit-test storefront token on the Strapi govinor instance

Good point, hadn't considered that solution.

I've implemented your suggestions, thank you for the context!

un_dev_block ✌🏻

frontend/layouts/default/getLayoutServerSideProps.ts
```diff
@@ -6,7 +7,7 @@ export async function getLayoutServerSideProps(): Promise<DefaultLayoutProps> {
    const [globalSettings, stores, currentStore] = await Promise.all([
       getGlobalSettings(),
       getStoreList(),
-      getStoreByCode('us'),
+      getStoreByCode(STORE_CODE),
```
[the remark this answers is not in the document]
masonmcelvain: Thanks for pointing this out. I've renamed the environment variable and added a clarifying comment as suggested.

deltuh-vee:
Carry over QA 🎬

sterlinghirsh:
CR :zap:

sterlinghirsh (approved):

PR #678, opened 2022-09-01
Create New Guide Footer

hcutrone:
## Background
We want to use the new `react-commerce` footer on our guide pages! The current footer, however, relies on lots of store-based information. This pull creates a new footer that is not reliant on this information.

## Before (bottom) and After (top)
<details>
  
</details>

## Summary of Changes
1. Made two footers in `packages/ui/footer` - one for cart and one for guide
2. Added default information to be used on guide pages (no store-base props passed in)
3. Redesigned footer

## QA
This footer is not being used anywhere, so I'm gonna mark it as qa_req 0. This will get proper QA once it is displayed on Guide pages, but for now I made sure that all links work, everything shows up as expected, and it is responsive to all screen sizes.

## CR
Commit-by-commit. Keep an eye on the commit messages, I tried to explain everything I was doing to make things easier to understand.

## Notes
1. I also styled the guide footer to look like the footer design we want, see [this figma](https://www.figma.com/file/sfj6r7Bn6eREMvuqo0f7Du/iFixit---Components?node-id=463%3A21820).
2. There's no matomo tracking yet. I would like to import the Matomo tracking files we have in iFixit/ifixit to the matomo package so both repos can use them, but I had a lot of issues trying. I decided to make that a separate pull (or maybe add onto this one if it's easier than I thought.
3. Yes, there are two footers. I did not want to affect the current footer at all, so I copied everything over and altered those. I tried to build the GuideFooter so that it would be simple to make that _the_ footer for all pages.


jarstelfox (approved):
CR 🌵 but deploy_block 🌵 on reducing the size of return blocks and creating some smaller components

hcutrone:
Force pushed to remove a duplicated commit
un_deploy_block :burrito: Broke up the return statements and made the components smaller :partying_face:

packages/ui/footer/guide/components/Partners.tsx
```diff
@@ -0,0 +1,101 @@
...
+   )
+);
+
+const FooterPartners = ({ partners }: { partners: Menu }) => {
+   if (!partners) {
+      return null;
+   }
+   const partnerIcons = partners.items.map((partner: MenuItem) => {
+      if (partner.type === MenuItemType.ImageLink) {
+         return <FooterPartner partner={partner} />;
```
jarstelfox, line 40: You should add a key prop here

packages/ui/footer/guide/components/Partners.tsx
```diff
@@ -0,0 +1,101 @@
...
+         position="relative"
+         p="0"
+      >
+         <PartnerImage partner={partner} />
+      </FooterPartnerLink>
+   );
+};
+
+const PartnerImage = ({ partner }: { partner: ImageLinkMenuItem }) => {
+   if (partner.image) {
```
jarstelfox, line 62: I'd invert this if and drop the else

packages/ui/footer/guide/components/StoreMenu.tsx
```diff
@@ -0,0 +1,107 @@
...
+type StoreMenuButtonProps = MenuButtonProps & {
+   icon?: React.ReactNode;
+};
+
+export const StoreMenu = ({
+   stores,
+}: {
+   stores: StoreListItem[] | undefined;
+}) => {
+   if (stores && stores.length > 0) {
```
jarstelfox, line 27: If you invert this if, you can early return null instead

jarstelfox (approved):
CR 🌵 with a deploy_block 🌵 on minor fixes

jarstelfox (approved):
CR :shipit:

hcutrone:
un_deploy_block :burrito: 
Inverted `if`s and added keys :+1:

PR #683, opened 2022-09-07
Redirects: Escape special characters for item types and tool categories

masonmcelvain:
Instead of hard coding escape slashes in our item type redirect json file, let's add them in JS. This is safer, in case we missed an escape value in the json somewhere. This also lets us use the redirects in PHP more easily.

## QA

Redirects for part item types and tool categories should still work as before.

## On deploy

Bump the react-commerce submodule in iFixit/ifixit.

## Background

https://nextjs.org/docs/api-reference/next.config.js/redirects#regex-path-matching
https://stackoverflow.com/questions/3561493/is-there-a-regexp-escape-function-in-javascript/3561711#3561711

Connects https://github.com/iFixit/ifixit/pull/44563

CC @danielbeardsley


frontend/legacyPartCollectionRoutes.js
```diff
@@ -14,12 +14,16 @@ function mapPartItemTypes() {
...
-      source: `/Parts/:slug/${oldTag}`,
+      source: `/Parts/:slug/${escape(oldTag)}`,
       destination: `/Parts/:slug`,
       permanent: true,
    }));
 
    return [...oldToNew, ...legacyToParent];
 }
 
+function escape(str) {
```
danielbeardsley, line 25: Do we have a convenient place (utils package?) to move this function so it can be shared with the other usage below? This will also end up giving it a more descriptive name `escapeStringForRegex()` or something.

danielbeardsley:
CR :+1: 

Looks great! 

deploy_block :+1:  on considering sharing that function.

frontend/legacyPartCollectionRoutes.js
masonmcelvain: We don't have an existing place, but we can make a new place.

masonmcelvain:
dev_block ✌🏻

masonmcelvain:
un_dev_block ✌🏻 un_deploy_block ✌🏻

frontend/next-config/redirects/part-collections.js
```diff
@@ -14,7 +15,7 @@ function mapPartItemTypes() {
    const legacyTagsWithNoItemType = legacyFilterSlugs['tagsWithNoItemType'];
    // Redirect tags we don't support anymore to the base product list page.
    const legacyToParent = legacyTagsWithNoItemType.map((oldTag) => ({
-      source: `/Parts/:slug/${oldTag}`,
+      source: `/Parts/:slug/${escapeStringForRegex(oldTag)}`,
```
danielbeardsley, line 18: next.js only claims to support regexes inside parens after parameters (like the other use in `next.config.js`). Here we have a regex outside parens. They don't say if they split the string on `/` then interpret a regex, nor even if they support regexes outside params:

image

Sounds like this will take experimentation to validate if it works. I'm still quite leery of using an or-list `x|y|z` with no surrounding parens.

danielbeardsley:
dev_block :+1: on responding to the above comment.

frontend/next-config/redirects/part-collections.js
masonmcelvain: Good catch, I hadn't noticed that discrepancy. Given that we didn't use parens before, it probably _works_ as is (we don't use an or-list here, that's only for tools). But I'd like to follow the suggested pattern, so I'm gonna add them.

I also noticed that I neglected to escape `legacyTagToItemType` above.

masonmcelvain: Hm after looking again, I noticed that the old tags are just used as strings here. Since they aren't actually an or-list or any other kind of regex, I don't think we should wrap them in parens.

masonmcelvain:
un_dev_block ✌🏻

danielbeardsley:
Hmm.. dev_block :+1:  for now on #654 cause it changes the list of item types. I think we should get that one merged, then re-copy the list and update the code that uses these in this pull.

CR :+1: Otherwise!

masonmcelvain:
un_dev_block ✌🏻 looks like #654 didn't end up changing the item types after all

danielbeardsley:
CR :+1: 

Though I'm curious. Is there a reason that we are using *many* redirect entries for `mapItemTypes()` and using a long regex with many ORed together expressions for the legacy slugs?

Two different approaches for the same problem seems a bit odd, but they are both functional. I bet the single regex is faster, but there aren't enough of them that it matters I bet.

k0rvusk0r4x:
I'm getting a 404 from these rather than a redirect. Is testing it via the "View Deploy" the way to go here, or does it need something else?

masonmcelvain:
> I'm getting a 404 from these rather than a redirect. Is testing it via the "View Deploy" the way to go here, or does it need something else?

Could you share a link? I can't reproduce a 404

masonmcelvain:
> CR +1
> 
> Though I'm curious. Is there a reason that we are using _many_ redirect entries for `mapItemTypes()` and using a long regex with many ORed together expressions for the legacy slugs?
> 
> Two different approaches for the same problem seems a bit odd, but they are both functional. I bet the single regex is faster, but there aren't enough of them that it matters I bet.

Yea it is a bit strange. The item types that just redirect to `/Parts/:device` could use a regex like we do for tools. I think they were just implemented at different times, I'm not aware of a reason they're different.

jordycosta:
QA 🟢 
deploy_block 🟨 

I ran a bulk redirect on each of these categories and didn't receive any 404 responses. I'm not entirely sure that I did these redirect checks correctly so that's the reason for the deploy_block, a quick check for validation on this would be nice!

[tagToItemType.csv](https://github.com/iFixit/react-commerce/files/9569085/tagToItemType.csv)
[tagsWithNoItemType.csv](https://github.com/iFixit/react-commerce/files/9569086/tagsWithNoItemType.csv)
[toolLegacyRoutes.csv](https://github.com/iFixit/react-commerce/files/9569088/toolLegacyRoutes.csv)

masonmcelvain:
`tagsWithNoItemType` is actually for `/Parts/device/tag` => `/Parts/device` , not for tools. The other 2 look great, thanks!

jordycosta:
> `tagsWithNoItemType` is actually for `/Parts/device/tag` => `/Parts/device` , not for tools. The other 2 look great, thanks!

un_deploy_block 🟩 
Thanks! I updated the csv in https://github.com/iFixit/react-commerce/pull/683/#issuecomment-1247176993

danielbeardsley (approved):
CR :+1:

PR #694, opened 2022-09-12
Update algolia instantsearch hooks

danielbeardsley:
Update them to the latest version and apply a patch to the result to work around [the bug](https://github.com/vercel/next.js/issues/40067)


sterlinghirsh:
dev_block :zap: since it seems like #700 will work instead.

danielbeardsley:
Yes, #700 is the answer, this failed on vercel cause `patch` isn't available!

PR #709, opened 2022-09-16
packages/ui: Update Dependencies

hcutrone:
To use the new footer, we are importing more packages and components into iFixit/ifixit, which meant there were more dependency issues to iron out.

## QA
Mainly focus on the react-commerce side, similar to [this pull](https://github.com/iFixit/react-commerce/pull/633#issuecomment-1223033565).

Connects: https://github.com/ifixit/ifixit/issues/43967


packages/ui/package.json
```diff
@@ -12,7 +12,8 @@
       "@ifixit/app": "workspace:*",
       "@ifixit/cart-sdk": "workspace:*",
       "@ifixit/icons": "workspace:*",
-      "@ifixit/newsletter-sdk": "workspace:*"
+      "@ifixit/newsletter-sdk": "workspace:*",
+      "regenerator-runtime": "*"
```
masonmcelvain, line 16: Can `regenerator-runtime` be a `devDependency`, or does it need to be a direct one?

packages/app/package.json
```diff
@@ -9,7 +9,8 @@
       "@types/react": "17.0.37",
       "@types/react-dom": "17.0.11",
       "@ifixit/tsconfig": "workspace:*",
-      "typescript": "4.5.3"
+      "typescript": "4.5.3",
+      "react": "16.14.0 || 17.0.2"
```
masonmcelvain, line 13: I think we only use `react 17.0.2` in this project. Where is `16.14.0` coming from, is that what we use in `iFixit/ifixit`?

masonmcelvain (dismissed):
CR ✌🏻 with some questions. Thanks for doing this!

packages/app/package.json
hcutrone: Yes, there are many places in `iFixit/ifixit` that use `16.14.0`. `16.14.0` is the preferred version on `iFixit/ifixit`, actually

packages/ui/package.json
hcutrone: You're right, it can be a `devDependency`! Fixed

masonmcelvain:
I think QA for this should include checking out the `react-commerce` submodule on cominor, pulling this branch `footer--update-deps`, and then running `rush update` in the code dir. That will verify that we'll be able to use this new version of `react-commerce`. As is, `rush update` still fails with the following output for me, so I'll leave the dev_block.

image

hcutrone:
Shoot, I know why that error is happening, @masonmcelvain. Should know how to fix it. Thanks for catching it!

hcutrone:
We are fine with this for now! It will cause errors in iFixit/ifixit only if we update the submodule. To my and @andyg0808 knowledge, we aren't using anything from react-commerce live on the site yet, so this shouldn't be an issue.

This fix for this issue is a part of the pull to use the react-commerce footer in iFixit/ifixit (WIP), which should be the first time we use react-commerce stuff in iFixit/ifixit that isn't hidden behind a feature flag. Ultimately, I can extract the fix and make a separate iFixit/ifixit pull if we are worried about the gap between deploys.

un_dev_block :burrito:

erinemay:
On the react side, the following continue to work on desktop and mobile (except for the mobile header menu - see below):

- Header and footer 
     - Links to other pages
     - User profile menu
     - Mini cart (modifying cart and checking out)
     - Newsletter signup is recording email to `ifixit_guide.newsletter`
- Toolbar
     - Guides/Answers links
     - Breadcrumbs and breadcrumb toolbar on desktop and mobile (/Parts and /Tools)
- Pagination
- Product cards
     - Title, description, flair, price, discounts
     - List view/tile view
- Navigating up and down the heirarchy
- Back button
- Search
- Filters
- Item Type collections

But dev_block 🧷 on the same issue as in https://github.com/iFixit/react-commerce/pull/633.
<details><summary>The mobile header menu breaks the page.</summary>

Click on the hamburger menu
mobile menu 09222022

This is what's shown in the UI
image

Console:
image

</details>

masonmcelvain:
@hcutrone Sounds good. One place we _do_ use a file from react-commerce in iFixit/ifixit is the legacyFilterSlugs.json file. I think that's used in LegacyProductTags.php (could have a different name, not sure). We use it to build links in PartCollectionSitemapGenerator that don't redirect, ie they point directly to a real collection. As long as that functionality still works, your plan sounds good to me.

hcutrone:
@masonmcelvain In that case, I'll go ahead and open a separate pull for iFixit/ifixit. It's small so I might as well do it just in case
iFixit/ifixit: https://github.com/iFixit/ifixit/pull/44775

masonmcelvain (dismissed):
CR ✌🏻 Thanks @hcutrone!

masonmcelvain (approved):
CR ✌🏻

hcutrone:
un_dev_block :burrito: , fixed the hamburger menu issue

hcutrone:
I reverted the changes I made to the `sentry` and `newsletter-sdk` packages, as it looks like we won't be using them in iFixit/ifixit.

I am going to deploy_block :burrito: , just in case we actually _do_ want to use these packages.

masonmcelvain (approved):
CR ✌🏻 Yea I'd say if we don't have a need for those packages in iFixit/ifixit, and they work as-is in this project, it's not worth going through the trouble of making them compatible. We can cross that bridge if we ever come to it, but it's hard to imagine we will. Great work!

k0rvusk0r4x:
Hamburger menu is working.

However:

Newsletter signup in the footer is now broken. It is giving an error on any email entered and not recording them in the db.

Otherwise, continues to function as [here](https://github.com/iFixit/react-commerce/pull/709#issuecomment-1255607183).

dev_block 😿 to get the newsletter signup in the footer working again.

hcutrone:
>Newsletter signup in the footer is now broken

I'm inclined to say this is not a big deal, at least for this pull. It only occurs on the `prod` preview, and seems to be happening on other pulls as well (happens on [this new pull](https://github.com/iFixit/react-commerce/pull/725)). It doesn't happen on the other preview nor locally.

un_dev_block :burrito: Keeping the deploy block

hcutrone:
un_deploy_block :burrito: - We are _not_ using the Newsletter packages

erinemay:
>It only occurs on the `prod` preview, and seems to be happening on other pulls as well (happens on [this new pull](https://github.com/iFixit/react-commerce/pull/725)). It doesn't happen on the other preview nor locally.

Yeah. It's my understanding that this will always be broken in the prod preview because don't want to be writing to the prod db. Not ideal for testing since it throws up false positives, bug-wise. But, we get the benefit of viewing the prod preview with prod data.

hcutrone:
This pull just needs to merge main and re-run `pnpm install:all`, but I gotta run to office hours. I will fix it tonight

dhmacs:
> I'm inclined to say this is not a big deal, at least for this pull. It only occurs on the `prod` preview, and seems to be happening on other pulls as well (happens on [this new pull](https://github.com/iFixit/react-commerce/pull/725)). 

@hcutrone I'm able to successfully subscribe  in #725 

Screenshot 2022-09-28 at 08 26 45

> It's my understanding that this will always be broken in the prod preview because don't want to be writing to the prod db.

@erinemay I don't think so, previews are using strapi production data, but the ifixit origin is set to `www.cominor.com`, so you should be able to subscribe via the staging api.

Btw I get a 500 error on this preview if I visit `/Parts`, and inspecting the Vercel logs it seems to be related to the Algolia instantsearch library update 🤔 :

Screenshot 2022-09-28 at 08 30 46

masonmcelvain:
The ReactDOMServer error was affecting all branches. We fixed it in https://github.com/iFixit/react-commerce/pull/752. I've merged the fix into this branch and ran `pnpm install:all` 👍🏻

packages/icons/package.json
```diff
@@ -4,13 +4,28 @@
...
    "license": "MIT",
+   "scripts": {
+      "clean": "",
+      "build": ""
+   },
    "peerDependencies": {
-      "@chakra-ui/react": ">= 1.8.3 < 2"
+      "@chakra-ui/react": ">= 1.8.3 < 2",
+      "@emotion/react": ">=11.0.0 <12.0.0",
+      "framer-motion": ">=3.0.0 <4.0.0 || >=4.0.0 <5.0.0 || >=5.0.0 <6.0.0 || >=6.0.0 <7.0.0",
```
sterlinghirsh, line 14: Would it have worked to say `>=3.0.0 <7.0.0`?  Also, why do we need these to be peerDeps and devDeps?

hcutrone: You're right, I didn't catch that. Fixed!

To answer your second question, I'll point you to this comment from @andyg0808 : https://github.com/iFixit/react-commerce/pull/633#discussion_r950431392

hcutrone:
Carryover QA 🌯 - change should have no effect on functionality

packages/icons/package.json
```diff
@@ -4,13 +4,28 @@
...
+      "clean": "",
+      "build": ""
+   },
    "peerDependencies": {
-      "@chakra-ui/react": ">= 1.8.3 < 2"
+      "@chakra-ui/react": ">= 1.8.3 < 2",
+      "@emotion/react": ">=11.0.0 <12.0.0",
+      "framer-motion": ">=3.0.0 <7.0.0",
+      "@emotion/styled": ">=11.0.0 <12.0.0",
+      "react-dom": ">=16.8.6 <17.0.0 || >=17.0.0 <18.0.0",
```
masonmcelvain, line 16: Similar to https://github.com/iFixit/react-commerce/pull/709#discussion_r982883122, we can probably simplify this requirement and the other `react-dom` requirement below.

```suggestion
      "react-dom": ">=16.8.6 <18.0.0",
```

packages/icons/package.json
```diff
@@ -4,13 +4,28 @@
...
       "@types/react": "17.0.37",
       "@types/react-dom": "17.0.11",
       "@ifixit/tsconfig": "workspace:*",
-      "typescript": "4.5.3"
+      "typescript": "4.5.3",
+      "react": "16.14.0 || 17.0.2",
+      "@emotion/react": ">=11.0.0 <12.0.0",
+      "framer-motion": ">=3.0.0 <7.0.0",
+      "@emotion/styled": ">=11.0.0 <12.0.0",
+      "react-dom": ">=16.8.6 <17.0.0 || >=17.0.0 <18.0.0",
```
masonmcelvain, line 28:
```suggestion
      "react-dom": ">=16.8.6 <18.0.0",
```

masonmcelvain (commented):
CR ✌🏻 With one more suggestion! So close!

masonmcelvain (approved):
CR ✌🏻 Thanks @hcutrone!

hcutrone:
I fixed these conflicts, but I'm getting an infinite loop trying to push. It seems to just keep running the "type-check" script indefinitely: https://ifixit.slack.com/archives/C01K32B4FPX/p1664479146641279

hcutrone:
Carryover CR :burrito: & Carryover the Carryover the Carryover QA :burrito: - merged main and fixed conflicts

_deploy_block :burrito: on issue I posted [above](https://github.com/iFixit/react-commerce/pull/709#issuecomment-1262792619). I went ahead and did `git push --no-verify` since it seems this is a bug on main. Deploy blocking to make sure this is okay

pnpm-lock.yaml
```diff
@@ -1,9 +1,4 @@
-lockfileVersion: 5.4
-
-patchedDependencies:
-  react-instantsearch-hooks-server@6.32.1:
-    hash: hiec5pcpgig2m3qcu64y3isqke
-    path: patches/react-instantsearch-hooks-server@6.32.1.patch
+lockfileVersion: 5.3
```
masonmcelvain, line 2: Downgrading the lockfile version will break our Vercel deployment (see https://github.com/iFixit/react-commerce/pull/752). Make sure `pnpm --version` shows you version 7, and then run `pnpm install:all` and commit the result.

masonmcelvain (changes requested):
CR ✌🏻 but dev_block ✌🏻 on using the right version of pnpm to generate the lockfile.

pnpm-lock.yaml
masonmcelvain: If you're not on version 7, you could do `npm install -g pnpm@7` to install it. https://pnpm.io/installation#using-npm

jarstelfox:
un_dev_block 🌵

masonmcelvain (approved):
CR ✌🏻 Thanks!

jyee27:
CR ☕

erinemay:
Got a problem with pro pricing. Log in as a pro test user, or set any user as pro by modifying their discount in customer manager `https://www.cominor.com/Admin/CustomerManager?userid=<userid>`

Main branch: https://react-commerce.vercel.app/Tools/Toolkits
image


This branch: https://react-commerce-git-footer-update-deps-ifixit.vercel.app/Tools/Toolkits
image

Otherwise, confirmed the following all look and behave the same as on main on desktop and mobile:

- Header and footer 
     - Links to other pages
     - User profile menu
     - Mini cart modifying cart 
     - Checkout is working again (goes to store.cominor.com)
     - Newsletter signup is recording email to `ifixit_guide.newsletter`
- Toolbar
     - Guides/Answers links
     - Breadcrumbs and breadcrumb toolbar on desktop and mobile (/Parts and /Tools)
- Pagination buttons (forward, backward, page 1, last page, single page buttons)
- Product cards
     - Title, description, flair, price, discounts
     - List view/tile view
     - Links to appropriate product page (live for prod preview, and react product page for dev preview)
- Navigating up and down the hierarchy
- Back/forward button with and without filters 
- Search
- Filters
- Item Type collections
- Refreshing with different filters
- Viewing as logged in/logged out

erinemay:
Jk. It was fixed already: https://github.com/iFixit/react-commerce/pull/760

QA 👍

PR #724, opened 2022-09-22
Ignore Typeerror: Network Request Failed In Sentry

danielcliu-ifixit:
## Overview

These ` Typeerror: Network Request Failed` seem hard to reproduce and not overly harmful. Lets ignore them from sentry logs so they don't take up room on Sentry.

## QA
I don't think this needs qa since we don't know how to reproduce the error specifically. qa_req 0;

Connects: iFixit/ifixit#42336

### DIFF
--- a/frontend/sentry.client.config.ts
+++ b/frontend/sentry.client.config.ts
@@ -21,5 +21,8 @@ Sentry.init({
          'next.runtime': 'client',
       },
    },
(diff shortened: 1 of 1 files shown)


sterlinghirsh (approved):
CR :zap: This is a good change (we've seen both versions of this error in react-commerce), but the intention of this issue was to add these ignores in ifixit's sentry config, not just in react-commerce.

Issue #695, opened 2022-09-13
Fix Typescript types regressions on `ui` package


dhmacs:
PR #639 has broken Typescript types for the cart component. Digging through the code I see that we're importing dependencies in the `ui` package that should only be visible in the `frontend` package, e.g.:

![Screenshot 2022-09-13 at 14 58 05](https://user-images.githubusercontent.com/4640135/189907180-f59f1d16-07f5-4edd-8c31-f738976357bd.png)

## Do this

`ui` package should not depend on the `frontend` package, so we need to refactor the code to remove these dependencies.
Part of the refactor is to put components into the ui package (e.g. `IFixitImage`), and part of the refactor is declare props locally (For example we don't want our `ui` package to depend on types exported from the models of our `frontend` app. Let's decouple the type definitions).

## Bonus

There is a git hook that should have prevented us from pushing code with TS errors. I'd like to learn how these errors weren't catched, @hcutrone did you force push or is the git hook not working on your machine?
I think it might be worth to create a Github action to perform the ts check.

hcutrone:
Hm, no I did not force push or by-pass any tests. It stopped me from pushing when there were other type errors, but it never warned me about those.

dhmacs:
@hcutrone are you running on Windows? I think we should have a Github action in place anyway 🤔

hcutrone:
@dhmacs I'm running Linux (Ubuntu). Agreed that a Github action is probably the right idea here. 
Also, we're working away on some dependency issues that arose trying to use the footer in the iFixit/ifixit repo. Do you think this could be causing extra issues?

dhmacs:
@hcutrone I think so. It's weird to me that it was good enough to pass QA, because the code is basically importing stuff that it shouldn't know anything about. Like `@models/menu` is something that only the `frontend` app knows how to resolve (since its a path alias set on the `frontend` `tsconfig.json`). It would really surprise me if you are able to use these components in other projects 🤔

PR #733, opened 2022-09-25
Check types in GitHub Actions

masonmcelvain:
Adds a GitHub action that runs the typescript compiler in the frontend, the backend, and every package with a tsconfig.json.

qa_req 0

CC @dhmacs @hcutrone

Connects #695


masonmcelvain:
Those look like real type failures in our project. I'm not very familiar with the code in `packages/ui/footer/guide`, so I'm inclined to merge this and fix the type failures in a follow up pull. Merging this as-is will illuminate existing type failures in our project, it won't introduce new ones.

hcutrone:
@masonmcelvain The code in `packages/ui/footer/guide` is not being used anywhere (in either repo). They are components I updated to be used in the new guide footer, but ended up not being used. I only kept them because they restyled the footer to match the new design (see description of [this pull](https://github.com/iFixit/react-commerce/pull/678)), so I thought they might be helpful when we updated the collections footer. 

All this to say, it might not be worth it to fix those type bugs.

masonmcelvain:
Thanks for the explanation @hcutrone. I agree that it's probably not worth fixing the type errors right now.

In order for this new workflow to pass on `main` (and all other branches) though, perhaps we can temporarily delete the code that is causing `tsc` errors and then reintroduce it when we update the collections footer later on. Since it's in git, we'll still have a copy of the code, but it won't cause distracting CI failures on other features. What do you think?

hcutrone:
Go for it! I only kept it since it would be easier than searching git

masonmcelvain:
I opted to fix the type errors in a separate pull so that we don't have to block this one on fixing the type errors: https://github.com/iFixit/react-commerce/pull/738

.github/workflows/tsc.yml
```diff
@@ -0,0 +1,18 @@
...
+         - name: Checkout
+           uses: actions/checkout@v2
+         - name: Install PNPM
+           run: corepack enable && corepack prepare pnpm@6.32.11 --activate
+         - name: Install workspaces
+           env:
+              FONT_AWESOME_NPM_TOKEN: ${{ secrets.FONT_AWESOME_NPM_TOKEN }}
+           run: pnpm install:all
+         - name: Run Typescript Compiler
+           run: bin/type-check.sh
```
dhmacs, line 18: I think we can just use the npm script, which is also the one that we run on git hook.

```suggestion
           pnpm type-check
```

You can extend it to check the backend folder using the same approach used for the dev script (`npm-run-all`)

dhmacs (changes requested):
Thanks for working on this 👍 I've commented on how to simplify this before we merge

.github/workflows/tsc.yml
masonmcelvain: Good idea, thanks for the suggestion.

I noticed that `pnpm type-check:frontend` overlooked the type errors in `@ifixit/ui` that are being fixed here https://github.com/iFixit/react-commerce/pull/738. Other times, it catches the type errors in `packages/`. Perhaps it's only checking code that is being imported from `packages/` to `frontend/`?

This motivated me to explicitly run `tsc` on all packages to always catch type errors (see https://github.com/iFixit/react-commerce/pull/733/commits/12374962062d76059e34ad191cedf30f420a5f8c). However, this can result in running `tsc` on the same code more than once, which slows things down. What are your thoughts on this?

dhmacs: @masonmcelvain Yeah I was actually surprised that type checking the frontend wasn't type checking also the dependent packages. I think your approach of type-checking every package is the way to go, and I've found a "more native" pnpm way to do it using this command:

```sh
pnpm run -r type-check
```

The `-r` flag tells pnpm to run the `type-check` script of every app or package defined in `pnpm-workspace.yaml`.
So it's sufficient to declare the `type-check` script on every package, we don't need to define the glue scripts ourselves.
Notably `backend` is not specified as part of the monorepo (due to issues installing Strapi dependencies as part of the monorepo), and that's fine we can just write an additional script for that

dhmacs (approved):
lgtm CR 👏

PR #734, opened 2022-09-25
No-index Vercel previews

### LINKED ISSUE #722: No-index all react-commerce.vercel.app pages
I don't know how Google found them but at least 60 are now indexed. It was 52 earlier today so it seems like more and more are getting indexed. https://www.google.com/search?q=site%3Areact-commerce.vercel.app

masonmcelvain:
Only index pages on ifixit.com. The previews and react-commerce-prod and react-commerce domains should all be no-indexed.

## QA

Make sure the previews are no-indexed.

## On deploy

Makes sure www.ifixit.com is indexed, and react-commerce-prod.vercel.app and react-commerce.vercel.app are no-indexed.

CC @danielbeardsley

Closes #722


masonmcelvain:
dev_block ✌🏻

masonmcelvain:
QA ✌🏻 the previews are no-indexed

masonmcelvain:
un_dev_block ✌🏻

dhmacs (dismissed):
I'd refactor this to directly send a `X-Robots-Tag` header in the response, instead of passing down a prop to the render method

dev_block 🔢

masonmcelvain:
Thanks for the suggestion @dhmacs, I wasn't aware of that header. Much simpler! un_dev_block ✌🏻

masonmcelvain:
QA ✌🏻 The response [header](https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag#xrobotstag) is present in the previews.

image

frontend/templates/product-list/index.tsx
```diff
@@ -70,6 +71,11 @@ export const getProductListServerSideProps = ({
          );
       }
 
+      const url = urlFromContext(context);
+      if (new URL(url).hostname !== PROD_HOSTNAME) {
+         context.res.setHeader('X-Robots-Tag', 'noindex, nofollow');
```
danielbeardsley, line 76: Given that we're Just looking at a url and setting a header, should we do this in a middleware and not limit it to just the `product-list`?

danielbeardsley:
dev_block :+1: on considering my comment.

frontend/templates/product-list/index.tsx
masonmcelvain: I hadn't considered that! I gave it a shot, but haven't used middlewares before, let me know what you think.

masonmcelvain:
un_dev_block ✌🏻

masonmcelvain:
QA ✌🏻 

We still get the `x-robots-tag` header with this approach

image

sterlinghirsh:
Does it make sense to be sending X-Robots-Tag blank? I mean to say maybe we should not set it at all in prod.

danielbeardsley:
> Does it make sense to be sending X-Robots-Tag blank? I mean to say maybe we should not set it at all in prod.

I agree. :-/ Sorry for all the back and forth.

masonmcelvain:
> Does it make sense to be sending X-Robots-Tag blank? I mean to say maybe we should not set it at all in prod.

Good call, probably not. I was assuming blank would be equivalent to not sending it at all, but since headers have very specific meanings I think that's not a safe assumption to make.

masonmcelvain:
I'm uncertain why the collection scroll cypress test keeps failing. All the tests pass locally for me.

frontend/middleware.ts
```diff
@@ -0,0 +1,16 @@
...
+type Header = [string, string] | null;
+
+export function middleware(request: NextRequest) {
+   const robotTag: Header =
+      request.nextUrl.hostname === PROD_HOSTNAME
+         ? ['X-Robots-Tag', 'noindex, nofollow']
+         : null;
+
+   const headers = [robotTag].filter(isNonNullable);
+   return NextResponse.next({ headers });
```
sterlinghirsh, line 15: This seems a little overcomplicated to me with the filter, especially since this is going to run on every single request. This way brings it down to a single conditional with no additional function calls, and in the prod case, it should be equivalent to calling the function with no params. Also, I flipped the conditional since that way it's written seems like it would noindex every page on prod *only*.
```suggestion
   const options =
      request.nextUrl.hostname !== PROD_HOSTNAME
         ? { headers: [['X-Robots-Tag', 'noindex, nofollow']] }
         : undefined;

   return NextResponse.next(options);
```
I couldn't find any docs explaining what NextResponse.next() takes, and it seemed all the examples took nothing. I saw that some next.js tests included a form like this, but that seems weird. Did you find any docs explaining what NextResponse.next takes or what?

sterlinghirsh:
CR :zap: dev_block :zap:

frontend/middleware.ts
masonmcelvain: I chose to implement it with `.filter()` in case we wanted to add other headers on different conditions in the future. The `isNonNullable` callback was to appease typescript (given an anonymous function, typescript still thought the result array could have null values). What you have will work too, I don't have a strong opinion.

Here are the types we're working with. I found them with Ctrl + click in VSCode 🚀 . I think they're also on GH [here](https://github.com/vercel/next.js/blob/canary/packages/next/server/web/spec-extension/response.ts).

<details>

```ts

// node_modules/next/dist/server/web/spec-extension/response.d.ts

export declare class NextResponse extends Response {
    [INTERNALS]: {
        cookies: NextCookies;
        url?: NextURL;
    };
    constructor(body?: BodyInit | null, init?: ResponseInit);
    get cookies(): NextCookies;
    static json(body: any, init?: ResponseInit): NextResponse;
    static redirect(url: string | NextURL | URL, init?: number | ResponseInit): NextResponse;
    static rewrite(destination: string | NextURL | URL, init?: ResponseInit): NextResponse;
    static next(init?: ResponseInit): NextResponse;
}
interface ResponseInit extends globalThis.ResponseInit {
    nextConfig?: {
        basePath?: string;
        i18n?: I18NConfig;
        trailingSlash?: boolean;
    };
    url?: string;
}

// node_modules/typescript/lib/lib.dom.d.ts

interface globalThis.ResponseInit {
    headers?: HeadersInit;
    status?: number;
    statusText?: string;
}

/* snip */

type HeadersInit = [string, string][] | Record<string, string> | Headers;

declare var Headers: {
    prototype: Headers;
    new(init?: HeadersInit): Headers;
};
```

</details>

sterlinghirsh: Ah cool. Thank goodness for inline docs. I think we should do this the simpler way, and if we decide we need more complicated header middleware in the future, we can always add it then.

masonmcelvain:
un_dev_block ✌🏻

frontend/middleware.ts
```diff
@@ -0,0 +1,11 @@
+import { NextResponse } from 'next/server';
+import type { NextRequest } from 'next/server';
+import { PROD_HOSTNAME } from '@config/constants';
+
+export function middleware(request: NextRequest) {
+   const headers =
+      request.nextUrl.hostname === PROD_HOSTNAME
+         ? { 'X-Robots-Tag': 'noindex, nofollow' }
+         : undefined;
+   return NextResponse.next({ headers });
```
sterlinghirsh, line 10: I think this is the format it wants
```suggestion
   const init =
      request.nextUrl.hostname === PROD_HOSTNAME
         ? { headers: ['X-Robots-Tag', 'noindex, nofollow'] }
         : undefined;
   return NextResponse.next(init);
```

masonmcelvain: I think they're equivalent based on the definition of HeadersInit.

`type HeadersInit = [string, string][] | Record<string, string> | Headers;`

But I'm happy to use the array syntax as well.

sterlinghirsh:
This looks good but I don't understand the test failure.

masonmcelvain:
Me neither. It's definitely related to this code change though - I tried opening another PR with the same change, and the test failed in the same way. https://github.com/iFixit/react-commerce/pull/758

danielbeardsley (approved):
CR :+1:

masonmcelvain:
QA ✌🏻 the header is added to both react-commerce and react-commerce-prod

PR #736, opened 2022-09-26
Footer: Move Newsletter and Update Styling

hcutrone:
There are some final changes that need to be made to the Footer before it can be used in iFixit/ifixit. First, since we are not using it in iFixit/ifixit, let's move the Newsletter component back into `/frontend`. Second, there are some global styles in iFixit/ifixit that affect a:visited and a:hover that are not present in react-commerce. This meant that a lot of our links had inconsistent styles on iFixit/ifixit, so I had to be more explicit with the styles.

deploy_block :burrito: on #709 - I based this branch off of `footer-update-deps` so that we would only need to checkout one branch in the react-commerce submodule to test the footer in iFixit/ifixit.

Connects: https://github.com/iFixit/ifixit/issues/43967


jordycosta:
dev_block 🟥 

test

Related to https://github.com/iFixit/react-commerce/pull/709#issuecomment-1258314997

hcutrone:
un_dev_block :burrito: , see https://github.com/iFixit/react-commerce/pull/709#issuecomment-1259827323

jordycosta:
QA 🟢 

Newsletter signup continues to work in the footer ✔️ 
Footer components haven't changed in appearance ✔️

ardelato (approved):
CR 🦖 but deploy_block ☁️ do any of the other `components` need to have the `textDecoration` explicitly set to none? For instance `Partners`?

https://github.com/iFixit/react-commerce/blob/83d3e736cb356afaf19984498af16d5f39675d4f/packages/ui/footer/cart/components/Partners.tsx#L40-L42

hcutrone:
> do any of the other components need to have the textDecoration explicitly set to none?

Good question, @ardelato. I considered this, but decided that since we are not using (and don't plan on using) any of the other components in the iFixit/ifixit footer, that I would leave them alone. I only wanted to change what was needed. That being said, it is a simple fix if we'd rather update them all for future assurance

ardelato:
un_deploy_block 👍

PR #737, opened 2022-09-26
Only append suffix to default titles

### LINKED ISSUE #735: Don't append `| iFixit` when there is a custom meta title in Strapi
## Problem

We want to give Strapi admin users total control over the `<title>` of our Next.js pages. Right now, our code automatically appends `| iFixit` to the `<title>` of all pages, even if there is a custom title from Strapi.

## Do this

Don't append `| iFixit` when there is a custom title from Strapi.

masonmcelvain:
Only append " | iFixit" to the titles of pages that don't have a custom title set in Strapi.

To give Strapi admin users full control over the title, we will use the custom title exactly as it exists on Strapi.

## QA

All product list pages should have ` | iFixit` appended to the title, unless there is a custom `meta title` set in Strapi. If that's the case, the meta title in Strapi is used as is. There should never be a case where it is double appended (like www.ifixit.com/Parts is right now).

Closes #735

### DIFF
--- a/frontend/templates/product-list/MetaTags.tsx
+++ b/frontend/templates/product-list/MetaTags.tsx
@@ -30,11 +30,11 @@ export function MetaTags({ productList }: MetaTagsProps) {
    const isFiltered = currentRefinements.items.length > 0 && !isItemTypeFilter;
    const itemType = useDevicePartsItemType(productList);
    let title =
-      productList.metaTitle || getProductListTitle(productList, itemType);
+      productList.metaTitle ||
+      getProductListTitle(productList, itemType) + ' | iFixit';
    if (!isFiltered && page > 1) {
       title += ` - Page ${page}`;
    }
-   title += ' | iFixit';
    const itemTypeHandle = itemType
       ? `/${encodeURIComponent(stylizeDeviceItemType(itemType))}`
       : '';


erinemay:
QA 🛩️ Looks good. 
- When there is a metaTitle present, it's set as the page title and " | iFixit" is not appended to it. 
- When there isn't a metaTitle present, the page title is the strapi `Title` field + " | iFixit"
- Adding and removing the metaTitle updates the page title appropriately, following the above two conventions.

sumukhipandey:
CR 🦋

sterlinghirsh (approved):
CR :zap: One weird thing is that this will put ` - Page 2` after the `| iFixit` part, i.e. `My Special Stuff | iFixit - Page 2`. I don't know if that's better or worse than `My Special Stuff - Page 2 | iFixit`, so I'm going to merge it as-is, but I figured I'd call it out just in case anyone has a stronger preference.

erinemay:
>this will put `- Page 2` after the `| iFixit` part, i.e. `My Special Stuff | iFixit - Page 2`. I don't know if that's better or worse than `My Special Stuff - Page 2 | iFixit`

I missed this. I don't have a personal preference, but it might be wrong?

masonmcelvain:
I'll fix it, I'm sure max won't like the inconsistency

erinemay:
Thanks @masonmcelvain. Sorry about that.

ifixitmaxb:
If I ever start seeing our paginated collections showing up or outranking our "page 1" collections, then I'll come back to this - but for now, I don't have concerns here. I'm pretty sure it's going to be quite rare that a paginated collection is showing up in the Google SERPs - and even then, Google often changes the Meta Title in the SERPs to whatever they want anyway.

PR #740, opened 2022-09-26
Build seeding api and interface

### LINKED ISSUE #739: Improve seeding of new instances
With #611 we started to seed new Strapi instances using production data. Although this improves the quality of the data that we have for QA, we still needed to change a couple of settings manually to account for differences between production and previews. In particular previews need the data of the `ifixit-test` Shopify store, and this has generated a little bit of confusion so far (e.g. #725 ).
Let's automate these settings so this would be one less thing to think about when we setup a new strapi instance.

dhmacs:
closes #739 

## Changelog

1. Added a UI and API to seed data from arbitrary strapi instances 
  Screenshot 2022-09-26 at 16 03 34
2. Default to import data from `main.govinor.com` instead of `strapi.ifixit.com`

This approach will enable us to customize the data we use for our previews (by editing `main.govinor.com`) while keeping it up to date with production (by using the seeding interface when we want to update `main.govinor.com` ).
Whenever `main.govinor.com` will be reset with production data, changes will need to be re-applied (right now this means just adding back a test store content type which is basically a duplicate of the us store with `ifixit-test` shopify credentials).
I have considered also hard coding the `test` store content type in the seed, but it seemed like a less flexible solution. 

@sterlinghirsh we should update `strapi.ifixit.com` when this gets merged. The UI for seeding shouldn't be available in production. It's only enabled if `STRAPI_ADMIN_ENABLE_ADDONS_DANGEROUS_ACTIONS` is set.

## QA

1. Open [govinor preview for this branch](https://add-seed-from-arbitrary-strapi-instance.govinor.com/admin/plugins/addons)
2. Verify that you can see the new seed UI
3. Visit `Content Manager` > `Store` and delete the stores
4. Go back to the seeding interface and seed with `main.govinor.com`
5. Head back to `Content Manager` > `Store` and verify that stores have been seeded according to main.govinor.com
6. Visit the Vercel preview and verify that the `Test` store is present on the footer region selector


masonmcelvain:
I redeployed strapi.ifixit.com successfully 👍🏻 

image

PR #745, opened 2022-09-27
Revert "Check types in GitHub Actions"

masonmcelvain:
Reverts iFixit/react-commerce#733

Let's wait until we fix the type errors to add this check.


dhmacs (approved):
CR 👍

PR #748, opened 2022-09-27
Bug: Fix repeated facets

### LINKED ISSUE #698: Collections: Bug: /Hard_Drives_ Item Type
Seems like maybe we added an additional Hard Drives Item Type facet on accident. For example, https://www.ifixit.com/Parts/iMac_Intel_17%22/Hard_Drives_.

image

I don't know how prevalent this is but we should find out and then fix it. @erinemay

sumukhipandey:
Previously, we  had some repeated facets being displayed. Although I don't know if this is the solution we are looking for here but this PR basically makes sure that we do not display any facets that have an `item.count` of 0.

## QA:
- Visit the preview and add `/Parts/iMac_Intel_17"/Hard_Drives_` 
- Make sure there are no repeated facets as mentioned in the issue.

Closes #698 

CC: @erinemay @sterlinghirsh

### DIFF
--- a/frontend/components/common/InstantSearchProvider.tsx
+++ b/frontend/components/common/InstantSearchProvider.tsx
@@ -174,11 +174,7 @@ export function InstantSearchProvider({
                .split('/')
                .filter((part) => part !== '');
             const deviceHandle = pathParts.length >= 2 ? pathParts[1] : '';
-            let itemType = pathParts.length >= 3 ? pathParts[2] : '';
-
-            if (itemType.endsWith('_')) {
-               itemType = itemType.substring(0, itemType.length - 1);
-            }
+            const itemType = pathParts.length >= 3 ? pathParts[2] : '';
 
             const { q, p, filter } = qsModule.parse(location.search.slice(1));
 
@@ -189,7 +185,7 @@ export function InstantSearchProvider({
 
             if (deviceHandle && itemType) {
                filterObject['facet_tags.Item Type'] = [
-                  destylizeDeviceItemType(decodeURIComponent(itemType)),
+                  destylizeDeviceItemType(decodeURIComponent(itemType)).trim(),
                ];
             }
 
--- a/frontend/components/common/InstantSearchProvider.tsx
+++ b/frontend/components/common/InstantSearchProvider.tsx
@@ -174,7 +174,11 @@ export function InstantSearchProvider({
                .split('/')
                .filter((part) => part !== '');
             const deviceHandle = pathParts.length >= 2 ? pathParts[1] : '';
-            const itemType = pathParts.length >= 3 ? pathParts[2] : '';
+            let itemType = pathParts.length >= 3 ? pathParts[2] : '';
+
+            if (itemType.endsWith('_')) {
+               itemType = itemType.substring(0, itemType.length - 1);
+            }
 
             const { q, p, filter } = qsModule.parse(location.search.slice(1));
 
--- a/frontend/templates/product-list/sections/FilterableProductsSection/RefinementMultiSelect.tsx
+++ b/frontend/templates/product-list/sections/FilterableProductsSection/RefinementMultiSelect.tsx
@@ -68,28 +68,31 @@ const MultiSelectItem = React.memo(function MultiSelectItem({
 }: MultiSelectItemProps) {
    const [isRefined, setIsRefined] = useDecoupledState(item.isRefined);
 
-   return (
-      <HStack key={item.label} justify="space-between">
-         <Checkbox
-            role="option"
-            value={item.value}
-            isChecked={isRefined}
-            onChange={() => {
-               setIsRefined((current) => !current);
-               refine(item.value);
-            }}
-            data-value={item.value}
-         >
-            {item.label}
-         </Checkbox>
-         <Text
-            size="sm"
-            fontFamily="sans-serif"
-            color="gray.500"
-            fontWeight={isRefined ? 'bold' : 'inherit'}
-         >
-            {item.count}
-         </Text>
-      </HStack>
-   );
+   if (item.count > 0) {
+      return (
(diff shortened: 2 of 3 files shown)


dhmacs:
Seems like we're trying to cure the symptoms and not the disease. I'd try to understand why we have repeated items and how we can prevent that. Also `useRefinementList` isn't supposed to send back `0` count items I think, so I wonder how we end up having this

jordycosta:
QA 🟢 

No longer seeing duplicate facets

[https://react-commerce-prod-git-bug-fix-repeated-facets-ifixit.vercel.app/Parts/iMac_Intel_17%22/Hard_Drives_](https://react-commerce-prod-git-bug-fix-repeated-facets-ifixit.vercel.app/Parts/iMac_Intel_17%22/Hard_Drives_)

---

sterlinghirsh:
dev_block :zap: I'm not seeing this bug either. Maybe it was bad data or another bug that got fixed already, but unless we have a repro, I don't want to merge this change. I agree with @dhmacs that if we're seeing this bug, we should figure out what that is rather than adding a bandaid fix that will stick around forever and cover up deeper problems. If we're not seeing the bug anymore, then let's move on and not worry about it.

erinemay:
@sterlinghirsh it's a weird url https://github.com/iFixit/react-commerce/issues/698#issuecomment-1260163346

sterlinghirsh:
Ah okay. Looks like any item type with an underscore will do this. There's another bug here for sure that is causing it to get added to the list, so if we're going to fix it, let's fix it there rather than filtering it out later.

sumukhipandey:
un_dev_block 🦋  on 725cd864b68bb5afbb91a9386b4bccd1609b6279
@sterlinghirsh @dhmacs I wasn't sure if there was a cleaner and better solution to this and/or if this is where we want to address the bug. I am un dev blocking for right now.

frontend/templates/product-list/sections/FilterableProductsSection/RefinementMultiSelect.tsx
```diff
@@ -68,28 +68,31 @@ const MultiSelectItem = React.memo(function MultiSelectItem({
...
-            size="sm"
-            fontFamily="sans-serif"
-            color="gray.500"
-            fontWeight={isRefined ? 'bold' : 'inherit'}
-         >
-            {item.count}
-         </Text>
-      </HStack>
-   );
+   if (item.count > 0) {
```
sterlinghirsh, line 71: Let's get rid of these ifs since we should be operating under the assumption that nothing should show up with item.count = 0 and this will hide it if that breaks.

frontend/components/common/InstantSearchProvider.tsx
```diff
@@ -174,7 +174,11 @@ export function InstantSearchProvider({
                .split('/')
                .filter((part) => part !== '');
             const deviceHandle = pathParts.length >= 2 ? pathParts[1] : '';
-            const itemType = pathParts.length >= 3 ? pathParts[2] : '';
+            let itemType = pathParts.length >= 3 ? pathParts[2] : '';
+
+            if (itemType.endsWith('_')) {
+               itemType = itemType.substring(0, itemType.length - 1);
+            }
```
sterlinghirsh, line 181: I'm not 100% sure what's going on with the itemType here. Does using an itemType in the filterObject cause it to be added to the facets list? I could see Algolia doing that just to show the selected facet even if it's 0, though it is a little odd. Maybe our own code is doing that in the place where we filter facet options? I'm not sure. In any case, I think a cleaner way to accomplish this is to leave the itemType as a const, but down on line 192 when we destylizeDeviceItemType, we can add a .trim() call like `destylizeDeviceItemType(decode...).trim()` will drop any leading or trailing whitespace.

sumukhipandey:
un_dev_block 🦋  on 503da96d26a0f2e1a458a6b4a8ccccdd18913412 and d8e2412efef6e68936c1fd6462b0c5b09fccfafd

frontend/templates/product-list/sections/FilterableProductsSection/RefinementMultiSelect.tsx
sumukhipandey: Resolved d8e2412efef6e68936c1fd6462b0c5b09fccfafd

frontend/components/common/InstantSearchProvider.tsx
sumukhipandey: Resolved 503da96d26a0f2e1a458a6b4a8ccccdd18913412

danielcliu-ifixit:
CR :v:

frontend/templates/product-list/sections/FilterableProductsSection/RefinementMultiSelect.tsx
```diff
@@ -92,4 +92,5 @@ const MultiSelectItem = React.memo(function MultiSelectItem({
          </Text>
       </HStack>
    );
+   return null;
```
danielcliu-ifixit, line 95: do we need these return nulls?

sumukhipandey: Thanks for pointing it out! You are right we do not need these anymore.
Resolved 5e36b200fbc107858ffc45e7501bab9f0d45b083

sumukhipandey:
un_dev_block 🦋 on 5e36b200fbc107858ffc45e7501bab9f0d45b083

danielcliu-ifixit:
CR :v:

erinemay:
QA 👍 
- item type collections with a trailing underscore now strip the underscore and loads without it
- there are no repeating item type facets
- item type filters with other filters still work
- refreshing with filters leaves the filters in place

danielcliu-ifixit (approved):

Issue #529, opened 2022-07-28
Redirect some more Product List tags 


erinemay:
https://github.com/iFixit/react-commerce/pull/524 was done, but there are others trickling in that weren't caught.

List:
| example old collection path | ifixit redirect | next redirect | non-existent item type |
| -- | -- | -- | -- |
| /Store/iPad/iPad-Pro/USB_Boards | /Parts/iPad_Pro/USB_Boards | /Parts/iPad_Pro/Boards | Boards |
| /Store/iPad/iPad-Pro/Gripping | /Parts/iPad_Pro/Gripping | none | Gripping |
| /Store/iPad/iPad-Pro/Prying_and_Opening | /Parts/iPad_Pro/Prying_and_Opening | none | Prying_and_Opening |
| /Store/iPad/iPad-Pro/Sensors | /Parts/iPad_Pro/Sensors | none | Sensors |
| *️⃣ /Store/Game-Console/Sony-Game-Consoles/I | /Parts/Sony_Game_Console/I | none | I |
| /Store/Game-Console/PlayStation-5/Optical_Drives | /Parts/PlayStation_5/Optical_Drives | none | Optical_Drives |
| /Store/Game-Console/Steam-Deck/Power_Supplies | /Parts/Steam_Deck/Power_Supplies | none | Power_Supplies |

*️⃣  - redirected from `/Store/Game-Console/Sony-Game-Consoles/I/O_Board` which was already broken on the old pages

CC: @ifixitmaxb @davidrans

erinemay:
I'm poking through some old collections a little more thoroughly now. Updating that table as I find problematic item types.

ifixitmaxb:
These legacy Collections still redirect to a /Logic_Boards Item Type and they should be redirecting to /Motherboards:
https://www.ifixit.com/Store/iPad/iPad/Logic_Boards
https://www.ifixit.com/Store/iPad/iPad-Air/Logic_Boards
https://www.ifixit.com/Store/iPad/iPad-Air-2/Logic_Boards
https://www.ifixit.com/Store/iPad/iPad-Pro/Logic_Boards
https://www.ifixit.com/Store/iPad/iPad-Pro-12.9”/Logic_Boards
https://www.ifixit.com/Store/iPhone/iPhone/Logic_Boards
https://www.ifixit.com/Store/iPhone/iPhone-6s/Logic_Boards
https://www.ifixit.com/Store/iPhone/iPhone-7/Logic_Boards
[...]

PR #755, opened 2022-09-28
Some More Legacy Redirect Fixes

danielcliu-ifixit:
## Overview

Basically #529 had comments floating around it in and there were some more complaints to address.
This PR

1. Handles incorrect item types where they have spaces instead of "_" so things like `Parts/iPhone_11/Adhesive%20Strips` will now correctly reroute to `/Adhesives`. This might seem like an unnecessary fix but we are already handling things like `iPhone%2011` correctly so at least this stays consistent.
2. Redirect both  `/SSD_Enclosures` and `/Upgrade_Kits` to  `/Hard_Drives`.
3. Reorganize the `tool-collections` redirect generation so its more analogous to the parts one.

## QA
It should be pretty easy to test 1.) and 2.) above. 
`/Parts/HTC_Vive/I%2FO%20Board` seems like a decent route to test with.

Connects: #529

### DIFF
--- a/frontend/next-config/redirects/part-collections.js
+++ b/frontend/next-config/redirects/part-collections.js
@@ -1,12 +1,12 @@
-const { escapeStringForRegex } = require('../helpers/redirects');
+const { escapeStringForRegex, destylizeURIComponent } = require('../helpers/redirects');
 const legacyFilterSlugs = require('./legacy-filter-slugs.json');
 
-function mapPartItemTypes() {
+function getLegacyPartItemTypeRedirects() {
    const legacyTagToItemType = legacyFilterSlugs['tagToItemType'];
    // Redirect tags we've renamed to the new item type name
    const oldToNew = Object.entries(legacyTagToItemType).map(
       ([oldTag, itemType]) => ({
-         source: `/Parts/:device/${escapeStringForRegex(oldTag)}`,
+         source: `/Parts/:device/${getRegexForTag(oldTag)}`,
          destination: `/Parts/:device/${itemType}`,
          permanent: true,
       })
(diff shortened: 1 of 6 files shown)


deltuh-vee:
QA 🎬 
* Item types with spaces get redirected to their proper routes
* Both /SSD_Enclosures and /Upgrade_Kits links redirect to /Hard_Drives

frontend/next-config/redirects/part-collections.js
```diff
@@ -27,9 +27,10 @@ function getLegacyPartItemTypeRedirects() {
 }
 
 function getRegexForTag(oldTag) {
-   return `(${escapeStringForRegex(
-      encodeURIComponent(destylizeURIComponent(oldTag))
-   )}|${escapeStringForRegex(oldTag)})`;
+   const encodedTag = encodeURIComponent(destylizeURIComponent(oldTag));
+   return encodedTag === oldTag
+      ? escapeStringForRegex(oldTag)
+      : `(${escapeStringForRegex(encodedTag)}|${escapeStringForRegex(oldTag)})`;
```
ardelato, line 33: Wouldn't this regex cover both cases so the ternary operator would not be needed?
- `encodedTag` **equals** `oldTag`
- `encodedTag` **does not equal** `oldTag`

ardelato (approved):
CR 👍 , but question about the regex matching

frontend/next-config/redirects/part-collections.js
danielcliu-ifixit: Yes this is to reduce the regex where we can. While `(Hello|Hello)` would technically work, i think we should just reduce it to `Hello` where we can.

masonmcelvain (approved):
CR ✌🏻

PR #758, opened 2022-09-29
feat: middleware to no-index non ifixit.com pages

masonmcelvain:
We only want to index ifixit.com urls, not vercel.app urls. This will no-index all pages that are not hosted on ifixit.com.


masonmcelvain:
dev_block 👍🏻 this is a test

PR #760, opened 2022-09-30
[Hotfix] Fix pro prices

### LINKED ISSUE #759: Pro prices are not correctly displayed on the product list page
When logged in as a pro user, pro prices are not correctly displayed

Screenshot 2022-09-30 at 18 24 06

## Do this

Turns out the `price_tiers` attribute has been updated with a different shape, so we need to update the frontend code accordingly

dhmacs:
fixes #759 

## QA

1. Visit Vercel preview
2. Visit `/Tools` product list
3. Verify that bug described by #759 is fixed

### DIFF
--- a/frontend/models/product-list/types.ts
+++ b/frontend/models/product-list/types.ts
@@ -4,7 +4,7 @@ export interface ProductSearchHit {
    handle: string;
    price_float: number;
    compare_at_price?: number;
-   price_tiers?: Record<string, string>;
+   price_tiers?: Record<string, PriceTier>;
    sku: string;
    image_url: string;
    short_description?: string;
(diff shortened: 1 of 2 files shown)


sterlinghirsh (approved):
CR :zap: thanks for getting this out so quick

frontend/models/product-list/types.ts
```diff
@@ -18,6 +18,12 @@ export interface ProductSearchHit {
    [attribute: string]: unknown;
 }
 
+export type PriceTier = {
+   default_variant_price: string | number;
+   min: string | number;
+   max: string | number;
```
masonmcelvain, line 24: I think these are always supposed to be floats in algolia, so I don't think you need `string` here. Have you seen strings in practice? (Maybe floats are serialized as strings from Algolia?)

masonmcelvain (approved):
CR ✌🏻 with a question

sterlinghirsh:
QA :zap: I checked this against main, and the products with $NaN now have a price.

frontend/models/product-list/types.ts
dhmacs: No I did this just to be safe.. I didn't want to fail just because the format is a string

PR #764, opened 2022-10-03
Icons: Depend on the icon lib we use

jarstelfox:
We never required the icons to be pulled in? Weird!

qa_req 0


andyg0808 (approved):
CR :+1:

PR #773, opened 2022-10-04
packages/ui: Pull out shared components

hcutrone:
We want to pull out shared components into their own package so that we don't end up having to deal with dependency issues for things that are not used in iFixit/ifixit.

qa_req 0 - Ran through the footer and tested that everything is looking/working like on master

Connects: https://github.com/iFixit/ifixit/issues/44935


aburke07 (approved):
CR :alien: Nice layout of commits.

PR #774, opened 2022-10-05
Handle disable_when_oos metafield

### LINKED ISSUE #701: Respect disable_when_oos
A variant can have a metafield called disable_when_oos. This relates to clearance products that we don't want to display once we run out, e.g. used products. If this metafield is true, the product should behave as if it's disabled when stock = 0.

This relates to a problem @federicobadini had when trying to add something to the cart.

federicobadini:
closes #701 

This PR disable product variants that have `disable_when_oos = true` and `variant.availability <= 0`.
Such disabled variants are filtered out from the active variant set.
Moreover, product options that are not associated to any active variant are filtered out too, since hinting to the existence of those variants is not useful anymore.

To summarize:

- Variants with availabilities are always active
- Variants with zero availability and `disable_when_oos = false` show the `Notify me` banner
- Variants with zero availability and `disable_when_oos = true` will not be shown on the website. 
**If a product option is associated only to disabled variants or no variants at all, it will be hidden**

### QA

1. Visit Vercel preview (you can also take [this product](https://react-commerce-git-respect-disable-when-oos-ifixit.vercel.app/Products/iphone-11-screen) as a reference since it has variants with 0 availability and `disable_when_oos = true`
2. Try playing with variant availabilities and their associated `disable_when_oos` metafield
3. The summary above should always be valid

### DIFF
--- a/frontend/models/product.ts
+++ b/frontend/models/product.ts
@@ -25,6 +25,18 @@ export async function findProduct(shop: ShopCredentials, handle: string) {
       return null;
    }
    const variants = getVariants(response.product);
+   const activeVariants = variants.filter(
+      (variant) =>
+         !variant.disableWhenOOS ||
+         (variant.quantityAvailable && variant.quantityAvailable > 0)
+   );
+   if (activeVariants.length === 0) {
+      return null;
+   }
+
+   const images = getImages(response.product.images, variants);
+   const options = getOptions(response.product.options, activeVariants);
+
    const variantSku = variants.find((variant) => variant.sku != null)?.sku;
    if (variantSku == null) {
       console.warn(`No sku found for product "${handle}"`);
@@ -47,8 +59,9 @@ export async function findProduct(shop: ShopCredentials, handle: string) {
       ...response.product,
       breadcrumbs,
       iFixitProductId,
-      variants,
-      images: response.product.images.nodes,
+      images,
+      options,
+      variants: activeVariants,
       prop65WarningType: response.product.prop65WarningType?.value ?? null,
       prop65Chemicals: response.product.prop65Chemicals?.value ?? null,
       productVideos: response.product.productVideos?.value ?? null,
@@ -90,8 +103,45 @@ function getVariants(shopifyProduct: NonNullable<FindProductQuery['product']>) {
          specifications: variant.specifications?.value ?? null,
          warranty: variant.warranty?.value ?? null,
          crossSellVariants: getCrossSellVariants(variant),
+         disableWhenOOS: variant.disableWhenOOS?.value === 'true',
       };
    });
+}
+
+function getImages(
+   shopifyImages: NonNullable<FindProductQuery['product']>['images'],
+   variants: ReturnType<typeof getVariants>
+) {
+   return shopifyImages.nodes.filter(
+      (image) =>
+         image.altText == null ||
+         !variants.find((variant) => {
+            const skuAltTextMatch = variant.sku === image.altText;
+            const variantDisabled =
+               variant.disableWhenOOS &&
+               variant.quantityAvailable != null &&
+               variant.quantityAvailable <= 0;
+            return skuAltTextMatch && variantDisabled;
+         })
+   );
+}
+
+function getOptions(
+   shopifyOptions: NonNullable<FindProductQuery['product']>['options'],
+   activeVariants: ReturnType<typeof getVariants>
+) {
+   return shopifyOptions.map((option) => ({
+      ...option,
+      values: option.values.filter((value) =>
+         activeVariants.find((variant) =>
+            variant.selectedOptions.find(
+               (selectedOption) =>
+                  selectedOption.name === option.name &&
+                  selectedOption.value === value
+            )
+         )
+      ),
+   }));
 }
 
 function getCrossSellVariants(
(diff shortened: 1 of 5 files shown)


frontend/models/product.ts
```diff
@@ -25,6 +25,15 @@ export async function findProduct(shop: ShopCredentials, handle: string) {
       return null;
    }
    const variants = getVariants(response.product);
+   const activeVariants = variants.filter(
+      (variant) =>
+         !variant.disableWhenOOS ||
+         (variant.quantityAvailable && variant.quantityAvailable > 0)
+   );
+
+   const images = getImages(response.product.images, variants);
```
masonmcelvain, line 34: Do we want to only get images for active variants here?

```suggestion
   const images = getImages(response.product.images, activeVariants);
```

frontend/models/product.ts
```diff
@@ -90,10 +100,46 @@ function getVariants(shopifyProduct: NonNullable<FindProductQuery['product']>) {
          specifications: variant.specifications?.value ?? null,
          warranty: variant.warranty?.value ?? null,
          crossSellVariants: getCrossSellVariants(variant),
+         disableWhenOOS: variant.disableWhenOOS?.value === 'true',
```
masonmcelvain, line 103: Oh my are these booleans serialized as strings? I thought we added a metafield definition that made `disable_when_oos` a boolean, and I would expect the storefront API to send you a boolean. @sterlinghirsh does that sound right?

frontend/models/product.ts
```diff
@@ -90,10 +100,46 @@ function getVariants(shopifyProduct: NonNullable<FindProductQuery['product']>) {
...
+) {
+   return shopifyImages.nodes.filter(
+      (image) =>
+         !variants.find(
+            (variant) =>
+               image.altText &&
+               variant.sku === image.altText &&
+               variant.disableWhenOOS &&
+               variant.quantityAvailable != null &&
+               variant.quantityAvailable <= 0
```
masonmcelvain, line 120: If we pass in only active variants, then we probably don't need these comparisons I think.

frontend/models/product.ts
```diff
@@ -90,10 +100,46 @@ function getVariants(shopifyProduct: NonNullable<FindProductQuery['product']>) {
...
    });
 }
 
+function getImages(
+   shopifyImages: NonNullable<FindProductQuery['product']>['images'],
+   variants: ReturnType<typeof getVariants>
+) {
+   return shopifyImages.nodes.filter(
+      (image) =>
+         !variants.find(
```
masonmcelvain, line 114: I might be misreading this boolean expression, but it seems like this will exclude any images with altText, which doesn't seem like what we want.

frontend/models/product.ts
```diff
@@ -90,10 +100,46 @@ function getVariants(shopifyProduct: NonNullable<FindProductQuery['product']>) {
...
+   );
+}
+
+function getOptions(
+   shopifyOptions: NonNullable<FindProductQuery['product']>['options'],
+   activeVariants: ReturnType<typeof getVariants>
+) {
+   return shopifyOptions.map((option) => ({
+      ...option,
+      values: option.values.filter((value) =>
```
masonmcelvain, line 131: What happens if `values` ends up being an empty array? Looking at `ProductOption.tsx`, it seems like we'd get some strange UI, or array indexing errors. Maybe there's a conditional that I'm missing somewhere that guards against this.

masonmcelvain (changes requested):
CR ✌🏻 Looks great! dev_block ✌🏻 on a few concerns, though I might just be misunderstanding how this code works. If that's the case, feel free to un_dev_block.

frontend/models/product.ts
federicobadini: Metafield values on storefront api seems to be always string: https://shopify.dev/api/storefront/2022-10/objects/metafield

frontend/models/product.ts
federicobadini: See [answer above](https://github.com/iFixit/react-commerce/pull/774#discussion_r988575778)

frontend/models/product.ts
federicobadini: I rewrote the condition to make it more readable.

We are keeping `altText = null` meaning the explicitly generic images
We are filtering out images with altText matching disabled variants
We still keep images with an altText not matching any variant skus - since these might be seo based texts inserted in generic images

masonmcelvain: Thank you for the explanation and the change!

masonmcelvain (approved):
CR ✌🏻 Thank for the responses and the changes 😄

masonmcelvain:
un_dev_block ✌🏻

erinemay:
QA 👍 It's clear when a variant is not available
image

When `disable_when_oos` is set to true, or a variant doesn't exist, that would-be variant is not available on the page:
- If it's a part/kit, the unavailable image is above and button isn't selectable. 
- If it's a variant represented in a dropdown, that option is not present in the dropdown. 

OOS alert email input continues to work for OOS variants with `disable_when_oos` set to false.

deploy_block ❔ I noticed this error in my console on this branch, and not on the [main preview](https://react-commerce.vercel.app/Products/iphone-11-pro-screen). It's also present on other new branch previews. I'm assuming it's a result of the preview setup, but I just want to be sure. 
image

As I was clicking through here, I found some minor UE issues, but none were introduced in this branch. I'll open up a couple of new issues.

sterlinghirsh:
un_deploy_block :zap: I'm not seeing that error. Looks like it might be related to a chrome plugin.

PR #775, opened 2022-10-05
Fix layout shift in gallery thumbnails

### LINKED ISSUE #762: Product Page: Images causing layout shift
Try loading https://react-commerce.vercel.app/Products/pro-tech-toolkit with your connection throttled down. 

(Capture from Chrome on Windows)

https://user-images.githubusercontent.com/72166715/193631624-35318c57-1dbf-495c-a3ae-a48859dd0129.mp4

federicobadini:
closes #762 

The problem seems to be due to the fact that before js kicks in, thumbnails are rendered with a width equal to 100% of the column width. I've inserted some css to resize images to the same scale that will be enforced as soon as js kicks in.

### QA

1. Visit Vercel preview
2. No thumbnail induced shift should be visible

### DIFF
--- a/frontend/templates/product/sections/ProductSection/ProductGallery.tsx
+++ b/frontend/templates/product/sections/ProductSection/ProductGallery.tsx
@@ -139,6 +139,7 @@ export function ProductGallery({
                         key={variantImage.id}
                         style={{
                            maxWidth: 'calc((100% - 60px) / 6)',
+                           marginRight: '12px',
                         }}
                      >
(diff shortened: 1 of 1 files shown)


jordycosta:
QA 🟢 

No longer seeing a thumbnail induced shift on page reload 👍 

---

### Demo (before/after)

firefox_fr3Q1g2RdQ

masonmcelvain (approved):
CR ✌🏻

PR #771, opened 2022-10-04
Noindex dev domains via middleware

masonmcelvain:
This is a simpler approach to noindexing dev domains. We couldn't use it before because it caused Cypress to fail the collection scroll test. Now that this passes tests we prefer to use it.

## QA

Make sure the previews are no-indexed. (see #734)

PR #779, opened 2022-10-06
Noindex previews via middleware

masonmcelvain:
Instead of noindexing dev domains in getServerSideProps, we can noindex all pages via a middleware.

We tried deploying this in #771, and it started noindexing www.ifixit.com. Let's log the `nextUrl` so we can see what values are on the client and the server.


masonmcelvain:
dev_block ✌🏻 for debugging

danielbeardsley:
Oh of course! The `Host` header *has* to be `react-commerce-prod.vercel.app` otherwise, vercel wouldn't know where to send the request. Bummer that we didn't catch it.

masonmcelvain:
QA ✌🏻 the previews are noindexed with `X-Robots-Tag`. un_dev_block ✌🏻

frontend/middleware.ts
```diff
@@ -0,0 +1,11 @@
+import { NextResponse } from 'next/server';
+import type { NextRequest } from 'next/server';
+import { PROD_USER_AGENT } from '@config/constants';
+
+export function middleware(request: NextRequest) {
+   const init =
+      request.headers.get('user-agent') === PROD_USER_AGENT
+         ? undefined
+         : { headers: { 'X-Robots-Tag': 'noindex, nofollow' } };
+   return NextResponse.next(init);
```
danielbeardsley, line 10: Can you link to something that indicates this is how to mutate the response?

The docs only say:

> next() - Returns a NextResponse that will continue the middleware chain 

The docs I've seen don't mention this function taking any arguments.

danielbeardsley (approved):
CR :+1:   But deploy_block :+1:  on a question.

frontend/middleware.ts
masonmcelvain: oh oopsies that's not what you're asking

masonmcelvain: Here's the source of `NextResponse.next()`

https://github.com/vercel/next.js/blob/24b600303f2e8074c32b79d9b799ade85772d75e/packages/next/server/web/spec-extension/response.ts#L80-L85

danielbeardsley: Thanks!

danielbeardsley:
un_deploy_block :+1:

masonmcelvain:
I don't see no-index header on [www.ifixit.com/Parts](http://www.ifixit.com/Parts), but it is present on react-commerce-prod.vercel.app :tada:

image

image

PR #778, opened 2022-10-05
Revert "Noindex dev domains via middleware"

masonmcelvain:
Reverts iFixit/react-commerce#771

This is noindexing www.ifixit.com/Parts, yikes!

image

PR #780, opened 2022-10-06
Stop noindexing prod

masonmcelvain:
Reverts #734 

Also adds some logging so we can find out what Next thinks the hostname is, because comparisons against `www.ifixit.com` seem to be false in prod.

## QA

Make sure the previews don't have `X-Robot-Tag` header

Connects #778


masonmcelvain:
I'm confused why the header is still there. I wonder if it's being added by code that isn't ours, since it's missing `nofollow` now.

image

PR #786, opened 2022-10-06
Have Single Select Refinement UI Change Look Smoother

### LINKED ISSUE #526: Default results appear when switching single-select facets

danielcliu-ifixit:
## Overview

Basically a reopening of #617 now that the `instant-search` update is working on vercel.

When we used to swap single select refinements we would sometimes have the no-refinements selection blip on the screen for a half second (see the issue #526 for an example). This pr should solve that problem by updating the full Index UI State at once (including the refinements) and not clearing out the refinements and then setting them which was why we would have a slight blip of wrong results.

## CR
Is there a more efficient way to organize the data for the IndexUIState than what I have?

## QA
Make sure that pages don't have the blip of wrong results when switching single select refinements (note that even on master this blip disappears on switching if youve already switched between the two refinements on a page load. it must be caching the refinement results in the browser). Make sure that refinements are working as intended. Also make sure that performance isn't down the toilet.

Closes #526

### DIFF
--- a/frontend/templates/product-list/sections/FilterableProductsSection/RefinementSingleSelect.tsx
+++ b/frontend/templates/product-list/sections/FilterableProductsSection/RefinementSingleSelect.tsx
@@ -3,15 +3,13 @@ import { faSort } from '@fortawesome/pro-solid-svg-icons';
 import { stylizeDeviceItemType } from '@helpers/product-list-helpers';
 import { FaIcon } from '@ifixit/icons';
 import { ProductList, ProductListType } from '@models/product-list';
-import {
-   RefinementListItem,
-   RefinementListRenderState,
-} from 'instantsearch.js/es/connectors/refinement-list/connectRefinementList';
+import { RefinementListRenderState } from 'instantsearch.js/es/connectors/refinement-list/connectRefinementList';
 import NextLink from 'next/link';
 import React from 'react';
 import {
    useCurrentRefinements,
    UseRefinementListProps,
+   useInstantSearch,
 } from 'react-instantsearch-hooks-web';
 import { useFilteredRefinementList } from './useFilteredRefinementList';
 import { useSortBy } from './useSortBy';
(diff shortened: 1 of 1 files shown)


danielcliu-ifixit:
Something is going on weird with our product metafields and thats making this not work rn. Im going to try to solve that issue first

masonmcelvain (approved):
CR 🔥

frontend/templates/product-list/sections/FilterableProductsSection/RefinementSingleSelect.tsx
```diff
@@ -106,7 +89,21 @@ const SingleSelectItem = React.memo(function SingleSelectItem({
...
+            setIndexUiState((prevUiState) => {
+               const refinementList = {
+                  ...prevUiState.refinementList,
+                  [attribute]: [item.value],
+               };
+               if (
+                  item.value === prevUiState?.refinementList?.[attribute]?.[0]
+               ) {
+                  delete refinementList[attribute];
+               }
```
masonmcelvain, line 101: This is the only other way I can think of, but I think what you have is fine

```suggestion
               const refinementList = item.value === prevUiState?.refinementList?.[attribute]?.[0]
                  ? {
                        ...prevUiState.refinementList,
                        [attribute]: undefined,
                  } : {
                        ...prevUiState.refinementList,
                        [attribute]: [item.value],
                  };
```

k0rvusk0r4x:
The blip of wrong results is gone.

However, the images are now taking a rather long time to load.

dev_block 😿

sterlinghirsh:
un_dev_block :zap: I am not seeing a blip nor particularly slow loading of images. Given that no image code was changed, I'm going to say let's merge this and if we see poor image performance going forward, we can look into it or revert.

sterlinghirsh:
QA :zap:

PR #802, opened 2022-10-09
Track product page with Matomo

### LINKED ISSUE #792: Matomo: Track product page activity
[`product-page-matomo`](https://github.com/iFixit/react-commerce/compare/product-page-matomo)

## Problem

We want to track page visits and ecommerce events on the product page.

## Do This

Refactor our matomo code to track visits to any page in Next.js. I wonder if we can add this in one global page so we don't have to add the script to all the templates.

Also determine which [ecommerce](https://developer.matomo.org/api-reference/tracking-javascript) events take place on the product page, and track them appropriately.

Our shopify-v2 [implementation](https://github.com/iFixit/shopify-v2/blob/master/src/scripts/analytics/matomo-cart.ts)
Our ifixit [implementation](https://github.com/iFixit/ifixit/tree/master/carpenter-frontend/Shared/Analytics/Matomo)

masonmcelvain:
Tracks product page views in Matomo and Google Analytics.

Tracks ecommerce product views in Matomo.

Tracks ecommerce carts in Matomo.

## QA

Make sure all of the above types of activities get tracked appropriately.

Closes #792


frontend/templates/product/index.tsx
```diff
@@ -23,6 +25,16 @@ export const ProductTemplate: NextPageWithLayout<ProductTemplateProps> = () => {
...
 
+   React.useEffect(() => {
+      trackMatomoEcommerceView({
+         productSku: selectedVariant.sku ?? selectedVariant.id,
+         productName:
+            selectedVariant.internalDisplayName?.value ?? product.title,
+         price: selectedVariant.price,
+      });
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+   }, []);
```
sterlinghirsh, line 36: Why doesn't this depend on selectedVariant?

packages/matomo/index.tsx
```diff
@@ -1,13 +1,110 @@
...
+   const totalPrice = sumMoney(items.map((i) => i.price));
+   trackCartUpdated(totalPrice);
+}
+
+/**
+ * @see https://developer.matomo.org/api-reference/tracking-javascript
+ * @see https://matomo.org/docs/ecommerce-analytics/#example-of-adding-a-product-to-the-order
+ */
+type AddToCartData = Omit<ProductData, 'price'> & {
+   price?: Money;
```
sterlinghirsh, line 73: Why do we omit and re-add price?

packages/matomo/index.tsx
```diff
@@ -1,13 +1,110 @@
...
+      product.categoryName,
+      product.price.amount,
+   ]);
+}
+
+export function trackMatomoCartChange(items: CartLineItem[]) {
+   if (!_paq) {
+      return;
+   }
+   trackClearCart();
```
sterlinghirsh, line 51: This seems wrong. Why do we clear and re-add every time the cart is changed? Seems like that's going to add a lot of fake events to Matomo.

sterlinghirsh:
CR :zap: dev_block :zap:

packages/matomo/index.tsx
masonmcelvain: This is how we do it in ifixit https://github.com/iFixit/ifixit/blob/master/carpenter-frontend/Shared/Analytics/Matomo/subcribeToMiniCart.ts.

@jarstelfox can you remind us why it's done this way in iFixit/ifixit?

packages/matomo/index.tsx
masonmcelvain: Price is a mandatory field for product view events (ie ProductData), and it's an optional field in add to cart events.

frontend/templates/product/index.tsx
masonmcelvain: As I understand it, in ifixit we just track the default selected option, instead of recording a product view every time the user changes the selected variant in the UI.
https://github.com/iFixit/ifixit/blob/5b354e81eace15e414f45b26595003a6c7ec6462/carpenter-frontend/Cart/product_view.js#L919

The empty array only tracks this product once. If we depend on selectedVariant, then we will track the product every time the user switches the variant.

packages/matomo/index.tsx
masonmcelvain: image
https://developer.matomo.org/api-reference/tracking-javascript

frontend/templates/product/index.tsx
masonmcelvain: @jarstelfox does that sound right?

packages/matomo/index.tsx
jarstelfox: Matomo is stateful: every time you call out to Matomo saying you are adding an item, it will add the state it it’s view.

The problem on the main site was keeping cart data in Matomo completely in sync with the cart state. Mainly, the hook I was using to grab the cart state just tells me a user has 1 of X and 1 of Y in their cart, not that the user only now just added Y to the cart.

Full example:

Cart empty. User views page X and adds X to cart. Fire cart={ X: 1 } to Matomo. Matomo’s cart: { X: 1 } User views page Y and adds Y to cart. Fire cart: { X: 1, Y: 1}. Matomo’s cart: { X: 2, Y: 1 }

In fact, the problem with the main repo’s code is the cart update event was fired on every page load since we grab that data over the wire.

In short, this could all be fixed with some state diffing, but this was the safest way to shop for V1

frontend/templates/product/index.tsx
jarstelfox: @masonmcelvain that does sound correct.

This comes down to a business decision. Do we want the view product to be the first view product xor do we want the viewed product plus any viewed variants?

packages/matomo/index.tsx
masonmcelvain: Thanks for the explanation! Since we're ok with this behavior on the current product page, I'm inclined to say that it's fine to keep the status quo for V1 of the Next.js product page. We can devote more time to this after we launch.

frontend/templates/product/index.tsx
masonmcelvain: Let's keep the status quo for V1 of the Next.js product page, and let folks that use Minerva decide if they want to change this later on.

masonmcelvain:
un_dev_block ✌🏻 Assuming we're ok with keeping the same Matomo behavior as ifixit to get this out the door for the launch

sterlinghirsh (approved):
CR :zap:

erinemay:
dev_block 🌵  on some things

### Page views
I see some page views (collection and product) in Matomo, but the event title(?) is iFixit for everything.
<details><summary>This is two page views here</summary>

image

</details>

<details><summary>Page views on the php product page for reference</summary>

image

</details>

If this pull is only for product pages, then this is out of scope. BUT, I'm not seeing page view tracking for children collection pages.

<details><summary>Here, I stepped down from /Parts to a distant child in macbook parts and entered a product page. All it recorded was /Parts and the final product page I got to.</summary>

image


</details>

### Cart tracking
I'm not seeing any cart tracking. When I visited some product pages, I added the items to my cart and eventually viewed my cart. This includes the single add to cart button and the frequently bought together button.

<details><summary>What saw after following those steps (no cart events)</summary>

image

</details>

<details><summary>I would expect it to be something like this</summary>

image

</details>

Also, I'm not 100% sure exactly what event triggers the abandoned cart record (maybe when you add something to the cart and don't go to checkout?), but it's not happening on the product page on this branch.

<details><summary>How it's recorded</summary>

image

</details>

### Questions/Details
(The visit screenshotted above https://matomo.ubreakit.com/index.php?date=today&module=Widgetize&action=iframe&visitorId=fcc404a2d83c27ce&idSite=1&period=day&moduleToWidgetize=Live&actionToWidgetize=getVisitorProfilePopup)


> Tracks product page views in Matomo and Google Analytics.

Do you know if this guide is still up to date for testing GA? https://overlord.ifixit.com/Wiki/Google_Analytics_Testing
Are we still using it?

> Tracks ecommerce product views in Matomo.

Does the "ecommerce" here mean that these events should show up under the ecommerce section in Matomo? 
They aren't. All the events I showed (and the visit I linked) are only showing up in the visitor log.

Is the spec here to get parity with current product page tracking in Matomo?

masonmcelvain:
Thanks for the QA @erinemay!

> iFixit is the event title for everything

I think that's because the page `<title>` is "iFixit" on the product pages. So this is expected. If we change the page titles in the future, the Matomo events will be automatically updated.

image

> children collection pages

https://github.com/iFixit/react-commerce/pull/802/commits/5ccf63fb914f9a645120cde199a37f05c1027b00 Looks like a matomo variable was undefined after clicking next links. I think this commit fixes it.

> cart tracking

I refactored the code to do cart tracking every time we modify the cart, which seems like a better solution than tracking it when the user views their cart. Hopefully this gets cart tracking working https://github.com/iFixit/react-commerce/pull/802/commits/a2604e7df052cc95ec8d543ffbc5db792ff3161f

### Answers

> Do you know if this guide is still up to date for testing GA?

It might be out of date. Here's a recent pull where we QA'd it. We are still using GA. https://github.com/iFixit/react-commerce/pull/594

> Does the "ecommerce" here mean that these events should show up under the ecommerce section in Matomo?

Hm based on scrolling through some recent matomo data, it appears that most (if not all) events in the ecommerce log interact with the cart in some way. Maybe when the user interacts with the cart, the entire sequence of events shows up in the ecommerce log?

> Is the spec here to get parity with current product page tracking in Matomo?

Yep that's the goal.

un_dev_block ✌🏻

sterlinghirsh (approved):
CR :zap: QA :zap: carryover. The merge was fairly simple.

masonmcelvain:
Yea I think it was a mistake, which led to https://github.com/iFixit/react-commerce/issues/855

masonmcelvain:
I appreciate you spearheading this effort! 😄 

> I don't see it ever properly working

We've never had great spec for what these things should look like in Matomo/GA dashboards. If you know the expected behavior, it would be really helpful to include that in future issues (ideally with screenshots of the analytics UI) and then I'm sure we can get it right. Matomo has historically taken indeterminate amounts of time to process events in the past (10 min+), which further complicates QA.

PR #812, opened 2022-10-11
Handing enable (enabled2) metafield

### LINKED ISSUE #784: Support variant enabled metafield
This is similar to disable_when_oos, but we should treat the variant as enabled only when the metafield is true, regardless of inventory.

federicobadini:
closes #784 

This PR adds handling of the `enabled` metafield (formally `enabled2`)
We assumed all active variant have the `enabled2` metafield set to true by default and that only disabled variants have falsey values in it.

### QA

1. Visit the Vercel preview
2. Try to disable explicitly some product variants via the `enabled2` metafield
3. Verify that the interface is working as expected locking or even hiding disabled product options if no active variants include them at all

### DIFF
--- a/frontend/lib/shopify-storefront-sdk/generated/sdk.ts
+++ b/frontend/lib/shopify-storefront-sdk/generated/sdk.ts
@@ -6435,6 +6435,7 @@ export type FindProductQuery = {
             warning?: Maybe<{ __typename?: 'Metafield'; value: string }>;
             specifications?: Maybe<{ __typename?: 'Metafield'; value: string }>;
             warranty?: Maybe<{ __typename?: 'Metafield'; value: string }>;
+            enabled?: Maybe<{ __typename?: 'Metafield'; value: string }>;
             disableWhenOOS?: Maybe<{ __typename?: 'Metafield'; value: string }>;
             crossSell?: Maybe<{
                __typename?: 'Metafield';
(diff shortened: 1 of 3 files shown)


sterlinghirsh (approved):
CR :zap:

erinemay:
QA 👍 After setting a variant's `enabled2` metafield to false that variant is no longer available.

deploy_block ❔ I'm pretty sure this is expected because it means that we don't want the product to be available on the site, but when all of the variants of a product are disabled (`enabled2`=false), the product page 404s?

When I set `enabled2` to false on a single variant product (below), it 404s on this branch. 
- product page: https://react-commerce-git-respect-enable-metafield-ifixit.vercel.app/Products/ifixit-brand-sticker-collection
- shopify admin: https://ifixit-test.myshopify.com/admin/products/6556186017882/variants/39333701714010

federicobadini:
Yes, it is by design.

According to what we discussed, the `enabled2` metafield is there to explicitly deactivate content that had become obsolete / irrelevant even for SEO reasons. 
So as we remove disabled variants, we also end up 404ing a page with no active variant.
This was also needed because the product page is expecting to find at least an active variant inside a product.

If we wanna deviate from this we would need to refactor a bit the page.
But given that disabled variants were simply removed, it seemed pretty straightforward to 404 a page with no active variant.

erinemay:
un_deploy_block 👍

PR #814, opened 2022-10-11
Footer: minor styling updates

hcutrone:
We have some more minor changes to be made to the footer for iFixit/ifixit.

First, we added a margin on top of the footer to separate it from the content on the page. We want the KPI banner to be flush with the footer, however. So let's add a negative margin to remove the gap.

Second, we had some pages with a different default text color. To make sure we have the same color text on every page, let's be explicit about the colors.

qa_req 0


jyee27 (approved):
CR ☕

Issue #818, opened 2022-10-12
Make RefinementSingleSelect and some other links not a NextLink


sterlinghirsh:
## Problem
When loading https://www.ifixit.com/Parts/iPhone_12 it looks like we're prefetching things that shouldn't be prefetched because of using NextLink. These links are meant to be handled with Algolia, so we shouldn't be prefetching them. We still want to include the url in the href for SEO reasons.

![image](https://user-images.githubusercontent.com/589425/195424986-547818ea-99f8-4abc-8ad8-23e219624203.png)

I've also noticed that some links in the header are attempting to prefetch.

![image](https://user-images.githubusercontent.com/589425/195427213-4e56c33b-d6fb-4e69-afc7-cabd140089df.png)

## Do This
Make these normal links so they don't waste a bunch of round trips prefetching.

Note: It's possible that this will fix some bugs we've had with algolia's history control conflicting with next's history control, so we might be able to get rid of the hack we have to fix the back button.

dhmacs:
We should be careful about changing url without Next being aware of it.. it will likely cause the page props to go out of sync with the page url (e.g. page title, description, sections, anything that is fetched server side). 
I'd say that unless we're only changing search params, it's a good idea to let Next.js manage url.

danielbeardsley:
> I'd say that unless we're only changing search params, it's a good idea to let Next.js manage url.

I'm not sure here, but I think we currently intercept the clicks on these links and algolia updates the url.

Also, the point of this issue is that we shouldn't be pre-fetching 10 other next.js pages that aren't actually needed. Note, he also mentioned that non-next.js pages are also being pre-fetched.

sterlinghirsh:
I am in favor of being careful, but it's pretty easy to try it out and then see if it breaks. It seems like it should work to let Algolia change the url and then have next resume control when we switch. But also if we link to 10 products from a collection using NextLink, is it going to preload all those? Seems like it could be a lot of stuff to load, but I guess we'll have to experiment. I don't know how practical it is, but if we run into a conflict with Algolia and Next, maybe it's possible that the first time on a page Algolia starts changing the url, we opt out of the NextLink optimization and handle any links as full pageloads. I'd rather not if it's not necessary, but let's experiment with it.

dhmacs:
@sterlinghirsh 

[...]

PR #826, opened 2022-10-13
Stop using middleware

### LINKED ISSUE #820: No-indexing Middleware: Put it back in code
### Facts

* Currently our tiny no-index middleware is being run as an `Edge Function` in vercel. This means that code is run in a different physical location from the main app
   - https://github.com/iFixit/react-commerce/blob/c7692458348abb34f589faaf59e4abb565ee66ed/frontend/middleware.ts#L5-L11
   - This code also runs on every cache hit, slowing them down
* Right when this change was deployed, the number of requests to our app jumped up 5x
  - Don't know why here
* Right when this change was deployed several errors about getting 503 from Strapi became 20x more frequent

### Hypothesis
????

### Do This

Implement the above functionality without using middleware because it [doesn't appear](https://github.com/vercel/next.js/discussions/38989) you can have vercel run your middleware in your normal app.

CC @sterlinghirsh @masonmcelvain

masonmcelvain:
Middleware is run in a different location than our app. It runs on every cache hit, which is slow. We also noticed that the number of requests to our app spiked 5x and Strapi 503s increased by 20x

We're not sure why these things are happening, but for now let's do this in code instead.

## QA

Make sure the previews have the header `X-Robots-Tag: noindex, nofollow`

## On deploy

Make sure www.ifixit.com/Parts doesn't have that header, but https://react-commerce-prod.vercel.app does.

CC @danielbeardsley

Closes #820


masonmcelvain:
QA ✌🏻 the previews have the header

danielbeardsley (approved):
CR :+1: 

:boom:

danielbeardsley:
> Did you experience slow response times due to middleware? 

I think he means comparatively slow. Vercel is saying invocations average in around 5ms. This is fast enough to not care from a users perspective. But the extra server process running, the more limited environment your code is running in, and the fact that the middleware runs on cache hits too is just extra complexity that we don't need. This code is equivalent in functionality, but involves way less infrastructure and special knowledge.

While this is the right task (set a header) for the idea of middleware, the fact that they run all middleware on a separate server process, on a different JS runtime, on a different continent is just way more complexity than is needed to run several lines of code.

dhmacs:
@danielbeardsley I agree, not necessary for this 👍 I was just trying to understand what issues we might have if for some reason we need to use them in the future

danielbeardsley:
FYI, using middleware *definitely* changed something. After disabling it (this pull), server invocations dropped back down 5x:

image

It sure seems like the middleware caused #818 which would explain the increase in number of requests.

danielbeardsley:
Turns out, it's a bug in next.js. Well, at least it's been reported:
https://github.com/vercel/next.js/issues/38845

PR #828, opened 2022-10-13
Enable product page in production

### LINKED ISSUE #770: Enable product page in production without linking to it
We want to start testing the product page in production by enabling it but without linking to it from product list pages

## Proposed solution

Right now we use the `PRODUCT_PAGE_ENABLED` flag to decide whether we should 404 or not. By removing this check the page will be visible in production, and the product lists still won't link to it.
Should we also return a no-index header to prevent search engine from indexing the page until we go live? 

cc @sterlinghirsh

dhmacs:
closes #770 

## QA

1. Visit [react-commerce-prod preview](https://react-commerce-prod-git-enable-product-page-in-pr-a1546e-ifixit.vercel.app/products/macbook-pro-14-2021-a2442-and-16-2021-a2485-antenna-cable-bracket)
2. Verify that the product page is visible
3. Verify that the product page is not indexable (`x-robots-tag` header set to `noindex, nofollow`
4. Go to product lists (e.g. [`/Parts`](https://react-commerce-prod-git-enable-product-page-in-pr-a1546e-ifixit.vercel.app/Parts)) and verify that products links still point to the legacy us product page

### DIFF
--- a/frontend/helpers/next-helpers.ts
+++ b/frontend/helpers/next-helpers.ts
@@ -4,7 +4,7 @@ import { setSentryPageContext } from '@ifixit/sentry';
 import * as Sentry from '@sentry/nextjs';
 import { PROD_USER_AGENT } from '@config/constants';
 
-export function serverSidePropsWrapper<T>(
+export function serverSidePropsWrapper<T extends { [key: string]: any }>(
    getServerSidePropsInternal: GetServerSideProps<T>
 ): GetServerSideProps<T> {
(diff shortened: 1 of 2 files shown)


sterlinghirsh (approved):
CR :zap:

sterlinghirsh:
un_deploy_block :zap: I think it's enough. We'd like to add a robots.txt to make it so google doesn't crawl these domains at all in the future, but I think this is fine for now.

PR #832, opened 2022-10-14
Rename Store to prevent 404 on live site

sterlinghirsh:
We have gotten 404s on the live site when clicking links from the react-commerce header since we pushed the Store page because of NextLink. This renames it to Store2.


dhmacs:
CR ⚡

dhmacs (approved):

PR #838, opened 2022-10-18
Product option switch improvements

### LINKED ISSUE #788: Product Page: Improve UX of trying to change variants when some are OOS
When selecting variants, the `Condition` option shows all the possible (in stock or (oos and `disable_when_oos` = true)) options. Of those options, users can only click on options that have available items matching all the other selected options. 
https://react-commerce.vercel.app/Products/iphone-11-pro-screen
image

For example, if I want that `Used, B stock` item ⬆️, it's unclear why that item is not clickable. My first assumption would be that it's not available. After clicking the Part option, there's no indication that the `Used, B stock` is suddenly available. 

I would expect to be able to select that `Used, B stock` option and have the appropriate variant availability load ⬇️ 
image

I would expect this for all the options available. The first option I choose is probably the most important to me, and I would want to see all the available combinations for that specific option. 

### Note:
This experience of choosing variants was improved by https://github.com/iFixit/react-commerce/pull/774, so I'm going off of that branch and not main.

federicobadini:
closes #788 

Allow to click on any variant option value. 
In case the chosen options combination has no corresponding variant, tweak it to match the nearest variant.

#### Note on how how the nearest variant is computed
In case there is no exact match for the currently selected options, we have to pick the nearest option set available.
In order to do so we have to compute a `distance-score`.

We take into account only variants that include the selection just performed by the user.
Among those variants we check how the other options changes wrt the previously selected values.

The first variant is generally considered the most important, the last is the least important.
So if we have to tweak the first variant value in order to find an existing variant we will add 3 to the distance score, 2 if we need to tweak the second variant and just 1 if we have to tweak the third and last variant.

The variant with the least distance score will be chosen

### QA

1. Visit the Vercel preview
2. Find a product that has some non-existing variant combination in its option matrix
3. Check that, whenever we try to chose a non-existing combination of options, the app tweak the selection to the nearest existing variant

### DIFF
--- a/frontend/templates/product/sections/ProductSection/ProductOptions.tsx
+++ b/frontend/templates/product/sections/ProductSection/ProductOptions.tsx
@@ -14,6 +14,7 @@ import { faImageSlash } from '@fortawesome/pro-duotone-svg-icons';
 import type { Product } from '@models/product';
 import * as React from 'react';
 import { FaIcon } from '@ifixit/icons';
+import { invariant } from '@ifixit/helpers';
 
 export type ProductOptionsProps = {
    product: Product;
@@ -59,16 +60,13 @@ export function ProductOptions({
                         }}
                      >
                         {option.values.map((value) => {
-                           const variant = findVariant(product.variants, {
-                              ...selectedOptions,
-                              [option.name]: value,
-                           });
+                           const variant = findVariant(
+                              product,
+                              selectedOptions,
+                              { [option.name]: value }
+                           );
                            return (
-                              <option
-                                 value={variant?.id}
-                                 key={value}
-                                 disabled={variant == null}
-                              >
+                              <option value={variant?.id} key={value}>
                                  {value}
                               </option>
                            );
@@ -78,14 +76,14 @@ export function ProductOptions({
                   {selectorType === SelectorType.IMAGE_RADIO && (
                      <SimpleGrid columns={2} spacing="2">
                         {option.values.map((value) => {
-                           const variant = findVariant(product.variants, {
-                              ...selectedOptions,
-                              [option.name]: value,
-                           });
+                           const variant = findVariant(
+                              product,
+                              selectedOptions,
+                              { [option.name]: value }
+                           );
                            return (
                               <ProductOptionValue
                                  key={value}
-                                 disabled={variant == null}
                                  isActive={variant?.id === selected}
                                  label={value}
                                  image={variant?.image}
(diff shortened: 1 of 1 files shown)


frontend/templates/product/sections/ProductSection/ProductOptions.tsx
```diff
@@ -113,19 +111,60 @@ enum SelectorType {
...
-type Variant = Product['variants'][0];
-
 type Option = Product['options'][0];
 
 function findVariant(
-   variants: Variant[],
-   selectedOptions: Record<string, string>
+   product: Product,
+   previousOptions: Record<string, string>,
+   optionsUpdate: Record<string, string>
```
masonmcelvain, line 119: Is it possible to ever update more than one option at a time? If not, I might prefer an `Option` to make the logic easier to follow.

```suggestion
   optionUpdate: Option
```

frontend/templates/product/sections/ProductSection/ProductOptions.tsx
```diff
@@ -113,19 +111,60 @@ enum SelectorType {
...
+   const productOptionsNames = options.map((option) => option.name);
+   const optionsUpdateNames = Object.keys(optionsUpdate);
+
+   const variantsMatchingUpdate = variants.filter((variant) => {
+      return optionsUpdateNames.every((name) => {
+         const correspondingVariantOption = variant.selectedOptions.find(
+            (option) => option.name === name
+         );
+         return correspondingVariantOption?.value === optionsUpdate[name];
       });
```
masonmcelvain, line 143: With only one option update, this could be simplified to

```suggestion
      return optionUpdate.value === variant.selectedOptions.find(
         (option) => option.name === optionUpdate.name
      )?.value;
```

masonmcelvain (approved):
CR ✌🏻 but deploy_block ✌🏻 on potentially simplifying the logic based on the use case.

frontend/templates/product/sections/ProductSection/ProductOptions.tsx
```diff
@@ -113,19 +111,60 @@ enum SelectorType {
...
+         for (let i = 0; i < productOptionsNames.length; i++) {
+            const name = productOptionsNames[i];
+            const variantOption = variant.selectedOptions.find(
+               (option) => option.name === name
+            );
+            if (
+               !optionsUpdateNames.includes(name) &&
+               previousOptions[name] !== variantOption?.value
+            ) {
+               distance += 3 - i;
```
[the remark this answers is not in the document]
federicobadini: Yes, at most 3 variants per product
> You can create up to 100 variants for a product. Each product can have up to 3 options.

But anyway if you feel it more readable with the `productOptionNames.length` or even introducing a const value `MAX_OPTIONS_PER_PRODUCT` I can switch to it

frontend/templates/product/sections/ProductSection/ProductOptions.tsx
federicobadini: At the moment no, it is not. 
No problem updating it

frontend/templates/product/sections/ProductSection/ProductOptions.tsx
federicobadini: See answer above

frontend/templates/product/sections/ProductSection/ProductOptions.tsx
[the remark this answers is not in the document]
masonmcelvain: Thanks for pointing me to the docs! I think using the literal `3` is fine then.

frontend/templates/product/sections/ProductSection/ProductOptions.tsx
masonmcelvain: Ah, I didn't realize `Option` was a different type than `{name: string, value: string}`. I like your idea to use `optionUpdate: {name: string, value: string}`.

masonmcelvain:
(@erinemay it sounds like we will make a change to the code, so I will dev_block ✌🏻 and we can QA after the change)

masonmcelvain (approved):
CR ✌🏻 Thanks for the change @federicobadini! un_dev_block ✌🏻 un_deploy_block ✌🏻

erinemay:
This is great! It's no longer misleading that an available product is unavailable. I'm never able to reach an unavailable combination. QA 👍 

I'm gonna open a follow up issue so as not to bog down this one to get it a little clearer that unavailable options are unavailable.

PR #839, opened 2022-10-18
Improve product page gallery alt tags

### LINKED ISSUE #823: Product Page Feature Image Alt Text
On this page, https://react-commerce.vercel.app/Products/iphone-11-screen, and I assume all other products, the alt text for the main feature image of the product is the item # (alt="IF426-000-2" in this case). Compared to our current product page where the alt text of the main feature image of the product is essentially the product variant name (for https://www.ifixit.com/Store/iPhone/iPhone-11-Screen/IF426-000?o=2, it's "alt="iPhone 11 Screen / Fix Kit / New"") - I actually don't know exactly what we're using but it's pretty close to the product variant name, I'm sure a Dev can see easily what we're using. 

Can we make the alt text for our main feature image of the product something better than the item #? It seems like sticking with what we do currently is fine for now. 

This is also relevant for any product image. It seems like we're using item # for any/all images in the gallery. We should at least just do what we're doing currently. We can come up with even better templates in v2.1 or later.

dhmacs:
closes #823 

## QA

1. Visit [Vercel preview](https://react-commerce-git-fix-product-page-gallery-alt-texts-ifixit.vercel.app/products/galaxy-a51-screen)
2. Verify that product gallery images have alt tags set to the product title, or product title + product variant title

### DIFF
--- a/frontend/templates/product/sections/ProductSection/ProductOptions.tsx
+++ b/frontend/templates/product/sections/ProductSection/ProductOptions.tsx
@@ -210,14 +210,7 @@ function ProductOptionImage({ image }: ProductOptionImageProps) {
          : null;
 
    if (ratio == null) {
-      return (
-         <Img
-            h="16"
-            src={image.url}
-            alt={image.altText ?? ''}
-            objectFit="contain"
-         />
-      );
+      return <Img h="16" src={image.url} alt="" objectFit="contain" />;
    }
    return (
       <Box h="16" position="relative" overflow="hidden" mb="1">
@@ -227,7 +220,7 @@ function ProductOptionImage({ image }: ProductOptionImageProps) {
             maxW="full"
             mx="auto"
             src={image.url}
-            alt={image.altText ?? ''}
+            alt=""
             objectFit="contain"
          />
       </Box>
--- a/frontend/lib/shopify-storefront-sdk/generated/sdk.ts
+++ b/frontend/lib/shopify-storefront-sdk/generated/sdk.ts
@@ -6400,6 +6400,7 @@ export type FindProductQuery = {
          nodes: Array<{
             __typename?: 'ProductVariant';
             id: string;
+            title: string;
             sku?: Maybe<string>;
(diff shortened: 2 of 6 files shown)


frontend/templates/product/sections/ProductSection/ProductOptions.tsx
```diff
@@ -210,14 +210,7 @@ function ProductOptionImage({ image }: ProductOptionImageProps) {
...
    if (ratio == null) {
-      return (
-         <Img
-            h="16"
-            src={image.url}
-            alt={image.altText ?? ''}
-            objectFit="contain"
-         />
-      );
+      return <Img h="16" src={image.url} alt="" objectFit="contain" />;
```
masonmcelvain, line 213: I'm curious why the `alt` is always an empty string now?

masonmcelvain (approved):
CR ✌🏻 Looks great! deploy_block ✌🏻 on a question to check my understanding

frontend/templates/product/sections/ProductSection/ProductOptions.tsx
dhmacs: This only applies to product options, and I think that it's better for the screen reader to focus on the label (e.g. "Part Only"), rather than on the image.. Setting `alt` to `""` is akin to say that the image is only for decorative purposes

Screenshot 2022-10-18 at 16 45 43

masonmcelvain:
un_deploy_block ✌🏻 Thanks for the explanation!

erinemay:
QA 👍 
- Product gallery images have alt tags set to the product title + product variant title and change appropriately when variant is changed (`/products/google-pixel-6-screen-genuine`)
   - `alt="Google Pixel 6 Screen - Genuine New Fix Kit"`
   -  `alt="Google Pixel 6 Screen - Genuine New Part Only"`
- Single item product  have alt tags set to the product title (`/products/fix-the-world-poster`)
   - `alt="Fix the World Poster"`

PR #840, opened 2022-10-18
Product page SEO fields

### LINKED ISSUE #824: Product Page: Add SEO fields
Add title, meta description, and canonical. There is a product metafield called `meta_title` to be used as the page's `<title>`. Product metafield `short_description` should be used as the meta description. The canonical url should be the url without any variant specifier.

federicobadini:
closes #824 

This PR adds the `title`, `description` and `canonical` meta tags described in the issue.

This PR however also focused on including `og:image` tags to the product page.

We decided to add multiple `og:image`s since some sharing previews (like the facebook one) allow to view a gallery of images. 

Regarding the order in which images are written onto the page it was given priority to generic images over variant specific ones. According to [the protocol](https://ogp.me/#array) this means putting the generic ones first in the markup.

The reason of this is that, according to the spec, the content will be indexed according to the canonical, so probably most of the share preview implementations will include the image associated to the canonical independently of the fact that the link with parameters represents a specific variant url.
So a generic image, if any, would be more accurate to represent the content.
> og:url - The canonical URL of your object that will be used as its permanent ID in the graph, e.g., "https://www.imdb.com/title/tt0117500/".

At first we also included many optional parameters like `og:image:width`, `og:image:height`, `og:image:alt` etc but it came out that the Next Head component is not taking particular care of preserving the tag order and this could have been problematic. So we reverted the code to include the `og:image` tags only.

I've attached a video that shows how tag order is modified by client side interactions: updated tags are concatenated to the bottom of the list, separating each one of them from their siblings

https://user-images.githubusercontent.com/7805759/196474910-d0989f0f-c277-4ef1-b048-c569e601a87e.mov

### QA

1. Visit Vercel preview
2. Check that all the described SEO tags are present on the page


frontend/templates/product/MetaTags.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+         {genericImages.length > 0 &&
+            genericImages.map((image) => (
+               <meta key={image.id} property="og:image" content={image.url} />
+            ))}
+
+         {selectedVariantImages.length > 0 &&
+            selectedVariantImages.map((image) => (
+               <meta key={image.id} property="og:image" content={image.url} />
+            ))}
```
masonmcelvain, line 63: > the Next Head component is not taking particular care of preserving the tag order

Is it possible, then, that the variant images could be ordered before the generic images? Is that a problem?

masonmcelvain (approved):
CR ✌🏻 Happy to merge as is, just had a question.

frontend/templates/product/MetaTags.tsx
danielbeardsley: Sure, likely and possible, but I don't want to consider that important unless we have evidence that it is.

federicobadini: In this particular case it should not happen though. 
The reason is that on first render I've never noticed any out of order content. The out of order content seems to be generated only by client side updates of existing tags which are enqueued to the existing content.

Client side updates are only induced by a change in the currently selected variant.
Given that the generic images are memoized and depends only on the product, `og:image` tags associated with them should never be updated and should always precede the ones associated with the selected variant

k0rvusk0r4x:
- Went to a product page and pulled up dev tools.
- Noted the tags in the description are in the head.

QA 😺

frontend/templates/product/MetaTags.tsx
```diff
@@ -0,0 +1,66 @@
...
+            </>
+         )}
+
+         <meta property="og:type" content="website" />
+
+         <link rel="canonical" href={canonicalUrl} />
+         <meta property="og:url" content={canonicalUrl} />
+
+         {genericImages.length > 0 &&
+            genericImages.map((image) => (
```
danielbeardsley, line 56: Does this somehow end up different than just `genericImages.map()`?

A `.map()` on an empty array will return an empty array and I think react just ends up rendering nothing. If I'm wrong here, I'd love to know and adjust my mental model.

masonmcelvain: I think you're right that `.map` will return an empty array

Issue #461, opened 2022-07-19
Show grayed out facet empty state when no facets are available


sterlinghirsh:
If your search ends up with no facets visible, it's a weird looking empty state, so let's at least show some facets grayed out.

jeffsnyder:
I think this looks worse: 
![image](https://user-images.githubusercontent.com/7053450/181072005-a977799b-88e4-4421-95ac-accc01b55359.png)

Maybe it's not so bad if we need to hide Tool Category and Worksin.

jeffsnyder:
Er, maybe that's what's being discussed in the pull. I can't follow this very well https://github.com/iFixit/react-commerce/pull/465

erinemay:
> I think this looks worse: ![image](https://user-images.githubusercontent.com/7053450/181072005-a977799b-88e4-4421-95ac-accc01b55359.png)
> 
> Maybe it's not so bad if we need to hide Tool Category and Worksin.

@jeffsnyder, I totally agree. I think they're separate issues (empty list vs the single item list). 

@sterlinghirsh @sumukhipandey Or is it all the same code and can be handled in the same PR? I made a new issue for this behavior. If it should all be handled in one PR, go ahead https://github.com/iFixit/react-commerce/issues/500

PR #845, opened 2022-10-18
Always have a facet box of some sort

danielcliu-ifixit:
## Overview
It was decided that the blank facet accordion is not what we want to display, so this implements a change so the `price_range` facet should always have some item returned so the accordion is not empty. The data sorted by the facet is not useful, but this looks better and isn't too confusing IMO.

## QA
Pages like `/Parts/Panasonic_PT-DX610` should now not have an empty accordion. Any product list should only have the price range as a facet choice if there is not multiple facet ranges/options for the returned products.

Closes: #461


sterlinghirsh (approved):
CR :zap: QA :zap: works perfectly. Thanks!

PR #846, opened 2022-10-18
Don't report react query retries to Sentry

### LINKED ISSUE #827: Error: fetch() HTTP error: 429 Unknown HTTP Status
React re-rendering is causing us to query the cart api so much that some users get 429s cause they hit our request rate limit!

Sentry Issue: [REACT-COMMERCE-62](https://sentry.io/organizations/ifixit/issues/3501047285/?referrer=github_integration)

```
Error: fetch() HTTP error: 429 Unknown HTTP Status
  at None (/_next/static/chunks/pages/_app-ddace1ac9a1ba5e5.js:131164:50)
```

masonmcelvain:
So I _think_ we were reporting Sentry exceptions even when we hadn't exhausted all available retries in react-query. This changes our api clients to only report to sentry when we've exhausted all retries.

## CR

By default, react query retries 3 times with a backoff of 1 sec, then 2 sec.

React query docs for [`QueryCache`](https://tanstack.com/query/v4/docs/reference/QueryCache#global-callbacks).

## QA

One way to QA this would be to modify the `store/user/cart` api on your subdomain on cominor to return a 400-level response.

Then, set the following env var in `frontend/.env.local`

```
NEXT_PUBLIC_IFIXIT_ORIGIN=https://your-subdomain.cominor.com
```

Then, running localdev, open up any page (like /Parts). Your cart will be fetched for the mini cart, and there should be an error in your console.

The error should also get reported to Sentry.

Closes #827


deltuh-vee:
dev_block ✊ 
I modified the api accordingly and got the 404 errors
image
but were we expecting this last error?
image
I also didn't see any Sentry errors for react-commerce-dev.

danielbeardsley:
CR :+1: 

I understand the code and see how it works. It's a little unclear if the error will have enough context and a stack trace, but QA can bear that out. @deltuh-vee ... maybe sentry wasn't configured in that environment?

masonmcelvain:
un_dev_block ✌🏻 QA ✌🏻 

Events now get reported to sentry _after_ we finish all available retries.

Example event where I forced the cart api to 429: https://sentry.io/share/issue/1ca4f3c24ec04cba93c0a5d8440facbc/

image

image

image

danielcliu-ifixit:
CR :v:

sterlinghirsh (approved):
I think @danielcliu-ifixit meant to approve since he left a CR so I'm doing it now.

PR #847, opened 2022-10-18
No-Index: Pro Only products

### LINKED ISSUE #825: Noindex when Pro Only tag is present
We're adding a `Pro Only` tag in https://github.com/iFixit/ifixit/issues/45079. When it's present, noindex the page.

danielbeardsley:
Yes, both branches do the same thing, but it'll make it easy to still just delete the "else" branch before we go live.

Note: I haven't updated the generated typescript file for the
FindProductQuery cause I couldn't get codegen to work. It just kept
failing with no useful errors. I was able to eventually update the generated shopify sdk graph schmea, only to find out that I didn't need to.

Please help @dhmacs or @federicobadini -- codegen is just failing with no useful errors:

image

Closes #825

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,8 +119,9 @@ import { LifetimeWarrantyIcon } from '@assets/svg';
 
 When you need to update the Shopify storefront GraphQL schema version, follow these steps:
 
-1. Update `NEXT_PUBLIC_SHOPIFY_STOREFRONT_VERSION` in `frontend/.env.development` and `frontend/.env.production`
-2. Run `pnpm codegen:download-shopify-storefront-schema`
+1. Install `graphqurl`: `npm install -g graphqurl@1.0.1`
+2. Update `NEXT_PUBLIC_SHOPIFY_STOREFRONT_VERSION` in `frontend/.env.development` and `frontend/.env.production`
+3. Run `pnpm codegen:download-shopify-storefront-schema`
 
 ### Troubleshooting
 
--- a/frontend/lib/shopify-storefront-sdk/generate-schema.js
+++ b/frontend/lib/shopify-storefront-sdk/generate-schema.js
@@ -14,30 +14,13 @@ if (shopName == null || shopName.length === 0) {
    throw new Error('CODEGEN_STOREFRONT_SHOP_NAME is not set');
 }
 
-(async () => {
-   try {
-      await spawn(
-         `gq https://${shopName}.myshopify.com/api/${version}/graphql.json --introspect -H 'X-Shopify-Storefront-Access-Token: ${apiKey}' -H 'Content-Type: application/json' > ${__dirname}/schema.graphql`
-      );
-   } catch (error) {
-      console.log('ERROR:', error);
+child.exec(
+   `gq https://${shopName}.myshopify.com/api/${version}/graphql.json --introspect -H 'X-Shopify-Storefront-Access-Token: ${apiKey}' -H 'Content-Type: application/json' > ${__dirname}/schema.graphql`,
+   {},
+   (error) => {
+      if (error) {
+         console.error('ERROR:', error);
+         process.quit(1);
+      }
    }
-})();
-
-function spawn(command) {
-   return new Promise((resolve, reject) => {
-      const commandProcess = child.spawn(command, undefined, {
-         shell: true,
-      });
-      commandProcess.on('close', (code) => {
-         if (code === 0) {
-            resolve();
-         } else {
-            reject(code);
-         }
-      });
-      commandProcess.on('error', (err) => {
-         reject(err);
-      });
-   });
-}
+);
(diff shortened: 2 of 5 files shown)


frontend/templates/product/index.tsx
```diff
@@ -86,6 +84,13 @@ export const getServerSideProps: GetServerSideProps<ProductTemplateProps> =
          },
          handle
       );
+      const proOnly = product?.tags.find((tag: string) => tag === "Pro Only");
+      if (proOnly) {
+         context.res.setHeader('X-Robots-Tag', 'noindex, nofollow');
```
sterlinghirsh, line 89: This doesn't make it very easy to QA. Maybe we can noindex, follow on proOnly but noindex nofollow otherwise.

sterlinghirsh:
dev_block :zap: Looks like you need to include the tags in the type for product. Also it's a comma-separated string, not an array. An array-based interface would be convenient though.

README.md
```diff
@@ -119,8 +119,9 @@ import { LifetimeWarrantyIcon } from '@assets/svg';
 
 When you need to update the Shopify storefront GraphQL schema version, follow these steps:
 
-1. Update `NEXT_PUBLIC_SHOPIFY_STOREFRONT_VERSION` in `frontend/.env.development` and `frontend/.env.production`
-2. Run `pnpm codegen:download-shopify-storefront-schema`
+1. Install `graphqurl`: `npm install -g graphqurl@1.0.1`
```
dhmacs, line 122: This shouldn't be needed, as `graphqurl` is already a project dependency

frontend/lib/shopify-storefront-sdk/generate-schema.js
```diff
@@ -14,30 +14,13 @@ if (shopName == null || shopName.length === 0) {
...
-   } catch (error) {
-      console.log('ERROR:', error);
+child.exec(
+   `gq https://${shopName}.myshopify.com/api/${version}/graphql.json --introspect -H 'X-Shopify-Storefront-Access-Token: ${apiKey}' -H 'Content-Type: application/json' > ${__dirname}/schema.graphql`,
+   {},
+   (error) => {
+      if (error) {
+         console.error('ERROR:', error);
+         process.quit(1);
+      }
```
dhmacs, line 24: What's the reasoning behind this refactor?

dhmacs (commented):
Schema generation is only relevant when we change the version of the Shopify Storefront graphql schema. It basically runs an introspection against the Shopify Storefront api to get the updated graphql schema used for types generation.
Whenever you add something to a query (like you're doing here) you shouldn't be doing anything special, if the dev server is running there is a file watcher that detects the changes and updates the operation generated types.

I've run your branch locally, and I can't reproduce the error.
Looking at the error screenshot more specifically, it seems that the generation is failing not due to the shopify sdk, but due to the strapi sdk. The introspection to get the strapi sdk graphql schema is done against the local version of Strapi, so Strapi must be running. 
Since if you run the `dev` command Strapi is supposed to be running, I'd look at errors in you local strapi instance. Are there errors when strapi boots up? Are you able to perform query using the graphql playground at `http://localhost:1337/graphql` ?

cc @danielbeardsley

sterlinghirsh:
Closing in favor of #851 since it has the types defined properly and doesn't have the other changes. I'm still curious if there's a reason for the refactor of the generate schema.

README.md
danielbeardsley: I hunted all over the codebase, then npm for a cli tool named `gq`. Took me a while to trace that to a package named `graphqurl`. Didn't realize it was already a dependency until you mentioned it.

Anyway, now that I realize this is a dependency, this docs change isn't needed.

frontend/lib/shopify-storefront-sdk/generate-schema.js
danielbeardsley: I was trying to get it to run and I was getting a failure with no output. I was trying to get to the root of why it was failing so I tried launching with a different node function thinking `spawn()` may not have been capturing the output correctly. In the end, I realized the output was being redirected at the end of the command and the failure message was in the generated file and my change had no affect. I left the refactor cause the resulting code was simpler.

That said, it doesn't change the result, so I'm ok dropping this change.

danielbeardsley:
> Looking at the error screenshot more specifically, it seems that the generation is failing not due to the shopify sdk, but due to the strapi sdk

OMG, I can't believe I missed that :facepalm: 

That said, I don't know why it still wasn't updating the types from the graphql query change.

Thanks for Re-opening this @sterlinghirsh!

I'm not sure what the lesson is here. I started at the bottom (finding and editing the graphql query), then typescript failed to validate, so I went searching for the typescript file to update, realized it was dynamically generated, codegen didn't work (still doesn't)

PR #848, opened 2022-10-18
Noindex when user agent does not match

sterlinghirsh:
We tried to noindex when user agent was not "Amazon CloudFront", but we accidentally inverted the condition. This re-inverts the condition.

qa_req 0


danielcliu-ifixit:
CR :v:

danielcliu-ifixit (approved):

Issue #843, opened 2022-10-18
Test adding product to cart from product page


sterlinghirsh:
This is not intended to be an api test. There should be separate api tests on ifixit/ifixit.

   - [ ] Click once on the `Add to cart` button
       - [ ] Open the cart and ensure the item quantity is 1.
   - [ ] Click multiple times (for example 5) on the `Add to cart` button
       - [ ] Open the cart and ensure the quantity is 5
   - [ ] Test increasing and decreasing the quantity of the item in the cart
   - [ ] Test removing items from the cart and re-adding them by clicking `Add to cart` on the product page.

PR #854, opened 2022-10-20
Add Cypress Cart Test To Product Page

danielcliu-ifixit:
## Overview

We want to add some tests to the repo to cover cart functionality. This adds several tests that make sure that several actions go as expected.

## QA
Should be covered by tests qa_req 0;

Closes: #843

### DIFF
--- a/frontend/templates/product/sections/ProductSection/AddToCart/index.tsx
+++ b/frontend/templates/product/sections/ProductSection/AddToCart/index.tsx
@@ -74,6 +74,7 @@ export function AddToCart({ product, selectedVariant }: AddToCartProps) {
                   isLoading={addToCart.isLoading}
                   disabled={inventory.remaining === 0}
                   onClick={handleAddToCart}
+                  data-testid="product-add-to-cart-button"
                >
                   Add to cart
                </Button>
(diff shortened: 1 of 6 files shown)


danielbeardsley:
CR :+1: 

But the CI failures are relevant.

danielbeardsley:
Very likely that it will never finish. Maybe there's an error with the cart api. Maybe there's a CORS problem and the request is denied. Any idea what the url looks like for these cypress tests?

danielcliu-ifixit:
@danielbeardsley i considered that but if its a cors problem i assume that the vercel previews wouldn't work either right? The dev one works and the prod one doesn't seem to be the one being tested against

danielbeardsley:
> i considered that but if its a cors problem i assume that the vercel previews wouldn't work either right?

CORS is about exempting certain domains from cross-origin request limitations. We have to explictly exempt certain origins: https://github.com/iFixit/ifixit/blob/master/iFixit/Auth/CORSMiddleware.php#L58-L72

Looks like the dev server is run at `http://localhost:3000` which should be trusted:
https://github.com/iFixit/react-commerce/blob/main/frontend/cypress.config.ts#L26

But it's hard to know. Do these pass locally?

danielcliu-ifixit:
@danielbeardsley yes they pass locally

danielcliu-ifixit:
un_deploy_block :v:

danielbeardsley (approved):
CR :+1: 

Good work!

PR #875, opened 2022-10-25
Reset cross sell section state on variant change

### LINKED ISSUE #860: Frequently Bought Together selection retained when returning to a product page
When you enter a new product page via a link on the page like the Frequently Bought Together and return to that page without prompting a full page refresh, the frequently bought together selections are retained.

Video:
Flipping back and forth between the portable anti-static mat and the magnetic mat: Each time I get to the new page, my previously selected frequently bought together options on that same page were still selected.

https://user-images.githubusercontent.com/72166715/197232413-b2c1ce6e-4c83-4f6e-8fd0-fe5950cf4608.mp4

dhmacs:
closes #860 

## QA

1. Open [Vercel preview](https://react-commerce-git-reset-cross-sell-section-state-86e57f-ifixit.vercel.app/products/iphone-se-2020-replacement-battery)
2. Verify that issue described by #860 is fixed

### DIFF
--- a/frontend/templates/product/index.tsx
+++ b/frontend/templates/product/index.tsx
@@ -64,6 +64,7 @@ export const ProductTemplate: NextPageWithLayout<ProductTemplateProps> = () => {
             <ReplacementGuidesSection product={product} />
             <ServiceValuePropositionSection />
             <CrossSellSection
+               key={selectedVariant.id}
                product={product}
                selectedVariant={selectedVariant}
             />


sterlinghirsh (approved):
CR :zap: QA :zap: Glad it was that simple.

danielbeardsley:
As sterling and I discussed, this is a symptom of a larger issue in that our product page top-level component doesn't have `key=X` and thus when click a next link (and the props are switched out) react sees it as the same page with different props and thus preserves state where structure ends up the same.

I think we could use a `key={productcode}` or `key={handle}` fairly high up in some component that wraps the page.

dhmacs:
@danielbeardsley I agree with you, but I decided to postpone that decision because we might end up having some sections in which preserving state when the variant changes might be desirable, so for now I've only applied this change to the section that requires it.

PR #879, opened 2022-10-25
Proposal active image always first

### LINKED ISSUE #861: Add scroll buttons to image thumbnail list
Without scroll indicators, it's unclear that there are more thumbnail images available to click on.

This product has more images than can fit in the row of thumbnails.
https://react-commerce.vercel.app/products/Precision-Tweezers-Set

image

federicobadini:
closes #861 

As proposed in [comment](https://github.com/ifixit/react-commerce/issues/861#issuecomment-1289278944) this PR helps the user in understanding that there are more thumbnails available.

### QA

1. Visit [Vercel preview](https://react-commerce-git-improvements-on-product-galler-0c72c4-ifixit.vercel.app/products/fix-the-world-t-shirt)
2. Verify that the currently active image is always positioned as the leftmost visible in the thumbnails carousel
3. Only exception is when the carousel cannot scroll anymore to the right

### DIFF
--- a/frontend/templates/product/hooks/useSwiper.ts
+++ b/frontend/templates/product/hooks/useSwiper.ts
@@ -22,7 +22,7 @@ export function useSwiper({
    const [mainSwiper, setMainSwiper] = React.useState<Swiper | null>(null);
    const [thumbsSwiper, setThumbsSwiper] = React.useState<Swiper | null>(null);
    const slideChangeCallbackRef = React.useRef<(swiper: Swiper) => void>();
-   const slideChangeTransitionStartCallbackRef =
+   const slideNextTransitionStartCallbackRef =
       React.useRef<(swiper: Swiper) => void>();
 
    const [{ realIndex, isBeginning, isEnd }, setNavigationConfig] =
@@ -38,7 +38,7 @@ export function useSwiper({
          setNavigationConfig({ realIndex, isBeginning, isEnd });
          onSlideChange?.(swiper.realIndex);
       };
-      const slideChangeTransitionStartCallback = (swiper: Swiper) => {
+      const slideNextTransitionStartCallback = (swiper: Swiper) => {
          const { realIndex } = swiper;
          if (thumbsSwiper && realIndex > thumbsSwiper.realIndex) {
             thumbsSwiper?.slideTo(realIndex);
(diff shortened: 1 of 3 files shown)


sterlinghirsh (approved):
CR :zap: This is an improvement for sure. I might want to come back later and use a transparent gradient to indicate when there are more images, or start scrolling the container when the active image is in the middle, but I'm okay with this for now.

deltuh-vee:
dev_block ✊ 
This seems to obscure that there's images to the left even more than before. Looking at the Precision Tweezers Set, 
image
you can't tell there's images to the left unless you notice the dot indicators.

sterlinghirsh:
Good catch @deltuh-vee. I didn't notice that we weren't starting with the first image. I think the behavior of keeping the selected image's thumbnail on the left is fine, but we'd like to start on the first image, even if it's a product-level instead of variant-level image. Starting in the middle of the list is too confusing.

federicobadini:
Thumbnails are now rendered with a gradient overlay on the sides where more images could be revealed by scrolling.
The currently selected image is always kept on the left but outside of the gradient overlay.
So it is initially positioned on the very left and then is it kept as second leftmost image.

deltuh-vee:
QA 🎬 
un_dev_block 👍 
* Selected images start at the left most image
* Gradients have been added to suggest there are more images

federicobadini:
I've noticed a slight regression and I've modified the code in order to continue keeping mobile and the desktop carousel in sync.

jordycosta:
Screenshot 2022-11-04 090546

I noticed after scrolling all the way to the right, the last image still appears that there are more images beyond that since the gradient is still present

last

Same goes for scrolling back to the first image in the carousel

first

federicobadini:
> I noticed after scrolling all the way to the right, the last image still appears that there are more images beyond that since the gradient is still present

Is this a drag without a select or an arrow click here right?

deltuh-vee:
> Is this a drag without a select or an arrow click here right?

Yes. Clicking in the middle will also result in double gradients after dragging.

deltuh-vee:
dev_block ✊ on resolving merge conflicts.

jordycosta:
QA 🟢 

Active images start at the left 👍 
The scroll gradient issue in https://github.com/iFixit/react-commerce/pull/879#issuecomment-1303815963 is fixed ✔️ 

firefox_9BblC3nPBC

federicobadini:
Do you see other details to refine here or we are ready to ship? 🙂

PR #880, opened 2022-10-25
Add a robots.txt file

### LINKED ISSUE #743: Search indexing: Add robots.txt
Attacking the concerns in #734 from a different angle, let's prevent search engines from even crawling our non-production pages.

### Do This
***Wait until a month or two after we merge #734***

Add a `robotos.txt` page that responds with plain text and just renders two lines:
```
User-agent: *
Disallow: /*
```

~~Maybe use a middleware.~~

Note: www.ifixit.com doesn't proxy this path to us, so this will only be accessible from non-production pages.

sumukhipandey:
This PR adds this file to the public folder and includes the 2 lines we want based on the issue.

Closes #743

### DIFF
--- a/frontend/public/robots.txt
+++ b/frontend/public/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Disallow: /*


jordycosta:
QA 🟢 

**robots.txt** is now present on the frontend 👍 
Contents are as shown in https://github.com/iFixit/react-commerce/issues/743

robots

danielbeardsley (approved):
The magic here is that in production, our main php app hides this url cause it serves it's own robots.txt and we don't have cloudfront forwarding requests to for this url.

CR :+1:

PR #885, opened 2022-10-25
Product Page: use high level key prop

danielbeardsley:
There's some state on product pages. When you click a next-link and go to a different product page, we shouldn't be preserving any of that state. Previously, we were and we ran into bugs like: https://github.com/iFixit/react-commerce/issues/860

This should avoid an entire class of bugs like that.

CC @sterlinghirsh @federicobadini


deltuh-vee:
QA 🎬 
Selected product state isn't carried over to other product pages.

Issue #859, opened 2022-10-21
Product review count in Featured Products inconsistent


erinemay:
<details><summary>In the Featured Products section, the PTT has 1361 reviews, and the Mako has 390</summary>
https://react-commerce.vercel.app/products/2-tb-ssd-hybrid-3-5-hard-drive

![image](https://user-images.githubusercontent.com/72166715/197233680-e6734e5d-c7ff-42ff-be51-111129dbbedb.png)

</details>

On the product page, the PTT has 1447, and the Mako has 420
https://www.cominor.com/Store/Tools/Pro-Tech-Toolkit/IF145-307#productReviews
![image](https://user-images.githubusercontent.com/72166715/197233779-0561290f-ef31-4f78-90f8-9b8b72100bd3.png)

https://react-commerce.vercel.app/products/mako-driver-kit-64-precision-bits#reviews
![image](https://user-images.githubusercontent.com/72166715/197233816-293895bc-0915-4ff3-b0b0-9f95bf6b6b06.png)

dhmacs:
Just for context: featured products get the review count from a Shopify metafield, while the product page review count comes from the iFixit api, so it's likely a data sync issue

dhmacs:
@masonmcelvain Is there anything that prevents us from keeping the data in sync?

erinemay:
I don't think it's worth it to keep the pages looking synced just in dev. It would be nice if we could keep ifixit-test more up to date. Is this a similar issue to the issue here https://github.com/iFixit/react-commerce/pull/829#issue-1407932357?

dhmacs:
> Is this a similar issue to the issue here [#829 (comment)](https://github.com/iFixit/react-commerce/pull/829#issue-1407932357)?

Not exactly, in that context it made sense to use more up to date data from the ifixit api as the user made changes to his cart.
Here instead we would have to make more requests to the ifixit api. Besides this, I remember that @sterlinghirsh was thinking about putting reviews into metafield. If we do so, the problem described by this issue would disappear entirely

masonmcelvain:
> @masonmcelvain Is there anything that prevents us from keeping the data in sync?

Just so I understand correctly, the metafield we're concerned with is `product.metafields.reviews.rating_count`, correct?

I think we would have to repush all product data every time a user makes a product review in order to keep this metafield up to date. This is easy to do, but I don't think we are currently doing it.

> Besides this, I remember that @sterlinghirsh was thinking about putting reviews into metafield.

[...]

PR #909, opened 2022-10-26
Make Featured Product Review Data More Accurate

danielcliu-ifixit:
## Overview
For the featured product sections at the bottom of a product page, we weren't always getting the most up to date data in terms of reviews. This change fixes that by having the featured products make calls to the ifixit api for review data instead of using the review data from shopify (where possible).

## CR
I'm not an expert in async, so I'm unsure if this is the best solution. I originally tried to use `useProductReviews()` but the new data never seemed to get fetched/never updated the existing data anyways, so I just use the api fetch.

## QA

Make sure that the issues on `/products/2-tb-ssd-hybrid-3-5-hard-drive` mentioned in the issue are no longer there.

Closes: #859

### DIFF
--- a/frontend/templates/product/sections/FeaturedProductsSection/index.tsx
+++ b/frontend/templates/product/sections/FeaturedProductsSection/index.tsx
@@ -23,9 +23,15 @@ import {
    PageContentWrapper,
    ProductVariantPrice,
 } from '@ifixit/ui';
-import { Product } from '@models/product';
+import {
+   fetchProductReviews,
+   Product,
+   ProductReviewData,
+} from '@models/product';
 import { ImagePlaceholder } from '@templates/product/components/ImagePlaceholder';
 import NextLink from 'next/link';
+import { useAppContext } from '@ifixit/app';
+import React, { useState, useEffect } from 'react';
 
 export type FeaturedProductsSectionProps = {
    product: Product;
@@ -74,6 +80,7 @@ export function FeaturedProductsSection({
                      proPricesByTier={variant.proPricesByTier}
                      oemPartnership={variant.product.oemPartnership}
                      warranty={variant.warranty}
+                     sku={variant.sku}
                   />
                );
             })}
@@ -94,6 +101,7 @@ type ProductGridItemProps = {
    isPro?: boolean;
    warranty?: string | null;
    oemPartnership?: string | null;
+   sku?: string | null;
 };
 
 function ProductGridItem({
@@ -108,6 +116,7 @@ function ProductGridItem({
    isPro,
    warranty,
    oemPartnership,
+   sku,
 }: ProductGridItemProps) {
    const discountPercentage = computeDiscountPercentage(price, compareAtPrice);
    const hasLifetimeWarranty =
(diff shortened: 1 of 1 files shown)


jordycosta:
QA 🟢 

Products in the featured products section now have accurate review counts 👍 
[Product page tested](https://react-commerce-git-sync-review-count-on-feature-products-ifixit.vercel.app/products/2-tb-ssd-hybrid-3-5-hard-drive)

ratings

---

pro-tech

mako

frontend/templates/product/sections/FeaturedProductsSection/index.tsx
```diff
@@ -108,13 +116,43 @@ function ProductGridItem({
...
+         sku
+      );
+      return reviewsData;
+   };
+
+   const [reviewsData, initReviews] = useState<
+      ProductReviewData | null | undefined
+   >();
+   useEffect(() => {
+      fetchData()
```
masonmcelvain, line 143: This is a little briefer and obviates the need for the `fetchData` callback

```suggestion
      if (!sku) {
         return;
      }
      fetchProductReviews(appContext.ifixitOrigin, sku)
```

frontend/templates/product/sections/FeaturedProductsSection/index.tsx
```diff
@@ -108,13 +116,43 @@ function ProductGridItem({
...
+         return null;
+      }
+      const reviewsData = await fetchProductReviews(
+         appContext.ifixitOrigin,
+         sku
+      );
+      return reviewsData;
+   };
+
+   const [reviewsData, initReviews] = useState<
```
masonmcelvain, line 139: Generally speaking, I think it's [convention](https://reactjs.org/docs/hooks-state.html#declaring-a-state-variable) to name the state update function like `setNameOfStateVar`. I think it makes it a little clearer later on that you're updating React state, as opposed to invoking some other kind of callback.

So here, I would personally name this
```suggestion
   const [reviewsData, setReviewsData] = useState<
```

frontend/templates/product/sections/FeaturedProductsSection/index.tsx
```diff
@@ -23,9 +23,15 @@ import {
...
-import { Product } from '@models/product';
+import {
+   fetchProductReviews,
+   Product,
+   ProductReviewData,
+} from '@models/product';
 import { ImagePlaceholder } from '@templates/product/components/ImagePlaceholder';
 import NextLink from 'next/link';
+import { useAppContext } from '@ifixit/app';
+import React, { useState, useEffect } from 'react';
```
masonmcelvain, line 34: I think in this repo we've been following the pattern of
```suggestion
import * as React from 'react';
```

And then using the hooks like `React.useState` and `React.useEffect`. It probably doesn't matter, but just thought I'd share the pattern we've been using.

frontend/templates/product/sections/FeaturedProductsSection/index.tsx
```diff
@@ -108,13 +116,43 @@ function ProductGridItem({
...
+   >();
+   useEffect(() => {
+      fetchData()
+         .then((res) => {
+            initReviews(res);
+         })
+         .catch(() => {
+            initReviews(null);
+         });
+   }, []);
```
masonmcelvain, line 150: Generally, the dependency array should include any local vars/props that are used in `useEffect`, though if ESLint didn't complain maybe it's not a problem here.

masonmcelvain (changes requested):
CR ✌🏻 Nice job glad it works!

I might have missed the conversation, but did we consider re-pushing product data to Shopify whenever a product review is made (in iFixit/ifixit) as opposed to making more api requests on every product page view?

dev_block ✌🏻 on that and a few other questions

danielbeardsley:
> I might have missed the conversation, but did we consider re-pushing product data to Shopify whenever a product review is made (in iFixit/ifixit) as opposed to making more api requests on every product page view?

I really think this is preferable. It's like a 1 or 2-line change to our php app and would mean this whole pull isn't needed and we wouldn't need to make additional api requests.

danielcliu-ifixit:
Closing for now in favor of https://github.com/iFixit/ifixit/pull/45303 which should fix the issue by updating shopify whenever a review is made.

PR #922, opened 2022-10-27
User API: cache response in local storage

danielbeardsley:
Previously, there were some UI components that couldn't be drawn correctly until we heard back from the user API about the pro status of the given user. This meant either showing the wrong UI or hiding the UI until the response came back.

No more! This will still happen on the first page load, but after that, we store the user api data in localStorage (if available). We then use that value as the "placeholder data" for react-query. This is effectively used as the data *before* the user api response comes back.

99% of the time, this wil get us the right data on the first render pass. In that 1% of the time when the data is out of date (user has logged out or logged in just now), the cache will update as soon as the response comes back.

We also memoize the retrieval of the value from the storage so it's only done once per component (not once per render).

This fixes the flash of the non-pro price on first render for pro users.

Co-Authored by @sterlinghirsh

### DIFF
--- a/packages/auth-sdk/user.tsx
+++ b/packages/auth-sdk/user.tsx
@@ -1,5 +1,6 @@
 import { invariant, isRecord } from '@ifixit/helpers';
 import { useQuery } from 'react-query';
+import { useMemo } from 'react';
 import { useIFixitApiClient, IFixitAPIClient } from '@ifixit/ifixit-api-client';
 
 export type User = {
@@ -16,16 +17,46 @@ const userKeys = {
    user: ['user'],
 };
 
+const hasLocalStorage = typeof localStorage !== 'undefined';
+const userDataLocalKey = 'user.data';
+
 export function useAuthenticatedUser() {
    const apiClient = useIFixitApiClient();
+   const cachedUserData = useMemo(
+      () =>
+         !hasLocalStorage
+            ? null
+            : (() => {
+                 const storedString = localStorage.getItem(userDataLocalKey);
+                 return storedString === null ? null : JSON.parse(storedString);
+              })(),
+      []
+   );
    const query = useQuery(
       userKeys.user,
-      () => fetchAuthenticatedUser(apiClient).catch(() => null),
+      () => {
+         const responsePromise = fetchAuthenticatedUser(apiClient).catch(
+            () => null
+         );
+         if (hasLocalStorage) {
+            responsePromise
+               .then((response) => {
+                  localStorage.setItem(
+                     userDataLocalKey,
+                     JSON.stringify(response)
+                  );
+               })
+               .catch(() => {});
+         }
+         return responsePromise;
+      },
       {
          retryOnMount: false,
          staleTime: Infinity,
+         placeholderData: cachedUserData,
       }
    );
+
    return query;
 }
 


packages/auth-sdk/user.tsx
```diff
@@ -16,16 +17,46 @@ const userKeys = {
    user: ['user'],
 };
 
+const hasLocalStorage = typeof localStorage !== 'undefined';
+const userDataLocalKey = 'user.data';
+
 export function useAuthenticatedUser() {
    const apiClient = useIFixitApiClient();
+   const cachedUserData = useMemo(
```
masonmcelvain, line 25: React also offers a `useCallback` hook that might fit this case better

packages/auth-sdk/user.tsx
```diff
@@ -16,16 +17,46 @@ const userKeys = {
...
+            () => null
+         );
+         if (hasLocalStorage) {
+            responsePromise
+               .then((response) => {
+                  localStorage.setItem(
+                     userDataLocalKey,
+                     JSON.stringify(response)
+                  );
+               })
```
masonmcelvain, line 48: I'm not super familiar with our user api. If `response` is ever undefined or an empty string or something, I think the `JSON.parse` above will throw. But unless we see problems in practice it probably doesn't matter

packages/auth-sdk/user.tsx
```diff
@@ -16,16 +17,46 @@ const userKeys = {
...
+               .catch(() => {});
```
masonmcelvain, line 49: I'm curious why this extra catch block is here

masonmcelvain (dismissed):
CR ✌🏻 Great idea! I just have some questions

packages/auth-sdk/user.tsx
danielbeardsley: I think *callback* might be a bit disingenuous here... Though it seems like I can pass a function to `placeholderData` and that makes the code a bit cleaner. I'll do it.

https://react-query-v3.tanstack.com/guides/placeholder-query-data#placeholder-data-as-a-function

packages/auth-sdk/user.tsx
danielbeardsley: Calling methods on `localStorage` sometimes throws exceptions if they have some browser security feature enabled:
https://sentry.io/organizations/ifixit/issues/3505592338/?query=is%3Aunresolved+localstorage&referrer=issue-stream&statsPeriod=14d

And I don't want those reported to sentry.

masonmcelvain: Makes sense!

packages/auth-sdk/user.tsx
danielbeardsley: Looks like just undefined is a problem:
image

I don't think the above function can ever return *undefined* (because typescript), and if it does, we have the `catch()` below to turn that into null.

packages/auth-sdk/user.tsx
masonmcelvain: Oh gotcha, the use of `useMemo` in the docs is interesting. Seems like it would work either way

masonmcelvain (dismissed):
CR ✌🏻

masonmcelvain (approved):
CR ✌🏻 seems sane

jordycosta:
dev_block 🟥 

> This will still happen on the first page load, but after that, we store the user api data in localStorage (if available). We then use that value as the "placeholder data" for react-query. This is effectively used as the data before the user api response comes back.

On initial load, the brief flash is shown as described above

However, when I refresh the page I am still seeing the temporary flash of the non-pro user and the new section becomes buggy

firefox_xCDXTxYwAO

masonmcelvain:
@jordycosta is this as an anonymous user visiting a pro-only page? Hard to tell from the GIF

danielbeardsley:
> is this as an anonymous user visiting a pro-only page? Hard to tell from the GIF

I've found the problem. React doesn't like it when the server-rendered DOM doesn't match the DOM generated by the first render pass.

I may try using `useEffect()` here.

danielbeardsley:
Still working on this in some down-time.

danielbeardsley:
I think I forgot to un_dev_block :+1: this after fixing it :-(.

masonmcelvain (approved):
CR ✌🏻 noice

k0rvusk0r4x:
- The non-pro section is no longer flashing.
- The add to cart section is no longer buggy.
- Otherwise, things continue to look good.

QA 😺

Issue #923, opened 2022-10-27
Product Page buy box: spacing issues


davidrans:
Found a couple spacing issues:
![image](https://user-images.githubusercontent.com/806152/198363534-621dfb94-ce11-4934-93ef-c13ef7556667.png)

1. Need more space between "Only 2 left" message and Add to Cart button
2. The value props bullets look funny to me. I think they ought to be separated by the same amount of vertical space even if some are only 1 line

https://www.cominor.com/products/macbook-pro-14-2021-a2442-and-16-2021-a2485-antenna-cable-bracket

sterlinghirsh:
Part 2 wontfix

PR #933, opened 2022-10-27
Small Styling Changes for the Buy Box.

danielcliu-ifixit:
## Overview
Pictures in here are the examples from our figma mock up.
In this pull we address the spacing for the limited products. It should look like 

We also change the icons in the buy box to be solid colors again. Ignore the line dividing the two sections.
Screenshot from 2022-10-27 15-32-17

## QA
Make sure the buy box looks right and that the Value proposition at the bottom looks right too.
Screenshot from 2022-10-27 15-32-36
Note the text will be different in the mock ups than on the site and that is fine. Just make sure everything else visually hasn't broken.

Connects: #923

### DIFF
--- a/frontend/templates/product/sections/ProductSection/AddToCart/index.tsx
+++ b/frontend/templates/product/sections/ProductSection/AddToCart/index.tsx
@@ -255,7 +255,13 @@ function InventoryMessage({
    const remaining = Math.max(0, quantityAvailable - quantityAddedToCart);
 
    return (
-      <Flex color="red.600" py="0" fontSize="sm" align="center">
+      <Flex
+         color="red.600"
+         mt="2.5"
+         fontSize="sm"
+         align="center"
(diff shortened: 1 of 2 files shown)


masonmcelvain (approved):
CR ✌🏻

erinemay:
QA 👍 "Only # left" looks good on desktop, mobile, tablet (iphone ipad + safari and android + chrome)

The icons look better returned to the solid color icons.

> I think they ought to be separated by the same amount of vertical space even if some are only 1 line

deploy_block 🦈 @davidrans I know you want the spacing to be uniform, but I think that would look weirder if the icons don't align with the lines of text...
image

danielcliu-ifixit:
un_deploy_block :v: this only connects the issue i'm not worried about the spacing for those point rn, or at least we shoudl get fixes out where we can before we deploy

PR #934, opened 2022-10-27
Cypress: Make tests faster and less flaky

masonmcelvain:
I noticed the cart test was failing sometimes, and then rerunning it would cause it to pass.

Cypress's claim to fame is that it's less flaky than Selenium, so I used some best practices to hopefully make this test and others less flaky and faster.

## CR

The commits have documentation links in them.

qa_req 0 this only modifies tests

### DIFF
--- a/frontend/test/cypress/integration/product/add_to_cart.cy.ts
+++ b/frontend/test/cypress/integration/product/add_to_cart.cy.ts
@@ -3,68 +3,61 @@ describe('product page add to cart', () => {
       cy.loadProductPageByPath(
          '/products/spudger-retail-3-pack?variantid=39419992408154'
       );
-      cy.findByTestId('cart-drawer-open').click().wait(10);
-      cy
-         .findAllByTestId('cart-drawer-item-count')
+      cy.findByTestId('cart-drawer-open').click();
+      cy.findAllByTestId('cart-drawer-item-count')
          .should('exist')
          .invoke('text')
          .should('eq', '0');
       cy.findByTestId('cart-drawer-quantity').should('not.exist');
-      cy.findByTestId('cart-drawer-close').click().wait(10);
+      cy.findByTestId('cart-drawer-close').click();
       cy.findByTestId('cart-drawer-close').should('not.exist');
    });
 
    it('Clicking Add To Cart Adds Items To Cart', () => {
       const genArr = Array.from({ length: 5 }, (v, k) => k + 1);
       cy.wrap(genArr).each((index) => {
-         cy.findByTestId('product-add-to-cart-button').click().wait(10);
-         cy
-            .findAllByTestId('cart-drawer-quantity')
+         cy.findByTestId('product-add-to-cart-button').click();
+         cy.findAllByTestId('cart-drawer-quantity')
             .invoke('text')
             .then(parseInt)
             .should('eq', index);
-         cy.findByTestId('cart-drawer-close').click().wait(10);
+         cy.findByTestId('cart-drawer-close').click();
       });
 
-      cy.findByTestId('cart-drawer-open').click().wait(10);
+      cy.findByTestId('cart-drawer-open').click();
       cy.removeItemsFromCartAndCloseDrawer();
    });
 
    it('Clicking + and - Buttons Changes Item Quantity in Cart', () => {
-      cy.findByTestId('product-add-to-cart-button').click().wait(10);
-      cy
-         .findAllByTestId('cart-drawer-item-count')
+      cy.findByTestId('product-add-to-cart-button').click();
+      cy.findAllByTestId('cart-drawer-item-count')
          .invoke('text')
          .then(parseInt)
          .should('eq', 1);
       const genArr = Array.from({ length: 5 }, (v, k) => k + 2);
       cy.wrap(genArr).each((index) => {
-         cy.findByTestId('cart-drawer-increase-quantity').click().wait(10);
-         cy
-            .findAllByTestId('cart-drawer-quantity')
+         cy.findByTestId('cart-drawer-increase-quantity').click();
+         cy.findAllByTestId('cart-drawer-quantity')
             .invoke('text')
             .then(parseInt)
             .should('eq', index);
       });
-      cy.findByTestId('cart-drawer-close').click().wait(10);
+      cy.findByTestId('cart-drawer-close').click();
 
-      cy.findByTestId('cart-drawer-open').click().wait(10);
+      cy.findByTestId('cart-drawer-open').click();
       const genArrRev = Array.from({ length: 5 }, (v, k) => k + 1).reverse();
       cy.wrap(genArrRev).each((index) => {
-         cy.findByTestId('cart-drawer-decrease-quantity').click().wait(10);
-         cy
-            .findAllByTestId('cart-drawer-quantity')
+         cy.findByTestId('cart-drawer-decrease-quantity').click();
+         cy.findAllByTestId('cart-drawer-quantity')
             .invoke('text')
             .then(parseInt)
             .should('eq', index);
       });
-      cy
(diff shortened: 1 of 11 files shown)


masonmcelvain:
dev_block ✌🏻 wait there's more

masonmcelvain:
un_dev_block :v: ok now it's zippy

frontend/test/cypress/integration/product/add_to_cart.cy.ts
```diff
@@ -1,138 +1,136 @@
 describe('product page add to cart', () => {
-   const user = cy;
```
danielcliu-ifixit, line 2: Should we do this for the rest of the cypress tests too?

frontend/test/cypress/integration/product/add_to_cart.cy.ts
```diff
@@ -3,132 +3,111 @@ describe('product page add to cart', () => {
...
-      cy
-         .findAllByTestId('cart-drawer-item-count')
+      cy.findByTestId('cart-drawer-open').click();
+      cy.findAllByTestId('cart-drawer-item-count')
          .should('exist')
          .invoke('text')
          .should('eq', '0');
       cy.findByTestId('cart-drawer-quantity').should('not.exist');
-      cy.findByTestId('cart-drawer-close').click().wait(10);
+      cy.findByTestId('cart-drawer-close').click();
```
danielcliu-ifixit, line 9: I have my reservations over this change, mainly because these were implemented to cut down on flakiness that i found https://github.com/iFixit/react-commerce/pull/854/commits/243c9f73dbf1b0c107f680e3028bfd3a818d796a

But if something else fixed this than im glad we could reduce the gross `wait()`s everywhere

frontend/test/cypress/integration/product/add_to_cart.cy.ts
masonmcelvain: yea probably. I only did this one since the others didn't seem flaky at the moment

frontend/test/cypress/integration/product/add_to_cart.cy.ts
```diff
@@ -4,108 +4,65 @@ describe('product page add to cart', () => {
...
       cy.findByTestId('cart-drawer-quantity').should('not.exist');
       cy.findByTestId('cart-drawer-close').click();
       cy.findByTestId('cart-drawer-close').should('not.exist');
    });
 
    it('Clicking Add To Cart Adds Items To Cart', () => {
-      const genArr = Array.from({ length: 5 }, (v, k) => k + 1);
-      cy.wrap(genArr).each((index) => {
+      for (let i = 1; i <= 5; i++) {
          cy.findByTestId('product-add-to-cart-button').click();
```
danielcliu-ifixit, line 15: theres a reason to use `cy.wrap()` here allegedly 

https://stackoverflow.com/questions/52212868/cypress-io-writing-a-for-loop
>This is the only reliable way, at least if the order of code execution is important. It's worth noting that - due to the asynchronous nature of cypress - a classical for loop will likely result in a non-deterministic execution order of the cy.* calls inside the loop.

danielcliu-ifixit:
dev_block :v: on addressing the `cy.wrap()` usage

frontend/test/cypress/integration/product/add_to_cart.cy.ts
masonmcelvain: Ah I see, thanks for the link. I think this is working because the find...() function waits until each element is available, so when each iteration of the for loop is over, all assertions have been made. Maybe `.get()` doesn't wait

frontend/test/cypress/integration/product/add_to_cart.cy.ts
masonmcelvain: Hm yea I'm not sure why the behavior is different all of a sudden. Maybe we changed some animations at some point.

I'm not aware of a way to stress test in cypress. It would be nice to run these tests many times in parallel to try to flesh out any flakiness.

---

Found some docs about it https://docs.cypress.io/guides/guides/parallelization#Overview

> While parallel tests can also technically run on a single machine, we do not recommend it since this machine would require significant resources to run your tests efficiently.

frontend/test/cypress/integration/product/add_to_cart.cy.ts
danielcliu-ifixit: in either case, i think its probably better form to use the `wrap()`

masonmcelvain: I'm not sure there's a reason to block on this if the tests pass though? I might be misunderstanding.

danielcliu-ifixit: im not convinced that this change is necessary or desirable.

danielcliu-ifixit: i think why it still works is that each loop isn't dependent on the others. its like a map()

danielcliu-ifixit:
have you tested this locally? I've had some flakiness on this test. Im not sure if thats worth weighing heavily against CI but it would be nice to factor it in to some degree

frontend/test/cypress/integration/product/add_to_cart.cy.ts
masonmcelvain: Ok let me take one more shot at convincing you, and if you're not convinced I'll change it back.

> this change is necessary

This change is definitely not necessary since there's no functional difference. I think, however, that it's easier to tell what the bounds of the loop are. Another way that might be just as clear: `cy.wrap([1,2,3,4,5]).each()`. I think it's the `Array.from()` method that is kinda confusing (is it 0 indexed? does it include or exclude 5?).

> the asynchronous nature of cypress

Cypress (and javascript) are only async when we want them to be async. If we want actions to be done sequentially, we can tell cypress to do that. Here in the for-loop, all the actions are sequential. They will wait til the element is found in the DOM, or the test will fail because it couldn't find those elements, and it will abort execution. If someone later on introduces async code in the for loop, they will have to handle that appropriately.

---

So I guess I just find the `Array.from()` syntax hard to read, since it's so JS-specific. I'm open to other suggestions for iterating in a way that is easy to read.

masonmcelvain:
> have you tested this locally?

Yea I've rerun this several times locally and they passed each time. Like you said, it doesn't mean much because it's hard to tease out flakiness locally. I think the only way we'll know is by deploying this and seeing if the test is flaky across different PR's. It would be nice to get CI running the tests several times in parallel.

masonmcelvain:
I cleared session state between tests. Sometimes the cart from a previous test was used, which would cause the test to fail.

I think that happened because when we hit the cart api, it uses a cookie to identify the user (I think). If a test leaves items in the cart for some reason, and then a later test talks to the cart api expecting there to be 0 items in the cart, the later test will fail.

masonmcelvain:
un_dev_block ✌🏻 for more CR

danielcliu-ifixit (approved):

frontend/test/cypress/integration/product/add_to_cart.cy.ts
danielcliu-ifixit: Ill say that i think using syntax that is language-specific isn't really a substantial reason to shy away from it because its easy to look up and well documented, but at this point im not going to continue to hold this up over 2 lines of inconsequential syntax

danielcliu-ifixit:
CR :v:

frontend/test/cypress/integration/product/add_to_cart.cy.ts
masonmcelvain: I see your point. I think it's ok to disagree here, since functionally it's the same. I'm going to merge as is for expediency. But if you wanna change it back in the future, I'll give you CR on it.

Issue #931, opened 2022-10-27
Handle long lists of worksins


erinemay:
https://www.cominor.com/products/HP-Pavilion-G4-G6-and-G7-Fan
<details><summary>yikes</summary>

![image](https://user-images.githubusercontent.com/72166715/198405180-307ac519-c597-4493-8571-20c0bf12a8cb.png)

</details>

We're not handling it on the legacy pages, but we should still fix it here.
https://www.ifixit.com/Store/PC-Laptop/HP-Pavilion-G4-G6-and-G7-Fan/IF273-014?o=1

<details><summary>yikes, but legacy</summary>

![image](https://user-images.githubusercontent.com/72166715/198405369-b1eacb64-5093-40bc-8676-a28beac18639.png)

</details>

erinemay:
After https://github.com/iFixit/react-commerce/pull/935 was merged, a [ux issue was brought up by Jackie](https://ifixit.slack.com/archives/C0252FETV2S/p1667837002655909). (product: https://www.ifixit.com/products/galaxy-note10-plus-replacement-battery)

![image](https://user-images.githubusercontent.com/72166715/200360884-d176214c-f604-406a-9a9d-d1cca463a59d.png)

Users expect the "And 4 other devices..." to be an expandable element that shows the rest of the devices. It is not clear that this is a truncated version of the full list lower on the page. On products with more than 3 device compatibilities, we get the see all compatible devices link that goes to the full list, but 3 or fewer devices don't show that link.
![image](https://user-images.githubusercontent.com/72166715/200362568-3310b3fa-5215-461b-a5e5-d7f9b1ec5150.png)
 
Maybe the "And 4 other devices..." could link to `#compatibility` on the page?

cc: @danielcliu-ifixit, @sterlinghirsh

sterlinghirsh:
use the term models instead of devices to refer to variants

PR #935, opened 2022-10-28
Truncate Device List In Compatibility Accordion 

danielcliu-ifixit:
## Overview
There are pages where theres a million devices under a compatible product list. This pull truncates that list so its only 3, and makes the page cleaner.

## QA
Does the page `/products/HP-Pavilion-G4-G6-and-G7-Fan` compatibility accordion now truncate into a readable form?
Screenshot from 2022-10-27 17-09-53

Connects: #931

### DIFF
--- a/frontend/components/common/CompatibleDevice.tsx
+++ b/frontend/components/common/CompatibleDevice.tsx
@@ -3,9 +3,16 @@ import { Product } from '@models/product';
 
 export type CompatibleDeviceProps = {
    device: NonNullable<Product['compatibility']>['devices'][number];
+   truncate?: number;
 };
 
-export function CompatibleDevice({ device }: CompatibleDeviceProps) {
+export function CompatibleDevice({
+   device,
+   truncate = 0,
+}: CompatibleDeviceProps) {
+   const variants = truncate
+      ? device.variants.slice(0, truncate)
+      : device.variants;
    return (
       <>
          <Img
@@ -32,7 +39,7 @@ export function CompatibleDevice({ device }: CompatibleDeviceProps) {
                {device.deviceName}
             </Text>
             <Flex mb="2px" flexDir="column">
-               {device.variants.map((variant) => (
+               {variants.map((variant) => (
                   <Text
                      key={variant}
                      lineHeight="short"
--- a/frontend/components/common/CompatibleDevice.tsx
+++ b/frontend/components/common/CompatibleDevice.tsx
@@ -1,4 +1,5 @@
-import { Flex, Img, Text } from '@chakra-ui/react';
+import { Flex, Img, Link, Text } from '@chakra-ui/react';
+import NextLink from 'next/link';
 import { Product } from '@models/product';
 
 export type CompatibleDeviceProps = {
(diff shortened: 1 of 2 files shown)


dhmacs:
@danielcliu-ifixit can we add below the list ".and 23 more" or something like that to hint at the fact that we're truncating the list?

jordycosta:
QA 🟢 

The compatible devices shown in the drop-down are now limited to display 3 devices maximum

sidebar


deploy_block 🟨 on [Massimo's comment](https://github.com/iFixit/react-commerce/pull/935#issuecomment-1294827375)

sterlinghirsh:
The rule should be such that we don't display "show more" if "show more" is only replacing one item. I hate when someone shows me a truncated list with a show more button and it's like why didn't you take the space that was used by the show more button and show me the one remaining item without me having to click.

Something like `if (variants.length > truncate + 1 ) { remaining = variants.length - truncate; variants = variants.slice(0, truncate) } `

sterlinghirsh:
dev_block :zap: un_deploy_block :zap:

danielcliu-ifixit:
un_dev_block :v:

deltuh-vee:
dev_block ✊ 
It looks like we're showing "And 1 other devices..." for a bunch of devices that shouldn't have them in the Compatibility section.
This branch:
image
Master:
image

deltuh-vee:
QA 🎬 
The compatibility section is now truncated to be more readable.
un_dev_block 👍

frontend/components/common/CompatibleDevice.tsx
```diff
@@ -1,11 +1,22 @@
...
 };
 
-export function CompatibleDevice({ device }: CompatibleDeviceProps) {
+export function CompatibleDevice({
+   device,
+   truncate = 0,
+}: CompatibleDeviceProps) {
+   let variants = device.variants;
+   let unlistedVariants = 0;
+   if (truncate > 0 && variants.length > truncate) {
```
masonmcelvain, line 16: Based on Sterling's comment https://github.com/iFixit/react-commerce/pull/935#issuecomment-1295396340, I'd expect this to look like `variants.length > truncate + 1` to handle the case where we only have one extra variant. Maybe I'm missing something and it's being handled elsewhere.

masonmcelvain (approved):
CR ✌🏻 Nice! deploy_block ✌🏻 on a question about the expected behavior

frontend/components/common/CompatibleDevice.tsx
danielcliu-ifixit: im not quite sure what you mean, my solution is just maxed at `truncate` variants, whiles sterlings is `truncate` + 1 variants.  I felt like `truncate` being the cap made more intuitive sense.

masonmcelvain: Oh I see, `truncate` doesn't indicate how many variants to show _after_ truncation, it indicates the max number of variants to show before we start truncating.

masonmcelvain:
un_deploy_block ✌🏻

PR #939, opened 2022-10-28
Improve error reporting for Shopify GraphQL SDK

### LINKED ISSUE #914: Better error reporting on product api failures
## Problem
When checking the product page live, the page would 404. I thought this was from a setting, but in fact it was a misconfiguration on our live storefront api permissions. When fetching the product returned null, we would silently 404 the product page.

## Do This
If there's an error from the storefront api, log it in the console and sentry. I expect that missing permissions to query product data would return a relevant message. Might be nice to also display another type of error page like a 500 level to make it easier to distinguish between wrong product handles and backend errors.

dhmacs:
closes #914 

This PR makes GraphQL call fail instead of silently return a successful result when there are GraphQL errors.
It also appropriately logs the error, as shown here:

Screenshot 2022-10-28 at 17 21 17

## QA

I've set an invalid storefront token in Strapi:

Screenshot 2022-10-28 at 17 28 12

1. Open [Vercel preview](https://react-commerce-git-improve-product-page-error-reporting-ifixit.vercel.app/products/iphone-7-replacement-battery)
2. Verify that product page now returns 500 error

If you want to get rid of the error you just need to log into the branch Strapi instance and insert the same storefront access token used by https://main.govinor.com/admin

### DIFF
--- a/frontend/lib/shopify-storefront-sdk/index.ts
+++ b/frontend/lib/shopify-storefront-sdk/index.ts
@@ -1,6 +1,7 @@
 import { SHOPIFY_STOREFRONT_VERSION } from '@config/env';
 import { getSdk, Requester } from './generated/sdk';
 import * as Sentry from '@sentry/nextjs';
+import { z } from 'zod';
 export * from './generated/sdk';
 
 export type ShopCredentials = {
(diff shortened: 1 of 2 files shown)


deltuh-vee:
QA 🎬 
Failed GraphQL calls on a Product page now return a 500 error and show up in Sentry.

masonmcelvain (approved):
CR ✌🏻

Issue #950, opened 2022-10-28
Redirect Sitemap For Products


danielcliu-ifixit:
## Overview

Max has requested a vanity redirect for the product sitemap. So basically we'll redirect so I need https://www.ifixit.com/products/sitemap.xml to https://dozuki-sitemaps.s3.amazonaws.com/2/en/products.xml|

Similar to #488 

@ifixitmaxb

ifixitmaxb:
I get a 404...

danielcliu-ifixit:
i think it hadn't deployed. Confirmed working now

PR #952, opened 2022-10-28
Add sitemap redirect for products

danielcliu-ifixit:
## Overview

Max has requested a vanity redirect for the products.xml, so this implements a redirect to ifixit when you try to access it. 

## QA
Covered by test and also this is internal facing so im ok with qa_req 0;

Closes: #950


sterlinghirsh (approved):
CR :zap:

Issue #949, opened 2022-10-28
iFixit API: use `IFixitAPIClient` everywhere


danielbeardsley:
### Problem
We make quite a few requests with a bare `fetch()` call. This doesn't log to sentry and doesn't generate a log for the async operation

### Do This
Get an instance of the api client and use it:
```
const client = new IFixitAPIClient({origin: IFIXIT_ORIGIN});
```   

Or wrap it like we've done with the devices api if you want more type info or are going to be using this endpoint from more than one place:
https://github.com/iFixit/react-commerce/blob/2ca75fa98a9fd28a1698615ac949fd1e2b9313d0/frontend/lib/ifixit-api/devices.ts#L7-L21

dhmacs:
I started working on this and I have an idea. Instead of having to remember to use `sentryFetch` everywhere, and disseminate Sentry dependencies throughout our codebase, why don't we intercept `globals.fetch` and `window.fetch` instead?

It'll look like this:

```ts
const originalFetch = window.fetch;

window.fetch = async (input, init) => {
  try {
    const response = await originalFetch(input, init);
    // Handle exceptions
    return response;
  } catch (error) {
    // Handle network error logging
    throw error;
  }
}
```

And pretty much the same thing for `globals.fetch`.

Thoughts?

cc @danielbeardsley @masonmcelvain

dhmacs:
This would address also #948

masonmcelvain:
So instead of explicitly capturing the exception on non-200 responses (current implementation shown below), we would throw and let Sentry implicitly record the exception (in contexts where we have Sentry set up, i.e. `frontend/`?

https://github.com/iFixit/react-commerce/blob/0ea458b75a33fa44cac8d816aef0204dad19b761/packages/sentry/index.tsx#L19-L31

That seems like it would work, and it would be very nice not to have to add Sentry as a dependency in so many of our packages.

dhmacs:
No, the snippet that I wrote above is a general implementation for how you might intercept fetch and inject custom code. The custom code that will go in there  is the current `sentryFetch` implementation that you just referenced.

The only difference is that won't depend explicitly on `sentryFetch`, we would just use the standard `fetch` as we're used to and the sentry logging would be injected at runtime.

For example, instead of doing:

```ts
   const response = await sentryFetch(`${STRAPI_ORIGIN}/graphql`, {
      // etc etc
   });
```

we can just use `fetch` as we'd normally do:

```ts
[...]

PR #955, opened 2022-10-28
PPv2: add logging

### LINKED ISSUE #944: PPV2: Add logs for async operations
New product pages are timing out at an alarming rate and we have little-to-no logging

Do This
###
* Add logs for all async operations like we do for other pages (I think we have a `logAsync` function to wrap them)
* Remove the logging of headers
image
* Remove log of: `Ifixit Origin:  https://www.ifixit.com`

danielbeardsley:
Add more logs and drop some verbose ones. Individual commits have more context.

Closes #944 
Connect #949 -- uses it in one more place

### DIFF
--- a/frontend/models/product.ts
+++ b/frontend/models/product.ts
@@ -24,7 +24,7 @@ export type ProductImage = ReturnType<typeof getFormattedImages>[0];
 
 export async function findProduct(shop: ShopCredentials, handle: string) {
    const storefront = getShopifyStorefrontSdk(shop);
-   const ifixitApiClient = new IFixitAPIClient({ origin: IFIXIT_ORIGIN });
+   const iFixitApiClient = new IFixitAPIClient({ origin: IFIXIT_ORIGIN });
 
    const response = await logAsync('shopify.findProduct', () =>
       storefront.findProduct({
@@ -54,7 +54,7 @@ export async function findProduct(shop: ShopCredentials, handle: string) {
    }
    const iFixitProductId = computeIFixitProductId(variantSku);
    const reviewsData = await fetchProductReviews(
-      ifixitApiClient,
+      iFixitApiClient,
       iFixitProductId
    );
    let breadcrumbs = parseBreadcrumbsMetafieldValue(
--- a/frontend/models/product.ts
+++ b/frontend/models/product.ts
@@ -16,6 +16,7 @@ import {
 } from '@lib/shopify-storefront-sdk';
 import { z } from 'zod';
 import shuffle from 'lodash/shuffle';
+import { IFixitAPIClient } from '@ifixit/ifixit-api-client';
 
 export type Product = NonNullable<Awaited<ReturnType<typeof findProduct>>>;
 export type ProductVariant = ReturnType<typeof getVariants>[0];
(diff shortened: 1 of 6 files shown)


frontend/models/product.ts
```diff
@@ -425,29 +430,16 @@ export type ReviewAuthor = {
 };
 
 export async function fetchProductReviews(
-   apiOrigin: string,
+   apiClient: IFixitAPIClient,
    productId: string
 ): Promise<ProductReviewData | null> {
-   const response = await fetch(
+   const response = await apiClient.get(
```
sterlinghirsh, line 436: Why don't we just instantiate the apiclient here rather than passing it in?

frontend/models/product.ts
```diff
@@ -15,17 +16,21 @@ import {
...
 import shuffle from 'lodash/shuffle';
+import { IFixitAPIClient } from '@ifixit/ifixit-api-client';
 
 export type Product = NonNullable<Awaited<ReturnType<typeof findProduct>>>;
 export type ProductVariant = ReturnType<typeof getVariants>[0];
 export type ProductImage = ReturnType<typeof getFormattedImages>[0];
 
 export async function findProduct(shop: ShopCredentials, handle: string) {
    const storefront = getShopifyStorefrontSdk(shop);
+   const ifixitApiClient = new IFixitAPIClient({ origin: IFIXIT_ORIGIN });
```
sterlinghirsh, line 27: This makes 3 different capitalizations of iFixit in this file (IFixitAPIClient ifixitApiClient iFixitProductId)

frontend/models/product.ts
sterlinghirsh: I wonder why we weren't using the ifixit api client here before. Weird.

frontend/models/product.ts
danielbeardsley: Oh dear. Three worlds collide. :boom: 

I think two is probably unavoidable because types / variables, but three is a bit much.

frontend/models/product.ts
sterlinghirsh: Any idea what the 401 handling was about? I didn't think product reviews would return that.

danielbeardsley: Likely the author didn't know it existed. I made #949

Totally can, same result. I left it out to stick to the pattern (the shopify sdk is also created at the top level) and this encourages re-using the client if we make more than one api call later.

sterlinghirsh (dismissed):
I left some style comments but I'm okay with merging this quickly in the interest of tracking down the timeout issues we were seeing.

sterlinghirsh:
CR :zap:

frontend/models/product.ts
danielbeardsley: I think it was vestigal. Perhaps this code was copied from the user-api request code?

Definitely doesn't return 401:
https://github.com/iFixit/ifixit/blob/master/Guide/UI/api/2.0/reviews.php#L39-L53

frontend/models/product.ts
danielbeardsley: There's actually 4!  `IFIXIT_ORIGIN`

danielbeardsley:
QA :+1: 

image
image

sterlinghirsh (approved):

PR #957, opened 2022-10-31
Add-to-cart: don't fail when there are no images

danielbeardsley:
Had some products that were impossible to buy because of this.

I don't think there's an issue.


jarstelfox (approved):
CR 🌵

deltuh-vee:
QA 🎬 
I was able to add a product to the cart that had its images removed from ifixit-test.
image

Issue #930, opened 2022-10-27
International Buy Box Banner


sterlinghirsh:
We want to let people know when they're viewing a product page that is available in a store close to them. e.g. If a Canadian customer looks at a product, they should get a message suggesting they purchase from the Canadian store.

We use GeoIP to detect where the customer is, but you can force it on our legacy product page by adding a country=ca parameter to the product url, e.g. 
https://www.ifixit.com/Store/iPhone/iPhone-6s-Battery/IF314-011?o=10&country=ca

This is what it looks like:
![image](https://user-images.githubusercontent.com/589425/198401148-5059c4a1-1f92-4966-b836-0037c7b96d0a.png)

This replaces the Add to Cart button and quantity selector.

The buy from Canada Store link takes you to https://canada.ifixit.com/products/iphone-6s-replacement-battery (though it looks like we add a url param for tracking, so it ends up as https://canada.ifixit.com/products/iphone-6s-replacement-battery?pk_vid=ba9f6a8e9107d2f1166690605321f03f). The buy from USA Store simply hides this section and re-adds the Add to Cart section.

The GeoIP api is documented at https://www.ifixit.com/api/2.0/doc/Internal#returns-info-about-whether-a-product-is-available-in-a-non-us-store
The request is to https://www.ifixit.com/api/2.0/internal/international_store_promotion/buybox?productcode=314011 (or to force the country, https://www.ifixit.com/api/2.0/internal/international_store_promotion/buybox?productcode=314011&country=ca)

The api response looks like
```json
{
    "countryCode": "CA",
    "storeid": "ca",
    "storeName": "Canada",
    "storeUrl": "https:\/\/canada.ifixit.com\/",
    "salesChannel": "shopify.ca",
    "products": {
        "9": {
            "sku": "3140119",
            "optionid": 9,
            "productURL": "https:\/\/canada.ifixit.com\/products\/iphone-6s-replacement-battery",
            "price": "CA$34.99",
            "currency": "CAD",
            "avail": 1,
            "salesChannel": "shopify.ca",
            "gaSku": "314011-9"
        },
        "10": {
            "sku": "31401110",
            "optionid": 10,
            "productURL": "https:\/\/canada.ifixit.com\/products\/iphone-6s-replacement-battery",
            "price": "CA$42.99",
            "currency": "CAD",
            "avail": 1,
            "salesChannel": "shopify.ca",
            "gaSku": "314011-10"
        }
    }
}
```
[...]

PR #962, opened 2022-10-31
International Stores: add buy box when appropriate

danielbeardsley:
This should function the same as it did on the previous product page.

* Makes an API request to ifixit.com from the browser for the given
  product
* That responds with "user is geolocated in country X, here's info for a
  banner for each of the variants"
  * OR it responds with null
* We use a react component to render the "Buy from our X Store" banner
  for the currently selected variant
* Shows a "or buy from the US Store" link that will dismiss the buy box
  for 7 days
   * The previous product page used a 7-day cookie, so I kept it the
     same.

Note: The UI differs from the old product page bannerl slightly and we
can definitely tweak it, but I might do that in a follow-up pull unless
there are bugs.

Also, matomo events will be in a follow up pull.

QA:
* Set your preferred store via the store dropdown in the header
* Visit product pages on the US store for products that are available in your preferred store
* Assert you see the "buy box" instead of the Add To Cart button
* Assert that clicking the button takes you to the correct store
* Assert that clicking the bottom option "buy from US store" makes the buy box disappear and you shouldn't see it on further page reloads (should stay away for 7 days)

Connect #930

### DIFF
--- a/frontend/lib/ifixit-api/international-buy-box.ts
+++ b/frontend/lib/ifixit-api/international-buy-box.ts
@@ -1,0 +1,33 @@
+import { IFixitAPIClient } from '@ifixit/ifixit-api-client';
+import { invariant } from '@ifixit/helpers';
+
+export type BuyBoxVariant = {
+   sku: string;
+   optionid: number;
+   productURL: string;
+   price: string;
+   currency: string;
+   avail: number;
+   salesChannel: string;
+   gaSku: string;
+};
+
+export type BuyBoxResponse = {
+   countryCode: string;
+   storeid: string;
+   storeName: string;
+   storeUrl: string;
+   salesChannel: string;
+   // Keyed by optionid
+   products: Record<string, BuyBoxVariant>;
+};
+
+export function getBuyBoxForProduct(
+   client: IFixitAPIClient,
+   productcode: string
+): Promise<BuyBoxResponse | null> {
+   // TODO pull country override from browser url
+   return client.get(
+      `internal/international_store_promotion/buybox?productcode=${productcode}`
+   );
+}
--- a/frontend/models/product.ts
+++ b/frontend/models/product.ts
@@ -7,6 +7,7 @@ import {
    isRecord,
    Money,
    logAsync,
+   parseItemcode,
 } from '@ifixit/helpers';
 import {
    FindProductQuery,
@@ -69,6 +70,7 @@ export async function findProduct(shop: ShopCredentials, handle: string) {
       ...response.product,
       breadcrumbs,
       iFixitProductId,
+      productcode: parseItemcode(variantSku).productcode,
       images: activeImages,
       options,
       variants: activeVariants,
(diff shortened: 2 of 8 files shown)


frontend/lib/ifixit-api/international-buy-box.ts
```diff
@@ -0,0 +1,33 @@
...
+   salesChannel: string;
+   // Keyed by optionid
+   products: Record<string, BuyBoxVariant>;
+};
+
+export function getBuyBoxForProduct(
+   client: IFixitAPIClient,
+   productcode: string
+): Promise<BuyBoxResponse | null> {
+   // TODO pull country override from browser url
```
sterlinghirsh, line 29: Should this be done? Not sure how we'll QA without this.

sterlinghirsh (dismissed):
CR :zap: though this is missing the country override code. There are also no QA instructions.

frontend/lib/ifixit-api/international-buy-box.ts
danielbeardsley: I'm pretty sure setting `country=` is just a shortcut to changing your preferred store. When we "geo-locate" you, we use the preferred store if it's set.

danielbeardsley:
QA instructions added

danielbeardsley:
Here's a link to the preferred store logic for reference:
https://github.com/iFixit/ifixit/blob/e7407b612e1efdd31d57645971608568c2938688/iFixit/Cart/Stores/StoreProvider.php#L304-L352

deltuh-vee:
If I clear my cookies on the react commerce product page that previously had a buy box, should I see the buy box appear again? This is after clicking "Buy from US Store"

deltuh-vee:
dev_block ✊ 
> When the International Buy Box is present / unresolved, we must hide the "Frequently Bought Together" section to ensure no "Add to Cart" action is available on the page.

This section is still available with the International Buy Box present.
image

sterlinghirsh:
Can we add the logic to specify country= in the url?

danielbeardsley:
> Can we add the logic to specify country= in the url?

That parameter doesn't actually work. It likely used to.

https://www.cominor.com/api/2.0/internal/international_store_promotion/buybox?country=CA&productcode=145198
https://www.cominor.com/api/2.0/internal/international_store_promotion/buybox?country=DE&productcode=145198

In practice, it's not much harder to just change your preferred store.

danielbeardsley:
un_dev_block :+1:    Thanks for catching that @deltuh-vee !  Fixed!

sterlinghirsh:
CR :zap: dev_block :zap: on https://react-commerce-git-intl-buy-box-ifixit.cominor.com/products/iopener we correctly hide "frequently bought together" but after clicking "buy from US store", it doesn't reappear. After refreshing the page, it reappears.

danielbeardsley:
un_dev_block :+1: 

Good catch @sterlinghirsh 

I wanted so much to use redux or make a little state store implementation (like https://github.com/iFixit/ifixit/blob/master/carpenter-frontend/Shared/use-shared-state.ts), but I resisted and just went with calling it higher up in the stack and passing props down.

sterlinghirsh (approved):
CR :zap:

sterlinghirsh:
QA ⚡

PR #965, opened 2022-11-01
Value Proposition: hide iFixit's for drop-shipped items

danielbeardsley:
We don't want to say all the "lifetime warranty" stuff for products that are dropshipped and thus not actually guaranteed / warrantied by us.

Note: I'd rather have this conditional on variant.fulfillmentService, but that info is literally not availble from the graphQL api. We set that value at the same time and from the same source as this meatfield, so they *will* stay in sync.

QA: View a product page who's `shopify_fulfillment_us` Akeneo attribute is not null. The "value proposition" sections should both be missing

Closes iFixit/ifixit#45298

### DIFF
--- a/frontend/templates/product/hooks/useInternationalBuyBox.tsx
+++ b/frontend/templates/product/hooks/useInternationalBuyBox.tsx
@@ -15,7 +15,7 @@ export function useInternationalBuyBox(product: Product) {
       null,
       7 // expire in days
    );
-   const [selectedVariant] = useSelectedVariant(product);
+   const selectedVariant = useSelectedVariant();
    const buyBox = useQuery(
       buyBoxKey(String(product.productcode)),
       () => {
--- a/frontend/templates/product/hooks/useSelectedVariant.ts
+++ b/frontend/templates/product/hooks/useSelectedVariant.ts
@@ -1,7 +1,7 @@
 import { invariant } from '@ifixit/helpers';
 import { Product, ProductVariant } from '@models/product';
 import { useRouter } from 'next/router';
-import React from 'react';
+import { useCallback } from 'react';
 
 type SetVariantIdFn = (variantId: string) => void;
 
@@ -25,7 +25,7 @@ export function useSelectedVariant(
       `Something went wrong, variant with id "${currentVariantId}" not found`
    );
 
-   const setVariantId = React.useCallback<SetVariantIdFn>(
+   const setVariantId = useCallback<SetVariantIdFn>(
       (variantId) => {
          const { variant, ...newQuery } = router.query;
          if (variantId !== defaultVariantId) {
--- a/frontend/templates/product/hooks/useSelectedVariant.ts
+++ b/frontend/templates/product/hooks/useSelectedVariant.ts
@@ -1,11 +1,22 @@
 import { invariant } from '@ifixit/helpers';
 import { Product, ProductVariant } from '@models/product';
 import { useRouter } from 'next/router';
-import { useCallback } from 'react';
+import {
+   useCallback,
+   createContext,
+   useContext,
+   PropsWithChildren,
+} from 'react';
 
 type SetVariantIdFn = (variantId: string) => void;
 
-export function useSelectedVariant(
+const SelectedVariantContext = createContext<ProductVariant | undefined>(
+   undefined
+);
+
+export const SelectedVariantProvider = SelectedVariantContext.Provider;
+
+export function useSelectedVariantFromUrl(
    product: Product
 ): [ProductVariant, SetVariantIdFn] {
    const router = useRouter();
@@ -43,6 +54,14 @@ export function useSelectedVariant(
    );
 
    return [variant, setVariantId];
+}
+
+export function useSelectedVariant(): ProductVariant {
+   const contextValue = useContext(SelectedVariantContext);
+   if (!contextValue) {
+      throw new Error('Component must be inside SelectedVariantProvider');
+   }
+   return contextValue;
 }
 
 function useDefaultVariantId(product: Product): string {
--- a/frontend/templates/product/index.tsx
+++ b/frontend/templates/product/index.tsx
@@ -68,7 +68,7 @@ export const ProductTemplate: NextPageWithLayout<ProductTemplateProps> = () => {
                onVariantChange={setSelectedVariantId}
             />
             <ReplacementGuidesSection product={product} />
-            <ServiceValuePropositionSection />
+            <ServiceValuePropositionSection variant={selectedVariant} />
             {isProductForSale && (
                <CrossSellSection
                   key={selectedVariant.id}
--- a/frontend/templates/product/index.tsx
+++ b/frontend/templates/product/index.tsx
@@ -72,7 +72,7 @@ export const ProductTemplate: NextPageWithLayout<ProductTemplateProps> = () => {
                onVariantChange={setSelectedVariantId}
             />
             <ReplacementGuidesSection product={product} />
-            <ServiceValuePropositionSection variant={selectedVariant} />
+            <ServiceValuePropositionSection />
             {isProductForSale && (
                <CrossSellSection
                   key={selectedVariant.id}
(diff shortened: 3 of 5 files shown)


sterlinghirsh:
dev_block :zap: Only hide the one about shipping for dropshipped products. Ideally, we can replace this with another item soon. Also, why not call useSelectedVariant inside the buybox instead of passing it in?

danielbeardsley:
> Also, why not call useSelectedVariant inside the buybox instead of passing it in?

That's what I did at first, but It's looking like `useSelectedVariant()` isn't meant to be called all over and is mean to be called once.

danielbeardsley:
un_dev_block :+1: 

We're now only hiding the shipping proposition and the `useSelectedVariant()` can be used ubiquitously.

dhmacs:
> why not call useSelectedVariant inside the buybox instead of passing it in?

Imo it's not worth it to use a context for the selected variant when you can just explicitly pass the variant as a prop to the section component. Having that prop also makes it easy to know which sections are variant dependent, and which one aren't. It will also make it easier to reuse sections with other pages should we need it.

I generally try to avoid contexts as they make the code harder to reuse and harder to optimize. My rule of thumb for using context is to use it for truly global stuff (e.g. theming, api client, page data) or when you're doing prop drilling and can't really use composition. 

https://reactjs.org/docs/context.html#before-you-use-context

Screenshot 2022-11-02 at 11 06 38

danielbeardsley:
Everything is tradeoffs. A context that provides the whole variant seems easier, though it has some cons too:

* Components can't be rendered in isolation very easily, you need a whole `ProductVariant` instead of just the one interesting attribute your component uses
   * Taking specific attributes is more verbose and makes it harder to change when you need more data, but makes it easier to test in isolation

dhmacs:
> What is the "cost" here? It seems that using a Context will result in fewer lines of code (vs an often-passed prop). Is there some other cost that this is referring to?
> Heh.. I would have made the opposite argument: Why bother passing in a prop into every component when we have a useful hook?

For the same reason that relying on global variables is not a good idea in general. Of course global variables are useful when used sparingly, but one should still prefer explicit dependencies.

It's also highlighted in the [React docs](https://beta.reactjs.org/learn/passing-data-deeply-with-context#before-you-use-context):

Screenshot 2022-11-02 at 18 41 32
 
> Though I'm not sure why that's important to know. 

https://github.com/iFixit/react-commerce/pull/875 is an instance where that has been useful. At a glance you know which sections you might want to reset state.

> Searching for `useSelectedVariant()` gives a more correct list than trying to search for "components that take a prop of type `ProductVariant`.

Think about context as global variables, would you make a case to use global variables because you can easily search every occurence of the global variable? Sure you can search every occurrence, but usually when you're trying to understand a component tree you read it top down, and in that situation it's helpful when the dependencies are explicit.

> Is it harder because every context needs to under a provider? That's true, but having to pass a prop down several levels isn't "easier".

Prop drilling is an issue, but the instance we're discussing here is not a case of prop drilling, as there aren't several levels of nesting to use the variant. A typical instance of prop drilling would be a page where a small component deep in the dom tree needs to access user details, and none of the ancestors needs to know about the user.

Screenshot 2022-11-02 at 19 14 52


Also it should be noted that composition should be preferred when one wants to avoid prop drilling.


> Hmm, I would think a context would help with that as the ancestor components don't have to re-render as only the component that uses the hook gets re-rendered when it changes (or so I thought). Is there something else I'm missing here?

That's not the case, every time the context value changes, everything under the provider is re-rendered. It's actually tricky to make efficient use of context, you need to rely on memoization.

---

Beside [React docs](https://beta.reactjs.org/learn/passing-data-deeply-with-context#before-you-use-context), there are many resources that discuss this. Here's a few:

- https://kentcdodds.com/blog/application-state-management-with-react
- https://kentcdodds.com/blog/prop-drilling
- https://kentcdodds.com/blog/how-to-use-react-context-effectively

sterlinghirsh:
> That's not the case, every time the context value changes, everything under the provider is re-rendered. It's actually tricky to make efficient use of context, you need to rely on memoization.

TIL

sterlinghirsh:
Regarding using the context or not, I could go either way in this case. I think it's true that we weren't passing the variant around so much that it was a burden, but I think it's a good call to be judicious in the use of contexts. I'd be surprised if that ended up being the cause of performance issues on this page. In any case, this is now failing tests, so it looks like we'll have to make some changes anyway.

sterlinghirsh:
CR :zap: dev_block :zap:

dhmacs:
> I'd be surprised if that ended up being the cause of performance issues on this page. 

True, most of the time you don't need to optimize for rendering perfs 👍

danielbeardsley:
> Think about context as global variables, would you make a case to use global variables because you can easily search every occurence of the global variable? Sure you can search every occurrence, but usually when you're trying to understand a component tree you read it top down, and in that situation it's helpful when the dependencies are explicit.

Hmm, I wouldn't put them in the same class as global variables. They are read-only and they effectively notify each reader when the value changes. What's often disliked about global variables is that they can be updated by anyone and no readers can know when the values might change. Hooks that provide the needed data solve both of those issues. So the major headache of global variables doesn't really apply. You get the convenient access to data you need with out the dangers.

Regarding testing in isolation, contexts can make that more verbose (requiring a `<Provider>` instead of a prop). Though I'd say we do far more reading and maintaining of code than we do testing components in isolation.

> It's actually tricky to make efficient use of context, you need to rely on memoization.

I understand now, thanks for clearing it up .This isn't a strike against contexts, It seems it results in the same amount of re-renders and memoization can help both techniques.

I think globally available hooks have some maintenance benefits:

* Moving a component is easier
   * Don't have to reason about adding a new prop and passing it down through new ancestors and removing the potentially unused prop from old ancestors
* Don't have to see the prop-passing noise in current ancestors, making the structure and layout easier to see

I'm doing all this questioning because I want to end up on the same page regarding the patterns we use in development, not because this particular use-case is a perfect fit for one technique vs the other. Thanks for taking the time @dhmacs !

dhmacs:
> Hmm, I wouldn't put them in the same class as global variables. They are read-only and they effectively notify each reader when the value changes. What's often disliked about global variables is that they can be updated by anyone and no readers can know when the values might change. Hooks that provide the needed data solve both of those issues. So the major headache of global variables doesn't really apply. You get the convenient access to data you need with out the dangers.
> 
> Regarding testing in isolation, contexts can make that more verbose (requiring a `<Provider>` instead of a prop). Though I'd say we do far more reading and maintaining of code than we do testing components in isolation.
> I think globally available hooks have some maintenance benefits:
> 
> * Moving a component is easier

I think the disagreement here is on the fact code using context is easier to read and to maintain. My experience so far has been that the components that are the easier to understand and maintain are those that are functionally pure. The less you use context and side effects, the easier it is to move the component around and reuse it.
 
>   * Don't have to reason about adding a new prop and passing it down through new ancestors and removing the potentially unused prop from old ancestors

Here you're describing the prop drilling anti-pattern, and you're right, we should avoid this. Although that's not a problem that we have here, so I don't think this justifies the use of a context. Also, as React docs suggests, if we're presented with an instance of a prop drilling problem, we should try to get rid of it by using composition, and use context only as a last resort.

> * Don't have to see the prop-passing noise in current ancestors, making the structure and layout easier to see

To me it's not noise, it's signal. It makes it easier to understand what a component does and what data it depends on.

> I'm doing all this questioning because I want to end up on the same page regarding the patterns we use in development, not because this particular use-case is a perfect fit for one technique vs the other. 

You're right, it's important to have these discussions so we understand why we write code the way we do, so I'm glad you're asking these questions 👍 
While writing this code I definitely considered using context, since the selected variant is kind of a global concept within the product page, so I definitely see why you did this refactor. The reason I kept using props is that over these years writing React code, I've learned (the hard way 😅 ) that it's always wise to avoid context if possible, as there have been many times where I regretted it because it made more difficult to reuse the code down the road.
I definitely experienced first hand why React docs suggest to stick with props, so that's why I think they have good reasons to warn against overusing it.
When deciding if we need to use a context, I think we need to find very good reasons to use it, and not the other way around.

danielbeardsley:
Thanks for sticking with me @dhmacs and taking the time!

I'm not totally convinced that passing oft-used props to a bunch of components is easier to maintain long term than using a context and a hook, but you have more experience maintaining wholly-React codebases so I'll stick with that convention!

I'll close this pull and re-open it without the context change.

Issue #961, opened 2022-10-31
Parallelize or make client-side-only product reviews api request


sterlinghirsh:
## Problem
We currently make two requests to generate the product page. One is the Shopify query to get the product data, and the other is an iFixit api query to get review data. These are currently sequential (waterfall) requests.

## Do This
Parallelize these requests or make the request on the client side. If we parallelize the requests, then https://github.com/iFixit/ifixit/issues/45373 is a prereq to this, since we need to be able to request by handle instead of productcode. If we do the request clientside-only, we can go ahead without it. The reviews are so far down the page, we could get away without loading them in the initial request.

In any case, if we continue to request the reviews on the server, we should also make sure we only send the first 3 reviews (i.e. the ones visible on page load). Looking at the source of https://www.ifixit.com/products/iphone-6s-replacement-battery, a big chunk of the bytes sent over the wire are reviews that aren't visible until someone clicks "see more reviews".

Issue #853, opened 2022-10-20
Add structured data on Product Page


sterlinghirsh:
Add jsonld

This doc lists "Required" properties, as well as some other properties that it'd be great to implement later: https://docs.google.com/document/d/1gprBDfp7ZQ7nG57DE45JOMG4YNr8F7uqr2lCQTBiMyA/edit#

ifixitmaxb:
In general, my hope is that we simply keep the same Schema that we have on legacy for v2.0. But I found at least one example of an issue on legacy Schema that we should try to fix, which would be an improvement/iteration, for v2.0. 

For the example page, https://www.ifixit.com/Store/Mac/MacBook-Pro-13-Inch-Retina-Early-2015-Battery/IF123-054?o=1, here is the Rich Results Test,
https://search.google.com/test/rich-results/result/r%2Fproduct?id=BEByUQRlYj_hNUl3ew5XbQ, the 2 "non critical issues" says we're missing review and aggregate review:
![image](https://user-images.githubusercontent.com/86265382/198345642-ff51bd83-c2ba-4211-85c0-0eee2e5eff54.png)

sterlinghirsh:
Any movement here? @davidrans said that we've lost "rich results" on the new pages, so we should get this working again asap.

davidrans:
Here's an example of what we're passing on the Pro Tech Toolkit page: https://search.google.com/test/rich-results/result?id=QqXl7U1WWc98tIXer8yAAA

ifixitmaxb:
I'll have more direction on structured data I can share on Tuesday next week.

ifixitmaxb:
https://docs.google.com/document/d/1gprBDfp7ZQ7nG57DE45JOMG4YNr8F7uqr2lCQTBiMyA/edit#

PR #972, opened 2022-11-02
Move Review API Calls To the Front End

danielcliu-ifixit:
## Overview

We would like to stop calling the reviews API from the backend and put it in the frontend to reduce the time to page load and the amount of data for the props we are passing. This makes it so the product pages call to the review API themselves when they load on the client.

## QA 
Does the review data still work (in the product section, the cross sell section, and the reviews section)? Are page load times faster now? 

`/products/xbox-one-x-fan-assembly` is a good url for no reviews.

Closes: #961


frontend/templates/product/index.tsx
```diff
@@ -34,6 +40,22 @@ import { useInternationalBuyBox } from '@templates/product/hooks/useInternationa
...
    const { product } = useProductTemplateProps();
    const [selectedVariant, setSelectedVariantId] = useSelectedVariant(product);
+   const apiClient = useIFixitApiClient();
+   const [reviewsData, setReviewsData] = useState<
+      ProductReviewData | null | undefined
+   >();
+   useEffect(() => {
+      if (!product.iFixitProductId) {
+         return;
+      }
```
sterlinghirsh, line 50: I think you can simplify the type to ProductReviewData | null if you init this with null.
```suggestion
   const [reviewsData, setReviewsData] = useState<
      ProductReviewData | null
   >(null);
   useEffect(() => {
      if (!product.iFixitProductId) {
         return;
      }
```

sterlinghirsh:
CR :zap: dev_block :zap: This does the thing, but it went a little too far. I think this was an ambiguity in the spec. There are two relevant chunks of reviews data. One is the rating+count, which comes from the Shopify api as a metafield and gets displayed at the top of the page and in the "frequently bought with" section. The second is the content of the reviews and the breakdown of scores, which gets shown in the reviews section lower down on the page. The intent of this issue was to load the second section client-side, but to leave the first section alone. On the one hand, it's good that these numbers will stay in sync, since it would look weird if the page loaded and said 10 reviews in one place but 11 in another. However, the way it's implemented right now causes a layout shift and a visible change as the page loads, since the user will initially see the page load without the change, and then when the reviews api request completes, everything below the review stars shifts down. I'd rather leave the ratings that used to come from Shopify metafields as they were, and move the useEffect and all uses of the reviews api down to the Reviews component. It's technically possible that the two values could briefly diverge, but since we're repushing to Shopify on every new review, I'm not too worried about it.

danielcliu-ifixit:
un_dev_block :v:

federicobadini:
⚡ dev_block ⚡
In a [separate issue](https://github.com/iFixit/react-commerce/issues/853) we are working to include `ld+json` structured data on product pages. That json includes product reviews. 
We are investigating whether we can handle those fields client side or if we are forced to handle them server side.
Given that the two issues seem pretty connected, I though it was better to pause this for a second in order to avoid merging and then having to revert things

frontend/templates/product/hooks/useProductReviews.ts
```diff
@@ -1,22 +0,0 @@
...
-import { fetchProductReviews, Product } from '@models/product';
-import { useQuery } from 'react-query';
-
-const productReviewsKeys = {
-   reviews(productId: string) {
-      return ['product-reviews', productId];
-   },
-};
-
-export function useProductReviews(product: Product) {
```
danielbeardsley, line 11: Why did we remove this usage of react-query? I'm pretty sure this is the exact use-case for it so we only end up querying the API one time instead of one time per component the data is is used in.

ifixitmaxb:
I'm investigating product page Schema but won't be done until Tuesday. I can't say that my investigation will answer this exact scenario though. Fwiw, from what I can see here, I agree with Danny that including actual review content in our json-ld doesn't seem necessary and that review count and aggregate rating is most common/important. Hopefully I'll know more Tuesday. 
I'm fine with it if we merge this and check in the Schema validator / rich results tester once it's live and make a call then if we need to redo it or if it's good to go.

frontend/templates/product/hooks/useProductReviews.ts
danielcliu-ifixit: yeah i think i got into my head this wasn't necessary because we weren't using initial data anymore

danielcliu-ifixit:
un_dev_block :v:

frontend/templates/product/hooks/useProductReviews.ts
```diff
@@ -12,11 +12,7 @@ export function useProductReviews(product: Product) {
    const apiClient = useIFixitApiClient();
    const query = useQuery(
       productReviewsKeys.reviews(product.iFixitProductId),
-      () => fetchProductReviews(apiClient, product.iFixitProductId),
-      {
-         initialData: product.reviewsData,
-         staleTime: Infinity, // Right now we're not interested in refreshing data client side
```
danielbeardsley, line 18: I think we still want this `staleTime` setting. Otherwise, the reviews may be refetched too often:
image

danielbeardsley (changes requested):
dev_block :+1:  on the one comment, but CR :+1:  otherwise.

I noticed you removed some unused imports and may have added some now-unused imports. I'm a little surprised. I thought we had some sort of hook that prevented committing or pushing in those cases... but I guess not. Not really that harmful, so I don't really care.

frontend/templates/product/hooks/useProductReviews.ts
danielcliu-ifixit: im not sure any of those cases are really that bad in this instance, but I can add it back in.

danielcliu-ifixit:
un_dev_block :v:

danielbeardsley (approved):

danielbeardsley:
CR :+1:

jordycosta:
QA 🟢 

Review data is working 💯 
Page load times are faster ✔️ 

> 604 ms vs. 892 ms avg load time (no reviews)
> 824 ms vs. 1054 ms avg load time (reviews)

### Testing Procedure:
1. Navigate to [a product without reviews](https://react-commerce.vercel.app/products/xbox-one-x-fan-assembly)
2. Navigate to [a product with reviews](https://react-commerce.vercel.app/products/moray-driver-kit#reviews)
3. Ensure review data is loading & is correct
4. Compare load times to normal preview

PR #973, opened 2022-11-03
International Buy Box: track clicks in matomo

### LINKED ISSUE #930: International Buy Box Banner

danielbeardsley:
Keep the same event names / categories we had on the previous product page.

Closes #930

### DIFF
--- a/frontend/templates/product/sections/ProductSection/InternationalBuyBox.tsx
+++ b/frontend/templates/product/sections/ProductSection/InternationalBuyBox.tsx
@@ -1,9 +1,11 @@
 import { Button, Link, Box } from '@chakra-ui/react';
+import { useCallback } from 'react';
 import {
    BuyBoxResponse,
    BuyBoxVariant,
 } from '@lib/ifixit-api/international-buy-box';
 import { Flag, FlagCountryCode } from '@ifixit/icons';
+import { trackInMatomoAndGA } from '@ifixit/analytics';
 
 type InternationalBuyBoxProps = {
    store: BuyBoxResponse;
@@ -16,6 +18,13 @@ export function InternationalBuyBox({
    selectedVariant,
    dismissBuyBox,
 }: InternationalBuyBoxProps) {
+   const trackClick = useCallback(() => {
+      const category = `${store.salesChannel} Store Buy Box`;
+      trackInMatomoAndGA({
+         eventCategory: category,
+         eventAction: `${category} - Click - ${selectedVariant.gaSku}`,
+      });
+   }, [selectedVariant, store]);
    return (
       <Box textAlign="center">
          <Button
@@ -24,6 +33,7 @@ export function InternationalBuyBox({
             as="a"
             w="full"
             colorScheme="brand"
+            onClick={trackClick}
             href={selectedVariant?.productURL}
          >
             <Flag mr={3} w={7} code={store.countryCode as FlagCountryCode} />{' '}


danielcliu-ifixit (approved):
CR :v:

danielbeardsley:
This is what is sent.
image

Does that seem right @ifixitmaxb or @jeffsnyder ? I copied this format from the previous product page and the events it would send.

danielbeardsley:
I'm nearly sure it's the correct stuff. Here is an existing event in matomo:
image

QA :+1:

Issue #741, opened 2022-09-26
Collections: Display 4 rows "devices" on /Parts/Electronics parent collection


ifixitmaxb:
On [/Parts/Electronics parent collection](https://www.ifixit.com/Parts/Electronics), display 4 rows instead of the default 3. We want all the current child collections to show and not be hidden by "Show more".

Similar to https://github.com/iFixit/react-commerce/issues/682

ifixitmaxb:
I saw your comment before, Federico, about this all depends on screen size. That's fair. I'm focused here on the large screen size where each row has 4 columns. The most common screen size viewing iFixit.com is 1900 wide so I'm focused on this for now. I'd like to show 4 rows for the largest/widest screen size (resolution?). Is it possible to code this change for the largest/widest resolution and not affect the other sizes/resolutions? @federicobadini

federicobadini:
Would a Strapi flag like `showAllChildrenCollectionsOnLargeScreen` solve this?
Just let me know if this would suffice or if you would like to have an even more granular control on this.

ifixitmaxb:
@jeffsnyder what do you think?

jeffsnyder:
@federicobadini Yes, that would solve the problem.

ifixitmaxb:
Are we going to default to not show all? And the toggle in Strapi would only be used if we want to override the default and show all? It seems like this new feature is causing some page creation/editing issues.

sterlinghirsh:
I don't think this needs any more spec. Let me know if there's anything missing here.

Issue #682, opened 2022-09-07
Collections: Display 4 rows "devices" on /Tools parent collection


ifixitmaxb:
Just for /Tools parent collection, display 4 rows instead of the default 3. We want all the current child collections to show and not be hidden by "Show more".

ifixitmaxb:
I'd like to propose doing this exact same thing but for https://www.ifixit.com/Parts/Electronics. (Imo, we should show Vacuum, Television and VR parts and not hide behind "Show more")

federicobadini:
How we are supposed to handle this for smaller screen sizes?
At the moment we are showing a different number of columns according to the screen size:
```
base: 1,
sm: 2,
md: 3,
lg: 4,
```
So small screens already show a much shorter list of items.

PR #974, opened 2022-11-03
Add a flag to control the show all default value of product list children on large screen (backend changes)

federicobadini:
Strapi changes linked to #682 and #741 

This PR covers only the Strapi related changes and will be completed by #983 
The aim of the two PR combined is the following:

A new product list flag is now giving the possibility to show all product list children on large screen.
If enabled the list is not showing the `Show more` / `Show less` button on `> lg` screen sizes.
On smaller screen sized nothing should change.

Here is a preview of the `/Tools` page with the flag enabled:

Screenshot 2022-11-03 at 08 51 52

The same page on mobile is showing the `Show more` / `Show less` button as before:

Screenshot 2022-11-03 at 09 03 54

This PR introduces only the flag into Strapi 

### QA

1. Visit Vercel preview
2. Verify that Strapi includes the new flag in the `productList` model

### DIFF
--- a/backend/src/api/product-list/content-types/product-list/schema.json
+++ b/backend/src/api/product-list/content-types/product-list/schema.json
@@ -177,6 +177,16 @@
             }
          },
          "type": "string"
+      },
+      "defaultShowAllChildrenOnLgSizes": {
+         "pluginOptions": {
+            "i18n": {
+               "localized": true
+            }
+         },
+         "type": "boolean",
+         "default": false,
+         "required": true
       }
    }
 }
--- a/frontend/lib/strapi-sdk/generated/sdk.ts
+++ b/frontend/lib/strapi-sdk/generated/sdk.ts
@@ -969,6 +969,7 @@ export type ProductList = {
    children?: Maybe<ProductListRelationResponseCollection>;
    childrenHeading?: Maybe<Scalars['String']>;
    createdAt?: Maybe<Scalars['DateTime']>;
+   defaultShowAllChildrenOnLgSizes: Scalars['Boolean'];
    description: Scalars['String'];
    deviceTitle?: Maybe<Scalars['String']>;
    excludeFromHierarchyDisplay: Scalars['Boolean'];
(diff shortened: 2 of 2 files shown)


frontend/lib/strapi-sdk/operations/getProductList.graphql
```diff
@@ -15,6 +15,7 @@ query getProductList($filters: ProductListFiltersInput) {
             description
             metaDescription
             metaTitle
+            defaultShowAllChildrenOnLgSizes
```
masonmcelvain, line 18: In the past, we've had to deploy Strapi API changes before deploying frontend changes to use that API. Otherwise, in the gap between when Vercel redeploys, and when a dev manually redeploys Strapi, this query will fail because the field doesn't exist and the page will 500 (you'll notice that the [prod deployment](https://react-commerce-prod-git-display-4-rows-devices-on-684edd-ifixit.vercel.app/Tools) for this branch is 500ing right now). This is unfortunate, and we'd like to solve this in the future by automatically redeploying Strapi ([ref](https://github.com/iFixit/ifixit/issues/42822)).

I would suggest making all the Strapi changes in one pull, merging and deploying that pull, then making the frontend changes.

masonmcelvain (changes requested):
CR ✌🏻 The code looks good, but dev_block ✌🏻 on splitting the Strapi changes into their own pull

frontend/lib/strapi-sdk/operations/getProductList.graphql
federicobadini: You are right, I'll split this PR in two

federicobadini:
I've isolated only the Strapi related changes in here

masonmcelvain:
un_dev_block ✌🏻

masonmcelvain:
For QA, let's just check that the new field `defaultShowAllChildrenOnLgSizes` shows up in the admin UI.

masonmcelvain:
There it is! QA ✌🏻 

image

https://display-4-rows-devices-on-selected-pages.govinor.com/admin/content-manager/collectionType/api::product-list.product-list/246?plugins[i18n][locale]=en

masonmcelvain:
Redeployed strapi.ifixit.com ✔️

Issue #884, opened 2022-10-25
Collection Pages: Add "Store" as top level breadcrumbs 


erinemay:
![image](https://user-images.githubusercontent.com/72166715/197894302-8a53b8e5-334f-4269-a039-7ec7ab8a698d.png)

On collection pages, when breadcrumbs are present, add a higher root level breadcrumb to "Store" which links to /store

PR #981, opened 2022-11-03
Product Lists: Add Home to all Breadcrumbs

danielcliu-ifixit:
## Overview

Apparently people want to be able to go back to `/Store` from the product-list pages. This makes sure that all `ancestor` lists for strapi productlists have "Home" as an ancestor by inserting them into all ancestor lists.

## CR
I wasn't able to fix out a switch statement to be as clean as i wanted (the `throw Error` line) but Im not sure if its supposed to as it can type-wise reach that case still.

## QA
Do all product-list breadcrumb navbars seem correctly? The only new functionality should be that `/Parts` and `/Tools` have breadcrumbs now and that `/Tools/<category>` should have the desktop view breadcrumbs now.

Closes: #884

### DIFF
--- a/frontend/models/product-list/types.ts
+++ b/frontend/models/product-list/types.ts
@@ -36,6 +36,7 @@ export enum ProductListType {
    AllTools = 'tools',
    ToolsCategory = 'tools-category',
    Marketing = 'marketing',
+   Store = 'store',
 }
 
 export enum RefinementDisplayType {
@@ -48,7 +49,8 @@ export type ProductList =
    | DevicePartsProductList
    | AllToolsProductList
    | ToolsCategoryProductList
-   | MarketingProductList;
+   | MarketingProductList
+   | StoreProductList;
 
 export interface BaseProductList {
    title: string;
@@ -89,6 +91,10 @@ interface ToolsCategoryProductList extends BaseProductList {
 
 interface MarketingProductList extends BaseProductList {
    type: ProductListType.Marketing;
+}
+
+interface StoreProductList extends BaseProductList {
+   type: ProductListType.Store;
 }
 
 export interface ProductListAncestor {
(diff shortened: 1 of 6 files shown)


deltuh-vee:
QA 🎬 
* /Parts and /Tools have a Breadcrumb to /Store
* /Tools/<category> also have breadcrumbs

frontend/models/product-list/types.ts
```diff
@@ -36,6 +36,7 @@ export enum ProductListType {
    AllTools = 'tools',
    ToolsCategory = 'tools-category',
    Marketing = 'marketing',
+   Store = 'store',
```
dhmacs, line 39: I doesn't "feel" right to add "store" as a product list type, since store is not a product list page. Can we refactor the `ProductListBreadcrumb` component to account for non-product list segments?

There's a more generic `PageBreadcrumb` component that has been used in the new product page and could be used here too, but maybe this would be a bigger refactor than it needs to be for the scope of this PR. 🤔

frontend/templates/product-list/ProductListView.tsx
```diff
@@ -31,17 +31,16 @@ export function ProductListView({
...
+      productList.type === ProductListType.AllTools ||
+      productList.type === ProductListType.ToolsCategory;
 
    return (
       <>
          <SecondaryNavbar
             display={{
                base: isToolPage ? 'none' : 'initial',
-               sm: isAllToolsPage ? 'none' : 'initial',
+               sm: isToolPage ? 'none' : 'initial',
```
dhmacs, line 43: I think we can drop this breakpoint, since it's already covered by `base`

dhmacs (changes requested):
dev_block 🔢 on a suggestion

danielcliu-ifixit:
un_dev_block :+1: implemented suggestions

deltuh-vee:
Carry Over QA 🎬

sterlinghirsh (approved):
CR :zap:

Issue #953, opened 2022-10-28
Replacement guide images are low res


jordycosta:
### Issue Overview:

![guides](https://user-images.githubusercontent.com/95656772/198717330-12218819-f531-40b9-887c-69728ba3d3d7.jpg)

vs. 

![featured](https://user-images.githubusercontent.com/95656772/198717344-82feb0aa-b5d7-48c8-a671-7617173a4e5c.jpg)

### Reproduction Steps:
1. Navigate to [any product page](https://www.ifixit.com/products/macbook-pro-13-retina-function-keys-late-2016-2017-battery)
2. Scroll down to the **Replacement Guides** section
3. Observe the low-res guide images

PR #982, opened 2022-11-03
Fix isGuideImage() Regex

danielcliu-ifixit:
## Overview

Various images on our site are much lower res than we'd like. This is likely an issue with the product urls.

## CR
It turns out it was an issue with our `isGuideImage()` function, as the regex wasn't catching `.guide-image` urls due to looking for two `/` back to back. This removes the extra `/` from the regex, and also makes our `isCartImage()` more like `isGuideImage`.

## QA
Are the images on `/products/macbook-pro-13-retina-function-keys-late-2016-2017-battery` for Replacement Guides now higher res? This page in particular works because weve replaced the proper metafield on `ifixit-test` with production data (see replacement guides metafield https://ifixit-test.myshopify.com/admin/products/6556174286938). 

Also do the cart images still work after the slight changes? Production cartimage url (https://cart-products.cdn.ifixit.com/cart-products/jFsGZHKTWUGOJCUZ.size250?width=180) and its equivalent on dev (https://emay.cominor.com/cart-products/jFsGZHKTWUGOJCUZ.standard). These are from a product list page that had `Thunderbolt 3 (USB-C) Cable .5 Meter` on them

Closes: #953

### DIFF
--- a/packages/ui/misc/IfixitImage.tsx
+++ b/packages/ui/misc/IfixitImage.tsx
@@ -52,7 +52,7 @@ export function IfixitImage({ alt = '', ...props }: ImageProps) {
 
 function isGuideImage(src: string) {
    return src.match(
-      /^https:\/\/(guide-images\.cdn\.ifixit\.com\/|([^/]+\.(ubreakit|cominor)\.com\/igi))\//
+      /^https:\/\/(guide-images\.cdn\.ifixit\.com|([^/]+\.(ubreakit|cominor)\.com\/igi))\//
    );
 }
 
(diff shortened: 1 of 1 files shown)


deltuh-vee:
QA 🎬 
* Replacement Guide images are no longer low res.
* Cart images still work

danielbeardsley:
Wow... `isGuideImage()` has just never worked because of this. This is a case where a small set of unit tests would have been worth it.

danielbeardsley (approved):
CR :+1:

PR #991, opened 2022-11-04
Revert "Add a flag to control the show all default value of product list children on large screen (backend changes)"

federicobadini:
Reverts iFixit/react-commerce#974


masonmcelvain (approved):
CR ✌🏻

PR #993, opened 2022-11-04
Value propositions: hide parts for dropshipped products

danielbeardsley:
Hide the "shipping" parts of our value propositions for products that are not shipped by us.

Originally: #965 but simplified here

Closes https://github.com/iFixit/ifixit/issues/45298

CC @dhmacs

### DIFF
--- a/frontend/templates/product/sections/ServiceValuePropositionSection/index.tsx
+++ b/frontend/templates/product/sections/ServiceValuePropositionSection/index.tsx
@@ -1,3 +1,4 @@
+import { ProductVariant } from '@models/product';
 import { Heading, Stack, Text, VStack } from '@chakra-ui/react';
 import React from 'react';
 import { List, ListIcon, ListItem } from '@chakra-ui/react';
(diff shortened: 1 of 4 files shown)


frontend/templates/product/sections/ServiceValuePropositionSection/index.tsx
```diff
@@ -12,10 +13,15 @@ import {
    faShieldCheck,
 } from '@fortawesome/pro-solid-svg-icons';
 import { FaIcon, FaIconProps } from '@ifixit/icons';
+import { useSelectedVariant } from '../../hooks/useSelectedVariant';
```
dhmacs, line 16: This is not being used

```suggestion
```

dhmacs (approved):
lgtm! I've suggested a minor change, but it's fine to merge as is 👍 

CR ⚡

dhmacs (approved):
CR ⚡

deltuh-vee:
dev_block ✊ 
I still see the value proposition sections for dropshipped items. As an example, I chose Asus C302CA Laptop Battery(IF335-093-1). 
image
image

danielbeardsley:
Good catch @deltuh-vee !  Turns out the bug wasn't in this code and this code works when the metafield is edited to have the right stuff:

image

image

un_dev_block :+1: 

QA :+1: 

I'll open up a pull on our php app with the fix at the source of the data.

Issue #817, opened 2022-10-12
Make sure we don't canonical redirect away from ?worksin filtered pages


sterlinghirsh:
These are pages we get to from workbench. I've seen page loads where after the page load completes, we canonical redirect. This might be related to algolia's url handling.

Example: https://www.ifixit.com/Parts/PlayStation_4?worksin=PlayStation%204%20CUH-11XXA (From workbench after adding a specific topic)

sterlinghirsh:
Note to self: example

masonmcelvain:
[Algolia url handling](https://www.algolia.com/doc/api-reference/widgets/history-router/react-hooks/) is stripping any query params that are not used by algolia.

The `worksin` param is not used by Algolia, so it is stripped. The syntax Algolia recognizes is https://www.ifixit.com/Parts/PlayStation_4?filter[worksin][0]=PlayStation%204%20CUH-11XXA. Notice that the filter is applied, and the url is unmodified.

![image](https://user-images.githubusercontent.com/52104630/200396215-9cb62e23-0788-4954-a2a4-e9383caf1717.png)

Should this issue be "don't strip unused query parameters"?

Or "detect the worksin parameter and map it to the syntax algolia recognizes"?

PR #997, opened 2022-11-07
Collections: Don't strip query params

masonmcelvain:
Fixes a bug where query params on collection pages that weren't used by Algolia were stripped from the url.

This keeps all query params in the url, even if they aren't used in Algolia search.

This means we don't have to manually forward the `_vercel_no_cache` param anymore.

I also noticed that we can pass an option to `qs` to ignore the leading '?' in the query, instead of slicing it out ourselves.
https://github.com/ljharb/qs

## QA

- Make sure filtering on collections pages still works.
- Make sure when visiting a collections page from an external link, all query parameters are persisted in the url.
  - Specifically, make sure `_vercel_no_cache=1` still works and doesn't get stripped.

You'll notice that www.ifixit.com/Parts?dogs=cats the `dogs=cats` gets stripped. On this branch, it shouldn't be stripped, and search should still work as usual.

CC @sterlinghirsh

Noticed in #817

### DIFF
--- a/frontend/components/common/InstantSearchProvider.tsx
+++ b/frontend/components/common/InstantSearchProvider.tsx
@@ -11,7 +11,6 @@ import algoliasearch from 'algoliasearch/lite';
 import { history } from 'instantsearch.js/es/lib/routers';
 import { RouterProps } from 'instantsearch.js/es/middlewares';
 import { UiState } from 'instantsearch.js/es/types';
-import { mapValues } from 'lodash';
 import { useRouter } from 'next/router';
 import * as React from 'react';
 import {
@@ -153,13 +152,14 @@ export function InstantSearchProvider({
                // Item Type is the slug on device pages, not in the query.
                delete filterCopy['facet_tags.Item Type'];
             }
-            const customQueryParams = new URLSearchParams(location.search);
+            const initialQuery = qsModule.parse(location.search, {
+               ignoreQueryPrefix: true,
+            });
             const queryString = qsModule.stringify(
                {
+                  ...initialQuery,
                   ...routeState,
                   filter: filterCopy,
-                  _vercel_no_cache:
-                     customQueryParams.get('_vercel_no_cache') || undefined,
                },
                {
                   addQueryPrefix: true,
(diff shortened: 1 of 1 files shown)


frontend/components/common/InstantSearchProvider.tsx
```diff
@@ -176,7 +176,9 @@ export function InstantSearchProvider({
             const deviceHandle = pathParts.length >= 2 ? pathParts[1] : '';
             const itemType = pathParts.length >= 3 ? pathParts[2] : '';
 
-            const { q, p, filter } = qsModule.parse(location.search.slice(1));
+            const { q, p, filter } = qsModule.parse(location.search, {
+               ignoreQueryPrefix: true,
+            });
```
sterlinghirsh, line 181: I wonder if it would make sense to do something like `const {q, p, filter, ...otherParams} = ` here and then reference otherParams in createURL. I worry that defaulting the query in createURL is going to prevent removing params that should be removed, since they only get overridden when they're set I think.

sterlinghirsh:
CR :zap: dev_block :zap: There's a bug with the param overrides.

https://react-commerce-git-keep-all-collections-query-params-ifixit.vercel.app/Parts/iPhone if you type in a search of "asdf", the url correctly becomes https://react-commerce-git-keep-all-collections-query-params-ifixit.vercel.app/Parts/iPhone?q=asdf, but when you delete the search query in the UI, the ?q parameter updates (down to ?q=a if you go slowly) but is never removed.

frontend/components/common/InstantSearchProvider.tsx
masonmcelvain: Good catch. We need to unset keys of `RouteState` (ie `q`, `p`, and `filter`) in the `initialQuery`, otherwise they'll never vanish. I'm trying to think of a way to not make `createURL` depend on `otherParams` from `parseURL`, but maybe that's fine and the simplest way to do it.

masonmcelvain:
un_dev_block ✌🏻 I think the bug @sterlinghirsh described is fixed

sterlinghirsh (approved):
CR :zap:

deltuh-vee:
> Make sure when visiting a collections page from an external link, all query parameters are persisted in the url. 
Is this something that has to be checked after merging? I'm not sure how to visit the collections pages in this pull as an external link.

masonmcelvain:
> > Make sure when visiting a collections page from an external link, all query parameters are persisted in the url.
> 
> Is this something that has to be checked after merging? I'm not sure how to visit the collections pages in this pull as an external link.

By that, I meant "paste the url in the address bar of your browser and hit enter", sorry for the confusion. An internal link would be a link that exists on one of our Next.js pages, for clarity.

deltuh-vee:
QA 🎬 
-  Filtering on collections pages still works.
-  When visiting a collections page from an external link, all query parameters are persisted in the url.

PR #998, opened 2022-11-07
Set itemType on Server

### LINKED ISSUE #783: Collections: Set item type explicitly on the server
## Problem

When you load an item type collection page, it initially says "Macbook Parts" and then flashes to "Macbook Batteries". We thought upgrading instantsearch would have resolved this, so we dropped the feature that parsed the url on the server and set the item type in the productlist prop.

## Do This

When on the server, set `productlist.itemType` if there is an item type in the url. Use this instead of `devicePartsItemType` in the hook.

CC @sterlinghirsh

sumukhipandey:
This PR is basically a revert of this commit https://github.com/iFixit/react-commerce/commit/5155bbcca4ecfc6e7d287e91f7ca53bf10c3ba70 We got rid of this feature and upgraded to instantSearch, but this was causing a weird loading pattern.

Closes #783 

CC: @masonmcelvain

### DIFF
--- a/frontend/components/common/InstantSearchProvider.tsx
+++ b/frontend/components/common/InstantSearchProvider.tsx
@@ -11,7 +11,6 @@ import algoliasearch from 'algoliasearch/lite';
 import { history } from 'instantsearch.js/es/lib/routers';
 import { RouterProps } from 'instantsearch.js/es/middlewares';
 import { UiState } from 'instantsearch.js/es/types';
-import { mapValues } from 'lodash';
 import { useRouter } from 'next/router';
 import * as React from 'react';
 import {
@@ -153,13 +152,17 @@ export function InstantSearchProvider({
                // Item Type is the slug on device pages, not in the query.
                delete filterCopy['facet_tags.Item Type'];
             }
-            const customQueryParams = new URLSearchParams(location.search);
+            const { q, p, filter, ...otherParams } = qsModule.parse(
+               location.search,
+               {
+                  ignoreQueryPrefix: true,
+               }
+            );
             const queryString = qsModule.stringify(
                {
+                  ...otherParams,
                   ...routeState,
                   filter: filterCopy,
-                  _vercel_no_cache:
-                     customQueryParams.get('_vercel_no_cache') || undefined,
                },
                {
                   addQueryPrefix: true,
@@ -176,7 +179,9 @@ export function InstantSearchProvider({
             const deviceHandle = pathParts.length >= 2 ? pathParts[1] : '';
             const itemType = pathParts.length >= 3 ? pathParts[2] : '';
 
-            const { q, p, filter } = qsModule.parse(location.search.slice(1));
+            const { q, p, filter } = qsModule.parse(location.search, {
+               ignoreQueryPrefix: true,
+            });
 
             const filterObject =
                typeof filter === 'object' && !Array.isArray(filter)
--- a/frontend/components/common/InstantSearchProvider.tsx
+++ b/frontend/components/common/InstantSearchProvider.tsx
@@ -130,6 +130,7 @@ export function InstantSearchProvider({
                .filter((part) => part !== '');
             const partsOrTools = pathParts.length >= 1 ? pathParts[0] : '';
             const deviceHandle = pathParts.length >= 2 ? pathParts[1] : '';
+            const itemTypeHandle = pathParts.length >= 3 ? pathParts[2] : '';
 
             let path = '';
             if (partsOrTools) {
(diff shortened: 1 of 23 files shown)


frontend/components/common/InstantSearchProvider.tsx
```diff
@@ -130,6 +130,7 @@ export function InstantSearchProvider({
                .filter((part) => part !== '');
             const partsOrTools = pathParts.length >= 1 ? pathParts[0] : '';
             const deviceHandle = pathParts.length >= 2 ? pathParts[1] : '';
+            const itemTypeHandle = pathParts.length >= 3 ? pathParts[2] : '';
```
masonmcelvain, line 133: This var doesn't get used anywhere, so I don't think we need it

masonmcelvain (changes requested):
CR ✌🏻 Nice work @sumukhipandey! This adds `productList.deviceItemType` on the server, but doesn't yet use it anywhere. We need to add it in the hook (like in this commit https://github.com/iFixit/react-commerce/commit/a5481b5c597546946dd30bb5c24bf05bdfcdae9b) so that the app can use it on the server.

masonmcelvain:
forgot to dev_block ✌🏻 on that

masonmcelvain (dismissed):
CR ✌🏻 I think this is right, nice work!

masonmcelvain:
un_dev_block ✌🏻

erinemay:
dev_block 🌱 
The breadcrumbs got a smidge messed up:
image
https://react-commerce-git-set-item-type-on-server-ifixit.vercel.app/Parts/Mac/Bluetooth_Board

Note: It looks like what you're trying to do worked. There's an intermediate point during page load that shows the item type name is getting loaded instead of "Parts". It just finishes loading and looks like it does in that first image ☝️ 
image

I attached a couple of screen recordings with the speed toggled way down so you can see how the page loads:

<details><summary>Branch</summary>

https://user-images.githubusercontent.com/72166715/201189058-88886c25-71cd-483d-ac67-0423b32d81f6.mp4

</details>

<details><summary>Main</summary>

https://user-images.githubusercontent.com/72166715/201189072-9e1b7697-d8d9-4e15-9d72-96223de03a51.mp4

</details>

sumukhipandey:
> dev_block 🌱 The breadcrumbs got a smidge messed up: image https://react-commerce-git-set-item-type-on-server-ifixit.vercel.app/Parts/Mac/Bluetooth_Board
> 
> Note: It looks like what you're trying to do worked. There's an intermediate point during page load that shows the item type name is getting loaded instead of "Parts". It just finishes loading and looks like it does in that first image ☝️ image
> 
> I attached a couple of screen recordings with the speed toggled way down so you can see how the page loads:
> 
> Branch
> Main

Thanks Erin! I am not sure what's causing the breadcrumbs bug here... @masonmcelvain do you have any idea as to why this might've happened?

erinemay:
I've seen this exact issue pop up on a pull semi-recently. I'm gonna hunt for it.

masonmcelvain:
Thank you @erinemay! I knew we had seen this before. I'm also looking into it.

masonmcelvain:
un_dev_block ✌🏻 Super strange behavior, I can't explain why the breadcrumbs were borked. Regardless, we solved the problem before by not using the SSR hack for the breadcrumbs, so let's reuse that solution. (@danielbeardsley)

k0rvusk0r4x:
- Breadcrumbs are no longer messed up.
- I am not seeing the incorrect loading behavior.

QA 😺

frontend/templates/product-list/sections/FilterableProductsSection/useDevicePartsItemType.tsx
```diff
@@ -17,6 +18,9 @@ export function useDevicePartsItemType<T extends ProductListAttributes>(
       // `Item Type` is a single select, so just use the first value if it exists.
       return itemTypeRefinement?.refinements[0]?.value;
    }, [items]);
+   if (typeof window === 'undefined' && productList.deviceItemType) {
```
danielbeardsley, line 21: I kinds feel like this whole routine is somewhat convoluted. I'm gonna try something.

danielbeardsley: The pushed commit seems to not create the bookmarks weirdness and I feel it's clearer to read.

danielbeardsley:
CR :+1:

danielbeardsley (approved):
CR :+1:

k0rvusk0r4x:
Continues to function the same as [here](https://github.com/iFixit/react-commerce/pull/998#issuecomment-1312030669).

QA 😺

PR #999, opened 2022-11-07
Analytics: Add empty build/clean scripts

jyee27:
This is for compatibility with the `ifixit` monorepo. Rush expects each package we depend on to have these scripts.

See https://github.com/iFixit/react-commerce/pull/633 for the last time we did this.

Connects: 
https://github.com/iFixit/ifixit/pull/45326
https://github.com/ifixit/ifixit/issues/44933

qa_req 0

### DIFF
--- a/packages/analytics/package.json
+++ b/packages/analytics/package.json
@@ -13,6 +13,8 @@
       "typescript": "4.5.3"
    },
    "scripts": {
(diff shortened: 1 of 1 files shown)


sterlinghirsh (approved):
CR :zap: seems weird but ok

PR #1006, opened 2022-11-08
String.replaceAll(): Use more available .replace()

### LINKED ISSUE #1003: TypeError: t.replaceAll is not a function
Sentry Issue: [REACT-COMMERCE-HP](https://sentry.io/organizations/ifixit/issues/3720377548/?referrer=github_integration)

```
TypeError: t.replaceAll is not a function
  at a (/_next/static/chunks/pages/_app-d4f12109fe98edcf.js:1:583768)
  at Ft (/_next/static/chunks/413-8b53305ede261f8e.js:1:51461)
  at oo (/_next/static/chunks/framework-ad8a2cfd097a3be5.js:1:59412)
  at Ku (/_next/static/chunks/framework-ad8a2cfd097a3be5.js:1:111712)
  at Li (/_next/static/chunks/framework-ad8a2cfd097a3be5.js:1:98953)
...
(28 additional frame(s) were not displayed)
```

danielbeardsley:
.replaceAll() isn't available on older browsers (Chrome 84 and below). Quite a fair number of users visit our site on older browsers (like web-TVs?) and report errors on this not being a thing.

Closes #1003

CC @sterlinghirsh

### DIFF
--- a/frontend/helpers/product-list-helpers.ts
+++ b/frontend/helpers/product-list-helpers.ts
@@ -17,7 +17,7 @@ export function computeProductListAlgoliaFilterPreset<
 
    if (filters && filters.length > 0) {
       // Algolia can't handle newlines in the filter, so replace with space.
-      conditions.push(filters.replaceAll('\n', ' '));
+      conditions.push(filters.replace(/(\n|\r)+/g, ' '));
    } else if (deviceTitle && deviceTitle.length > 0) {
       conditions.push(`device:${JSON.stringify(deviceTitle)}`);
    }


jarstelfox (approved):
CR 🌵 I think we should have a discussion around polyfilling.

I noticed some Sentry errors regarding Safari not having intersection observer and found that it gets removed in incognito browsing even though the browser supports it.

Should we start using polyfill.io more? Remember it only downloads what it thinks may be needed

sterlinghirsh (approved):
CR :zap: QA :zap: I tested the regex

sterlinghirsh:
@jarstelfox it's a good discussion topic but I would open a new discussion rather than having it here.

PR #1007, opened 2022-11-08
Revert "refactor(product page):  Use first variant as default variant"

### LINKED ISSUE #1000: Out-of-stock Default Variant: show next in-stock variant
If the default variant is out of stock, show the next variant in line that is *in stock*.

This seems to exactly revert #899  - So make sure we're not just reverting to some behavior we don't want.

CC @ifixitmaxb @jeffsnyder

danielbeardsley:
This reverts commit 262cc4af7157a10a61247533e68cd6c66a977eb8.

Specifically, after ordering by orderby, choose the first in-stock variant to show as the default instead of always showing the default even if it's out of stock.

Closes #1000

CC @ifixitmaxb @jeffsnyder

### DIFF
--- a/frontend/templates/product/hooks/useSelectedVariant.ts
+++ b/frontend/templates/product/hooks/useSelectedVariant.ts
@@ -46,7 +46,11 @@ export function useSelectedVariant(
 }
 
 function useDefaultVariantId(product: Product): string {
-   return product.variants[0].id;
+   const variant =
+      product.variants.find(
+         (variant) => variant.quantityAvailable && variant.quantityAvailable > 0
+      ) ?? product.variants[0];
+   return variant.id;
 }
 
 function useSearchParamVariantId(): string | null {


sterlinghirsh (approved):
CR :zap: Just to be clear, by this point have we already filtered out all the disabled variants?

danielbeardsley:
> Just to be clear, by this point have we already filtered out all the disabled variants?

Good question. I didn't know. but I found that yes we filter them out here: 

https://github.com/iFixit/react-commerce/blob/c2bacfb7faaa984c738a32a3a9df5935d776be9a/frontend/models/product.ts#L38

https://github.com/iFixit/react-commerce/blob/c2bacfb7faaa984c738a32a3a9df5935d776be9a/frontend/models/product.ts#L168-L171

erinemay:
dev_block ❔ Do we want to change the product associated with the canonical url every time a default product is OOS, or just redirect to the next in stock variant?

When IF451-015-2 is OOS:
| Variant | URL |
| - | - |
| IF451-015-2 (default product as set by `orderby` in Akeneo) | https://react-commerce-git-out-of-stock-show-available-variant-ifixit.vercel.app/products/galaxy-note10-replacement-battery?variant=40036258283610 |
| IF451-015-1 | https://react-commerce-git-out-of-stock-show-available-variant-ifixit.vercel.app/products/galaxy-note10-replacement-battery |

When I "re-stock" IF451-015-2 so both variants are in stock:
| Variant | URL |
| - | - |
| IF451-015-2 (default product as set by `orderby` in Akeneo) | https://react-commerce-git-out-of-stock-show-available-variant-ifixit.vercel.app/products/galaxy-note10-replacement-battery |
| IF451-015-1  | https://react-commerce-git-out-of-stock-show-available-variant-ifixit.vercel.app/products/galaxy-note10-replacement-battery?variant=39333889998938 |

danielbeardsley:
> Do we want to change the product associated with the canonical url every time a default product is OOS, or just redirect to the next in stock variant?

Man, I don't know.

I'll put the question to @ifixitmaxb 

When a default option is unavailable, do we:

  * Keep the same url (no `variant=X` param) and display the next available option?
    * Keeps the same canonical, but changes it's content 
  * Pick the next available option and *redirect* the user to that url?
    * Canonical temporarily redirects to url with `?variant=x` param

erinemay:
Talked to @ifixitmaxb, and it isn't super clear cut. 
Concerns:
- If redirect (Google can ignore our suggested canonical if they think we're using it incorrectly or manipulatively)
   - Having current canonical url being removed from the index because google thinks that the new variant is now the page to index
   - might get both indexed
- If we keep the same url, but swap content on the page
   - Google can see this as a different page (unsure of the exact implications)
   - Confusion of merchant id things and structured data

danielbeardsley:
My gut is that keeping the canonical url stable is more important. Temporarily redirecting to a non-canonical url seems the less crime (as opposed to keeping the url, but showing different product data). That said. Default optionids change too and products are phased in / out.

I feel like both options are better than showing someone an out of stock product.

sterlinghirsh:
To be clear, we wouldn't have to do a true redirect, it would be just changing the url to include ?variant=1234. This would probably want to go through our normal variant selector code like `if (showing default variant) selectVariant(first in-stock variant)` 

I think we did something similar on the old product pages where we'd add ?o=1 in many cases, even though the canonical didn't include it. Maybe this was a redirect? Or maybe we just linked it that way most of the time? But it didn't seem to prevent google listings if I'm remembering right.

In any case, I agree both options are better than an out of stock product, so I'm inclined to merge this simple fix and then address it later if we'd rather do something more complicated.

erinemay:
Just for clarity

>I think we did something similar on the old product pages where we'd add ?o=1 in many cases, even though the canonical didn't include it. Maybe this was a redirect? Or maybe we just linked it that way most of the time? But it didn't seem to prevent google listings if I'm remembering right.

It wasn't a redirect. I set the product page feature flag on my subdomain to false while investigating with Max. 
Going from /Store/Android/Galaxy-Note10-Battery/IF451-015 to /Store/Android/Galaxy-Note10-Battery/IF451-015?o=2 isn't a redirect. It just 200ed, and appended that param after the page had almost finished loading. 

<details><summary>Also, unless we made changes to the php logic, we didn't have this behavior on the old pages. If the default variant was OOS, we stayed on that variant.</summary>

https://user-images.githubusercontent.com/72166715/200937487-1b2566f5-56ef-42b5-ae1f-fe70e2b04a5a.mp4

</details>

danielbeardsley:
Ok, pushed a commit that updates the selected variant client-side on page load if the one we were going to show is out of stock, It's *slightly* annoying to have it switch on page load and it would be nice if the out-of-stock item looked out of stock in the chooser, but switching is still better than landing on an out-of-stock item:

image

un_dev_block :+1:

sterlinghirsh:
I think I'd rather have it not flash but still change the url on load. The flash will mess with our lighthouse score and have worse performance, and changing the url has unknown but probably negligible effect on search. I think the only reason we care about it has something to do with the canonical. But if the canonical url's data changes slightly but the url stays the same, that should be fine, right @ifixitmaxb ? I feel like I'm going back and forth but I think having a flash is worse than just leaving the url without the variant id.

ifixitmaxb:
"if the canonical url's data changes slightly but the url stays the same, that should be fine, right...?"
My initial thought is yes, but then I think about the price and other structured data/Schema and also how this might impact Merchant Center data and potentially even SEM ads where the ad copy says it's going to the fix kit, but since it's OOS, it changes to the part only and now the user doesn't get what they thought. I don't know. I might be overthinking but I just have fears. I'd like Snyder to chime in with thoughts on Merchant Center and also SEM implications. @jeffsnyder

sterlinghirsh:
There's no rule against pricing and copy changing right? We can keep the structured data on the default variant or switch it to the displayed in-stock variant.

sterlinghirsh:
After talking with @ifixitmaxb we're going to go with the original solution of keeping the same url and just picking the first available variant. Sorry for the confusion.

danielbeardsley:
Ok, I just removed the most recent commit. I think that should do it.

ifixitmaxb:
For posterity, this article does a good job of articulating why it's sort of risky and not advisable to implement any of the classic options (404, 301, 302) because it's difficult to time Google's crawling: https://moz.com/blog/how-to-handle-temporarily-out-of-stock-product-pages

Additionally, my main concern with the solution we're going with of picking the first available variant, is that I don't know how fast Merchant Center will update on our end and how fast Google will receive the changes in our feed, so it's possible our Shopping ads and also the "free listing" search result type (see screenshot) will have incorrect ad title and prices. 
image
We need to keep an eye out for the following warning/error message for Merchant Listings: "[Mismatched value (page crawl): (price [price])](https://support.google.com/merchants/answer/6098334)". "A common cause for mismatched prices is the time difference between updates on your website and updates of your data in Merchant Center." (I will set an alert for myself to check the [Merchant listings report in GSC](https://search.google.com/search-console/r/merchant-listings?resource_id=https%3A%2F%2Fwww.ifixit.com%2F) next Friday). 
We also need to maybe notify Smarketer of these changes in case they notice any issues in our ad performance, perhaps Google will pause or shut off our ads if our ad information doesn't match what is on the page. I will notify Smarketer now via Slack.

ifixitmaxb:
One more question: for example, the Galaxy Note 10 Battery, will the fix kit variant now "inherit" the exact same variant parameter as what the part only has/had? https://www.ifixit.com/products/galaxy-note10-replacement-battery?variant=39372052824167. Will the fix kit now be at that exact variant URL or will the variant parameter be just slightly different? If slightly different, then will https://www.ifixit.com/products/galaxy-note10-replacement-battery?variant=39372052824167 301 or 302 redirect to the canonical? I hope it doesn't 404. Particularly for our Google Shopping ads and free listing results that might be delayed in being updated. We would have ads and links pointing to new 404s. I hope we can redirect it to the canonical. Or maybe the exact variant parameter is just an exact clean switch and a redirect isn't necessary. @sterlinghirsh

erinemay:
> Ok, I just removed the most recent commit. I think that should do it.

@danielbeardsley How can I trigger a re-deploy? It looks like the preview hasn't updated since you pushed your second commit.
image

erinemay:
> Will the fix kit now be at that exact variant URL or will the variant parameter be just slightly different?

@ifixitmaxb, It will be slightly different.

### More details (AFAIK): 
The number after `?variant=` is a static, unique variant identifier from Shopify. 
For example, the fix kit's variant id is 40036258283610, and the part only's is 39333889998938. These values don't change. You can think of them as hard coded variant id numbers. (NOTE: these are our test store's variant identifiers, so they won't work with the same products in production)

https://react-commerce.vercel.app/products/galaxy-note10-replacement-battery?variant=39333889998938 will always show the part only.
https://react-commerce.vercel.app/products/galaxy-note10-replacement-battery?variant=40036258283610 will always show the kit.

Since the default variant is the one we show for the canonical url, there's just no need ever to append it's unique variant identifier, so we leave it off. The default variant gets no variant param and all the others get their variant params. 

This is only changing which variant we choose to use the canonical for. Every other variant will get `?variant=<that variant's correct unique identifier>`

ifixitmaxb:
Got it. Thanks, Erin. So then any ads or links pointing to the part only variant param url version will still arrive at a 200 status code page and for the correct variant (it's just not the canonical version of the url, which is fine here). No reason or need to redirect.

sterlinghirsh (dismissed):
CR :zap:

danielbeardsley:
I just amended the commit with no changes so that it would be rebuilt.

erinemay:
Ok. QA 👍 Just did a final look around. Everything looks good:
- If none of the variants are OOS, they behave the same as on main
- If a link has a url param attached, the page will always load that specific variant - same as on main
- No redirects involved - the product displayed on the canonical URL is just different if a higher `orderby` is OOS
- If all variants are OOS, it reverts to original behavior where the first `orderby` is shown for the canonical URL
- Checked with products that have 1, 2, and 4 variants. Behavior stayed uniform.

But deploy_block 🧷 on making sure we've concluded this discussion

>"if the canonical url's data changes slightly but the url stays the same, that should be fine, right...?"
My initial thought is yes, but then I think about the price and other structured data/Schema and also how this might impact Merchant Center data and potentially even SEM ads where the ad copy says it's going to the fix kit, but since it's OOS, it changes to the part only and now the user doesn't get what they thought. I don't know. I might be overthinking but I just have fears. I'd like Snyder to chime in with thoughts on Merchant Center and also SEM implications. @jeffsnyder

sterlinghirsh:
un_deploy_block ⚡ yeah we're good with it

sterlinghirsh (approved):
CR :zap:

Issue #855, opened 2022-10-20
Matomo track product page activity pt 2


erinemay:
I think https://github.com/iFixit/react-commerce/pull/802 was merged a little prematurely. 

AFAICT, the following aren't registering in Matomo, or at least the way I would expect. Events listed have screenshots of the same event on cominor and how it's recorded there.

<details><summary>Add to cart & Mini cart clicks</summary>

![image](https://user-images.githubusercontent.com/72166715/197071464-044817d8-57d8-4642-b097-642e3730f39d.png)

Clicking on View cart does register, but not as a mini-cart click. It just looks like a link click?
![image](https://user-images.githubusercontent.com/72166715/197072541-77fe9724-916f-4080-a599-ae62876c9ded.png)

</details>

<details><summary>Add to cart Frequently Bought Together</summary>

![image](https://user-images.githubusercontent.com/72166715/197071230-a74a3c65-868e-44c0-ae46-9355948c4896.png)

</details>

<details><summary>Featured Products click</summary>

![image](https://user-images.githubusercontent.com/72166715/197071320-b49cd12e-22fd-48fd-9b3d-f5ef488b75b4.png)

</details>

<details><summary>Newsletter footer submit</summary>

![image](https://user-images.githubusercontent.com/72166715/197071394-c440fe33-5695-49cd-a7cd-dc7190bf3b49.png)

</details>

<details><summary>Footer link click (is registering somewhat, but not quite the same as on cominor)</summary>

On cominor, it's getting registered as its own event.
![image](https://user-images.githubusercontent.com/72166715/197072084-efaf47db-a64b-4279-9f2e-15ffd55f905e.png)

On the react pages, it's registering as a sub-event(?)
![image](https://user-images.githubusercontent.com/72166715/197072266-2c8d75e5-ed41-4844-a102-a82edb3559bf.png)

</details>

From what I can tell, these are registering as expected:
- Warranty clicks
- Abandoned cart

Testing details:
<details><summary>Visit 6 on this user profile has the details that were recorded when I performed each of the above actions on the main branch</summary>

https://matomo.ubreakit.com/index.php?date=today&module=Widgetize&action=iframe&visitorId=dd9162939bce75c8&idSite=1&period=day&moduleToWidgetize=Live&actionToWidgetize=getVisitorProfilePopup
![image](https://user-images.githubusercontent.com/72166715/197072424-5a7f8a40-f807-484f-b1a1-be9928e8e1b1.png)

</details>

[...]

PR #1009, opened 2022-11-08
Footer: Move click tracking to context

andyg0808:
We don't want to use the same click tracking function in every location we use the footer. In particular, we don't want to use `@ifixit/analytics` from anywhere but this repo. This pull switches the footer component to fetch its click tracking function from a context, which enables us to swap it out per-use. It also adds that new context around the use of the `Footer` component in this repo, so we continue to use the same click tracking function within this repo.

## QA Suggestion
Please confirm that the footer click tracking seems to still be working as described in https://github.com/iFixit/react-commerce/issues/855 (only the footer parts; the rest should not have been affected).

CC @jyee27 @jordycosta

### DIFF
--- a/packages/footer/hooks/TrackingContext.ts
+++ b/packages/footer/hooks/TrackingContext.ts
@@ -1,0 +1,41 @@
+import { createContext, useContext } from 'react';
+
+/**
+ * @see https://matomo.org/docs/event-tracking/
+ * @see https://developer.matomo.org/api-reference/tracking-javascript
+ */
+export type TrackEventMatomo = {
+   /**
+    * Describes the type of events you want to track.
+    * For example, Link Clicks, Videos, Outbound Links, and Form Events.
+    */
+   eventCategory: string;
+   /**
+    * The specific action that is taken.
+    * For example, with a Video category, you might have a Play, Pause and Complete action.
+    */
+   eventAction: string;
+   /**
+    * Usually the title of the element that is being interacted with, to aid with analysis.
+    * For example, it could be the name of a Video that was played or the specific
+    * form that is being submitted.
+    */
+   eventName?: string;
+   /**
+    * A numeric value and is often added dynamically. It could be the cost of a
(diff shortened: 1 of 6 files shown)


jyee27:
CR ☕ looks like `pnpm-lock.yaml` needs to be updated

packages/footer/hooks/TrackingContext.ts
```diff
@@ -0,0 +1,41 @@
...
+type TrackContext = {
+   trackClick: Track;
+};
+
+const defaultTracker: TrackContext = {
+   trackClick: (_trackData: TrackEventMatomo) => {},
+};
+
+export const TrackingContext = createContext(defaultTracker);
+export const ClickTracker = TrackingContext.Provider;
```
masonmcelvain, line 41: Minor naming question:

I'm curious, do we ever anticipate tracking events that are not the result of a click? Like perhaps a specific kind of page view, or a keystroke? Perhaps the names `trackEvent` or `EventTracker` are a little ambiguous though.

masonmcelvain (approved):
CR ✌🏻 Nice work! Minor question about naming, but I don't feel strongly about it. Happy to merge as is.

packages/footer/hooks/TrackingContext.ts
jyee27: Renaming this to `EventTracker` sounds reasonable. The `useTrackedOnClick` hook is already named appropriately, and I can see this being used for other events.

jyee27: [bb94039](https://github.com/iFixit/react-commerce/pull/1009/commits/bb94039321b97f744de0d5a8c3f64b4ac4ddf612)

masonmcelvain (approved):
CR ✌🏻 Thanks!

deltuh-vee:
QA 🎬 
Footer clicks are still registered in Matomo.
image

PR #1019, opened 2022-11-10
Algolia: Bump connect timeout to 5 seconds

### LINKED ISSUE #1016: Try increasing Algolia timeouts to avoid unreachable hosts errors
Docs: https://www.algolia.com/doc/api-reference/api-methods/configuring-timeouts/?client=javascript
Slack ref: https://ifixit.slack.com/archives/C6QL70G9Z/p1667227831078949

Algolia's suggestion is to increase out timeouts. The default connectTimeout in js is 2 seconds. Let's try 5 and see if that fixes it. I'm also curious if we're auto-retrying if it does timeout after 2 seconds, which would mean that we wouldn't want to go too long with the timeout, but 5 seconds is probably fine.

After deploying, check sentry and see if the frequency of these errors has decreased.

masonmcelvain:
This is an attempt to increase Algolia timeouts to avoid unreachable hosts errors.

## QA

Make sure collection pages work normally.

CC @sterlinghirsh @danielbeardsley

Closes #1016

### DIFF
--- a/frontend/components/common/InstantSearchProvider.tsx
+++ b/frontend/components/common/InstantSearchProvider.tsx
@@ -1,5 +1,6 @@
 import { useSafeLayoutEffect } from '@chakra-ui/react';
 import { ALGOLIA_APP_ID } from '@config/env';
+import { CLIENT_OPTIONS } from '@helpers/algolia-helpers';
 import {
    destylizeDeviceItemType,
    stylizeDeviceItemType,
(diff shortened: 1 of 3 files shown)


sterlinghirsh (approved):
CR :zap:

danielbeardsley:
CR :+1:

masonmcelvain:
QA ✌🏻 Collection pages aren't broken

PR #1025, opened 2022-11-14
CORS: Use relative urls in dev

### LINKED ISSUE #1022: CORS error on subdomain, not on www
My subdomain is on the latest master, and I can't view or add to my cart.

https://www.cominor.com/Parts
image

https://me.cominor.com/Parts
image

masonmcelvain:
When we're proxying our Next app on cominor, use relative urls for external links and api requests. This will prevent CORS errors in dev, and keep developers on the same domain when clicking on links.

The only place we are still using a hard coded origin is in frontend/pages/_document.tsx, since I don't think we have access to the request context in that file.

If the app is being visited at *.vercel.app or ifixit.com, then the IFIXIT_ORIGIN value will be used as before, so there should be no change in behavior.

## QA

I'm not sure how to reproduce the original problem with localhost or the previews.

However, let's be sure we're still using `IFIXIT_ORIGIN` on *.vercel.app.

Closes #1022

### DIFF
--- a/frontend/helpers/path-helpers.ts
+++ b/frontend/helpers/path-helpers.ts
@@ -1,3 +1,4 @@
+import { IFIXIT_ORIGIN } from '@config/env';
 import { invariant } from '@ifixit/helpers';
 import {
    ProductList,
@@ -45,4 +46,8 @@ export function productListPath(
          throw new Error(`unknown product list type: ${productList.type}`);
       }
    }
+}
+
+export function ifixitOriginFromHost(host: string): string {
+   return host.match(/(?:\w*\.)*cominor\.com/) ? '' : IFIXIT_ORIGIN;
 }
--- a/frontend/helpers/path-helpers.ts
+++ b/frontend/helpers/path-helpers.ts
@@ -55,7 +55,9 @@ export function ifixitOriginFromHost(
    const headers = context.req.headers;
    const host = (headers['x-ifixit-forwarded-host'] ||
       headers['x-forwarded-host']) as string | undefined;
-   const isDevProxy = !!host?.match(/\.(cominor\.com|ubreakit\.com)$/);
+   const isDevProxy = !!host?.match(
+      /^(?!react-commerce).*(\.cominor\.com|\.ubreakit\.com)$/
+   );
    const hostIfProxy = typeof window === 'undefined' ? `https://${host}` : '';
    return isDevProxy ? hostIfProxy! : IFIXIT_ORIGIN;
 }
--- a/frontend/helpers/path-helpers.ts
+++ b/frontend/helpers/path-helpers.ts
@@ -50,6 +50,6 @@ export function productListPath(
 
 export function ifixitOriginFromHost(host?: string): string {
    const isDevProxy = !!host?.match(/\.(cominor\.com|ubreakit\.com)$/);
-   const hostIfProxy = typeof window === 'undefined' ? host : '';
+   const hostIfProxy = typeof window === 'undefined' ? `https://${host}` : '';
    return isDevProxy ? hostIfProxy! : IFIXIT_ORIGIN;
 }
--- a/frontend/helpers/path-helpers.ts
+++ b/frontend/helpers/path-helpers.ts
@@ -48,6 +48,6 @@ export function productListPath(
    }
 }
 
-export function ifixitOriginFromHost(host: string): string {
-   return host.match(/(?:\w*\.)*cominor\.com/) ? '' : IFIXIT_ORIGIN;
+export function ifixitOriginFromHost(host?: string): string {
+   return host?.match(/\.(cominor\.com|ubreakit\.com)$/) ? '' : IFIXIT_ORIGIN;
 }
--- a/frontend/helpers/path-helpers.ts
+++ b/frontend/helpers/path-helpers.ts
@@ -49,5 +49,7 @@ export function productListPath(
 }
 
 export function ifixitOriginFromHost(host?: string): string {
-   return host?.match(/\.(cominor\.com|ubreakit\.com)$/) ? '' : IFIXIT_ORIGIN;
+   const isDevProxy = !!host?.match(/\.(cominor\.com|ubreakit\.com)$/);
+   const hostIfProxy = typeof window === 'undefined' ? host : '';
+   return isDevProxy ? hostIfProxy! : IFIXIT_ORIGIN;
 }
--- a/frontend/helpers/path-helpers.ts
+++ b/frontend/helpers/path-helpers.ts
@@ -6,6 +6,7 @@ import {
    iFixitPageType,
    iFixitPage,
 } from '@models/product-list';
+import { GetServerSidePropsContext } from 'next';
 import { stylizeDeviceTitle } from './product-list-helpers';
 
 type ProductListPathAttributes = Pick<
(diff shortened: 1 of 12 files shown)


masonmcelvain:
dev_block ✌🏻 on trying it out

masonmcelvain:
un_dev_block ✌🏻

frontend/templates/product-list/index.tsx
```diff
@@ -79,10 +80,12 @@ export const getProductListServerSideProps = ({
       let shouldRedirectToCanonical = false;
       let canonicalPath: string | null = null;
 
+      const ifixitOrigin = ifixitOriginFromHost(context.req.headers.host ?? '');
```
danielbeardsley, line 83: I think this will always be `*.vercel.app` because the host header *has* to be set to one of our app's domains so vercel knows which app to send the request to. However, the original host (that the browser requested) is often available on a different header

frontend/models/product-list/index.ts
```diff
@@ -412,8 +414,6 @@ function createPublicAlgoliaKey(appId: string, apiKey: string): string {
    return publicKey;
 }
 
-function createIFixitAPIClient() {
-   return new IFixitAPIClient({
-      origin: IFIXIT_ORIGIN,
-   });
+function createIFixitAPIClient(origin: string) {
```
danielbeardsley, line 417: This function is now kinda not worth having.

frontend/helpers/path-helpers.ts
```diff
@@ -46,3 +47,7 @@ export function productListPath(
       }
    }
 }
+
+export function ifixitOriginFromHost(host: string): string {
+   return host.match(/(?:\w*\.)*cominor\.com/) ? '' : IFIXIT_ORIGIN;
```
danielbeardsley, line 52: I'm not sure what's being attempted here with the first part of this regex:

`(?:\w*\.)*`  The star there indicates this part is effectively optional and since we don't capture it, this part of the pattern seems unused.

I might test a suffix: `/\.(cominor\.com|ubreakit\.com)$/` instead because we know subdomains can be several levels deep.

danielbeardsley:
I'm excited about this change! It'll make testing a lot easier and more intuitive.

I was fairly partial to the suggestion by @sterlinghirsh... that unfortunately was only sent to me:
image

masonmcelvain:
Thanks for the review! I like your suggestions.

> I don't see how the currently solution ends up talking to the forwarded host on the backend

When you say "backend" do you mean the server? Otherwise I'm not sure what you're referring to.

frontend/templates/product-list/index.tsx
masonmcelvain: Cool I'll check the forwarded host like you suggest below.

frontend/helpers/path-helpers.ts
masonmcelvain: Yea the first part does seem kind of useless. I'll use your regex

masonmcelvain:
Oh I think I see. Relative urls only work in the browser, but right now the code uses them on the server too.

frontend/models/product-list/index.ts
masonmcelvain: https://github.com/iFixit/react-commerce/pull/1025/commits/56cd5fcf4723cc7bafa50503d292cd568c53929d

masonmcelvain:
dev_block ✌🏻

masonmcelvain:
These were the only url-like fields in `context.req.headers` when I visited https://react-commerce-mvx2f9oq7-ifixit.cominor.com/Parts:
```json
{
"host": "react-commerce-mvx2f9oq7-ifixit.vercel.app",
"x-forwarded-host": "react-commerce-mvx2f9oq7-ifixit.vercel.app",
"x-vercel-deployment-url": "react-commerce-mvx2f9oq7-ifixit.vercel.app",
}
```
Maybe it's different in prod cuz we're using cloudfront, but we're not logging the headers in prod and I don't know how else to check them.

danielbeardsley:
> `"x-forwarded-host"`

That's the one. If it looks like our dev server, use it as the origin, otherwise, fall back to the env var. That's set if the app is behind a proxy (or at least it *should* be set).

masonmcelvain:
Thanks for the help @danielbeardsley! un_dev_block ✌🏻 I'm still not sure if we can QA the feature in dev, but we can at least be sure we're still using `IFIXIT_ORIGIN` when indicated.

danielbeardsley:
CR :+1: 

This looks good to me! But I'm having trouble QAing it... Rather it doesn't seem to work. I've also validated that it doesn't seem to issue a request to my domain.

image

dev_block :+1: 

We might need to also tell vercel not to cache these responses in dev? i.e. an ENV var that tells our app to set a response header indicating that this shouldn't be cached.

masonmcelvain:
Hm it works locally, can't figure out why it's not working on the preview...

image

> We might need to also tell vercel not to cache these responses in dev? i.e. an ENV var that tells our app to set a response header indicating that this shouldn't be cached.

Good idea. What if we compared the detected host against `www.cominor.com`, and don't cache it if it doesn't match? I tried that in https://github.com/iFixit/react-commerce/pull/1025/commits/d08c4ac274fbdc2030330dfb17e266781a5c5a51

danielbeardsley:
> can't figure out why it's not working on the preview...

Maybe let's dump the headers again. I bet the headers aren't making it into vercel (maybe vercel is replacing this header). If this doesn't work, we can try setting a custom header in our proxying layer. (`X-ifixit-forwarded-host`)

masonmcelvain:
Yea I'm still seeing 
image
for 
image

danielbeardsley:
OK, I'll make a pull to add `X-ifixit-forwarded-host` to our proxying layer.

frontend/helpers/path-helpers.ts
```diff
@@ -46,3 +47,9 @@ export function productListPath(
       }
    }
 }
+
+export function ifixitOriginFromHost(host?: string): string {
```
danielbeardsley, line 51: Let's have this take in a `context` or a `request` have it do:

`const host = headers['x-ifixit-forwarded-host`] || headers['x-forwarded-host']`

Does that seem reasonable?

masonmcelvain: Yea seems like an improvement, so the caller doesn't have to drill down in to the context

masonmcelvain:
image

image

Woohoo!

frontend/helpers/path-helpers.ts
masonmcelvain: https://github.com/iFixit/react-commerce/pull/1025/commits/6cb45de738a3463eb273e19b39e725017a6bd274

masonmcelvain:
un_dev_block ✌🏻

danielbeardsley:
CR :+1:

danielbeardsley (approved):
CR :+1:

danielbeardsley:
You likely already QAed this yourself, right @masonmcelvain ?  Borwsing around the prod preview deployment seems to work fine.

masonmcelvain:
> You likely already QAed this yourself

Yep!

QA ✌🏻 

react-commerce*.cominor.com makes requests to www.cominor.com

image

We use the forwarded host when we're supposed to: https://github.com/iFixit/react-commerce/pull/1025#issuecomment-1317619178

masonmcelvain:
Tried to update the submodule in https://github.com/iFixit/ifixit/pull/45577 but it seems like we're still trying to hit www.cominor.com on subdomains 🤔

PR #1036, opened 2022-11-17
Optimize shopify images

### LINKED ISSUE #967: Product Page loads full-res image
We are loading a full-res 2000x2000 image, but displaying it much smaller. It says it's showing at 470x472, which is odd that it's not quite square. 
image

https://www.ifixit.com/products/ifixit-opening-picks-set-of-6

We should take advantage of Shopify's image resizing to load a smaller image. The full image is https://cdn.shopify.com/s/files/1/0045/4092/4007/products/fDsOubGiL2QUMeOR_b6bb1238-5e21-4860-9a7a-47bc002c7745.jpg?v=1667237563, but adding _470x to the url before the file extension like https://cdn.shopify.com/s/files/1/0045/4092/4007/products/fDsOubGiL2QUMeOR_b6bb1238-5e21-4860-9a7a-47bc002c7745_470x.jpg?v=1667237563 returns a much smaller image. (7.9kB vs 51.1kB). Thumbnails using _65x are even smaller (2.4kB).

It may make sense to use the small images for the initial load but preload the larger sizes. We may also want to include a srcset attr here to support high dpi screens.

federicobadini:
closes #967 

### QA

1. Visit Vercel preview
2. Verify that images are appropriately scaled (take also into account that images may be upscaled wrt expectation depending on the device pixel ratio)


sterlinghirsh:
CR :zap: I know this is a draft but it looks reasonable.

deltuh-vee:
QA 🎬 
* Images are loading smaller images with the `width` parameter set
* `width` parameter is set depending on the size of the window.

federicobadini:
Is this ready to be merged? 
It would be nice to start evaluating the performance improvement in production

sterlinghirsh (approved):
CR :zap:

PR #1038, opened 2022-11-17
debug: log forwarded host headers

masonmcelvain:
We tried to get around CORS errors in https://github.com/iFixit/react-commerce/pull/1025 by using the detected host for API requests when the app is proxied on cominor.

However, in https://github.com/iFixit/ifixit/pull/45577 it seems like the detected host isn't being used.

This logs the relevant headers so we can debug the problem.

qa_req 0

CC @danielbeardsley


danielbeardsley (approved):
CR :+1:

PR #1040, opened 2022-11-17
Reset page to zero when deselecting facets

### LINKED ISSUE #1034: Collections: Reset `page` when deselecting facets
## Problem

~~When a multi select is selected/deselected, the `?page=` param should disappear from the url and the view should reset to the first page.~~

It also seems that deselecting a single select using the X in the blue boxes doesn't reset the page param.

[Screencast from 11-16-2022 02:33:13 PM.webm](https://user-images.githubusercontent.com/52104630/202276808-375fc75b-ab0f-4aaa-83ab-e943647a6a41.webm)

## Background 

#1010

CC @sterlinghirsh @danielbeardsley @rickisXP

rickisXP:
Closes #1034 

Previously, when removing facet filters using the pill's `x` button, the page number would not reset to zero. This pulls fixes that by setting the page to `NaN` when a facet filter is removed using the `x` button.
<details>

image
</details>

QA
--

1. Visit the preview for this pull
2. Add filters to narrow your search
3. Go to any page other than the first page
4. Remove the filter by using the `X` button next to the filter.
5. Observe that the page is now on the first page
6. Check that the same behavior is on Tools

### DIFF
--- a/frontend/templates/product-list/sections/FilterableProductsSection/CurrentRefinements.tsx
+++ b/frontend/templates/product-list/sections/FilterableProductsSection/CurrentRefinements.tsx
@@ -49,12 +49,7 @@ export function CurrentRefinements() {
                                  aria-label={`Remove ${formattedFacetName}: ${refinement.label}`}
                                  onClick={() => {
                                     item.refine(refinement);
-                                    setIndexUiState((uiState) => {
-                                       return {
-                                          ...uiState,
-                                          page: NaN,
-                                       };
-                                    });
+                                    pagination.refine(0);
                                  }}
                               />
                            </Tag>
--- a/frontend/templates/product-list/sections/FilterableProductsSection/CurrentRefinements.tsx
+++ b/frontend/templates/product-list/sections/FilterableProductsSection/CurrentRefinements.tsx
@@ -11,11 +11,13 @@ import * as React from 'react';
 import {
    useClearRefinements,
    useCurrentRefinements,
+   useInstantSearch,
 } from 'react-instantsearch-hooks-web';
 
 export function CurrentRefinements() {
    const currentRefinements = useCurrentRefinements();
    const clearRefinements = useClearRefinements();
+   const { setIndexUiState } = useInstantSearch();
 
    return (
       <Collapse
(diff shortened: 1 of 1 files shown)


deltuh-vee:
QA 🎬 
Removing facet filters using the pill's x button, resets the page number to zero.

frontend/templates/product-list/sections/FilterableProductsSection/CurrentRefinements.tsx
```diff
@@ -11,11 +11,13 @@ import * as React from 'react';
 import {
    useClearRefinements,
    useCurrentRefinements,
+   useInstantSearch,
 } from 'react-instantsearch-hooks-web';
 
 export function CurrentRefinements() {
    const currentRefinements = useCurrentRefinements();
    const clearRefinements = useClearRefinements();
+   const { setIndexUiState } = useInstantSearch();
```
masonmcelvain, line 20: This works, but I think we should use [`usePagination`](https://www.algolia.com/doc/api-reference/widgets/pagination/react-hooks/#hook) instead, as it is intended for setting the current page.

```suggestion
   const pagination = usePagination();
```

frontend/templates/product-list/sections/FilterableProductsSection/CurrentRefinements.tsx
```diff
@@ -11,11 +11,13 @@ import * as React from 'react';
 import {
    useClearRefinements,
    useCurrentRefinements,
+   useInstantSearch,
```
masonmcelvain, line 14:
```suggestion
   usePagination,
```

masonmcelvain (changes requested):
CR ✌🏻 Nice job @rickisXP! dev_block ✌🏻 on using a slightly more specific hook

frontend/templates/product-list/sections/FilterableProductsSection/CurrentRefinements.tsx
dhmacs: ⚠️ Yeah let's not use `useInstantSearch`, it has been the source of memory leaks and many issues. We should avoid it, and generally speaking it shouldn't be needed as it's a relatively low level hook.

masonmcelvain (approved):
CR ✌🏻 un_dev_block ✌🏻 Thanks!

deltuh-vee:
Carry over QA 🎬

Issue #1020, opened 2022-11-10
Community page no header


sterlinghirsh:
## Problem
Visiting https://www.ifixit.com/Community directly works as expected, but is served from our php app. However going to a nextjs page like https://www.ifixit.com/Parts and then clicking the link to /Community (Header > Community > Get Involved) causes the page to load without header or footer.

## Do This
Always use the header/footer. Let's pick whether we're going to serve this page from php or nextjs and then fix the routing in CloudFront / haproxy and remove the other version to avoid confusion.

andyg0808:
@djmetzle suggested using a new name for the page. I think that's a good place to start thinking about this: can we somehow find a way to go about adding new pages which allows us to avoid causing this problem and gives us a straightfoward way to convert new pages to Next.

danielbeardsley:
I think this is happening because `/Community` is a page that next.js serves (though it is incomplete). So if you navigate directly, you get cloudfront and that sends you to the php app. If you click a link, next.js realizes that's a URL that it knows about and makes an API request to itself to get the props / JS and renders the page without directly requesting the url.

Effectively, we're seeing a not-ready-yet version of the community page from next-js. We've run into this problem probably 3 or 4 times with different urls. I'm not sure there's much we can do about it other than to temporarily disable that next.js page (by renaming the `Community` directory?)

djmetzle:
That's correct. One nextjs is booted, we will route internal to the app. If we want to continue serving the community page from PHP, we should probably use an external link (`a` tag, not next/link). Ref: https://upmostly.com/nextjs/how-to-link-to-an-external-page-in-nextjs

dhmacs:
@sterlinghirsh was considering to use lowercase urls (just like we did for the new product page), so maybe using lowercase paths (e.g. `/community`) for the next.js page might be a viable option.

danielbeardsley:
> If we can use something like CommunityNew or something else to clearly distinguish it

I'd imagine this would jut be temporary until we're ready to actually replace the page.

aburke07:
It sounds like we have two options:
1. Rename the `Community` directory to `CommunityNew` or `CommunityNext` or something else.
2. Serve the link in this directory using an external link `<a>` tag over the nextjs link.

Which should we do?

andyg0808:
Strong preference for option 1. I'd probably use `CommunityNext`, personally.

djmetzle:
Second for option #1

PR #1047, opened 2022-11-18
Rename /Community to /CommunityNext

aburke07:
## Overview
Currently, our nextjs store routes links to ifixit.com/Community to the nextjs version of /Community, which is not ready yet and does not show the header or footer. 

Simply renaming this directory in react-commerce will force nextjs to use an external link rather than finding a path in the repo.

## QA
- On a nextjs product page ([you can use this one!](http://localhost:3000/products/2-tb-ssd-hybrid-3-5-hard-drive)), use the header to navigate to the community page (header > Community > Get Involved)
- Check that the page is loaded with the header and footer

Closes: #1020


sctice-ifixit (approved):
CR :flags:

deltuh-vee:
QA 🎬 
Navigating to the Community Page from a nextjs product page includes the header and footer.

PR #1048, opened 2022-11-18
Disable caching for `react-commerce`

masonmcelvain:
Turns off caching when the `CACHE_DISABLED` environment variable is set.

### Motivation

@danielbeardsley and I believe that caching in dev can lead to inconsistent results when QA'ing, and thus outweighs the benefits of snappy responses.

## QA

Make sure caching is turned on react-commerce, but not react-commerce-prod.

Connects #1022

### DIFF
--- a/frontend/helpers/next-helpers.ts
+++ b/frontend/helpers/next-helpers.ts
@@ -20,11 +20,6 @@ export function serverSidePropsWrapper<T extends { [key: string]: any }>(
          ...context.params,
          ...context.query,
       });
-      console.log({
-         'x-forwarded-host': context.req.headers['x-forwarded-host'],
-         'x-ifixit-forwarded-host':
-            context.req.headers['x-ifixit-forwarded-host'],
-      });
       return logAsync('getServerSideProps', () =>
          getServerSidePropsInternal(context)
       )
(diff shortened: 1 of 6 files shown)


masonmcelvain:
✔️ Added the env var on react-commerce

image

frontend/helpers/next-helpers.ts
```diff
@@ -18,17 +20,23 @@ export function serverSidePropsWrapper<T extends { [key: string]: any }>(
...
-         throw err;
-      });
+      )
+         .then((result) => {
+            if (CACHE_DISABLED || context.query._vercel_no_cache === '1') {
+               context.res.setHeader(
+                  'Cache-Control',
+                  'no-store, no-cache, must-revalidate, stale-if-error=0'
+               );
+               clearCache();
```
danielbeardsley, line 32: We'll actually want to do this line *before* we call the `getServerSidePropsInternal` otherwise, we've already used the cache by this point. That said... seems a bummer that we're not caching strapi in dev at all. I don't remember the motivation for that... do you?

danielbeardsley (changes requested):
CR :+1: 

dev_block :+1:  on moving a line I think.

frontend/helpers/next-helpers.ts
masonmcelvain: This is where it was added I think: https://github.com/iFixit/react-commerce/pull/659 @danielcliu-ifixit

masonmcelvain: I ended up clearing the cache before and after. Let me know if you think that is incorrect, I wasn't sure.

masonmcelvain:
un_dev_block ✌🏻

danielbeardsley (approved):
CR :+1:

masonmcelvain:
QA ✌🏻 

image

image

Issue #1043, opened 2022-11-18
Product Page: Add tests for Product title, price and title


mlahargou:
- [ ] Make sure the product title, item code, and price are visible at all times.
   - [x] Login as pro user, and check that the price matches the pro user's tier.

PR #1049, opened 2022-11-18
Pro users: Add a test

mlahargou:
This verifies that pro users get a discount on products. We are using
the repair business toolkit and mocking the logged in user to verify
that they get the discount.

Because we are not mocking the product itself, I didn't want to get
too specific with prices. Instead, we just verify that the price
for a regular user is *higher* than it is for a pro user.

Connects: #1043 

cc @batbattur

### DIFF
--- a/frontend/test/cypress/integration/product/pro_user.cy.ts
+++ b/frontend/test/cypress/integration/product/pro_user.cy.ts
@@ -1,0 +1,31 @@
+describe('Pro user test', () => {
+   it('will give pro users a discount', () => {
+      cy.loadProductPageByPath('/products/repair-business-toolkit');
+
+      // Get price from page
+      cy.findAllByTestId('product-price')
+         .first()
+         .invoke('text')
+         .then((originalPriceString) => {
+            // Login as pro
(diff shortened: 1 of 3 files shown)


batbattur (approved):
CR 🚛

frontend/test/cypress/integration/product/pro_user.cy.ts
```diff
@@ -0,0 +1,31 @@
+describe('Pro user test', () => {
+   it('will give pro users a discount', () => {
+      cy.loadProductPageByPath('/products/repair-business-toolkit');
+
+      // Get price from page
+      cy.findAllByTestId('product-price')
+         .first()
```
ardelato, line 7: Is there a possibility of the price for a recommended product being added to the list first? In the case, the actual product price does not render or is slower at rendering?

ardelato (approved):
CR 👍 with question

jordycosta:
QA 🟢 

https://github.com/iFixit/react-commerce/actions/runs/3500010244/jobs/5862219029

test

PR #1050, opened 2022-11-18
Tests: Check for visible product title, price and sku

batbattur:
## Summary
We recently added [pro user product page test](https://github.com/iFixit/react-commerce/pull/1049) which compares the product prices.

This pull adds additional assertions to check product title, sku and price. Added assertions 
in `pro_user` test since we already visit product page (instead of creating a new one). 

Also refactored `interceptLogin()` command to accept additional/optional params, 
instead of only passing the argument for `discount_tier`.

## QA 
Make sure the Cypress tests are passing in CI. 

cc: @mlahargou 
Closes: https://github.com/iFixit/react-commerce/issues/1043

### DIFF
--- a/frontend/test/cypress/integration/product/pro_user.cy.ts
+++ b/frontend/test/cypress/integration/product/pro_user.cy.ts
@@ -10,6 +10,14 @@ describe('Pro user test', () => {
             expect(productTitle).to.equal('Repair Business Toolkit');
          });
 
+      // Assert product sku is visible
+      cy.findByTestId('product-sku')
+         .should('be.visible')
+         .invoke('text')
+         .then((productSku) => {
+            expect(productSku).to.match(/IF\d*\-\d*/g);
+         });
+
       // Get price from page
       cy.findAllByTestId('product-price')
          .first()
--- a/frontend/test/cypress/integration/product/pro_user.cy.ts
+++ b/frontend/test/cypress/integration/product/pro_user.cy.ts
@@ -2,31 +2,11 @@ describe('Pro user test', () => {
    it('will give pro users a discount', () => {
       cy.loadProductPageByPath('/products/repair-business-toolkit');
 
-      // Assert product title is visible
-      cy.findByTestId('product-title')
-         .should('be.visible')
-         .invoke('text')
-         .then((productTitle) => {
-            expect(productTitle).to.equal('Repair Business Toolkit');
-         });
-
-      // Assert product sku is visible
-      cy.findByTestId('product-sku')
-         .should('be.visible')
-         .invoke('text')
-         .then((productSku) => {
-            expect(productSku).to.match(/IF\d*\-\d*/g);
-         });
-
       // Get price from page
       cy.findAllByTestId('product-price')
          .first()
-         .should('be.visible')
          .invoke('text')
          .then((originalPriceString) => {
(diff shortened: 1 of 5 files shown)


frontend/test/cypress/integration/product/pro_user.cy.ts
```diff
@@ -2,13 +2,33 @@ describe('Pro user test', () => {
...
+            expect(productTitle).to.equal('Repair Business Toolkit');
+         });
+
+      // Assert product sku is visible
+      cy.findByTestId('product-sku')
+         .should('be.visible')
+         .invoke('text')
+         .then((productSku) => {
+            expect(productSku).to.match(/IF\d*\-\d*/g);
+         });
```
mlahargou, line 19: This doesn't really have to do with a pro user (the name of this file). How about we pull this out into a `product_page.cy.ts` file?

frontend/test/cypress/support/commands.ts
```diff
@@ -38,15 +38,19 @@ Cypress.Commands.add('loadProductPageByPath', (path: string) => {
...
    cy.intercept(
       { method: 'GET', url: '/api/2.0/user' },
       {
-         userid: 1,
-         algoliaApiKeyProduct: null,
-         username: 'john',
-         unique_username: 'john123',
-         discount_tier: discount_tier,
+         ...defaultParams,
+         ...optional,
```
mlahargou, line 53: Will the optional params override the default ones?

mlahargou (dismissed):
CR 📱 with a couple optional suggestions.

frontend/test/cypress/support/commands.ts
batbattur: Yup, looks like the optional params will override the default:

Screen Shot 2022-11-18 at 3 28 52 PM

frontend/test/cypress/integration/product/pro_user.cy.ts
batbattur: We can create a new test, but it will create additional few seconds to load up the same product page. 
I added it in this test since we already visit a product page so we can save up the time for tests to run.

batbattur: I suppose it's only `2-4` seconds, [ref](https://github.com/iFixit/react-commerce/actions/runs/3500832448/jobs/5863890797#step:6:205).

mlahargou: Yeah, I guess it doesn't seem so bad right now, but if we start adding a ton of tests then I think we will want separate logic tested in separate places.

This is an extreme example, but if I wanted to check if campaigns were tested but the test code was in a file named `answersTest`, I wouldn't think campaigns were tested. Let's split this up even if it adds a bit of time.

batbattur: Sounds good! Split them in https://github.com/iFixit/react-commerce/pull/1050/commits/5d3fd0af93406ec51159c34a033b10c77cd39b8d.

mlahargou: Thank you! Looks great!

mlahargou (approved):
CR 📱

deltuh-vee:
QA 🎬 
Cypress tests are still passing in CI.

PR #1051, opened 2022-11-19
Tests: Add tests for Fix kit / Part Only selection

### LINKED ISSUE #1045: Product Page: Add tests for Fix kit / Part Only
- [ ] If `Fix kit` is selected, the kit contents section should display.  
   - [ ] If `Part only is selected`, the kit contents section should not display. 
   - [ ] Verify that the product image, price, sku changes when switching between `Fix kit` / `Part Only`.

batbattur:
## Summary
Added test for `Fix Kit`/`Part Only` product selection which tests the following:
- If `Fix kit` is selected, verifies that the `kit contents` section is displayed.
- If `Part only` is selected/clicked, verifies that the `Assembly contents` section is displayed.
- Verifies that the product images and skus change when switching between `Fix kit` / `Part Only`.

Note: didn't test for the product price change since they could potentially be the same price. Sometimes, the fix kit is cheaper than the part only, for example: [when the fix kit is on sale](https://react-commerce.vercel.app/products/iphone-6s-plus-replacement-battery).

## QA notes
Make sure the Cypress tests are passing in CI.

Closes #1045

### DIFF
--- a/frontend/test/cypress/integration/product/fix_kit.cy.ts
+++ b/frontend/test/cypress/integration/product/fix_kit.cy.ts
@@ -1,0 +1,29 @@
+describe('Fix Kit and Part Only test', () => {
+   beforeEach(() => {
+      cy.loadProductPageByPath('/products/iphone-6s-plus-replacement-battery');
+   });
+
+   it('test kit contents and product skus', () => {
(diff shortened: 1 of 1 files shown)


mlahargou (approved):
CR 📱

Issue #1015, opened 2022-11-09
Product Page v2.0: page load speed, LCP "needs improvement"


ifixitmaxb:
[Google Search Console is reporting that all our new pages have high LCP (longer than 2.5s on mobile)](https://search.google.com/search-console/core-web-vitals/summary?resource_id=https%3A%2F%2Fwww.ifixit.com%2Fproducts%2F&device=2)
[Google Search Console was reporting all URLs were "Good" but then as of Nov 4, all pages "Need improvement"](https://search.google.com/search-console/core-web-vitals/summary?resource_id=https%3A%2F%2Fwww.ifixit.com%2Fproducts%2F&device=1), again for high LCP

danielbeardsley:
I've got a way to defer loading youtube videos that I'm working on.

danielbeardsley:
Man, youtube is really hurting us here. I'll see what I can do about that:
![image](https://user-images.githubusercontent.com/26855/202331626-26652e42-7b8f-4030-968c-2557404d9f6e.png)

danielbeardsley:
There's a few small changes that are deployed and a few small ones that are in open pulls. The stats on the product page are still.... not great. 

Note sure where to go from here. 


[The analysis stats on vercel](https://vercel.com/ifixit/react-commerce-prod/analytics?device=mobile&timePeriod=7d&percentile=90&timeInterval=1h&sort=datapoints&search=) are very nice, but the changes we have merged haven't moved that needle at all and the numbers I'm getting locally for mobile lighthouse runs aren't promising (Lots of JS to parse and evaluate).

Potential roads for improvement:

* Enable the old product page (with a certain get param) and we can compare these same stats on the old vs new page
   * Would help with know if we've gone backwards or forwards
* Cache the results of the shopify graphQL query somewhere else cause that's the biggest piece of our server-side time
   * in memory (limited space)
   * on some storage service, Cloudfront backed by S3 (could even re-make the request client-side if the value we got from CF is too old)
      * Imagine exposing this query as an API
   * Have iFixit API make the query and cache the result however we need
* Turn up page caching a bit cause we only have a ~7% hit rate:
![image.png](https://images.zenhubusercontent.com/54ff41794f7a08fbb7f24e2c/6205d798-96ff-40c8-8182-32a8a0c1cc37)
* So much unused JS is downloaded on the product page
[...]

PR #1052, opened 2022-11-21
Youtube: use lite embed

danielbeardsley:
Youtube iframes slow down a page a suprising amount, many requests, 1MB
of resources, slow JS, ...

A way to combat this is to show the thumbnail and the play button and
only create the iframe when the play button or thumbnail is clicked.

This is the most depended upon package that does this. It's only 3KB
minified+gzipped.

See individual commits for more context

Connect #1015 

Consumes the metafield added here: https://github.com/iFixit/ifixit/pull/45611


masonmcelvain (dismissed):
CR ✌🏻 Great work!

> Typescript couldn't find the index.d.ts file anymore cause I wasn't importing the main pacakge, but a particular file

I wonder if adding the file to the `include` array of `tsconfig.json` would work. Though you mentioned that you already played with the tsconfig, so you might have tried this already.

danielbeardsley:
>  Though you mentioned that you already played with the tsconfig, so you might have tried this already.

Puzzactly. Though the type definitions aren't *global* in those files, so I don't know how it would have associated it with the later `import` of a specific file in a specific package.

erinemay:
@danielbeardsley should this speed up page load as is or is there a follow up to this? For qa: validate page load is faster, validate the video loads when clicked on, no errors or usability issues...

deltuh-vee:
Is this blocked on https://github.com/iFixit/ifixit/pull/45611 merging first?

danielbeardsley:
> Is this blocked on 45611 merging first?

I don't think so, if the metafield is empty it's just ignored, though it'd be much easier to validate when that pull is merged cause there will be values in it.

danielbeardsley:
> should this speed up page load as is or is there a follow up to this?

This will only affect page load speed on products that have videos by replacing the video with a lite version once the other pull is merged and the products are re-synced.

sterlinghirsh (approved):
CR :zap:

deltuh-vee:
QA 🎬 
Youtube iframe is only loaded and created after clicking play on the video.

PR #1054, opened 2022-11-21
Widen zoomed image in product page

### LINKED ISSUE #966: Product Page: Zoomed image aspect ratio
The long narrow zoomed image is kind of weird looking. Let's try to make it take up the rest of the page. The height is okay, but if we could widen it so it goes to the right edge of the screen (or close), it would be easier to see.

Example: https://ifixit.slack.com/files/UBF7UQNC9/F04924B1L10/image__11_.png

https://www.ifixit.com/products/ifixit-opening-picks-set-of-6

We also should make sure not to blow up the image past its intrinsic size when zooming. If original size of the image smaller than the gallery size, we should  not bother with the zoom at all.

federicobadini:
closes #966 

### QA

1. Visit Vercel preview
2. Verify that the the zoom container is now taking all the space available in the right side of the screen (with a 24px space on all sides)


sterlinghirsh:
CR :zap:

deltuh-vee:
QA 🎬 
The zoom container is now taking all the space available in the right side of the screen

masonmcelvain (approved):
CR ✌🏻

PR #1057, opened 2022-11-22
Hotfix: Handle malformed collection menu params

### LINKED ISSUE #1056: Algolia refinement library error
## Problem

It seems that all of the urls in this Sentry event are 500ing. Examples:

https://www.ifixit.com/Parts/iPhone_SE_2020?filter[worksin][0]=iPhone%20SE%202020%20A2275%20US%2FCanada
https://react-commerce-prod.vercel.app/Parts?filter%5Bfacet_tags.Item%20Type%5D%5B0%5D=Batteries

These errors started happening the day we merged https://github.com/iFixit/react-commerce/pull/1021. Perhaps it's related to the use of the new `useMenu` hook that was introduced in that pull.

Facet tags it seems to happen for:
- Tool category
- Item type
- Part or Kit
- Device Brand
- worksin

## Do this

Research why this is happening

Author: @masonmcelvain

---

Sentry Issue: [REACT-COMMERCE-MK](https://sentry.io/organizations/ifixit/issues/3739967695/?referrer=github_integration)

```
TypeError: currentRefinement[0].split is not a function
  File "/var/task/node_modules/.pnpm/algoliasearch-helper@3.11.0_algoliasearch@4.13.1/node_modules/algoliasearch-helper/src/SearchResults/index.js", line 532, in <anonymous>
    if (currentRefinement.length === 0 || currentRefinement[0].split(separator).length < 2) {
  File "/var/task/node_modules/.pnpm/algoliasearch-helper@3.11.0_algoliasearch@4.13.1/node_modules/algoliasearch-helper/src/SearchResults/index.js", in Array.forEach
    'use strict';
  File "/var/task/node_modules/.pnpm/algoliasearch-helper@3.11.0_algoliasearch@4.13.1/node_modules/algoliasearch-helper/src/SearchResults/index.js", line 525, in new SearchResults
    state.getRefinedHierarchicalFacets().forEach(function(refinedFacet) {
  File "/var/task/node_modules/.pnpm/react-instantsearch-hooks@6.32.1_w7dlztxainugwieacvqvd4x65u/node_modules/react-instantsearch-hooks/dist/es/lib/createSearchResults.js", line 5, in createSearchResults
    return new algoliasearchHelper.SearchResults(state, [{
  File "/var/task/node_modules/.pnpm/react-instantsearch-hooks@6.32.1_w7dlztxainugwieacvqvd4x65u/node_modules/react-instantsearch-hooks/dist/es/lib/getIndexSearchResults.js", line 12, in getIndexSearchResults
    createSearchResults(helper.state);
...
(5 additional frame(s) were not displayed)
```

masonmcelvain:
Prevents 500s when the menu param is malformed.

## CR

Multiselect (aka refinement) facets use array indeces in the query params (like `[facet_tags.Price Range][0]=10-20`);

Single selects don't have that (e.g. `[worksin]=Device Name`).

However, if a user visits a malformed url that uses array indices on a menu facet, and we tell Algolia that the menu has an array of values, Algolia's code will throw.

The solution: detect array indices on these facets and ignore them before sending the state object to Algolia.

## QA

Visit /Parts?filter[facet_tags.Item Type][0]=Batteries, and make sure the item type filter is set correctly and that the page doesn't 500, and the url is quickly reverted to /Parts?filter[facet_tags.Item Type]=Batteries

Also make sure that any invalid filter params are ignored. You'll have to do some creative url hacking.

Closes #1056

### DIFF
--- a/frontend/components/common/InstantSearchProvider.tsx
+++ b/frontend/components/common/InstantSearchProvider.tsx
@@ -217,6 +217,16 @@ export function InstantSearchProvider({
                   ? filter
                   : {};
 
+            Object.entries(filterObject).forEach(([attribute, value]) => {
+               const widgetType = getFacetWidgetType(attribute);
+               if (
+                  widgetType === FacetWidgetType.Menu &&
+                  Array.isArray(value)
+               ) {
+                  filterObject[attribute] = value[0];
+               }
+            });
+
             if (deviceHandle && itemType) {
                filterObject['facet_tags.Item Type'] = destylizeDeviceItemType(
                   decodeURIComponent(itemType)
--- a/frontend/components/common/InstantSearchProvider.tsx
+++ b/frontend/components/common/InstantSearchProvider.tsx
@@ -4,6 +4,7 @@ import { CLIENT_OPTIONS } from '@helpers/algolia-helpers';
 import {
    destylizeDeviceItemType,
    getFacetWidgetType,
+   isValidRefinementListValue,
    stylizeDeviceItemType,
 } from '@helpers/product-list-helpers';
 import { cypressWindowLog } from '@helpers/test-helpers';
@@ -11,6 +12,7 @@ import { useAuthenticatedUser } from '@ifixit/auth-sdk';
 import { assertNever } from '@ifixit/helpers';
 import { usePrevious } from '@ifixit/ui';
 import { FacetWidgetType } from '@models/product-list';
+import { useFacets } from '@templates/product-list/sections/FilterableProductsSection/facets/useFacets';
 import algoliasearch from 'algoliasearch/lite';
 import { history } from 'instantsearch.js/es/lib/routers';
 import { RouterProps } from 'instantsearch.js/es/middlewares';
@@ -59,6 +61,7 @@ export function InstantSearchProvider({
    }, [algoliaApiKey]);
 
    const router = useRouter();
+   const facets = useFacets();
 
    // Currently, Algolia routing does not play well with Next.js routing, since Next.js
    // is not aware of url changes that happens without interacting with its builtin router.
@@ -218,12 +221,29 @@ export function InstantSearchProvider({
                   : {};
 
             Object.entries(filterObject).forEach(([attribute, value]) => {
+               if (!facets.includes(attribute)) {
+                  delete filterObject[attribute];
+                  return;
+               }
                const widgetType = getFacetWidgetType(attribute);
-               if (
-                  widgetType === FacetWidgetType.Menu &&
-                  Array.isArray(value)
-               ) {
-                  filterObject[attribute] = value[0];
+               switch (widgetType) {
+                  case FacetWidgetType.Menu: {
+                     if (isValidRefinementListValue(value)) {
+                        filterObject[attribute] = (value as string[])[0];
+                     } else if (typeof value !== 'string') {
+                        delete filterObject[attribute];
+                     }
+                     break;
+                  }
+                  case FacetWidgetType.RefinementList: {
+                     if (!isValidRefinementListValue(value)) {
+                        delete filterObject[attribute];
+                     }
+                     break;
+                  }
+                  default: {
+                     return assertNever(widgetType);
+                  }
                }
             });
 
--- a/frontend/helpers/product-list-helpers.ts
+++ b/frontend/helpers/product-list-helpers.ts
@@ -1,4 +1,5 @@
 import { FacetWidgetType, ProductListType } from '@models/product-list';
+import { z } from 'zod';
 
 type ProductListAttributes = {
    filters?: string | null;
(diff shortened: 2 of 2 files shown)


frontend/components/common/InstantSearchProvider.tsx
```diff
@@ -217,6 +217,16 @@ export function InstantSearchProvider({
...
 
+            Object.entries(filterObject).forEach(([attribute, value]) => {
+               const widgetType = getFacetWidgetType(attribute);
+               if (
+                  widgetType === FacetWidgetType.Menu &&
+                  Array.isArray(value)
+               ) {
+                  filterObject[attribute] = value[0];
+               }
+            });
```
danielbeardsley, line 228: This does fix the really specific case of submitting an array-like value to a param that doesn't accept an array, but there are many more ways this could 500. I'm more inclined to validate each param and ignore it if it's not the right format rather than auto-correcting a specific failure mode.

Thoughts?

masonmcelvain:
QA ✌🏻 

When visiting https://react-commerce-r7abyxhg4-ifixit.vercel.app/Parts?filter[facet_tags.Item%20Type][0]=Batteries, the page appears correctly filtered and the url quickly changes to https://react-commerce-r7abyxhg4-ifixit.vercel.app/Parts?filter[facet_tags.Item%20Type]=Batteries

frontend/components/common/InstantSearchProvider.tsx
masonmcelvain: see https://github.com/iFixit/react-commerce/pull/1057/commits/f7e1d7f4bee2053bd0ec82a6da7e3f482573774f and lmk what you think

frontend/helpers/product-list-helpers.ts
```diff
@@ -75,6 +76,12 @@ export function getFacetWidgetType(attribute: string) {
    return facetWidgetTypeMap[attribute] ?? FacetWidgetType.Menu;
 }
 
+const RefinementListValueSchema = z.string().array().nonempty();
+
+export function isValidRefinementListValue(value: unknown) {
```
danielcliu-ifixit, line 81: This is actually checking if it is a string array, not that it is a valid refinement list value necessarily as its not allowing strings. So maybe we could rename this?

masonmcelvain: Since `RefinementList`s are multi-selects in algolia, they will always be an array of strings even if there is just one value.

That being said, I could be convinced that `nonempty` isn't necessary, I wasn't really sure about that piece.

danielcliu-ifixit: Also could specify that it returns a `boolean`

frontend/components/common/InstantSearchProvider.tsx
```diff
@@ -217,6 +220,33 @@ export function InstantSearchProvider({
...
+               if (!facets.includes(attribute)) {
+                  delete filterObject[attribute];
+                  return;
+               }
+               const widgetType = getFacetWidgetType(attribute);
+               switch (widgetType) {
+                  case FacetWidgetType.Menu: {
+                     if (isValidRefinementListValue(value)) {
+                        filterObject[attribute] = (value as string[])[0];
+                     } else if (typeof value !== 'string') {
```
danielcliu-ifixit, line 233: value should never be string here

masonmcelvain: If it's a `Menu`, then the expected value is `string` (e.g. `[worksin]=Device Name`). However, in case the user visits a url with the refinement list array syntax (e.g. `[worksin][0]=Device Name`), we know they probably want the worksin filter, so that first `if` block will transform it into a string.

masonmcelvain: *the expected **type** is `string`

jordycosta:
QA 🟢 

`/Parts?filter[facet_tags.Item Type][0]=Batteries` navigates to `/Parts?filter[facet_tags.Item Type]=Batteries`

firefox_EBboIsh5aE

frontend/components/common/InstantSearchProvider.tsx
```diff
@@ -217,6 +220,33 @@ export function InstantSearchProvider({
                   ? filter
                   : {};
 
+            Object.entries(filterObject).forEach(([attribute, value]) => {
```
danielbeardsley, line 223: Would it be simpler to define the type of the whole `filterObject` using `zod` here instead of writing code that does the same?

danielbeardsley (approved):
CR :+1: 

I'm not blocking on my question.

danielcliu-ifixit (approved):

frontend/components/common/InstantSearchProvider.tsx
masonmcelvain: That would be nice. I had considered that, but I struggled to imagine what the implementation would look like.

**Option A:** Hard code a zod schema with all the facet names
```js
zod.object({
   "facet_tags.Item Type": z.string(),
   "facet_tags.Price Range": z.string().array(),
   ...,
});
```
Con: This schema could get out of sync with `useFacets()`.

**Option B:** Dynamically generate a zod schema with the result of `useFacets()`

Context: The `filterObject` will include any other garbage that happens to be under the `filter` namespace in the query params. For example, `filter[asdf]=qwer` in the params would generate `{ asdf: "qwer" }` in the filter object.

Algolia ignores any filter params whose key (aka attribute) don't match a real facet.

However, in order to dynamically generate the schema we'd need to be able to determine:
- if each attribute in `filterObject` is a valid facet
- if each valid attribute in `filterObject` is a Menu or a RefinementList

All that seems like it would require just as much code as validating the `filterObject` ourselves, plus we get the benefit of transforming the invalid menu params so that they are correctly applied when the page loads.

I'd be happy to implement this in a followup PR if we can come up with a solution tho

PR #975, opened 2022-11-03
Upgrade to react 18

dhmacs:
⚠️ WIP

PR #1058, opened 2022-11-22
Fix Jest CI

dhmacs:
This has been extracted from #975 

There are a few issues that break tests with updated dependencies, such as a version of the pnpm action that does not support pnpm v7 and a wrong command to install dependencies.

qa_req 0

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
       "start:frontend": "cd frontend && pnpm run start",
       "start": "cross-env FORCE_COLOR=1 npm-run-all -l -p start:*",
       "strapi": "cd backend && pnpm run strapi",
-      "test": "echo \"Error: no test specified\" && exit 1",
+      "test": "pnpm run -r test",
       "transform-svg": "cd frontend && pnpm run transform-svg",
       "cypress:open": "cd frontend && pnpm run cypress:open",
       "cypress:run": "cd frontend && pnpm run cypress:run",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -19,7 +19,7 @@
       "start:frontend": "cd frontend && pnpm run start",
       "start": "cross-env FORCE_COLOR=1 npm-run-all -l -p start:*",
       "strapi": "cd backend && pnpm run strapi",
-      "test": "echo \"Error: no test specified\" && exit 1",
```
danielbeardsley, line 22: Huh...  I thought `run test` effectively runs the command specified by `package.json['scripts']['test']`, so this appears circular. I'm obviously wrong... So I guess `-r` means recursive but *not* including the current package?

batbattur: >  So I guess -r means recursive but not including the current package?

In CI, we run `pnpm test` which runs for the `pnpm run test -r`.
I believe the `pnpm run test -r` includes the current package, from https://pnpm.io/cli/run#--recursive--r:
> This runs an arbitrary command from each package's "scripts" object. If a package doesn't have the command, it is skipped. If none of the packages have the command, the command fails.

danielbeardsley: > I believe the pnpm run test -r includes the current package

If that's the case, wouldn't that mean this setup is infinitely recursive?

i.e. `pnpm run test -r` runs the "scripts.test" command in each package... which is `pnpm run test -r` in the top level package.

I must be missing something.

batbattur: Whoops, I might have been wrong. The docs don't explicitly say it's excluded but it looks like the root dir **is** excluded. Fount the relevant issues https://github.com/pnpm/pnpm/issues/2769 and https://github.com/pnpm/pnpm/pull/3647

PR #403, opened 2022-07-06
Tests: Collections parts page add tests

batbattur:
## Summary

Adds a new test to the collection `/Parts` page. 
- ~The first test visits devices/models on `/Parts` page and asserts that the search filter and `filterable-products-section` section is visible (copy of `WDStorePagesSmokeTest -> testAllToplevelPartsPages` in iFixit/ifixt).~
- The second test clicks on a device (hard coded to `Mac Parts` for the moment), then navigates until the last device page and asserts that there is at least one product (`WDStorePagesSmokeTest -> testDrillDownUntilPartFound`).

## QA 

Please re-run the `E2E tests / cypress-run` ci check and make sure it passes.

Connects: https://github.com/iFixit/ifixit/issues/43590 and https://github.com/iFixit/react-commerce/issues/387

PR #1063, opened 2022-11-25
Fix flaky Cypress test

### LINKED ISSUE #1064: Fix flaky Cypress test
`parts page devices` test becomes flaky once we update to React 18.

## Do this

Improve the test so it becomes more resilient

dhmacs:
extracted from #975 

closes #1064

This PR fixes a test that becomes flaky once we upgrade to React 18

qa_req 0

### DIFF
--- a/frontend/test/cypress/integration/product-list/parts_page_devices.cy.ts
+++ b/frontend/test/cypress/integration/product-list/parts_page_devices.cy.ts
@@ -6,57 +6,53 @@ describe('parts page devices', () => {
    it('should navigate until the last device page', () => {
       assertVisibleFilterAndProducts();
       assertAvailableProducts();
+      const navigateUntilLastDevice = () => {
+         cy.get('body').then(($body) => {
+            const $children = $body.find(
+               '[data-testid="product-list-children"]'
+            );
+            if ($children.length <= 0) return false;
 
-      cy.get('body').then(($body) => {
-         const device = $body.find('a[href="/Parts/Mac"]');
-         if (!device.is(':visible')) {
-            cy.get('button').contains('Show more').should('be.visible').click();
-         }
-         device[0].click();
-      });
+            cy.findByTestId('product-list-children')
+               .findAllByRole('link')
+               .first()
+               .as('childLink');
 
-      cy.location('pathname').should('include', '/Parts/Mac');
-      cy.get('h1').contains('Mac Parts').should('be.visible');
+            cy.get('@childLink').then(($childLink) => {
+               const childHref = $childLink.attr('href');
+               cy.location('pathname').should('not.equal', childHref);
+
+               cy.get('@childLink').click();
+
+               cy.location('pathname').should('equal', childHref);
+
+               const childTitle = $childLink.text();
+               const childTitleRegexp = new RegExp(`^${childTitle}$`, 'i');
+               cy.findByRole('heading', {
+                  level: 1,
+                  name: childTitleRegexp,
+               }).should('exist');
+
+               navigateUntilLastDevice();
+            });
+         });
+      };
 
       navigateUntilLastDevice();
-      assertVisibleFilterAndProducts();
-      assertAvailableProducts();
    });
-
-   function assertVisibleFilterAndProducts() {
-      cy.findByTestId('filterable-products-section').should('be.visible');
-      cy.findByTestId('facets-accordion').scrollIntoView().should('be.visible');
-   }
-
-   // Makes sure there is at least 1 product available
-   function assertAvailableProducts() {
-      cy.findByTestId('list-view-products')
-         .children('article')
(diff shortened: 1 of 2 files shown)


frontend/test/cypress/integration/product-list/parts_page_devices.cy.ts
```diff
@@ -6,57 +6,53 @@ describe('parts page devices', () => {
...
+               }).should('exist');
+
+               navigateUntilLastDevice();
+            });
+         });
+      };
 
       navigateUntilLastDevice();
-      assertVisibleFilterAndProducts();
-      assertAvailableProducts();
```
masonmcelvain, line 23: It seems like the test has changed a little, since we no longer perform these assertions after navigating to the last device and we no longer call `assertVisibleFilterAndProducts` in `navigateUntilLastDevice`. Were these assertions redundant/unnecessary?

masonmcelvain (approved):
CR ✌🏻 with a question

frontend/test/cypress/integration/product-list/parts_page_devices.cy.ts
dhmacs: Great question. The name of the test is `should navigate until the last device page` so when refactoring the test I was focusing on that. Do you think these 2 assertions are important for the goal of the test? They seem a little bit out of scope and I'm not really sure what they're trying to prove here. 
What are you thoughts on this?

davidrans:
deploy_block :hand: on https://github.com/iFixit/react-commerce/pull/1063#discussion_r1032946537

frontend/test/cypress/integration/product-list/parts_page_devices.cy.ts
masonmcelvain: I agree that `assertVisibleFilterAndProducts` doesn't seem important to the goal of the test IMO.

It seems that the authors of this test (@batbattur @mlahargou) wanted to assert that the last device page has at least one product (based on the PR comment of https://github.com/iFixit/react-commerce/pull/403), so I'd like to get their input on this.

> The second test clicks on a device (hard coded to Mac Parts for the moment), then navigates until the last device page and asserts that there is at least one product (WDStorePagesSmokeTest -> testDrillDownUntilPartFound).

I know we put in effort to prevent child device links to empty collections, and I think that's this test is trying to check (@batbattur please check me on this assumption). However, we're following the first child device each time, so if something broke and there _were_ empty collections for child device links, this test would probably fail to catch that regression since we're only checking one leaf device. The only comprehensive way to check for that regression would be to follow all device links, which would be a very slow test.

TLDR; I agree with @dhmacs, these assertions both seem unnecessary. That being said, is this entire test even necessary if we're just following child device links without performing any assertions along the way?

dhmacs: > assert that the last device page has at least one product

Yep, but this seems to test the correctness of the data rather than the correctness of the code. Not sure we want to fail CI because we have errors in the data.

davidrans: Agreed. Unless we're able to set up a particular set of data at the beginning of the test then that seems like a recipe for flakiness

masonmcelvain:
un_deploy_block ✌🏻 We've agreed that certain assertions were not necessary, and the test will no longer be flaky, which is the scope of this PR.

frontend/test/cypress/integration/product-list/parts_page_devices.cy.ts
batbattur: This test was the replica of [`WDStorePagesSmokeTest -> testDrillDownUntilPartFound`](https://github.com/iFixit/ifixit/blob/564aae46f8d0cef69d5a3202ca66bbac3808affc/Tests/Smoke/WDStorePagesSmokeTest.php#L65-L82) where the old test was trying to click on the first [category](https://github.com/iFixit/ifixit/blob/564aae46f8d0cef69d5a3202ca66bbac3808affc/Tests/Objects/UIExtensions/CartUIExtension.php#L224) displayed on the collection page recursively until the last first child and click on the product. 

> I know we put in effort to prevent child device links to empty collections, and I think that's this test is trying to check 

Yes, I think that is what we wanted to check in this test. I think the old collection pages didn't show the link if there were no products, but if we do this we are relying on the data. 

> Yep, but this seems to test the correctness of the data rather than the correctness of the code. Not sure we want to fail CI because we have errors in the data.
 
I agree with this. I think we were trying to get tests started for collection page at that time so I added it without realizing it heavily relies on the data. We can skip this test or remove it if unnecessary.

mlahargou: I think this is fine 👍

PR #1067, opened 2022-11-28
Revert image widths in guideImageSizeMap

### LINKED ISSUE #1039: Guide image size map is obviously wrong
This is the guide image size map:
image

e.g. 140x105, the width is 140, not 110. This is almost certainly a copy/paste error, but we should make sure the guide images sizes are correct.

federicobadini:
closes #1039 

### QA

1. Visit Vercel preview
2. Verify that images are now generated according to the default imageSizes + deviceSizes rounded up to the nearest available format offered by the ifixit api

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -121,7 +121,6 @@ const moduleExports = {
          'guide-images.cdn.ifixit.com',
          process.env.STRAPI_IMAGE_DOMAIN,
       ].filter((domain) => domain),
-      imageSizes: [41, 70, 110, 170, 250, 400, 600],
    },
    i18n: {
       locales: ['en-US'],
--- a/packages/ui/misc/ResponsiveImage/iFixitUtils.ts
+++ b/packages/ui/misc/ResponsiveImage/iFixitUtils.ts
@@ -7,13 +7,13 @@ type SizeMap = {
 
 // The following lists should be sorted in order of ascending width.
 export const guideImageSizeMap: SizeMap = [
-   { name: 'mini', width: 41 },
-   { name: 'thumbnail', width: 70 },
-   { name: '140x105', width: 110 },
-   { name: '200x150', width: 170 },
-   { name: 'standard', width: 250 },
-   { name: '440x330', width: 400 },
-   { name: 'medium', width: 600 },
+   { name: 'mini', width: 56 },
+   { name: 'thumbnail', width: 96 },
+   { name: '140x105', width: 140 },
+   { name: '200x150', width: 200 },
+   { name: 'standard', width: 300 },
+   { name: '440x330', width: 440 },
+   { name: 'medium', width: 592 },
    { name: 'large', width: 800 },
    { name: 'huge', width: 1600 },
 ];


k0rvusk0r4x:
The correct image sizes are now being generated.

QA 😺

danielbeardsley:
I'm not confident in this change in reference to the long thread in the issue.

Can you give links and mention particular images who's situations have improved? Now that cart image sizes don't line up with next's image sizes, I'm worried that cart images will behave poorly.

danielbeardsley:
>  is just a list of breakpoints that are used in image generation 

As far as I'm aware, `next/image` snaps all incoming widths to one of these values before passing the `width` to the loader. Here's where the behavior diverges. Cart and Guide loaders are passed a pixel width (that's one of the values from `imageSizes`) and we get to choose which size name to provide a url to.

This pull will result in double rounding for guide and cart images as documented in [the comment here](https://github.com/iFixit/react-commerce/issues/1039#issuecomment-1329804778)

Here's an example of this change being worse-off: Go to `/Parts` and change to "Gallery" mode

Production:
image

This branch:
image

You're right in that for loaders that aren't *themselves* quantized, the arrangement of next's `imageSizes` array isn't critical. But for those that are (cart / guide images), the double-quantization ends up using larger-than-necessary images.

That said, I may be missing something or just wrong in that there are more counter-examples than I realize, so if you find images that are *better off* with this change, please highlight them so we can compare the tradeoffs.

federicobadini:
> As far as I'm aware, `next/image` snaps all incoming widths to one of these values before passing the `width` to the loader.

It is not working exactly like that. There are two ways in which next image generate `srcSet`. If `layout=fill` is specified it works in a way, if explicit width and height are provided it works in another.

In the former case it starts from imageSizes + deviceSizes and generate urls via the loader. So there is no double roundup since we start from the values in those lists and we round up to the nearest format available via the api

In the latter imageSizes are ignored. The passed width is the only measure to be rounded up to the nearest format available.

The intrinsic size that can be seen in those popup is the result of what the browser chooses to download according to its internal heuristics, the `sizes` props and the offered options in `srcSet`. 

In your example the code instructs the browser via sizes that a 250px wide image is requested. The srcSet info generated with custom sizes include that specific value - so the 250x250 image is picked. In this branch `srcSet` is generated for size like `[16, 32, 48, 64, 96, 128, 256, 384]` so the browser would pick `256w` which corresponds to a 400x400 image
So this is an example in which the custom imageSizes are a perfect fit for the request.

But let's for instance say that we want a 16x16 shopify image or a 256 one. The custom imageSizes still have an impact. 
In those cases the customization will lead to download 41x41 and 400x400 images, quite larger than the request.

So either we have a vision in which we know that most of the visual in the website will have specific form factors, or optimizing imageSizes for one specific scenario may end up damaging another one.

danielbeardsley:
Thanks for refreshing my memory about the two different rendering modes.

Though I think it's not quite right. If you specify width and height, the main `src` is set to the return value of calling the loader with your exact width as you said. However, the 1x and 2x sizes in the srcset are taken directly from the `imageSizes` array. Here I've logged the `src, width, baseWidth, sizename` for a given image that has it's height/width explicitly set.

image

>  So there is no double roundup since we start from the values in those lists and we round up to the nearest format available via the api

I think we're saying the same thing here and using different terms. By double rounding, I meant that if the browser displays an image at 130px, it has to choose a url among the `srcset` provided (the list of `imageSizes`) and *then* the url chosen for each of those effectively is rounded up to the nearest size we have available. With the default image sizes, this becomes `130px -> 256px -> 400px`.

That said, yes, this is complex, and yes optimizing too much for one case may damage another, which is why I'd rather practically look at how it shakes out (i.e. examples of cases where one choice is better / worse)

sterlinghirsh (approved):
CR :zap: un_dev_block :zap: I think this is the right move. I noticed that the `sizes` is still wrong on the child productlist images. They're shown at a constant 80x60 but sizes=20vw, so it's pulling a much bigger sizes (440 or even 800 depending on window size). I'll make a separate issue to address that.

danielbeardsley:
So far as we can tell this pull made twenty instances of the problem worse (every image on the product list page is now bigger than it needs to be as (documented here](https://github.com/iFixit/react-commerce/pull/1067#issuecomment-1333034408)

image

Did it make some instances better?

Why did we merge this when it definitely made things worse and we hadn't found an example where it improved things?

sterlinghirsh:
I missed the comment about it making that worse.

federicobadini:
I think a little more clarification is needed here:

We all agree that the sizes in the `guideImageSizeMap` were wrong right? 
The api was returning sizes that were different from the one listed in the map so I think we all agree on fixing that part.

Screenshot 2022-12-12 at 18 12 52

Said that, I proposed to revert also the imageSizes customization for three reasons:

1. We have 2 "non-continous" APIs which have different breakpoints among them. It is clear that adapting imageSizes on cart api breakpoints will perform well for that api. But if we will end up requiring a 256 pixel image on the guide api, being the breakpoint at 250px we will end up serving a 440px image, skipping the 300px one which would have been a valid option.
2. Whenever we drive away from default we are stepping aside from possible future improvements so generally I tend to move away from defaults only in situation of clear advantage. 
3. These images are almost all lazy loaded so they are not that impacting on performance (and the KB diff is very small too)

These were the reasons but as said I do not belive that all this will have a strong impact on performance so if you feel it valuable and wanna keep the customization please let me know.

> Though I think it's not quite right. If you specify width and height, the main `src` is set to the return value of calling the loader with your exact width as you said. However, the 1x and 2x sizes in the srcset are taken directly from the `imageSizes` array. Here I've logged the `src, width, baseWidth, sizename` for a given image that has it's height/width explicitly set.

Just for completeness, I'd like also to point out that the loader, whenever `width` is specified, is effectively called with the specified width and its `2x` multiple by default. As far as I know the only case in which this does not happen is when the same image is present on the page without an explicit width. 
Product list images for instance are required currently at 180px width, which is not a value of neither the default not the custom imageSizes. So in most cases in which width is specified, `imageSizes` is not involved.

sterlinghirsh:
We might want to make a custom image component that handles the predefined
sizes more appropriately. It's unfortunate but after reading through the
docs, it doesn't seem like it's possible to specify image sizes separately
for different types of image (guide, cart, Shopify...). Also it says that
image sizes should all be less than device sizes anyway. If we can override
the srcset then that might be good enough to still use next image but I
don't want to jump through hoops just to use the default next image
component if it's not well suited to our case.

sterlinghirsh:
I made #1131 / #1135 to deal with the weird sizes on the child productlist images. Haven't tried to deal with the responsive images for products yet though.

Issue #976, opened 2022-11-03
Product Page: set up `hreflang`s with other stores


davidrans:
Problem
---
We should add `hreflang`s on every product page, for every product that we sell in a non-US store. Now that we use the same URL structure/handle across all our stores this should be a bit easier. 

Notes
---
I think we need the hreflangs going both directions for Google to trust them, so we also need to add hreflangs pointing at the US store to the Shopify theme. Is there a significant amount of products we don't sell in the US store that we sell in other stores, or is it safe to just hreflang back to the US store on every non-US product page @jeffsnyder?

What's the best way to look up which other stores a product is sold in? Is there already an API we're using server-side that has this info? @sterlinghirsh 

Related: https://github.com/iFixit/react-commerce/issues/969

ifixitmaxb:
YES. Should I close this? https://github.com/iFixit/ifixit/issues/43739

sterlinghirsh:
Use a metafield with a list of storecodes where the product is available.

sterlinghirsh:
Use a list of domains like `["canada.ifixit.com", "australia.ifixit.com"...]` and then we don't have to do any mapping on the client side. This can be a product metafield and we just won't worry about mapping variants to other variants.

ifixitmaxb:
I don't think we should have a lot of 404ing hreflangs. If there is no equivalent in another store then I'd rather not hreflang to the non-existent equivalent.

masonmcelvain:
We finished adding hreflangs to the product pages 🎉

PR #1068, opened 2022-11-28
Add `hreflang`s to product page

masonmcelvain:
Adds hreflangs to all other stores (including the US store) to the `<head>`, like we do on the shopify theme stores.

## QA

Make sure we get the hreflangs in the head for products that are enabled on other stores. Verify that the urls are correctly formatted by clicking on one of them.

Connects #976

### DIFF
--- a/frontend/templates/product/MetaTags.tsx
+++ b/frontend/templates/product/MetaTags.tsx
@@ -52,6 +52,15 @@ export function MetaTags({ product, selectedVariant }: MetaTagsProps) {
          <link rel="canonical" href={canonicalUrl} />
          <meta property="og:url" content={canonicalUrl} />
 
+         {product.enabledDomains?.map((store) => (
+            <link
+               key={store.domain}
+               rel="alternate"
+               hrefLang={store.locale}
+               href={`${store.domain}/products/${product.handle}`}
+            />
+         ))}
+
          {genericImages.length > 0 &&
             genericImages.map((image) => (
                <meta key={image.id} property="og:image" content={image.url} />
--- a/frontend/lib/shopify-storefront-sdk/generated/sdk.ts
+++ b/frontend/lib/shopify-storefront-sdk/generated/sdk.ts
@@ -6405,6 +6405,7 @@ export type FindProductQuery = {
       metaTitle?: { __typename?: 'Metafield'; value: string } | null;
       shortDescription?: { __typename?: 'Metafield'; value: string } | null;
       oemPartnership?: { __typename?: 'Metafield'; value: string } | null;
+      enabledDomains?: { __typename?: 'Metafield'; value: string } | null;
       featuredImage?: { __typename?: 'Image'; id?: string | null } | null;
       images: {
          __typename?: 'ImageConnection';
@@ -6685,6 +6686,9 @@ export const FindProductDocument = `
       value
     }
     oemPartnership: metafield(namespace: "ifixit", key: "oem_partnership_json") {
+      value
+    }
+    enabledDomains: metafield(namespace: "ifixit", key: "enabled_domains") {
       value
     }
     featuredImage {
--- a/frontend/lib/shopify-storefront-sdk/operations/findProduct.graphql
+++ b/frontend/lib/shopify-storefront-sdk/operations/findProduct.graphql
@@ -72,6 +72,9 @@ query findProduct($handle: String) {
       ) {
          value
       }
+      enabledDomains: metafield(namespace: "ifixit", key: "enabled_domains") {
+         value
+      }
       featuredImage {
          id
       }
--- a/frontend/models/product.ts
+++ b/frontend/models/product.ts
@@ -106,6 +106,9 @@ export async function findProduct({ handle, storeCode }: FindProductArgs) {
       oemPartnership: parseOemPartnership(
          response.product.oemPartnership?.value
       ),
+      enabledDomains: parseEnabledDomains(
+         response.product.enabledDomains?.value
+      ),
    };
 }
 
@@ -432,6 +435,31 @@ function parseOemPartnership(
       return result.data;
    }
    logParseErrors(result.error, 'oem partnership metafield');
+   return null;
+}
+
+const EnabledDomainsSchema = z
+   .object({
+      domain: z.string().url(),
+      locale: z.string(),
+   })
+   .array()
+   .optional()
+   .nullable();
+
+function parseEnabledDomains(value: string | null | undefined) {
+   if (value == null) {
+      return null;
+   }
+   const rawJson = JSON.parse(value);
+   if (rawJson == null) {
+      return null;
+   }
+   const result = EnabledDomainsSchema.safeParse(rawJson);
+   if (result.success) {
+      return result.data;
+   }
+   logParseErrors(result.error, 'enabled domains metafield');
    return null;
 }
 


frontend/templates/product/MetaTags.tsx
```diff
@@ -52,6 +52,15 @@ export function MetaTags({ product, selectedVariant }: MetaTagsProps) {
          <link rel="canonical" href={canonicalUrl} />
          <meta property="og:url" content={canonicalUrl} />
 
+         {product.enabledDomains?.map((store) => (
```
danielbeardsley, line 55: Hmm... Looks like this will also include the current store. I think that's bad cause it'll be like a second canonical. Should we filter that out on the php app side, or try to here?

danielbeardsley (changes requested):
dev_block :+1: 

I think we should probably filter out the current store on the php app side.

CR :+1:  otherwise

frontend/templates/product/MetaTags.tsx
masonmcelvain: I agree that we should probably do it in PHP.

I didn't realize we needed to exclude the current store, because I saw this sort of self-referencing on the EU store.

image

https://eustore.ifixit.com/products/iphone-6s-replacement-battery

masonmcelvain:
https://github.com/iFixit/ifixit/pull/45677

masonmcelvain:
un_dev_block ✌🏻

danielbeardsley (approved):
CR :+1: 

Though I think we should deploy_block :+1:  on re-pushing all US products perhaps.

masonmcelvain:
> Though I think we should deploy_block +1 on re-pushing all US products perhaps.

Good call, I'll start that now.

deltuh-vee:
QA 🎬 
Now appear in the head for products
image

deploy_block ❓ 
The links have an extra "/". Not sure if we wanna address that here or somewhere else. The links still work with the extra "/".

Also, should the australia and canada stores have an hreflang of "en-US"? Should "en-GB" be the hreflang for EU store?

masonmcelvain:
> The links have an extra "/". Not sure if we wanna address that here or somewhere else. The links still work with the extra "/".

This was fixed by https://github.com/iFixit/ifixit/pull/45677, and it's likely that the product you were testing with hadn't been repushed to shopify (if you wanted to check it again, run the `console shopify:save-products --force` command)

un_deploy_block ✌🏻 

> Also, should the australia and canada stores have an hreflang of "en-US"? Should "en-GB" be the hreflang for EU store?

These locales come from our [`StoreProvider`](https://github.com/iFixit/ifixit/blob/master/iFixit/Cart/Stores/StoreProvider.php) class, so it's what the rest of our app uses. If we decide to change those locales in the future, these hreflangs will be updated automatically (when the product data is repushed).

danielbeardsley:
> These locales come from our [`StoreProvider`](https://github.com/iFixit/ifixit/blob/master/iFixit/Cart/Stores/StoreProvider.php) class, so it's what the rest of our app uses. If we decide to change those locales in the future, these hreflangs will be updated automatically (when the product data is repushed).

True, but now we're telling Google potentially something confusing... that there's an alternate version of this page that's *also* in US-English:

image

I can't help but think Google's not going to like that. Thoughts @ifixitmaxb ?

ifixitmaxb:
I'm a little confused by some of the comments. I'm sorry. I'm going to add thoughts and I hope it's helpful:

For example, https://www.ifixit.com/products/iphone-6-plus-replacement-battery. 
- We should ideally self-reference, so <link rel="alternate" hrefLang="en-us" href="https://www.ifixit.com/products/iphone-6-plus-replacement-battery"/>
- Australia Store should be en-au, so <link rel="alternate" hrefLang="en-au" href="https://australia.ifixit.com/products/iphone-6-plus-replacement-battery"/>
- Canada Store should be en-ca, so <link rel="alternate" hrefLang="en-ca" href="https://canada.ifixit.com/products/iphone-6-plus-replacement-battery"/>
- Since German Store doesn't only target Germany, German Store should just be language and no region, so <link rel="alternate" hrefLang="de" href="https://store.ifixit.de/products/iphone-6-plus-replacement-battery"/>
- Since French Store doesn't only target France, French Store should just be language and no region, so<link rel="alternate" hrefLang="fr" href="https://store.ifixit.fr/products/iphone-6-plus-replacement-battery"/>
- UK Store looks good
- EU Store is a tough situation since hreflang only accepts ISO country code designation and "EU" isn't a country. I actually haven't yet determined the right thing to do here but I don't think we should say EU Store is targeting the UK (en-gb) because that will conflate and potentially hurt our UK store rankings.

ifixitmaxb:
If it's helpful, I've found some global DTC brands that seem to be doing hrerlangs correctly. You can check the hreflang tags of the following for inspiration and gut-checking:
- https://www.patagonia.com.au/
- https://www.adidas.com.au/men
- https://www.etsy.com/
- https://www.vrbo.com/es-es/alquiler-vacacional/argentina/buenos-aires

danielbeardsley:
> If it's helpful, I've found some global DTC brands that seem to be doing hrerlangs correctly.

I noticed that Patagonia actually has different urls for each locale they list though:
image

I understand that having a bunch of hreflangs pointing at the same page might make technical sense, but I'm far less confident that Google recommends it in this situation.

masonmcelvain:
It looks like vrbo uses a plain `en` hreflang alongside `en-ca` et al. Perhaps using a single `en` hreflang for the EU won't confuse Google after all? (Max and I talked about this in DM's on slack)

image

ifixitmaxb:
As Mason and I discussed via Slack: I don't think it makes sense to use EN for EU Store. I think it would conflate with our other English Stores. With our Stores, we are targeting countries because we only ship to certain countries from certain Stores...which as I type this it makes me realize that we probably need to tweak our hreflangs for FR and DE Store...
@masonmcelvain 

- Similar to my previous comment, we should actually hreflang our DE Store as follows, sorted by top traffic to store.ifixit.de according to GA:
        - de-de
        - de-at
        - de-ch
        - de-nl
        - de-be
- And we should hreflang our FR Store as follows, sorted by top traffic to store.ifixit.de according to GA:
        - fr-fr
        - fr-be
        - fr-ch

masonmcelvain:
Thanks Max! I found some Google [documentation](https://developers.google.com/search/docs/specialty/international/localized-versions?hl=en&visit_id=638055078357788459-726003542&rd=1#html) that might serve as another data point for us (new to me, perhaps you're familiar with it).

The docs aren't clear about whether multiple hreflangs for the same url is ok. However I found an (old, 2018) [article](https://www.searchviu.com/en/multiple-hreflang-tags-one-url/) that seems to indicate it is.

Additionally, it confirms that you should indeed include a self-referencing hreflang.

> one link for each page variant including itself

> Each language version must list itself as well as all other language versions.

Google shows some examples of a generic language code alongside a region-specific language code, like

```html
<link rel="alternate" hreflang="en-gb" href="https://en-gb.example.com/page.html" />
<link rel="alternate" hreflang="en-us" href="https://en-us.example.com/page.html" />
<link rel="alternate" hreflang="en"    href="https://en.example.com/page.html"    />
```

> If you have several alternate URLs targeted at users with the same language but in different locales, it's a good idea to also provide a catchall URL for geographically unspecified users of that language. For example, if you have specific URLs for English speakers in Ireland (en-ie), Canada (en-ca), and Australia (en-au), provide a generic English (en) page for searchers in the US, UK, and all other English-speaking locations. It can be one of the specific pages, if you choose.

Also, do we want an `x-default` hreflang as a catch-all for unmatched languages?

ifixitmaxb:
Regarding x-default, my initial thought was that we shouldn't use it. From how I understand hreflang, the x-default hreflang makes sense for www. for all things except the Store. www. is definitely the default site and certainly the default English site. But the US Store doesn't ship outside the US so we don't necessarily want to indicate that it is globally relevant/targeted. However, I've read at least one reputable article that says you should always have a default backup hreflang. So maybe we should do the x-default to US Store and then allow all our other systems and logic in place, like IP detection, Store selectors, and shipping messaging that indicates where the US Store ships to, etc. to clarify to any non-US user who may happen to land on the US "default" Store that we don't ship outside the US. And to be fair, as I write this and keep reading articles, it's possible that the EU Store should be the x-default...I'm sorry, this is a tough topic and I'm not an absolute expert here. 

Fwiw, in case it's relevant, another thing I read is that we should only implement hreflangs in one place, like if we are doing this in our page header then we don't also want to do it in our xml sitemaps and http response header. 

[This article](https://www.searchenginejournal.com/international-seo-best-practices/252015/) is a good scan. The things I want to emphasize which we haven't technically discussed in this pull comment thread is that "all alternate attributes need to point back to their canonical" and "all alternate attributes need to be bidirectional" - so our hreflangs should point to canonical URLs and if we hreflang to a page on a different store, it should have the same hreflang(s) pointing back. I hope that makes sense. 

The most common hreflang mistakes include:

1. Failing to include a self-referential link in the alternate tag.
2. Directing an alternate URL to the wrong canonical.
3. Using the wrong ISO value.
4. Forgetting to insert return links on all alternative URLs.

ifixitmaxb:
Mason and I spoke. Here's what we decided:

- We're going to do just "en" for EU Store. It does seem perfectly viable, and perhaps even more correct to list the 29 en- variations I've listed above in a previous comment, but ultimately it seems like just "en" will work how we want it to as well so we're going to try it first (For posterity: my thought process is that since we have en-us, en-ca, en-gb, and en-au, Google should honor those explicitly for any user searching from those countries, and then otherwise any/all English speaking users (outside US, CA, UK and AU) will be shown the EU Store.)
- We're not going to list an x-default for now, mostly because it's difficult to decide whether US or EU Store should be the default ("The reserved x-default value is used when no other language/region matches the user's browser setting. This value is recommended for specifying the fallback page for users whose language settings don't match any of your site's localized versions.")
- We're going to do the specific lang/country variations for DE and FR as I outlined above in a previous comment (For posterity: it seems like this is the right approach since the DE Store is only a German-European Store and not any/all German speakers globally so it doesn't actually make sense to just do "de" for language and no country. Same goes for FR Store.)

ifixitmaxb:
A good keyword check for after this is merged is to see if the eustore.fixit.com ranks in a US SERP for "iphone repair kit". Currently, eustore.ifixit.com ranks #5 in the US.

ifixitmaxb:
@masonmcelvain it's weird/concerning to me that this page, https://www.ifixit.com/products/macbook-air-13-late-2010-2017-replacement-battery, (just an example), has fr-ch and de-be but not the other fr and de country specific combinations we outlined above...

masonmcelvain:
Good catch! That's a bug in the code, should be quick to fix.

PR #1072, opened 2022-11-28
Product pages: cache them

danielbeardsley:
Cache product pages for a minute and set "stale-while-revalidate" to 10 minutes. Product page data *doesn't* change much, so caching is fairly sensible. Let's start with 1/10 mins and see how we feel.


masonmcelvain (approved):
CR ✌🏻

danielbeardsley:
QA :+1: 
image

PR #1073, opened 2022-11-29
Product images: use srcset for thumbnails

danielbeardsley:
Product Thumbnails: use ResponsiveImage.

Previously, this ended up using  tags with urls of the 2000px
images even though the visible size was only 70px.

ResponsiveImage uses a srcset and results in the browser choosing the
correct size.

Co-author @danielcliu-ifixit 

Connect #1015

### DIFF
--- a/frontend/templates/product/sections/ProductSection/ProductOptions.tsx
+++ b/frontend/templates/product/sections/ProductSection/ProductOptions.tsx
@@ -267,22 +267,7 @@ function ProductOptionImage({ image, exactMatch }: ProductOptionImageProps) {
          </Flex>
       );
    }
-   const ratio =
-      image.width != null && image.height != null
-         ? image.width / image.height
-         : null;
 
-   if (ratio == null) {
-      return (
-         <Img
-            h="16"
-            src={image.url}
-            alt=""
-            objectFit="contain"
-            opacity={exactMatch ? 1 : 0.4}
-         />
-      );
-   }
    return (
       <Box
          h="16"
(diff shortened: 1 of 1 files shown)


frontend/templates/product/sections/ProductSection/ProductOptions.tsx
```diff
@@ -291,14 +277,12 @@ function ProductOptionImage({ image, exactMatch }: ProductOptionImageProps) {
...
       >
-         <Img
-            w="auto"
-            maxH="full"
-            maxW="full"
-            mx="auto"
+         <ResponsiveImage
             src={image.url}
             alt=""
+            layout="fill"
```
federicobadini, line 283: I wrote down some consideration regarding how srcSet are generated [here](https://github.com/iFixit/react-commerce/issues/1039#issuecomment-1330443421)

For this image we are saying that we want to fill a container whose height is 64px with an image with a suggested width of 64px. So this is some sort of fixed size image.

If instead of declaring it responsive we remove `layout=fill` and `sizes` and we provide `width` and `height`, `srcSet` will only contain pixel density specific variation instead of the whole set of `imageSizes` + `deviceSizes`

federicobadini (commented):
dev_block 👍 on a possible improvement

frontend/templates/product/sections/ProductSection/ProductOptions.tsx
danielbeardsley: Thanks for enlightening me. I took your advice.

danielbeardsley:
un_dev_block :+1:  -- implemented the suggestions from @federicobadini

danielbeardsley:
QA: Seems like the correct image is being shown now:
image

danielcliu-ifixit:
CR :v:

danielcliu-ifixit (approved):

PR #1075, opened 2022-11-29
Remove sources of FOUC

### LINKED ISSUE #1074: Fix FOUC on product list and product pages
Chakra under the hood uses Emotion which during SSR injects `style` elements in the DOM. When the app is rehydrated Emotion removes those tags and puts them inside the head.
On our server rendered markup those elements might conflict with our CSS rules and thus can potentially result in FOUC (flash on unstyled content) in the page.

To make the problem obvious to see, you can disable JS and you'd see something like this:

Screenshot 2022-11-29 at 15 48 47

As we upgrade to React 18 (#975) this problem is even more visible:

Screenshot 2022-11-29 at 15 55 58

## Do this

Make sure that CSS rules are not targeting `style` elements

dhmacs:
fixes #1074 

## QA

1. Open [Vercel preview](https://react-commerce-git-fix-fouc-ifixit.vercel.app/Parts)
2. Disable Javascript (you can do so using browser dev tools)
3. Verify that there's no unstyled content on product list and product pages (you can use #1074 screenshots as reference)

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -26,7 +26,6 @@
    "dependencies": {
       "@chakra-ui/icons": "^1.0.15",
       "@chakra-ui/react": "1.8.8",
-      "@chakra-ui/system": "1.12.1",
       "@core-ds/primitives": "2.4.5",
       "@emotion/react": "11.7.1",
       "@emotion/styled": "11.6.0",
--- a/frontend/components/common/PageBreadcrumb.tsx
+++ b/frontend/components/common/PageBreadcrumb.tsx
@@ -46,7 +46,7 @@ export function PageBreadcrumb({ items, ...otherProps }: PageBreadcrumbProps) {
             sm: 0,
          }}
          sx={{
-            '& > *': {
+            '& > :not(style)': {
                display: 'flex',
                flexWrap: 'nowrap',
             },
--- a/frontend/components/product-list/ProductListCard.tsx
+++ b/frontend/components/product-list/ProductListCard.tsx
@@ -69,7 +69,7 @@ export const ProductListCard = forwardRef<ProductListCardProps, 'div'>(
                            />
                         </Box>
                      </Flex>
-                     <Divider orientation="vertical" />
+                     <Divider orientation="vertical" borderColor="gray.200" />
                   </>
                )}
                <Flex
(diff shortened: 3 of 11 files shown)


masonmcelvain (approved):
CR ✌🏻

k0rvusk0r4x:
I am not seeing the FOUC when loading the pages with javascript turned off or on.

QA 😺

frontend/package.json
```diff
@@ -26,7 +26,6 @@
    "dependencies": {
       "@chakra-ui/icons": "^1.0.15",
       "@chakra-ui/react": "1.8.8",
-      "@chakra-ui/system": "1.12.1",
```
davidrans, line 29: This looks unrelated to the rest of these changes. Was it intentional?

dhmacs: It is intentional 👍 It's an unnecessary dependency, and I've removed it while fixing this behaviour

danielbeardsley:
Interesting to learn how emotion works with SSR. Hmm, does that mean the styles aren't applied until the app is hydrated? i.e. if the `<style>` tags render as regular elements, does that mean they don't actually have the intended affect until JS moves them into the `<head>`? If so, I'd think that's a huge blow to the promise of SSR delivering a page that is renderable before JS is evaluated.

Regarding this specific issue, it seems like it'd be better to use a global tag to hide all style tags: `<style> style { display:none }</style>`.  Thoughts?

dhmacs:
@danielbeardsley 

> Interesting to learn how emotion works with SSR. Hmm, does that mean the styles aren't applied until the app is hydrated? i.e. if the <style> tags render as regular elements, does that mean they don't actually have the intended affect until JS moves them into the <head>? If so, I'd think that's a huge blow to the promise of SSR delivering a page that is renderable before JS is evaluated.

Good question! `style` tags work even if they're in the body, so CSS is correctly applied 👍 The problem here is subtle, and it's that these emotion tags interfere with some css rules that **we** might use. To give you a concrete example, you might want to apply some style to the first child of an element `& > :first-child { color: red }`, but during server side rendering this would be applied to the `style` tag inserted by emotion not to the actual first child. This went mostly unnoticed until now because there are only a handful of cases where this happens, and rehydration quickly fixes the problem. 
I agree with you that this is a little bit of a bummer because it means the we need to be aware of this quirk when we style, and we need to write more specific selectors (e.g. `& > :not(style):first-of-type { color: red }`) 

> Regarding this specific issue, it seems like it'd be better to use a global tag to hide all style tags: <style> style { display:none }</style>. Thoughts?

Yeah we might do this, but we still need to be mindful about how we write child selectors

danielbeardsley:
> Good question! 

Thanks so much for all the insight. I understand now!

I see there's no real way to get around it given how chakra/emotion currently works on the server-side.

Issue #969, opened 2022-11-02
International Store Selector 


jeffsnyder:
Similar to International Buy Box Banner #930 but different. We also need to make sure that switching countries in the country selector takes you to the correct page.

Using the header selector in https://canada.ifixit.com/products/essential-electronics-toolkit takes you to https://www.ifixit.com/products/essential-electronics-toolkit.

![image](https://user-images.githubusercontent.com/7053450/199556845-734febfe-706d-4a99-805a-8f7ce4ae3328.png)

Using the footer selector in https://www.ifixit.com/products/essential-electronics-toolkit takes you to https://canada.ifixit.com/, _not_ the corresponding product page.

![image](https://user-images.githubusercontent.com/7053450/199556940-33505482-3aa5-49b5-8a13-3197535e199b.png)

CC @sterlinghirsh

masonmcelvain:
Though not explicitly specified in this issue, I'd imagine this is only applicable to product pages, and not collection pages?

canada.ifixit.com/collections/parts links to www.ifixit.com/store, and www.ifixit.com/Parts links to canada.ifixit.com currently. Though it would be trivial to link /Tools and /Parts correctly, it would be pretty complex to map between different collections across stores.

jeffsnyder:
We currently do not do anything special on Collections pages AFAIK, nor did we with the previous ProductList page. This only needs to be fixed on Product Detail Page. (But we will need to address this when we consolidate international stores onto the react framework.)

PR #1078, opened 2022-11-29
Add store selector product links

masonmcelvain:
Makes the links in the store selector product links for stores that the current product is enabled in. If the product is disable/doesn't exist, then we link to the homepage of the store as before.

## CR

I chose to augment the `stores` prop instead of introduce a new prop. Given that the `stores` prop could in theory be used for components besides the store selector, I'm open to adding a new prop like `storeSelectorUrlOverrides` (or any other implementation you can think of).

## QA

Pick a product page. Make sure the `product.enabled_domains` metafield has a value with store codes on ifixit test. Scroll to the footer, open the store selector, and verify that the links point directly to the product page of other stores (if the product is enabled on that store) or to the homepage of the store otherwise.

## On deploy

Change the store code of the United Kingdom store from `gb` to `uk` on strapi.ifixit.com

Closes https://github.com/iFixit/react-commerce/issues/969

### DIFF
--- a/frontend/templates/product/index.tsx
+++ b/frontend/templates/product/index.tsx
@@ -110,7 +110,7 @@ export const getServerSideProps: GetServerSideProps<ProductTemplateProps> =
       noindexDevDomains(context);
       const { handle } = context.params || {};
       invariant(typeof handle === 'string', 'handle param is missing');
-      const layoutProps = await getLayoutServerSideProps({
+      const { stores, ...otherLayoutProps } = await getLayoutServerSideProps({
          storeCode: DEFAULT_STORE_CODE,
       });
       const product = await findProduct({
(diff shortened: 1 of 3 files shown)


masonmcelvain:
deploy_block ✌🏻 until I finish repushing the US product data to have the store code in the metafield

masonmcelvain:
un_deploy_block ✌🏻

k0rvusk0r4x:
Using [this product](https://react-commerce-prod-4jci8q52e-ifixit.vercel.app/products/macbook-air-13-late-2010-2017-replacement-battery), most of the stores are linked to the correct page.

However, the UK link does not. The meta field shows it is enabled in the UK store, but the link goes to the home page.

dev_block 😿 on seeing what's up with the UK store and telling it to behave itself.

masonmcelvain:
QA ✌🏻 un_dev_block ✌🏻 

When the store code of the United Kingdom store is `uk` in Strapi, the flag still shows in the store selector and the link now points to the UK products page.

image

/products/macbook-air-13-late-2010-2017-replacement-battery

image

https://in8l-store-selector.govinor.com/admin/content-manager/collectionType/api::store.store?page=1&pageSize=10&sort=name:ASC

danielcliu-ifixit:
> Given that the `stores` prop could in theory be used for components besides the store selector, I'm open to adding a new prop like `storeSelectorUrlOverrides` (or any other implementation you can think of).

This seems like a good idea, try to separate pure store urls from the product urls, even if product urls are just store url + `/products/<handle>`. I think its a good idea to distinguish from product agnostic data and data that isn't. If that change adds a lot of extra work, lets just merge this though.

danielcliu-ifixit:
I've thought about this more and I feel less strongly that there is a need to separate the stores url from the store product urls. I think its a decent idea but def not necessary.

CR :v:

danielcliu-ifixit (approved):

erinemay:
I think https://github.com/iFixit/react-commerce/issues/1174 stemmed from this.

frontend/templates/product/index.tsx
```diff
@@ -129,8 +129,27 @@ export const getServerSideProps: GetServerSideProps<ProductTemplateProps> =
...
+      const codeToDomain =
+         product.enabledDomains?.reduce((acc, { code, domain }) => {
+            acc[code] = domain;
+            return acc;
+         }, {} as Record<string, string>) ?? {};
+      const storesWithProductUrls = stores.map((store) => {
+         const domain =
+            store.code === DEFAULT_STORE_CODE
+               ? new URL(store.url).origin
+               : codeToDomain[store.code];
```
masonmcelvain, line 141: We probably don't need to special case the default store code anymore, since all enabled store codes are included in the metafield.

```suggestion
         const domain = codeToDomain[store.code];
```

PR #1079, opened 2022-11-29
Cross Sell: Cards should be links

### LINKED ISSUE #924: Frequently Bought together: products should be links
I don't know if this was discussed when the feature was made, but I think at least *some* portion of the product card ought to be a link that takes you to a product page, and it should be obvious which part.

image

I really doubt that people will add something to their cart without even being able to go to the product page

CC @sterlinghirsh @federicobadini

masonmcelvain:
Makes the cross sell product cards links. Clicking the checkbox will select/deselect the cross sell product.

Note that `onKeyUp` was changed to `onKeyDown`. Without this change, clicking "Enter" with the checkbox focused caused the link to be followed (and the checkbox to be toggled). The expected behavior is to toggle the checkbox and not follow the link, and using `onKeyDown` achieves that.

Next link docs: https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-custom-component-that-wraps-an-a-tag

## CR

Easiest to view with whitespace hidden.

## QA

Make sure the cards are now links, and that you are still able to toggle the selections as before using the checkbox.

Closes #924

### DIFF
--- a/frontend/templates/product/sections/CrossSellSection/index.tsx
+++ b/frontend/templates/product/sections/CrossSellSection/index.tsx
@@ -438,9 +438,12 @@ function CrossSellItem({
                         >
                            <FaIcon
                               icon={faCircleCheck}
-                              color={isSelected ? 'brand.500' : 'gray.300'}
+                              color={isSelected ? 'brand.500' : 'gray.300'}                        
                               h="6"
                               transition="color 300ms"
+                              _hover={ isSelected ? {color: 'gray.300'} :
+                                 {color: 'brand.500'}
+                              }
                            />
                         </Center>
                      </Box>
--- a/frontend/templates/product/sections/CrossSellSection/index.tsx
+++ b/frontend/templates/product/sections/CrossSellSection/index.tsx
@@ -97,7 +97,7 @@ export function CrossSellSection({
          reviewsCount: product?.reviewsCount,
       };
    }, [
-      product?.handle,
+      product.handle,
       product?.rating?.value,
       product?.reviewsCount,
       product.title,
@@ -265,6 +265,7 @@ export function CrossSellSection({
                         </Box>
                      </Box>
                      <Button
+                        disabled={!selectedCrossSellVariantIds.length}
                         colorScheme="brand"
                         minW="240px"
                         onClick={handleAddToCart}
--- a/frontend/templates/product/sections/CrossSellSection/index.tsx
+++ b/frontend/templates/product/sections/CrossSellSection/index.tsx
@@ -443,9 +443,7 @@ function CrossSellItem({
                               h="6"
                               transition="color 300ms"
                               _hover={
-                                 isSelected
-                                    ? { color: 'gray.300' }
-                                    : { color: 'brand.500' }
+                                 isSelected ? { color: 'brand.700' } : undefined
                               }
                            />
                         </Center>
--- a/frontend/templates/product/sections/CrossSellSection/index.tsx
+++ b/frontend/templates/product/sections/CrossSellSection/index.tsx
@@ -3,6 +3,7 @@ import {
    Badge,
    Box,
    Button,
+   Center,
    Circle,
    Divider,
    Flex,
@@ -30,6 +31,7 @@ import {
 } from '@ifixit/ui';
 import { MoneyV2 } from '@lib/shopify-storefront-sdk';
 import { Product, ProductVariant } from '@models/product';
+import NextLink from 'next/link';
 import React from 'react';
 
 export type CrossSellSectionProps = {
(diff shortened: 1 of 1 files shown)


deltuh-vee:
dev_block ✊ 
Attempting to add out of stock items to the cart results in a console error:
image

https://user-images.githubusercontent.com/5289715/204681516-473934ea-002f-49a4-a5a9-779a84f70a05.mp4


I was still able to add the cross sold item on master, even though the main item was out of stock.

masonmcelvain:
I'm seeing the same behavior on main (react-commerce.vercel.app on the right) for `/products/xbox-series-x-power-and-eject-button-board` as this branch (on the left).

image

[Screencast from 11-30-2022 09:48:46 AM.webm](https://user-images.githubusercontent.com/52104630/204827191-07cb9d16-c977-4c63-a874-1b1aed659a66.webm)

un_dev_block ✌🏻 Lmk if you're still seeing the same behavior, it's not obvious from your console screenshots what went wrong there 🤔

deltuh-vee:
QA 🎬 
Cross sell cards are now links, and you are still able to toggle the selections as before using the checkbox.

ardelato (approved):
CR 👍 but deploy_block 🎨 on minor UI suggestion

We should add a hover effect around the checkboxes to inform the user that a different action will be performed when clicking the checkbox vs clicking the card.


Here I just set the `opacity` to `50%` when hovering over the checkbox.

https://user-images.githubusercontent.com/22064420/205997446-9d0573f2-ea67-421a-9996-b3975934c89a.mp4

frontend/templates/product/sections/CrossSellSection/index.tsx
```diff
@@ -89,11 +91,17 @@ export function CrossSellSection({
...
    const currentProduct = React.useMemo<CrossSellProduct>(() => {
       return {
+         handle: product.handle,
          title: product.title,
          rating: product?.rating?.value,
          reviewsCount: product?.reviewsCount,
       };
-   }, [product?.rating?.value, product?.reviewsCount, product.title]);
+   }, [
+      product?.handle,
```
danielcliu-ifixit, line 100: do we need the `?`

erinemay:
>We should add a hover effect around the checkboxes to inform the user that a different action will be performed when clicking the checkbox vs clicking the card.
>
>Here I just set the opacity to 50% when hovering over the checkbox.

I agree that the usability here is a little odd, but I think the universal this-is-a-clickable-action cue is that a button becomes darker. I think making it more opaque usually sends the message that that item is non-clickable.

deltuh-vee:
dev_block ✊ 
It might be better to have the product title be an underlined link and have the rest of the card toggle the selection for a product. Thoughts?

danielcliu-ifixit:
CR :v:

danielcliu-ifixit (approved):

erinemay:
> dev_block ✊ It might be better to have the product title be an underlined link on hover and have the rest of the card toggle the selection for a product so we don't deviate too much from the current behavior. Thoughts?

@deltuh-vee, I agree, but... From the issue: 
> Looking at Amazon, the entire card links to the product, and the only the checkbox let's you deselect the product. We might do the same. Thoughts?

>E-commerce wants to match Amazon closely.

erinemay:
>1.) awkwardly opting them into second purchase by selecting it be default 2.) having it default to selecting the current product as well by default. This is fine in concept but i could easily see someone adding to cart at the top, then scrolling down and seeing a useful addition in the Cross Sell section and hitting the button there too, without realizing they now have 2 of the current product in their cart.

I agree here too. I think I remember a conversation about this. It worked ok on the old product pages because cross sells were way higher up. When you got to the page, you could make your cross-sell decision without scrolling. 

From what I remember, we put a pin in either bumping up the cross-sell on the page or mucking with the behavior. @sterlinghirsh, do you remember the outcome of that decision?

frontend/templates/product/sections/CrossSellSection/index.tsx
```diff
@@ -303,137 +313,155 @@ function CrossSellItem({
...
+                           onClick={(event) => {
+                              onChange(!isSelected);
+                              event.preventDefault();
+                              event.stopPropagation();
                            }}
-                           top={{
-                              base: 'auto',
-                              md: 4,
+                           onKeyDown={(event) => {
+                              if (['Enter'].includes(event.code)) {
```
danielbeardsley, line 427: Is there a reason for this odd condition? It seems an odd way to do an equality test.

masonmcelvain: Good eye I didn't notice that. Looking at other examples of this pattern in this repo, it seems to be used when we expect to add another value to the array in the future. Given that we probably wouldn't add any other keypress to this array, I agree that checking equality with `===` makes more sense.

masonmcelvain:
Wow lots of activity! Thanks for all the feedback.

I ended up following the default checkbox hover pattern that Chakra uses.

un_dev_block ✌🏻 un_deploy_block ✌🏻

deltuh-vee:
QA 🎬 
* Cross sell Cards are now links
* You are still able to toggle the selections as before using the checkboxes
* Checkboxes will dim a little bit when hovered over

PR #1083, opened 2022-11-30
Implement fetch middleware for Sentry logging

### LINKED ISSUE #948: Sentry: use `sentryFetch()` over `fetch()`  everywhere
### Problem
`sentryFetch()` is a wrapper around fetch that adds context and reports errors appropriately. We don't use it everywhere.

### Do This
Use it everywhere unless there is good reason not to.

dhmacs:
closes #948 

Instead of spreading `sentryFetch` everywhere in the codebase I'm proposing here to add a sentry fetch middleware that transparently tracks all fetch requests. 

## QA

@masonmcelvain suggestions on how to QA this?


masonmcelvain:
Great work @dhmacs! QA suggestion:

Since we have control over the ifixit api, let's force the cart api to return a 4xx level response, and make sure it's logged in Sentry.

* Checkout `master` on your cominor subdomain.
* Go to the `store/user/cart` api controller, and force it to return a 4xx level response (401's and 404's won't go to Sentry, so don't use those).
* Set `NEXT_PUBLIC_IFIXIT_ORIGIN=https://your-subdomain.cominor.com` in `frontend/.env.local`.
* Start the Next.js dev server and visit any page (we make cart requests on every page).
* Verify that there is a console error about the cart request 4xx, and check react-commerce-dev in Sentry for the corresponding event.

masonmcelvain (approved):
CR ✌🏻 Great work @dhmacs!

deltuh-vee:
QA 🎬 
Errors from the ifixit API can be sent to Sentry from react-commerce.

packages/sentry/index.tsx
```diff
@@ -30,14 +59,26 @@ export const sentryFetch: typeof fetch = async (resource, options) => {
...
          throw error;
-      });
+      }
+   };
+
+const shouldSkipReporting: SkipRequestFn = (input, init) => {
+   const url = getRequestUrl(input);
+   // We have custom logic in place for reporting errors only after React Query retries have failed,
+   // so we don't want to report errors for the cart API here.
+   return url.includes('/store/user/cart');
```
danielbeardsley, line 73: I think I'd be more inclined to have our overridden `fetch()` take a `reportErrors: false` in it's `init` options instead of a regex on urls. I think the explicitness of that choice in the calling code makes it easier to understand.

PR #1085, opened 2022-11-30
Product Page: Transition to multiple hreflang locales

masonmcelvain:
We're transitioning the product enabled_domains metafield to support multiple locales per store. Prefer the `locales` field if available, otherwise fallback to the `locale` field.

## QA

Make sure we still get the same hreflang link tags in the `<head>` as on main.

Connects https://github.com/iFixit/ifixit/issues/45707

### DIFF
--- a/frontend/models/product.ts
+++ b/frontend/models/product.ts
@@ -440,8 +440,10 @@ function parseOemPartnership(
 
 const EnabledDomainsSchema = z
    .object({
+      code: z.string(),
       domain: z.string().url(),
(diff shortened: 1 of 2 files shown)


danielbeardsley (approved):
CR :+1:

masonmcelvain:
I was just gonna deploy_block this, oops my bad. Shouldn't matter much since the original behavior is preserved. If we decide not to use multiple locales per store, I'll revert this.

PR #1086, opened 2022-11-30
Use iFixit client for every iFixit related request

### LINKED ISSUE #949: iFixit API: use `IFixitAPIClient` everywhere

dhmacs:
closes #949 

The only instance I could find where iFixit client was not used is the newsletter sdk

## QA

I don't think is necessary since newsletter is already covered by integration tests

qa_req 0

### DIFF
--- a/packages/newsletter-sdk/index.tsx
+++ b/packages/newsletter-sdk/index.tsx
@@ -1,7 +1,6 @@
 import { trackInMatomoAndGA } from '@ifixit/analytics';
-import { useAppContext } from '@ifixit/app';
 import { isError } from '@ifixit/helpers';
-import { sentryFetch } from '@ifixit/sentry';
+import { useIFixitApiClient } from '@ifixit/ifixit-api-client';
 import * as React from 'react';
 
 /**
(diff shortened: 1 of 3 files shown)


packages/newsletter-sdk/index.tsx
```diff
@@ -66,27 +67,3 @@ export function useSubscribeToNewsletter(): [Subscription, SubscribeFn] {
...
-   const response = await sentryFetch(
-      `${apiOrigin}/api/2.0/cart/newsletter/subscribe`,
-      {
-         method: 'POST',
-         body: JSON.stringify({
-            email,
-         }),
-      }
-   );
-   if (response.status >= 300) {
```
danielbeardsley, line 87: I looked at the source and this api always returns 200, so I'm fine that we dropped this code-specific error handling.

danielbeardsley (approved):
CR :+1:

PR #1091, opened 2022-11-30
Collections: hide children that link to empty collections

### LINKED ISSUE #822: ProductList: Do not include "empty" ProductLists in gallery of Children
The E-commerce Team doesn't have a good way to know when we should publish new ProductList Collections.

When we create a Product for a new Device, we create the Device and we create the ProductList. We then leave the ProductList disabled until the compatible Products are activated in our store. If we enabled the ProductList as soon as we create it, we'd have empty ProductList pages super prominently linked in lists of Children. If we wait to enable the ProductList when we enable the Product, we need to keep track of which new/disabled Products will require ProductList enabling when the Product is enabled ... that's a rather annoying thing to keep track of. 

We'd like to be able to create, fully configure, and publish ProductLists in one step. And then have the ProductList "show up" in a list of Children once it has Products enabled. 

image

We already noindex ProductLists with 0 Products. 

Please drop ProductLists with 0 Products from the list of Children. 

CC @dhmacs @sterlinghirsh

masonmcelvain:
Greys out parts product list children links that link to collections that are empty.

Intersects the available device facets with the current product list children to determine which children are non-empty. This will not work for pages that have >200 device facets.

## QA

On parts collection pages (/Parts/**), links to child device collection pages that are empty should be greyed out and sorted to the end. Sorting within each group (nonempty and empty) should be preserved.

When selecting/deselecting filters, the empty children should be dynamically updated.

Closes #822


jordycosta:
https://www.ifixit.com/Parts/Garmin_Nuvi_2595LMT is an example

jordycosta:
QA 🟢 

Collection(s) children that are empty are no longer displayed 👍 

### Testing Procedure:
1. Navigate to the Vercel preview
2. Go to an parent page that contains an empty collection*
3. Ensure that the empty collection is not displayed anymore


> *Note: An example would be, **[Epson Projector Parts](https://www.ifixit.com/Parts/Epson_Projector)**, or **[Garmin GPS Parts](https://www.ifixit.com/Parts/Garmin_GPS)**

---

### Screenshots:

main_new

preview

frontend/templates/product-list/sections/ProductListChildrenSection.tsx
```diff
@@ -33,9 +38,21 @@ export function ProductListChildrenSection({
       [isShowingMore]
    );
 
+   const { items: deviceMenuItems } = useMenu({
+      attribute: 'device',
+      limit: PRODUCT_LIST_MAX_FACET_VALUES_COUNT,
+   });
```
dhmacs, line 44: I think it's better if we use our `useMenuFacet` here to make sure that the Algolia widget gets created with the same parameters everywhere so we avoid weird behaviors

dhmacs (changes requested):
dev_block 🔢 on a suggested change

frontend/templates/product-list/sections/ProductListChildrenSection.tsx
masonmcelvain: I had considered that, but doing so would display at most 10 empty children links (which defeats the purpose of the PR), since `limit=10` in `useMenuFacet` for the `device` facet.

This implementation intersects the device facet items with the children links to determine which children links are empty, so we want all possible `device` facet items that are applicable to the current collection. The assumption is that if the `device` facet item is not present, then there are no products in the collection that work in that device, and so we should exclude the device from the child link section.

However, now I realize this assumption does not hold if other facets are selected, since the list of applicable device facets will dwindle as the collection is filtered.

image

masonmcelvain: @sterlinghirsh It doesn't seem like Algolia offers a way to ask "Give me all the facet item values (i.e. device titles) for the unfiltered state of the current collection", meaning that this solution might not work.

sterlinghirsh: Seems we have two options. One is that we do a second query for the full unfiltered list. The other is that we accept that other filters may have an effect on the model list. I'm okay with the latter as long as the UI doesn't jump around too much.

dhmacs: We should probably do this filtering server side only, in the product list model so we're not affecting in any possible way client side filtering

masonmcelvain: > One is that we do a second query for the full unfiltered list

To keep this request separate from the instantsearch state, we could use `fetch` or install algolia's [JS api client](https://www.algolia.com/doc/api-client/getting-started/install/javascript/?client=javascript) to get the list of device titles on the server.

> The other is that we accept that other filters may have an effect on the model list.

This feels strange given that we clear all filters when users click on a child device link.

> as long as the UI doesn't jump around too much

The layout shift doesn't seem like great UX.

[Screencast from 12-02-2022 10:43:50 AM.webm](https://user-images.githubusercontent.com/52104630/205330992-7f1f92d3-d5dd-4cb3-8ccd-d8f90ceed765.webm)

sterlinghirsh: One option is that when we remove child productlist options that were previously visible, we gray them out and sort to the bottom instead of hiding them. Not quite as clean, but it reduces the layout shift.

masonmcelvain: > One option is that when we remove child productlist options that were previously visible, we gray them out and sort to the bottom instead of hiding them. Not quite as clean, but it reduces the layout shift.

@sterlinghirsh I tried this suggestion in https://github.com/iFixit/react-commerce/pull/1091/commits/43f3bba03d48519237ecd975d23ee37d651be639. I think the UI looks good, but it appears there are sometimes inconsistencies between the device titles on Algolia and the device titles on Strapi.

For example, on https://react-commerce-git-no-empty-child-collections-ifixit.vercel.app/Parts/Electronics "VR Gear" is greyed out, but it has plenty of products when you click on it. I'm not sure if it's just out of date data in Algolia, or an actual discrepancy between the two.

sterlinghirsh: I see that too, though I think this is a data issue. On electronics parts, searching for vive doesn't return all the vive parts. I think on cominor Virtual Reality Headset is not a child of Electronics or something.

sterlinghirsh: But on the live site it appears to work.

masonmcelvain:
un_dev_block ✌🏻 

> We should probably do this filtering server side only, in the product list model so we're not affecting in any possible way client side filtering

@dhmacs could you elaborate on this? I am not sure how to implement what you describe.

erinemay:
I was previewing this theme just to check out the new behavior, and I noticed a couple of errors roll in on Sentry. I saw a couple of them in my console when I returned to the filtered collection page via the back button, but I wasn't paying too close attention to my steps, and I can't seem to reproduce them right now. 

https://sentry.io/organizations/ifixit/issues/3823702396/?project=6475315
https://sentry.io/organizations/ifixit/issues/3823702400/?project=6475315
https://sentry.io/organizations/ifixit/issues/3823702425/?project=6475315

Gonna try a little more at reproducing.

jordycosta:
QA 🟢 

It appears that the new behavior is now implemented:

> One option is that when we remove child productlist options that were previously visible, we gray them out and sort to the bottom instead of hiding them. Not quite as clean, but it reduces the layout shift.

gray

### Screen Capture:

<details>
<summary> Toggle View </summary> <br>

firefox_9lM7Dkv0jo
</details>

---

deploy_block 🟨 on the new Sentry errors in https://github.com/iFixit/react-commerce/pull/1091#issuecomment-1358581362

I also found a bug that _appears_ unrelated, but new from the last time I checked out this PR:

firefox_HnGZyLUHwq

erinemay:
> I see that too, though I think this is a data issue. On electronics parts, searching for vive doesn't return all the vive parts. I think on cominor Virtual Reality Headset is not a child of Electronics or something.

https://github.com/iFixit/react-commerce/pull/1091#discussion_r1051253289

I'm seeing a bunch grayed out in the prod preview:
https://react-commerce-prod-git-no-empty-child-collections-ifixit.vercel.app/Parts/Electronics
image

erinemay:
Note: I'm also getting [repeated Failed to fetch errors](https://sentry.io/organizations/ifixit/issues/3789683451/?project=6475315) in the prod preview, but it's [also happening on the main prod preview](https://react-commerce-prod.vercel.app/Parts/Electronics).

sterlinghirsh:
I think this is another case of a mismatch between the product list tree and the device tree. We show Drone parts under Electronics parts, but the Drone device is under Vehicle > Aircraft. https://www.ifixit.com/Device/Drone

masonmcelvain:
> I think this is another case of a mismatch between the product list tree and the device tree. We show Drone parts under Electronics parts, but the Drone device is under Vehicle > Aircraft. https://www.ifixit.com/Device/Drone

@sterlinghirsh if we are to go forward with this solution, how should we handle these special devices without hardcoding them?

> I noticed a couple of errors roll in on Sentry

@erinemay do you think those errors are different than the hydration errors we are seeing in production? I can't tell from viewing the Sentry event.

dhmacs:
> > We should probably do this filtering server side only, in the product list model so we're not affecting in any possible way client side filtering
> 
> @dhmacs could you elaborate on this? I am not sure how to implement what you describe.

@masonmcelvain I've missed the notification about this. What I meant is that in `getServerSideProps` for the product list template we use the Algolia api to compute this data, and then we just send it as props. Which means that this would be a separate query from the product list query, can be cached server side, won't be affected by client side filters and won't require extra client-side queries.

masonmcelvain:
Thanks for the clarification @dhmacs. I'm going to close this in favor of https://github.com/iFixit/react-commerce/pull/1252, which attempts to implement the solution you describe.

PR #1092, opened 2022-11-30
Footer: defer to client-side

danielbeardsley:
The footer is a lot of html nearly (40KB) and since it's usually far below the fold, let's defer it to rendering on the client side.

To support this pattern I've added a hook that should help:

* Returns false on the server side
* Returns false during hydration on the client
  * React requires the first client render produce the exact same DOM state during hydration
* Re-renders right after hydration
* Returns true on every further re-render/re-mount, without returning false before `useEffect()` runs

Note: the footer is only mounted once, but this will be useful for other components that are unmounted when switching products.

Connect: #1015


danielbeardsley:
I'm interested in feedback from @iFixit/shopify and @dhmacs / @federicobadini

packages/ui/hooks/index.ts
```diff
@@ -74,6 +74,27 @@ export function useIsMounted() {
    return isMounted;
 }
 
+const isClient = typeof window !== 'undefined';
+let hydrationComplete = false;
```
masonmcelvain, line 78: What is the purpose of the `hydrationComplete` variable?

packages/ui/hooks/index.ts
```diff
@@ -74,6 +74,27 @@ export function useIsMounted() {
...
+
+/**
+ * Returns true on all client-side renders *after* hydration is complete.
+ * Useful for deferring some components to the client-side while allowing later
+ * mounts to render immediately (instead of waiting for useEffect()).
+ */
+export function useIsClientSide() {
```
masonmcelvain, line 85: I don't understand the benefit of this hook over checking `typeof window !== 'undefined'` in the Footer itself. Is the intent to let components above the fold render initially (and the footer is blank), and defer rendering the real Footer until after the first render?

masonmcelvain (approved):
Seems like it will work, I will hold off on CR to let others review.

packages/ui/hooks/index.ts
danielbeardsley: React mandates that the first render (the hydration step) produce exactly the same DOM structure that is already present in the DOM. So, the first pass (hydration) has to result in `false` even when on the client.

The `useEffect()` is run right after the hydration finishes and that changes both the `isClientSide` (triggering a re-render) and the `hydrationComplete = true`. This ensures re-renders result in `true`, and the `hydrationComplete` ensures that re-mounts also result in `true` instead of `false` before `useEffect()`.

packages/ui/hooks/index.ts
danielbeardsley: I think the below comment actually addresses this.

packages/ui/hooks/index.ts
masonmcelvain: Thanks for the explanation! I misunderstood hydration to mean "attaching event listeners after putting the html on the page" and I hadn't considered remounting.

danielbeardsley: > misunderstood hydration to mean "attaching event listeners after putting the html on the page"

Practically that's the end result, but it accomplishes this by performing a full render of your page, evaluating all your components, connecting existing html elements with the structure returned by the first render. If the structures don't match 100%, react complains.

danielbeardsley:
FYI, there's an open GH on react's repo about this very thing:

https://github.com/facebook/react/issues/23068

dhmacs:
@federicobadini will look into this

k0rvusk0r4x:
Is there anything we can do here beyond making sure the footer still renders?

danielbeardsley:
QA:

* Footer still renders and the height is reasonable with JS disabled (load the page with it disable / enabled in separate tabs and compare
* Expect this branch to have blanked footer and blanked "rating stars" without JS (mostly visible in the "Featured Products" section at the bottom of the product page
* Expect scroll height to not change much as the footer placeholder should be the same height at different browser widths.

federicobadini:
I was thinking that this may have a series of implications: 

1. The footer has many important links in it, so I'm a bit concerned this may have SEO implications
2. The component is still on page, so it will be bundled together with the rest of the page code. This means that transfer time, parsing time and evaluation time will not be affected. 

Moreover, inspecting Vercel analytics, we have currently better performances for mobile (low computation power) than desktop (higher compute power) so I would say that there is something else that is affecting load time independently from the amount of computing power available.

If you agree, I would suggest to dev_block ⏸ this and keep the focus on the CLS + LCP issues that are still affecting both product pages and collection pages. We have corrected some of them but not all of them.
In parallel we have to investigate if there are other delays along the response chain

danielbeardsley:
I merged in main.

I think this is worth considering.

This removes 40KB of html from the initial page download. I don't think content that is deferred to be rendered in JS is invisible to search engines. If so, then every website using react that doesn't also use next.js would be invisible and I'm sure that's not the case.

That said, after gzipping, the footer is much smaller, so this is removing 40KB of html, but only ~10kb of network bytes. I'm also not 100% confident in the exact timing of react's hydration step: does hydration happen for the whole page at the same time or do some pieces get hydrated, have their `useEffect()` calls fired, then other pieces get hydrated.

dhmacs:
> This removes 40KB of html from the initial page download. I don't think content that is deferred to be rendered in JS is invisible to search engines. If so, then every website using react that doesn't also use next.js would be invisible and I'm sure that's not the case.

The information I have on this topic is that the content is visible to Google's crawler, but it is penalized because it is more difficult for the crawler to parse. Specifically, the crawler initially analyzes the downloaded HTML and then, in a second pass, uses a tool like puppeteer to evaluate the JS. We previously tested this on a customer's website and found that implementing server-side rendering resulted in an improvement in ranking.

It's worth noting that reducing the size of the HTML does not have a significant impact on website performance. However, the size of the JS bundle does affect performance as it must be parsed and evaluated by the browser, which can slow down the page. It appears that this change does not save any bytes in the JS.

sterlinghirsh:
I think you're right that the html size is not the most impactful compared with the js, though it is always nice to save bytes if they're wasted, and a 10-20% reduction in html size seems like it would be worthwhile if there aren't significant downsides (there may yet be). I would rather bring down the js size, but unless there's a good way to do that which is mutually exclusive to reducing the html. Maybe a server component would do that? I don't really know the tradeoffs yet. We should run some lighthouse tests and see if we detect any changes in the scores between this branch and main, but I'd be surprised if a 10-20% reduction in html size had no measurable effect (good or bad), despite the ~170b extra js.

While it's true the content in the footer will link to other pages on the site, which has SEO implications, I'm not sure whether client-rendering that would matter (much?) since our pages typically wouldn't want to get indexed with the content of the footer anyway.

danielbeardsley:
I re-read the commit messages and this ends up deferring 40kb for the header and 60kb for SVG stars. So 100kb less html which is about 20% of the document.

danielbeardsley:
Maybe it's worth trying this again as an A/B test, but I'll close it for now.

PR #1106, opened 2022-12-06
Make hreflang key unique

masonmcelvain:
Fixes a bug where only one locale per store was showing up in the hreflangs.

## QA

Make sure all the different locales are showing up for the FR and DE stores in the hreflang tags in teh `<head>`. (list is [here](https://github.com/iFixit/react-commerce/pull/1068#issuecomment-1332965850))

Connects #1068


masonmcelvain:
QA ✌🏻 

Feature:
image

Main:
image

Issue #1111, opened 2022-12-07
Migrate from `Cypress` to `Playwright`


ardelato:
## Description

This was originally proposed as an idea in [ifixit.slack.com/archives/C089J162J/p1669838499336999](https://ifixit.slack.com/archives/C089J162J/p1669838499336999) where a core benefit of switching over to `Playwright` was a speed increase. It should be noted, there are of course other benefits to switching to `Playwright` besides a speed increase.
 - https://mtlynch.io/notes/cypress-vs-playwright/
 - https://twitter.com/housecor/status/1597954687705104385
 - Better cross-browser support: Playwright allows us to easily run tests on multiple browsers, including Chrome, Firefox, and Edge. This will help ensure that our tests are more comprehensive and less susceptible to browser-specific issues.
 - Easier to use: Playwright has a more intuitive API and simpler syntax than Cypress, which will make it easier for new team members to get up to speed with our testing setup.

After some experimentation in #1090, we found that it would be relatively easy to migrate our current `Cypress` tests to `Playwright` without significant changes to our test suite. 

## To Do:

That said, we will need to do the following over the next few sprints:

- [x] #1112
- [x] #1113
- [x] #1114
- [x] Testing the migrated tests to ensure that they are working properly with Playwright.
- [x] #1239

PR #1115, opened 2022-12-07
Playwright: Add Supporting Infrastructure 

ardelato:
## Description

This updates our test infrastructure to support writing and running Playwright tests. We will address integrating Playwright into our CI workflow in later pulls, for now, this will add the basic setup to begin writing tests and running them locally. As we continue the migration process, #1111, we will modify the `Playwright` config file to fit our testing needs. 

## QA Notes:
You can add the following test to `Playwright` to assert the expected behavior:
```Typescript
import { test, expect } from '@playwright/test';

test('homepage has title and links to intro page', async ({ page }) => {
  await page.goto('https://playwright.dev/');

  // Expect a title "to contain" a substring.
  await expect(page).toHaveTitle(/Playwright/);

  // create a locator
  const getStarted = page.getByRole('link', { name: 'Get started' });

  // Expect an attribute "to be strictly equal" to the value.
  await expect(getStarted).toHaveAttribute('href', '/docs/intro');

  // Click the get started link.
  await getStarted.click();

  // Expects the URL to contain intro.
  await expect(page).toHaveURL(/.*intro/);
});

```

- [ ] Confirm `Cypress` and `Jest` do not pick up tests in the `playwright` subfolder
- [ ] Confirm `Cypress` and `Jest` tests still work
- [ ] Confirm running `pnpm playwright:run` will pick up and run tests in the `playwright` subfolder
	- [ ] Confirm that you do not need the server running for the `Playwright` tests to be ran
	- [ ] Confirm that if you do have a server running, the `Playwright` tests will still work

Closes: #1112


frontend/playwright.config.ts
```diff
@@ -108,10 +108,14 @@ const config: PlaywrightTestConfig = {
...
-   //   command: 'npm run start',
-   //   port: 3000,
-   // },
+   webServer: {
+      cwd: '../',
+      command: 'pnpm run dev',
+      port: 3000,
+      timeout: 120 * 1000,
+      /* Reuse the same server if on local dev */
+      reuseExistingServer: !process.env.CI,
```
batbattur, line 117: Is this `webServer` going to start the `frontend` app (assuming the port is 3000)? If yes, then we need to have both frontend and backend to be running. If not, then I believe the port for the backend is different.

Also if a dev server is already running for the frontend app, does this mean Playwright won't start a new one and reuse an existing server?

batbattur (dismissed):
CR 🚛 
deploy_block 🚚 on question https://github.com/iFixit/react-commerce/pull/1115/commits/c3b4ce40a0a94bbaacf1193d6cf7500d79ad45cb#r1042771490

batbattur:
un_deploy_block 🚒

frontend/playwright.config.ts
```diff
@@ -0,0 +1,121 @@
...
+   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
+   reporter: [
+      [
+         'html',
+         {
+            open: 'never', // will not try to automatically open the report in the browser if test fails
+            outputFolder: 'tests/playwright/test-results/reports/',
+         },
+      ],
+   ],
```
batbattur, line 41: Just curious if we will ever need the `html` reporter? I think the list reporter might be faster (I'm just guessing this since it might take longer to generate the html). 

https://playwright.dev/docs/test-reporters#list-reporter

It looks like we can even change the reporter depending on the `process.env` like this `reporter: process.env.CI ? 'dot' : 'list'`.

batbattur: I think we can address this later on when we set up the workflow for Playwright.

dhmacs (changes requested):
LGTM 👍 Only one minor change

dev_block 🔢

ardelato:
>Also I'm curious, are you moving from test to tests because you want to use test folder for Playwright?

We will use `tests` for Playwright as well. The reason for moving `test` to `tests` was more a grammatical choice than anything else. We currently have `Jest` and `Cypress` tests, and will soon have `Playwright` tests, within that folder so it seemed to make more sense to rename it to `tests` to indicate that.

ardelato:
un_dev_block ⚡ , deploy_block ☁️ on Massimo's question being answered

deltuh-vee:
QA 🎬 
un_dev_block 👍 
- [x] Confirmed Cypress and Jest do not pick up tests in the playwright subfolder
- [x] Confirmed Cypress and Jest tests still work 
    * Cypress tests are failing for me due to timeouts, but that's also on master and we're moving away from Cypress anyways. 
- [x] Confirmed running pnpm playwright:run will pick up and run tests in the playwright subfolder
    - [x] Confirmed that you do not need the server running for the Playwright tests to be ran
    - [x] Confirmed that if you do have a server running, the Playwright tests will still work

        * I wasn't able to run the safari version, but the other Playwright tests passed 😢 
image

batbattur:
dev_block 🚒 on resolving merge conflicts

ardelato:
un_dev_block ⚡ merge conflicts resolved

batbattur (approved):
CR 🚛

deltuh-vee:
Carry over QA 🎬

ardelato:
> > Also I'm curious, are you moving from test to tests because you want to use test folder for Playwright?
> 
> We will use `tests` for Playwright as well. The reason for moving `test` to `tests` was more a grammatical choice than anything else. We currently have `Jest` and `Cypress` tests, and will soon have `Playwright` tests, within that folder so it seemed to make more sense to rename it to `tests` to indicate that.

@dhmacs does this answer your question?

dhmacs:
@ardelato yes! un_deploy_block 🟢

mlahargou (approved):
CR 📱 

This looks great!

ardelato:
QA 🎠 carry-over

PR #1122, opened 2022-12-08
Handle top level shopify graphql api errors

### LINKED ISSUE #1014: Shopify GraphQL: Report exception if `response.errors`
We recently dealt with a situation where the graphql response didn't contain `userErrors` but a higher-level `.errors` array and our client just ignores that and passes `null` to the requestor.  See pull #1011 We were missing out on this graphql error and are liking missing other errors (or breaking the page instead of handling the error).

### Do This
* Gracefully handle a null response to all shopify graphql queries
* Alter graphql client to check for and report errors to sentry
   * Consider throwing an exception? Though a nullable response ought to be easier to enforce the checking of
   * There may be some false positives, but we'll deal with those as they come

CC @sterlinghirsh

masonmcelvain:
Handles an undocumented top level `errors` key that is sometimes returned by the Shopify API. When this is the case, the field of interest (e.g. `checkoutCreate`) can be null.

Example event reported by this new code: https://sentry.io/share/issue/15b33ce210ca49e0bd64328b4994e15d/

## QA 

Make sure you can still make two purchases as described in https://github.com/iFixit/react-commerce/issues/1002  and fixed by https://github.com/iFixit/react-commerce/pull/1011

CC @danielbeardsley 

Closes #1014

### DIFF
--- a/packages/cart-sdk/hooks/use-checkout.ts
+++ b/packages/cart-sdk/hooks/use-checkout.ts
@@ -151,10 +151,12 @@ function useStandardCheckout() {
       const localCheckout = getLocalCheckout();
       let checkoutUrl: string | null = null;
       if (localCheckout != null) {
-         const checkout = await updateCheckout(localCheckout.id, lineItems);
-         if (checkout != null) {
-            checkoutUrl = checkout.webUrl;
-         }
+         try {
+            const checkout = await updateCheckout(localCheckout.id, lineItems);
+            if (checkout != null) {
+               checkoutUrl = checkout.webUrl;
+            }
+         } catch (error) {}
       }
       if (checkoutUrl == null) {
          const checkout = await createCheckout(lineItems);
--- a/packages/cart-sdk/hooks/use-checkout.ts
+++ b/packages/cart-sdk/hooks/use-checkout.ts
@@ -1,7 +1,7 @@
 import { trackInMatomoAndGA } from '@ifixit/analytics';
 import { useAppContext } from '@ifixit/app';
 import { useAuthenticatedUser } from '@ifixit/auth-sdk';
-import { assertNever, invariant, isError } from '@ifixit/helpers';
+import { assertNever, isError } from '@ifixit/helpers';
 import { useIFixitApiClient } from '@ifixit/ifixit-api-client';
 import { useShopifyStorefrontClient } from '@ifixit/shopify-storefront-client';
 import * as React from 'react';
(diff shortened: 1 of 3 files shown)


packages/cart-sdk/hooks/use-checkout.ts
```diff
@@ -151,12 +151,10 @@ function useStandardCheckout() {
       const localCheckout = getLocalCheckout();
       let checkoutUrl: string | null = null;
       if (localCheckout != null) {
-         try {
-            const checkout = await updateCheckout(localCheckout.id, lineItems);
-            if (checkout != null) {
-               checkoutUrl = checkout.webUrl;
-            }
-         } catch (error) {}
```
danielbeardsley, line 159: Any idea why this catch was here to begin with?

I don't think we should drop it without knowing why it was added and if that's still sensible.

danielbeardsley (approved):
CR :+1: 

depoy_block :+1:  on blaming the catch-all. I certainly don't like catch-alls, but it's worth looking into why it is there.

packages/cart-sdk/hooks/use-checkout.ts
deltuh-vee: This was the pull that added the `catch` https://github.com/iFixit/react-commerce/pull/1011

masonmcelvain: Thanks for finding the commit. I think you're right, we should keep the catch-all so that we don't skip creating a new checkout if the update fails.

Since we're doing our best to handle API errors in the storefront client, we should still hear about interesting updateCheckout failures, and the catch-all will filter out network failures and weird browser incompatibilities.

packages/cart-sdk/hooks/use-checkout.ts
```diff
@@ -151,12 +151,10 @@ function useStandardCheckout() {
...
+         const checkout = await updateCheckout(localCheckout.id, lineItems);
+         if (checkout != null) {
+            checkoutUrl = checkout.webUrl;
+         }
```
masonmcelvain, line 157:
```suggestion
         try {
            const checkout = await updateCheckout(localCheckout.id, lineItems);
            if (checkout != null) {
               checkoutUrl = checkout.webUrl;
            }
         } catch (error) {}
```

masonmcelvain:
un_deploy_block ✌🏻

danielbeardsley (approved):
CR :+1:

deltuh-vee:
QA 🎬 
I'm still able to complete the checkout with two orders from https://github.com/iFixit/react-commerce/issues/1002#issuecomment-1307568834

PR #1125, opened 2022-12-09
Product Reviews: treat rating as float

danielbeardsley:
Rating is an average of a bunch of ints and thus is a float!

Otherwise, 4.9 is rounded down to 4!

Closes iFixit/ifixit#45799


danielbeardsley:
QA :+1: 

image
image

PR #1127, opened 2022-12-09
Hide Cross Sell when current variant is sold out

### LINKED ISSUE #1124: Error: Bad Request when adding multiple to cart if one is out of stock
Given these two validators both return 400, I really want to see the input skus and they currently aren't captured in this error.

Do This
====
Hide the Frequently Bought Together section when the main product is OOS.

Improve the error handling on Sentry so when we get a 400, we capture the inputs to this api so we can see what bad values are being sent.

Some values are captured in Sentry, but they look like this:

image

These validators can result in 400-level responses, but they might not be what is causing a problem in this case.
https://github.com/iFixit/ifixit/blob/90d7f9ffb5aed5448c9338956c9b0b4edc3456d8/Guide/UI/api/2.0/store/user/cart/product.php#L233-L260

Sentry Issue: [REACT-COMMERCE-EN](https://sentry.io/organizations/ifixit/issues/3702563162/?referrer=github_integration)

```
Error: Bad Request
  at None (/_next/static/chunks/pages/_app-985921f110daf96f.js:1:578736)
  at l (/_next/static/chunks/main-e7de20e05a14787c.js:1:111705)
  at Generator._invoke (/_next/static/chunks/main-e7de20e05a14787c.js:1:111493)
  at forEach.e.<computed> [as next] (/_next/static/chunks/main-e7de20e05a14787c.js:1:112128)
  at r (/_next/static/chunks/pages/_app-985921f110daf96f.js:1:736015)
...
(1 additional frame(s) were not displayed)
```

masonmcelvain:
- Hides the cross sell section when the main product is OOS.
- Stringifies mutation data that is sent to Sentry when errors happen.

## QA

Out of stock variants should have the cross sell section hidden.

Closes #1124

### DIFF
--- a/frontend/components/common/AppProviders.tsx
+++ b/frontend/components/common/AppProviders.tsx
@@ -44,8 +44,8 @@ const queryClientConfig = shouldIgnoreUserAgent
               Sentry.captureException(error, (scope) => {
                  scope.setTag('react-query', 'mutation');
                  scope.setContext('Mutation', {
-                    variables,
-                    context,
+                    variables: JSON.stringify(variables, null, 2),
+                    context: JSON.stringify(context, null, 2),
                     mutationKey: mutation.options.mutationKey,
                  });
                  return scope;
--- a/frontend/templates/product/sections/CrossSellSection/index.tsx
+++ b/frontend/templates/product/sections/CrossSellSection/index.tsx
@@ -165,7 +165,10 @@ export function CrossSellSection({
       }
    };
 
-   if (crossSellVariantsForSale.length === 0) {
+   if (
+      !selectedVariant.quantityAvailable ||
+      crossSellVariantsForSale.length === 0
+   ) {
       return null;
    }
 


jordycosta:
QA 🟢 

On OOS products, the **Frequently bought together** section is not shown ⛔ [example](https://react-commerce-prod-git-hide-xsell-when-oos-ifixit.vercel.app/products/gopro-hero3-plus-silver-partial-assembly)
On products in stock, the  **Frequently bought together** section is shown ✅ [example](https://react-commerce-prod-git-hide-xsell-when-oos-ifixit.vercel.app/products/macbook-air-13-late-2010-2017-replacement-battery)

danielbeardsley:
Is this what we want though?

I would think the cross-sells might be relevant when the main product is out of stock. Though if they are almost always companion products, then perhaps not. I guess the main product is in that list, so we'd have to remove it in order for add-to-cart to work. Hard to know how important that is.

danielbeardsley:
CR :+1:   For fixing the error I suppose.

masonmcelvain:
@sterlinghirsh ^ deploy_block ✌🏻

sterlinghirsh (approved):
CR :zap: un_deploy_block :zap: I think it's a lot simpler to just hide the cross sell when OOS rather than try to just remove the main item, and this fixes the error we saw. Cross sell items are usually companion products (e.g. battery links to battery adhesive). If we want to revisit it later, we can, but I think it's not worth the effort in this narrow case.

Issue #1026, opened 2022-11-15
Git hooks: improve their performance


danielbeardsley:
I appreciate we have ways to verify some things about our code and that we can inform devs about errors in our code fairly early. **But** the cost is high and both the pre-commit hook and pre-push hook are slow and annoying.

### Do This
* pre-commit: Only reformat files that have changed!
   * This hook takes ~ 6s and is just long enough to be annoying, but not long enough to warrant doing something else with that time
   * ![image](https://user-images.githubusercontent.com/26855/201796547-b536708b-8deb-4bb9-85e7-c45035bae311.png)
* pre-push: Make the typescript check faster
   * I feel like we're doing something wrong; I expect there to be a better way than running one `tsc` process per package
   * This takes 1m4s and I mostly end up doing `--no-verify`
   * ![image](https://user-images.githubusercontent.com/26855/201797120-39450832-c474-48e5-a320-79be59ca211d.png)
   * This is a blocker for bringing this code into the monorepo

CC @sterlinghirsh @sctice-ifixit @davidrans @iFixit/shopify

dhmacs:
Agree we should do better in this regard, it's starting to be a little bit annoying to wait while pushing. I've your same feeling, that tsc should be taking way less time (although on my machine it takes around 30-40 seconds, it's still a lot).
Maybe there's some config that we can adjust to improve this, but probably the fastest thing to do is to disable it entirely and rely more on CI for type checking

sterlinghirsh:
I am in favor of disabling the hook and relying on CI for type checking. If anyone wants to do the type check, maybe they can put a custom commit or push hook. Or just run the check manually. I think a hook like that should take no more than 2 seconds to run.

masonmcelvain:
We have CI tasks for formatting and type checking, so I'm personally in favor of disabling both hooks.

sterlinghirsh:
Can this be closed?

danielbeardsley:
I would like to have the pre-commit format hook if we can limit it to the files that have changed. I'm sure it's possible, but I haven't looked into it. Meanwhile, I've had my hooks turned off entirely.

sterlinghirsh:
We did it

masonmcelvain:
> I would like to have the pre-commit format hook if we can limit it to the files that have changed. I'm sure it's possible, but I haven't looked into it. Meanwhile, I've had my hooks turned off entirely.

https://github.com/okonet/lint-staged looks like it's built for that and is compatible with husky

PR #1128, opened 2022-12-09
Git hooks: drop format and type-check

danielbeardsley:
Auto formatting is not setup to work on just changed files and so takes way too long. Typechecking is already done in CI and currently takes 30 - 60s. This is quite annoying and hampers productivity and interrupts development. I find myself constantly using `--no-verify` anyway.

We can re-add them if we improve their performance.

Connect #1026


masonmcelvain (approved):
CR 🙏🏻

masonmcelvain:
I don't think we need qa for this?

PR #1123, opened 2022-12-08
Hot Fix: Update mocked product properties to match new product type definition

ardelato:
## Description

The Product type was recently updated so we started throwing a type error when running `pnpm type-check`. This commit updates the mocked product properties to match the new type definition.

## QA Notes:
- [x] The type check CI action should pass

qa_req 0 - CI is enough

PR #1130, opened 2022-12-09
Fix: Updated newly added mocked product types

ardelato:
### Description

Similar to https://github.com/iFixit/react-commerce/pull/1123, this pull which added new mocked products was using the prior type definition of Product. 

qa_req 0 CI should pass now

**cc:** @davidrans


davidrans (approved):
CR :coffee:

Issue #1136, opened 2022-12-13
Fix responsive images loading too big


sterlinghirsh:
Related to #1131 and #1067 

We have some images that are displayed quite small, but the loaded image is huge, causing unnecessarily poor performance.

Example: On /Parts/iPhone, at screen-width 997px, in grid view, the first product image showing at 301px wide is actually downloading a 1000x1000 image. According to the `sizes` attr, we have `(max-width: 1000px) 400px`, meaning that the image will download the best size to show a 400px image. 
![image](https://user-images.githubusercontent.com/589425/207208571-3b794a60-dce9-4cd8-85d8-8cbe749c0f0a.png)

However, in the srcset, the image that would be shown at 400px is the size1000 image, due to the large gap between size400 and size1000.
![image](https://user-images.githubusercontent.com/589425/207208917-f947db48-ef59-4ce9-9a43-9bf4df2826d3.png)

Try out making a custom image component that uses correct sizes from our image size map.

danielbeardsley:
I think this was mostly fixed with https://github.com/iFixit/ifixit/pull/52942

PR #1137, opened 2022-12-13
Add FixedSizesResponsiveImage for cart images

sterlinghirsh:
This handles the existing sizeMap in a way that generates the srcset with the correctly sized images. This only applies when sizes is specified, and it doesn't handle every prop possible to pass to the next image component.

Example:
image
This is at screen-size 997px wide on /Parts/iPhone. The live site currently downloads a size1000 image for this 301px wide image, but this branch correctly downloads the size400 image. See the connected issue #1136 for more details on this example on the live site.

I'm creating this as a draft to get feedback. This does seem to cause us to load the right image size for cart images where we didn't before, though the new custom image component isn't as full-featured as next/image.

Connects #1136 

@danielbeardsley @federicobadini @dhmacs @masonmcelvain


federicobadini:
I understand the effort to serve the minimal image size that suffice to fulfill the browser request. 
However I'm quite convinced we are overestimating the advantages of this. 

Consider that adding this customization means that:

1. We are generating srcSet even though the component has an explicit width. So we leave more options available and some browser may decide to pick larger than needed images
2. We are using an img component so the only way to not impact highly on performance is adding the lazy img attribute
3. Next.JS image implements resizeObserver to avoid loading images out of screen or hidden. We have seen that it works better in understanding which images are shown to defer the others, especially in carousels. With lazy we will probably end up downloading more images than we do right now.
4. If we happen to have one of these images as LCP element we would need to manually add srcSet based preloading headers which is handled out of the box right now.

On the other side we do are saving some KB of data but those are saved on lazy loaded, mid-low priority content. So I am not convinced that the savings outweigh the cons.

sterlinghirsh:
Not sure I understand your first point. As far as I can tell, this doesn't come into play for explicitly sized images. I wrapped it in an `if (typeof props.sizes === 'string')` so that it would only apply to responsive images but not 1x or 2x srcsets.

For the second/third points, I think we could add the lazy attr support or other optimizations as necessary. These images are only used in a handful of places, and the carousel isn't one of them. Not sure about resizeObserver, but I think intersectionObeserver might be important to make lazy loading work, though I noticed that next13 starts using browser-native lazy loading, so there might be a simple way to make that work with this. The main place we see these images is in the ProductList filtered product area, so the first 2 images should load at the start of the page, and the rest load later. However, with the improved image sizes, preloading more images has less of a performance impact.

This does slightly improve the loading time for the first few images, though I think the main benefit is when we scroll down the page as the later images get lazy loaded faster and are more likely to be ready by the time they come into view.

erinemay:
@sterlinghirsh, the last time you touched this was in 2022. Is it still on your radar?

sterlinghirsh:
It is now.

mlahargou:
Erm, I didn't recognize this branch in [my branches page](https://github.com/iFixit/react-commerce/branches/yours)...
image

Sorry, we can totally restore this...

PR #1138, opened 2022-12-13
Fix minicart image size

sterlinghirsh:
It used to be a responsive image, and due to the default `sizes` attr being 100vw, it was downloading a huge (sometimes 2000px) image that would display at 62x62px.

Before:
image

After:
image

Probably won't matter to most of us on gigabit connections, but for mobile users it's nice to be a little more efficient.

Connects #1136


danielcliu-ifixit:
CR :v:

deltuh-vee:
QA 🎬 
The intrinsic size of the mini cart images are now 70 x 70 px.

danielbeardsley (approved):
CR :+1:

PR #1147, opened 2022-12-13
breakpoints: use consistent units

danielbeardsley:
The docs say chakra breakpoints should all be the same units because the JS must sort these by their literal values.

Thankfully, this didn't break for us, but if any other breakpoint was not in PX (besides base), it completely borks `useBreakpoint()`

See this codepen:
https://codesandbox.io/s/sparkling-wildflower-o2ww7z?file=/src/index.tsx

qa_req 0 -- These are equivalent values and this just works around what is effectively a bug in chakra.

CC @jarstelfox

### DIFF
--- a/packages/ui/theme/foundations/breakpoints.ts
+++ b/packages/ui/theme/foundations/breakpoints.ts
@@ -2,7 +2,7 @@ import type { ThemeOverride } from '@chakra-ui/react';
 import primitives from '@core-ds/primitives';
 
 export const breakpoints: ThemeOverride['breakpoints'] = {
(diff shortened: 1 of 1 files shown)


sterlinghirsh (approved):
CR :zap: lucky 0em = 0px

sterlinghirsh:
Is this connected to any issue?

danielbeardsley:
No, @jarstelfox told me about it cause he fixed a similar issue, after hours of investigation. So I figured I'd capitalize on his work by making the same fix.

Issue #1114, opened 2022-12-07
Update our CI infrastructure to support running Playwright tests


ardelato:

batbattur:
Found this article very interesting https://kontent.ai/blog/next-js-playwright-tests-github-action/.

They are using the deployed vercel pages to run the e2e tests:
>Now, as I mentioned above, we need to have a built site before we can run the tests. We could, of course, use a GitHub Action to build the site on every PR, but:
> - We’d still need to deploy it somewhere so that Playwright can access it.
> - It would be a waste of resources, as Vercel is already building and hosting the site for us.

>Instead, we’ll just wait for Vercel to build the preview and use it for our tests. We’ll create a GitHub Action that will:
> - Wait for Vercel and grab the preview URL
> - Set up Playwright environment
> - Run the tests

ardelato:
If we use the site built by Vercel, we would be able to inject `MSW` as the server?

batbattur:
> If we use the site built by Vercel, we would be able to inject `MSW` as the server?

I'm not sure how it will work with `MSW`. 

I guess the downside of using the Vercel site is that we have to wait for it to be done (which may take longer than building it in the gh workflow). Also, if something goes wrong with Vercel, the workflow will be useless since we would rely on Vercel to always work. So I think building it in the workflow might be better since we won't depend on external service.

ardelato:
If it is easier and faster to integrate `Playwright` into our CI pipeline by using the `Vercel` site, then we should start with that. Afterward, we can start working on building the site within the workflow. 

Basically, we are also trying to run `Playwright` in parallel with `Cypress` as soon as possible to address any issues sooner than later.

Issue #1113, opened 2022-12-07
Migrating our existing tests to the Playwright syntax.


ardelato:
### Description

This will probably include some continuation work from #1112 as we create utility functions similar to what we have for `Cypress` 

## To Do:

- [x] Port `product` tests
- [ ] Port `product-list` tests
- [ ] Port `redirect` test

Issue #1090, opened 2022-11-30
Experiment Writing E2E Tests with Playwright


ardelato:
### Description

It seems that `Playwright` is 2-6x faster than `Cypress`. In addition, it has other pros that we might consider reason enough to see if we should port our `Cypress` tests to `Playwright`

## To Do

- [ ] See how much work would be required to port our `Cypress` tests to `Playwright`
- [ ] Validate if and by how much `Playwright` is faster than `Cypress`

**Reference:**
- https://ifixit.slack.com/archives/C089J162J/p1669838499336999
- https://twitter.com/housecor/status/1597954687705104385

andyg0808:
https://ifixit.slack.com/archives/C97UC89AR/p1643096585147600
https://chlorinated-yellowhorn-f6b.notion.site/Playwright-vs-Cypress-726c9f55aeab499697d13443397102e5
https://ifixit.slack.com/archives/C01K32B4FPX/p1637102921113900
https://ifixit.slack.com/archives/C01K32B4FPX/p1629309309021000

![playwright-demo](https://user-images.githubusercontent.com/1283490/204926917-ffc884e8-d371-4968-9263-4ba9456bd58b.gif)
![playwright-trace-viewer](https://user-images.githubusercontent.com/1283490/204926929-02707cbd-f510-4d00-8bc5-6f7252fcb404.gif)
![playwright-debug-view](https://user-images.githubusercontent.com/1283490/204926942-7bacaa8b-8921-4d09-8ca8-80d4afbd99c3.gif)

ardelato:
Porting it might be easier than we thought 👀 https://contra.com/p/PWBcPYZc-rewriting-tests-from-cypress-to-playwright-using-gpt-3

ardelato:
After experimenting with Playwright https://github.com/iFixit/react-commerce/compare/testing--playwright-port it looks relatively easy to migrate our current `Cypress` tests to `Playwright`. While porting some of the `Cypress` tests I found it a lot simpler to work with `Playwright`'s `API` than `Cypress`'s. 

I also noted that the overall time (start to finish) of running a `Playwright` test was a lot faster than running a `Cypress` test.

|Cypress | Playwright |
|:---:|:---:|
[...]

PR #1149, opened 2022-12-14
Playwright Migration: Migrate first batch of Cypress tests to Playwright

ardelato:
## Description

We will slowly port over the `Cypress` tests to `Playwright`. We will start with the `product` tests so we have some tests to work with while addressing #1114.

## CR Notes:

The Playwright tests are pretty much the same as the Cypress tests.  Of course, there were some necessary changes like using Playwright syntax and how we select elements on the page. 

## QA Notes:

- [ ] Confirm running the Playwright tests will pass locally
	-  ⚠️ You might need to run `npx playwright install` to install the browsers
	- Run `(cd frontend && pnpm playwright:run)` 
- [ ] Stress test the Playwright tests by using the following command
	- `for i in {1..5}; do (cd frontend && pnpm playwright:run --trace on --retries 2); done` 
- [ ] Confirm the Playwright tests execute the same tasks as the Cypress tests
	-  You can add the following diff and then the following command to record videos of Playwright's interactions and then compare that to Cypress's interactions
		```diff
		diff --git a/frontend/playwright.config.ts b/frontend/playwright.config.ts
		index a87beadc..0d92ca1b 100644
		--- a/frontend/playwright.config.ts
		+++ b/frontend/playwright.config.ts
		@@ -48,6 +48,7 @@ const config: PlaywrightTestConfig = {
		 
		       /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
		       trace: 'on-first-retry',
		+      video: 'on',
		 
		       viewport: { width: 1400, height: 1000 },
		```
		- `(cd frontend && pnpm playwright:run --project=chromium)` 

Connects: #1113


frontend/playwright.config.ts
```diff
@@ -13,13 +13,13 @@ import { devices } from '@playwright/test';
...
    /* Maximum time one test can run for. */
-   timeout: 30 * 1000,
+   timeout: 60 * 1000,
    expect: {
       /**
        * Maximum time expect() should wait for the condition to be met.
        * For example in `await expect(locator).toHaveText();`
        */
-      timeout: 5000,
+      timeout: 10000,
```
mlahargou, line 22: Cool! On our phpunit WebDriver tests in ifixit/ifixit we wait for 30 seconds. We may need to expand this in the future.

frontend/tests/playwright/product/fix_kit.spec.ts
```diff
@@ -0,0 +1,42 @@
+import { test, expect } from '@playwright/test';
+
+test.describe('Fix Kit and Part Only test', () => {
+   test.beforeEach(async ({ page }) => {
+      await page.goto('/products/iphone-6s-plus-replacement-battery');
+   });
+
+   test('test kit contents and product skus', async ({ page }) => {
```
mlahargou, line 8: [minor]

```suggestion
   test('kit contents and product skus', async ({ page }) => {
```

Same with other tests.

frontend/tests/playwright/utils.ts
```diff
@@ -0,0 +1,21 @@
+import { Page } from '@playwright/test';
+
+export const interceptLogin = async (page: Page, userParams?: Object) => {
```
mlahargou, line 3: sweet!

frontend/playwright.config.ts
```diff
@@ -50,6 +50,9 @@ const config: PlaywrightTestConfig = {
       trace: 'on-first-retry',
 
       viewport: { width: 1400, height: 1000 },
+      launchOptions: {
+         slowMo: 1000,
```
mlahargou, line 54: Interesting. How much does this slow down the test?

frontend/tests/playwright/product/product_option.spec.ts
```diff
@@ -1,69 +1,75 @@
...
-      cy.findByTestId('cart-drawer-body')
-         .should('be.visible')
-         .then((el) => {
-            // Parse out the skus from the skuTexts which are in form of "Item # IF145-278-14"
-            const sku1 = this.firstOptionSkuText.match(/IF\d*\-\d*\-\d*/g)[0];
-            const sku2 = this.secondOptionSkuText.match(/IF\d*\-\d*\-\d*/g)[0];
+      await expect(page.getByTestId('cart-drawer-body')).toBeVisible();
+
+      // Parse out the skus from the skuTexts which are in form of "Item # IF145-278-14"
+      const sku1 = firstOptionSku?.replace('Item # ', '') ?? '';
```
mlahargou, line 56: I'm a little concerned that if it doesn't find the firstOptionSku, `sku1` will be an empty string. Then down below, we will grab an element like `getByText('');` which will probably grab anything.

I wonder if we should add an extra assertion to assert `sku1` is not an empty string.

mlahargou (approved):
CR 📱 with a few comments/questions deploy_block 📱 

Also, should we be dropping the old cypress tests?

frontend/playwright.config.ts
ardelato: Is the `30 seconds` for the assertion or for the test case?

frontend/tests/playwright/product/fix_kit.spec.ts
ardelato: Gotcha, and just to note these test descriptions were just copied over from the Cypress test descriptions.

ardelato:
>Also, should we be dropping the old cypress tests?

Not at the moment. That is the fifth step in the process (https://github.com/iFixit/react-commerce/issues/1111) since we have not integrated Playwright into our CI and have not confirmed that the Playwright tests are as reliable as our Cypress tests. 

@batbattur made a good suggestion to run them in parallel first and then drop the Cypress tests later.  https://github.com/iFixit/react-commerce/issues/1090#issuecomment-1339778829
>I like it! I think we can convert the existing cypress tests to Playwright and keep both of them for a week or so to observe any issues. Then we could remove Cypress related code if Playwright is working fine.

frontend/playwright.config.ts
mlahargou: Assertion. Basically, if we call `waitForPageLoadAfter` (or any of the wait calls) we default to 30 seconds. It may be excessive for these tests.

frontend/tests/playwright/product/fix_kit.spec.ts
mlahargou: > Gotcha, and just to note these test descriptions were just copied over from the Cypress test descriptions.

Totally, just something I saw while CRing. No need to change.

deltuh-vee:
QA 🎬 
I was able to confirm the tests pass locally and the Playwright tests execute the same tasks, but I wasn't able to stress test the tests. The tests would time out on my 2012 Macbook and fail on my Framework laptop running WSL2.

frontend/playwright.config.ts
ardelato: Without this the tests fail, but if were to just run the tests on Chromium with 4 parallel workers. Then it slows it down by about 20 seconds.

ardelato: Test Failure@2x

frontend/tests/playwright/product/product_option.spec.ts
ardelato: Addressed in [7593275](https://github.com/iFixit/react-commerce/pull/1149/commits/75932750cedac27415b6846b3991eeada354a00e)

frontend/tests/playwright/product/fix_kit.spec.ts
ardelato: Addressed in [11b3b86](https://github.com/iFixit/react-commerce/pull/1149/commits/11b3b86ed60932ff3778b4bb9431083cfec8a429)

ardelato:
un_deploy_block ⚡

mlahargou (approved):
CR 📱 
Nice!

deltuh-vee:
Carry Over QA 🎬

PR #1134, opened 2022-12-12
Product Page: Add Discounted Product Tests

ardelato:
## Description

This adds Jest tests to address #1132

## CR Notes:

I opted out of writing a `Jest` test to confirm the cart price matches the listed price when adding to the cart because it will be much simpler to test this through a `Cypress`, or `Playwright`, than through a `Jest` test. 

## QA Notes:

- [ ] CI should pass

Connects: #1132

PR #1150, opened 2022-12-14
Unskip tests for Product Section

ardelato:
## Description

While testing the tests, I left the `only` keyword in a test. As a result, we were skipping all other tests in the suite. This removes that keyword and therefore unskips all other tests in the file.

|Before|After|
|:---:|:---:|
|  No longer skipping tests     |   Number of tests with changes  |

## QA Notes:
- [ ] CI should pass and no longer have skipped tests

**cc:** @batbattur 

**Reference:** https://github.com/iFixit/react-commerce/pull/1134#discussion_r1048842900

### DIFF
--- a/frontend/tests/jest/tests/ProductSection.test.tsx
+++ b/frontend/tests/jest/tests/ProductSection.test.tsx
@@ -139,7 +139,7 @@ describe('ProductSection Tests', () => {
          (expect(price.textContent) as any).toBe('$29.99');
       });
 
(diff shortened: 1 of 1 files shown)


ardelato:
QA 👍 

We the `Jest` CI check passes and no longer skips tests

|Before|After|
|:---:|:---:|
|     Jest Before| Jest Now       |

batbattur (approved):
CR 🚛

PR #1155, opened 2022-12-15
Upgrade frontend prettier version to v2.7.1

dhmacs:
qa_req 0

The frontend package is not using the Prettier version of the project root `package.json`. This can cause mismatching when formatting with VSCode and the cli, because they can pick up different versions.


masonmcelvain (approved):
CR ✌🏻

PR #1160, opened 2022-12-15
GitHub workflow: add next bundle analysis

danielbeardsley:
Using this package: https://github.com/hashicorp/nextjs-bundle-analysis Add the github workflow to build our next bundle and compare the results to the bundle on the main branch.

The workflow may fail until it's run once on the main branch.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,7 +3,6 @@ name: E2E tests
 on: [push]
 env:
    npm_config_userconfig: './.npmrc'
-   npm_config_globalconfig: './.npmrc'
 jobs:
    cypress-run:
       runs-on: ubuntu-latest
@@ -28,7 +27,7 @@ jobs:
               path: ~/.cache/Cypress
               key: cypress-cache-v2-${{ runner.os }}-${{ hashFiles('pnpm-lock.yaml') }}
          # now let's install Cypress binary
-         - run: npx cypress install
+         - run: cd frontend && npm exec cypress install
          - name: Cypress run
            uses: cypress-io/github-action@v4.2.0
            env:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,8 +25,8 @@ jobs:
            id: cache-cypress
            uses: actions/cache@v1
            with:
-             path: ~/.cache/Cypress
-             key: cypress-cache-v2-${{ runner.os }}-${{ hashFiles('pnpm-lock.yaml') }}
+              path: ~/.cache/Cypress
+              key: cypress-cache-v2-${{ runner.os }}-${{ hashFiles('pnpm-lock.yaml') }}
          # now let's install Cypress binary
          - run: npx cypress install
          - name: Cypress run
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,6 +21,14 @@ jobs:
            env:
               FONT_AWESOME_NPM_TOKEN: ${{ secrets.FONT_AWESOME_NPM_TOKEN }}
            run: pnpm install
+         - name: Cache Cypress
+           id: cache-cypress
+           uses: actions/cache@v1
+           with:
+             path: ~/.cache/Cypress
+             key: cypress-cache-v2-${{ runner.os }}-${{ hashFiles('pnpm-lock.yaml') }}
+         # now let's install Cypress binary
+         - run: npx cypress install
          - name: Cypress run
            uses: cypress-io/github-action@v4.2.0
            env:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,11 +15,8 @@ jobs:
            env:
               JWT_SECRET: 'ci-secret'
               API_TOKEN_SALT: 'Not_A-s3Cr3t-/Qr5iGP0g=='
-         - name: Install pnpm
-           uses: pnpm/action-setup@v2.0.1
-           with:
-              version: 7
-              run_install: false
+         - name: Setup and cache pnpm
+           uses: ./.github/actions/pnpm-setup
          - name: Install workspaces
            env:
               FONT_AWESOME_NPM_TOKEN: ${{ secrets.FONT_AWESOME_NPM_TOKEN }}
--- a/.github/workflows/nextjs_bundle_analysis.yml
+++ b/.github/workflows/nextjs_bundle_analysis.yml
@@ -1,0 +1,113 @@
+name: 'Next.js Bundle Analysis'
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main # change this if your default branch is named differently
+  workflow_dispatch:
+
+defaults:
+  run:
+    # change this if your nextjs app does not live at the root of the repo
+    working-directory: ./
+
+jobs:
+  analyze:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up node
+        uses: actions/setup-node@v1
+        with:
+          node-version: '14.x'
+
+      - name: Install dependencies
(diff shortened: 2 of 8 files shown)


dhmacs (dismissed):
CI is failing

dev_block 🔢

danielbeardsley:
un_dev_block :+1: 

CI is passing. The bundle-analysis CI task is expected to fail the first time on the "compare" phase because there isn't something from the main branch to compare to.

QA :+1:

.github/workflows/main.yml
```diff
@@ -21,6 +21,14 @@ jobs:
            env:
               FONT_AWESOME_NPM_TOKEN: ${{ secrets.FONT_AWESOME_NPM_TOKEN }}
            run: pnpm install
+         - name: Cache Cypress
+           id: cache-cypress
+           uses: actions/cache@v1
+           with:
+             path: ~/.cache/Cypress
+             key: cypress-cache-v2-${{ runner.os }}-${{ hashFiles('pnpm-lock.yaml') }}
```
ardelato: If this is an isolated machine that will run on `ubuntu` does `runner.os` need to be included for the cache key?

danielbeardsley: I copied this little chunk from the cypress github action docs and it seemed reasonable. I'm inclined to leave it because it's safer (cache's won't be shared if we change OSes).

.github/workflows/nextjs_bundle_analysis.yml
```diff
@@ -0,0 +1,117 @@
...
+           run: npx -p nextjs-bundle-analysis report
+
+         - name: Upload bundle
+           uses: actions/upload-artifact@v2
+           with:
+              name: bundle
+              path: .next/analyze/__bundle_analysis.json
+
+         - name: Download base branch bundle stats
+           uses: dawidd6/action-download-artifact@v2
```
ardelato, line 61: TIL! @mlahargou apparently this GH Action allows you to download artifacts from other workflows.  https://github.com/dawidd6/action-download-artifact

> Let's suppose you have a workflow with a job in it that at the end uploads an artifact using actions/upload-artifact action and you want to download this artifact in another workflow that is run after the first one. Official actions/download-artifact does not allow this. That's why I decided to create this action.

mlahargou: Cool!

.github/workflows/nextjs_bundle_analysis.yml
```diff
@@ -0,0 +1,117 @@
...
+           with:
+              # if you use a custom build directory, replace all instances of `.next` in this file with your build directory
+              # ex: if your app builds to `dist`, replace `.next` with `dist`
+              path: .next/cache
+              # change this if you prefer a more strict cache
+              key: ${{ runner.os }}-build-${{ env.cache-name }}
+
+         - name: Build next.js app
+           # change this if your site requires a custom build command
+           run: ./node_modules/.bin/next build
```
masonmcelvain, line 44: Probably not necessary since we use `next build` in our package.json script, but this is simpler and will automatically pick up changes if we edit our build command down the line.

```suggestion
           run: pnpm build
```

masonmcelvain (approved):
CR ✌🏻 Nice job! I don't think we need to change the build command, I just noticed it.

ardelato:
deploy_block ☁️

.github/workflows/nextjs_bundle_analysis.yml
```diff
@@ -0,0 +1,117 @@
...
+           if: success() && github.event.number
+           run: |
+              body=$(cat .next/analyze/__bundle_analysis_comment.txt)
+              body="${body//'%'/'%25'}"
+              body="${body//$'\n'/'%0A'}"
+              body="${body//$'\r'/'%0D'}"
+              echo ::set-output name=body::$body
+
+         - name: Find Comment
+           uses: peter-evans/find-comment@v1
```
ardelato, line 96: I think this template is a bit outdated, we currently use `v2` for this action in `iFixit/ifixit` https://github.com/iFixit/ifixit/blob/master/.github/actions/issue-comment/action.yml#L16

```suggestion
           uses: peter-evans/find-comment@v2
```

.github/workflows/nextjs_bundle_analysis.yml
danielbeardsley: I intentionally didn't use that command because this task is explicitly about building next.js and if we change that command to do additional tasks (not sure what they would be) they may not be relevant. That said, I'm not sure which is more likely: changing our setup in a way that `next build` doesn't work, or changing that command in a way that breaks this analysis.

danielbeardsley:
un_deploy_block :+1:

masonmcelvain (dismissed):
CR ✌🏻

.github/workflows/nextjs_bundle_analysis.yml
masonmcelvain: Yea good points. I guess I was thinking "we should analyze the bundle the exact same way we build the app for production", but it's hard to imagine how we would augment that command to affect the bundle size. This is fine as is.

.github/workflows/nextjs_bundle_analysis.yml
```diff
@@ -0,0 +1,117 @@
...
+
+         - name: Find Comment
+           uses: peter-evans/find-comment@v2
+           if: success() && github.event.number
+           id: fc
+           with:
+              issue-number: ${{ github.event.number }}
+              body-includes: '<!-- __NEXTJS_BUNDLE -->'
+
+         - name: Create Comment
```
ardelato, line 103: The `peter-evans/create-or-update-comment` action allows for the `creation` and `updating` steps to be combined based on the output of `Find Comment`. [Link](https://github.com/peter-evans/create-or-update-comment#:~:text=If%20the%20find%2Dcomment%20action%20output%20comment%2Did%20returns%20an%20empty%20string%2C%20a%20new%20comment%20will%20be%20created.%20If%20it%20returns%20a%20value%2C%20the%20comment%20already%20exists%20and%20the%20content%20is%20replaced.)

We do this in `iFixit/ifixit` as well https://github.com/iFixit/ifixit/blob/master/.github/actions/issue-comment/action.yml#L23

> If the find-comment action output comment-id returns an empty string, a new comment will be created. If it returns a value, the comment already exists and the content is replaced.

```yaml
    - name: Create or Update comment
        uses: peter-evans/create-or-update-comment@v2
        if: success() && github.event.number
        with:
           issue-number: ${{ github.event.number }}
           comment-id: ${{ steps.fc.outputs.comment-id }}
           edit-mode: replace
           body: ${{ steps.get-comment-body.outputs.body }}
```

ardelato: This action version is also a bit outdated, we are currently using `v2` in `iFixit/ifixit`.

ardelato (dismissed):
dev_block ☁️ on updating the commenting steps of the workflow

.github/workflows/nextjs_bundle_analysis.yml
danielbeardsley: FYI, the bulk of this workflow was generated by: https://github.com/hashicorp/nextjs-bundle-analysis

So, while updating it seems like a good idea, diverging from the generated file makes updates harder. Thoughts?

ardelato: I did see that most of this workflow was generated from the `npx -p nextjs-bundle-analysis generate` command. However, it was generated from their template file where this particular section was last updated more than a year ago.

Git Blame@2x

The commenting part does not really have anything to do with the logic associated with how the GitHub action reports or compares the bundle size. Therefore, I think updating this section is fine.

danielbeardsley:
un_dev_block :+1: 

Updated the comment step!

masonmcelvain (approved):
CR ✌🏻

PR #1162, opened 2022-12-16
Stats: send some stats to graphite

### LINKED ISSUE #960: Stats tracking from react-commerce
## Problem
Similar to #959, we now have a fair amount of traffic coming through vercel. It would be nice to be able to collect stats for use in grafana like we do in our php app.

## Do This
Set something up so we can collect stats like we do in php so that things are visible in grafana and so that we can get alerts using our existing systems.

A few examples of things we'd like to track:
* RED metrics on Strapi, iFixit api, Shopify
* Cache hit rates by type of request

It may be a lot of work to get these stats into Grafana, but we already have a DataDog integration set up. The easiest path may be to start adding some of the data we currently console.log to a structured output for DD if possible, then try to get some graphing going in DD. Ultimately it would be nice to have everything in Grafana, but I'm more interested in getting something working with low effort that gives us visibility into timing and cache rates.

danielbeardsley:
I looked into exposing stats to our metrics infrastructure, in
particular to prometheus. There are some difficulties:

* Prometheus is a pull-based protocol so prom we'd have to open
  a particular endpoint in our next.js app
* Each "page" and each api "endpoint" are run as a separate
  "function" on vercel so they don't share memory state
   * This means that we would need a persistent and separate data store
     service to aggregate the stats across the app
* We'd still need to aggregate times in a useful way because there are
  multiple page requests in each reporting period

So, since we have statsd, and it solves these problems, let's start with
that. `hot-shots` has 500 dependents and is downloaded 1.4M times per
week. This makes it the most popular statsd package by several orders of
magnitude.

I've needed to separate server and client code in several places otherwise
building fails trying to require a node.js package on the client side.
This process of discovering which files need to be split and why is
tricky and not straightforward. I've done it only where the building process failed.

Closes #960


danielbeardsley:
I've rebased this on top of #1179

frontend/helpers/next-helpers.ts
```diff
@@ -7,6 +7,7 @@ import { clearCache } from '@lib/cache';
 import { CACHE_DISABLED } from '@config/env';
 
 export function serverSidePropsWrapper<T extends { [key: string]: any }>(
+   pageName: string,
    getServerSidePropsInternal: GetServerSideProps<T>
 ): GetServerSideProps<T> {
    return async (context) => {
```
masonmcelvain: > next.js doesn't have any way that I've seen to get the pathname like
"product/[handle]" Though there's a GH issue to add this feature.

Is [`context.resolvedUrl`](https://nextjs.org/docs/api-reference/data-fetching/get-server-side-props#context-parameter) not what you're looking for here? Like this:

https://github.com/iFixit/react-commerce/blob/66ae2a62c117aa219321e69a981a8dbada230958/packages/helpers/nextjs.tsx#L3-L6

masonmcelvain (commented):
CR ✌🏻 With a question about getting the pathname from the context

erinemay:
Is there a good way to go about testing this in development?

frontend/helpers/next-helpers.ts
danielbeardsley: Sorry, no, I want a canonicalized route name. Effectively the name of the file. Turns out a human generated name is probably better anyway.

danielbeardsley:
> Is there a good way to go about testing this in development?

I tried this:

* Cloned the [statsd repo](https://github.com/statsd/statsd)
* Installed deps, tweaked the example config to use the logging backend
* Started it
* Checked out this branch
* Edited `.env.local` to add `STATSD_HOST=127.0.0.1`
* Ran our app.... and....  Found a bug!

I pushed a commit and now the values are being reported as I expect:
image


QA :+1:

packages/stats/stats.ts
```diff
@@ -33,10 +33,5 @@ export function timeAsync<T>(
    statName: string,
    promiseFunc: () => Promise<T>
 ): Promise<T> {
-   const t = Date.now();
-   const result = logAsync(statName, promiseFunc);
```
masonmcelvain: Did you intentionally drop this `logAsync` call? I thought you wanted to log it and send stats.

If we don't want to log, we could just `export stats.asyncTimer` from this module, since the wrapper doesn't really add anything. But not worth blocking on if we're not making any other changes.

masonmcelvain (approved):
CR ✌🏻 with a question

packages/stats/stats.ts
danielbeardsley: Oh snap, you're right. I did drop that on accident. I noticed that the `asyncTimer()` removed the need for much of this code, but I went too far.  I'll make a follow up.

PR #1171, opened 2022-12-19
Bundle Analysis: add frontend/ for each step that takes a path

danielbeardsley:
We have working_directory: frontend/ as a default for all the steps but I think that doesn't apply for these tasks that take a "path" input. It hopefully *does* apply for all the steps with a "run" input.

This has meant that the cache of the .next directory is empty and the bundle from main isn't being uploaded.


masonmcelvain (approved):
CR ✌🏻

danielbeardsley:
QA :+1: 

Previously:
image

Now it actually uploads:
image

PR #1173, opened 2022-12-20
Product Page: Test Review Section

### LINKED ISSUE #1146: Product Page: Add Jest tests for Reviews
## Description

This is an ongoing effort in addressing https://github.com/iFixit/react-commerce/issues/452

### Requirements

- [ ] Make sure the review count and stars at the top of the page match the review count and stars at the bottom (Customer reviews) section.
- [ ] Check products with no reviews
	- [ ] Reviews section should not show up if the product doesn't have any reviews.

batbattur:
## Summary
Added the following tests for checking the reviews:
- Checks if the review section renders when the product has reviews.
- Checks if the review section renders with `No reviews yet` text when the product has no reviews.
- Checks if the review count and average at the top of the page (from `ProductRating` component) match the review count and average at the bottom (from `ReviewSection` component).

## QA notes 
Passing CI is enough. (qa_req 0)

closes #1146

### DIFF
--- a/frontend/tests/jest/__mocks__/products.ts
+++ b/frontend/tests/jest/__mocks__/products.ts
@@ -1,7 +1,7 @@
 import { MenuItemType } from '@models/menu';
 import { ProductTemplateProps } from '@templates/product/hooks/useProductTemplateProps';
 import { ProductSearchHit } from '@models/product-list';
-import { Product, ProductVariant } from '@models/product';
+import { Product, ProductVariant, ProductReview } from '@models/product';
 import { CurrencyCode } from '@lib/shopify-storefront-sdk';
 
 export const mockedProduct: Product = {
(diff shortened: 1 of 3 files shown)


frontend/tests/jest/__mocks__/products.ts
```diff
@@ -3174,3 +3174,124 @@ export const mockedLayoutProps: Pick<ProductTemplateProps, 'layoutProps'> = {
...
+         avatar:
+            'https://www.cominor.com/static/images/avatars/User/ifixit/avatar-9.thumbnail',
+         url: 'https://www.cominor.com/User/3081773/MARVIN+STABLER',
+         canEdit: true,
+      },
+   },
+   {
+      reviewid: 37196,
+      rating: 5,
+      headline: null,
```
sterlinghirsh, line 3202: Looks like all the headlines are null. We may want to include at least one review with a headline.

frontend/tests/jest/utils.tsx
```diff
@@ -157,3 +163,29 @@ export const getProductWithWarranty = (
...
+
+   if (type === 'empty') {
+      reviews = [];
+   }
+
+   return {
+      reviews: reviews,
+      count: reviews.length,
+      average:
+         // @ts-ignore
```
sterlinghirsh, line 180: Why do we need the ts-ignore here?

frontend/tests/jest/__mocks__/products.ts
batbattur: Got it! I'll add a headline to one of them.

frontend/tests/jest/utils.tsx
batbattur: I was getting `possibly 'undefined'` error:
Screen Shot 2022-12-20 at 12 10 35 PM

frontend/tests/jest/utils.tsx
```diff
@@ -157,3 +163,29 @@ export const getProductWithWarranty = (
...
+         reviews.reduce((total, next) => total + next.rating, 0) /
+         (reviews.length == 0 ? 1 : reviews.length),
```
sterlinghirsh, line 182: This is a minor point and I don't think you need to change it just for this, but I think when handling the empty case to avoid dividing by 0, it makes more sense to put the whole reduce in the non-zero case rather than just conditionalizing the divisor when the dividend is also going to be meaningless.
```suggestion
         reviews.length == 0 ? 0 : reviews.reduce((total, next) => total + next.rating, 0) / reviews.length,
```

sterlinghirsh (commented):
CR :zap: dev_block :zap: on a minor change to include at least one headline in the test data

frontend/tests/jest/__mocks__/products.ts
batbattur: Updated in https://github.com/iFixit/react-commerce/pull/1173/commits/ef3794b238643703f10ff9dc5a40f9bb6037dce4.

frontend/tests/jest/utils.tsx
batbattur: Thanks @sterlinghirsh for great the suggestion! Updated in https://github.com/iFixit/react-commerce/pull/1173/commits/2c391f6f02c82b8e20474529de48350a679c8cdc.

batbattur:
un_dev_block 🚛 updated the review test data to include a headline

frontend/tests/jest/utils.tsx
```diff
@@ -157,3 +163,30 @@ export const getProductWithWarranty = (
...
+      reviews = [];
+   }
+
+   return {
+      reviews: reviews,
+      count: reviews.length,
+      average:
+         reviews.length == 0
+            ? 0 // @ts-ignore
+            : reviews.reduce((total, next) => total + next.rating, 0) /
```
sterlinghirsh, line 182: I would rather we not use the ts-ignore in general if possible. It feels sloppy and can ignore more errors than the one you were trying to ignore. A simple workaround can make this pass the type check, though. This suggestion is a very simple but maybe slightly wrong way to do it, since it will create a 0-rating review that might not be right, but that might be just fine in a test that already ignores that problem. It may be the case that reviews can't actually have a 0 rating. We'd have to check the api that returns these and see if a null/undefined rating is possible, but if not, we can adjust the type. If we must support reviews without ratings and if we must handle that case in the test, then the most accurate way to handle it is to filter out reviews without ratings before this calculation. But I think we would be close enough to just accept the suggestion and move on.
```suggestion
            ? 0
            : reviews.reduce((total, next) => total + (next.rating || 0), 0) /
```

sterlinghirsh (approved):
CR :zap:

PR #1183, opened 2022-12-21
Don't reject null compatibility variants

### LINKED ISSUE #1182: Products are incorrectly hiding their compatibility display
Slack Ref https://ifixit.slack.com/archives/C0252FETV2S/p1671638641526619
## Problem
From Jeff Stephens:
> I stumbled on these two products that are not displaying compatibility. They have a worksin attached in Akeneo. I went looking for other cases but I came up empty. The only thing I see in common between the two worksin is the checked variants are “default.”
> 209021-1  [Nintendo DS Lite Touchscreen](https://www.ifixit.com/products/nintendo-ds-lite-touchscreen)   No compatibility displayed on the right below the description or below the reviews. The product has [worksin 583](https://www.ifixit.com/Admin/Worksin/worksin.php?worksin=583) attached.
> 258027-2  [Nintendo 3DS XL Battery](https://www.ifixit.com/products/nintendo-3ds-xl-replacement-battery)   No compatibility displayed on the right below the description or below the reviews. The product has [worksin 2028](https://www.ifixit.com/Admin/Worksin/worksin.php?worksin=2028) attached.

Upon investigating, it looks like the issue is that we don't handle `null` as a valid case in the list of variants. We do handle empty string properly, but in our DB, the "default" variant values in topic_variants are about evenly split between null and empty string. The variant name makes it into metafields.ifixit.compatibility_json, and my hunch is that the field fails validation since we expect it to be string, not nullable string.

## Do This
The simplest fix is probably to adjust the validation to support null or string. Alternatively we could adjust the values we save to Shopify (From ProductBodyFactory) to always be empty strings instead of nulls, but then we'd have to repush a bunch of data, though the data would be more consistent in Shopify going forward.

masonmcelvain:
- Fixes a bug where the compatibility section was hidden if any device had a `null` variant.
- Memoizes variant calculations in the compatibility section.
- Prevents empty variant `<p>` tags from being rendered in the compatibility section.
- Properly uses the plural for the hidden variant indicator in the compatibility section.

@sterlinghirsh 

Closes #1182

### DIFF
--- a/frontend/components/common/CompatibleDevice.tsx
+++ b/frontend/components/common/CompatibleDevice.tsx
@@ -4,12 +4,12 @@ import { Product } from '@models/product';
 
 export type CompatibleDeviceProps = {
    device: NonNullable<Product['compatibility']>['devices'][number];
-   truncate?: number;
+   maxModelLines?: number;
 };
 
 export function CompatibleDevice({
    device,
-   truncate = 0,
+   maxModelLines = 0,
 }: CompatibleDeviceProps) {
    const variants = React.useMemo(
       () => device.variants.filter(Boolean),
@@ -17,10 +17,13 @@ export function CompatibleDevice({
    );
    const [visibleVariants, hiddenVariantCount] = React.useMemo(
       () =>
-         truncate > 0 && variants.length > truncate
-            ? [variants.slice(0, truncate), variants.length - truncate]
+         maxModelLines > 0 && variants.length > maxModelLines
+            ? [
+                 variants.slice(0, maxModelLines - 1),
+                 variants.length - maxModelLines + 1,
+              ]
             : [variants, 0],
-      [truncate, variants]
+      [maxModelLines, variants]
    );
    return (
       <>
--- a/frontend/components/common/CompatibleDevice.tsx
+++ b/frontend/components/common/CompatibleDevice.tsx
@@ -1,5 +1,5 @@
-import { Flex, Img, Link, Text } from '@chakra-ui/react';
-import NextLink from 'next/link';
+import * as React from 'react';
+import { Flex, Img, Text } from '@chakra-ui/react';
 import { Product } from '@models/product';
 
 export type CompatibleDeviceProps = {
@@ -11,12 +11,17 @@ export function CompatibleDevice({
    device,
    truncate = 0,
 }: CompatibleDeviceProps) {
-   let variants = device.variants;
-   let unlistedVariants = 0;
-   if (truncate > 0 && variants.length > truncate) {
-      variants = device.variants.slice(0, truncate - 1);
-      unlistedVariants = device.variants.length - variants.length;
-   }
+   const variants = React.useMemo(
+      () => device.variants.filter(Boolean),
+      [device.variants]
+   );
+   const [visibleVariants, hiddenVariantCount] = React.useMemo(
+      () =>
+         truncate > 0 && variants.length > truncate
+            ? [variants.slice(0, truncate), variants.length - truncate]
+            : [variants, 0],
+      [truncate, variants]
+   );
    return (
       <>
          <Img
@@ -43,7 +48,7 @@ export function CompatibleDevice({
                {device.deviceName}
             </Text>
             <Flex flexDir="column">
-               {variants.map((variant) => (
+               {visibleVariants.map((variant) => (
                   <Text
                      key={variant}
                      lineHeight="short"
@@ -54,9 +59,10 @@ export function CompatibleDevice({
                   </Text>
                ))}
             </Flex>
-            {unlistedVariants !== 0 && (
+            {hiddenVariantCount > 0 && (
                <Text mb="2px" lineHeight="short" fontSize="xs" color="gray.600">
-                  And {unlistedVariants} other models...
+                  And {hiddenVariantCount} other model
+                  {hiddenVariantCount > 1 ? 's' : ''}...
                </Text>
             )}
          </Flex>
--- a/frontend/models/product.ts
+++ b/frontend/models/product.ts
@@ -383,7 +383,7 @@ const CompatibilityMetafieldSchema = z
             imageUrl: z.string(),
             deviceUrl: z.string(),
             deviceName: z.string(),
-            variants: z.array(z.string()),
+            variants: z.array(z.string().nullable()),
          })
       ),
       hasMoreDevices: z.boolean(),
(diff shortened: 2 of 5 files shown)


frontend/components/common/CompatibleDevice.tsx
```diff
@@ -11,12 +11,17 @@ export function CompatibleDevice({
...
-      unlistedVariants = device.variants.length - variants.length;
-   }
+   const variants = React.useMemo(
+      () => device.variants.filter(Boolean),
+      [device.variants]
+   );
+   const [visibleVariants, hiddenVariantCount] = React.useMemo(
+      () =>
+         truncate > 0 && variants.length > truncate
+            ? [variants.slice(0, truncate), variants.length - truncate]
```
danielcliu-ifixit, line 21: I personally still prefer the `truncate - 1` as `truncate` then refers the the total number of lines in a compatibility section, not how many variants can be shown.

danielcliu-ifixit:
CR :v: but i still like the original `truncate` meaning

danielcliu-ifixit (approved):

frontend/components/common/CompatibleDevice.tsx
masonmcelvain: It isn't clear to me how "the total number of lines in a compatibility section" is different from the number of variants to show, could you elaborate? Does the total number of lines include the "and X more models..."?

I made this change because it seemed like the one caller was passing in a value that represented "the number of variants to show".

https://github.com/iFixit/react-commerce/blob/26b01d5123b21df836e4ad9bfabe92096a48c098/frontend/templates/product/sections/ProductSection/CompatibleDevices.tsx#L35

masonmcelvain: I'm now realizing this is also one of those cases where instead of showing "and 1 more model..." we should just show the last model.

masonmcelvain:
@erinemay I'm gonna dev_block ✌🏻 on sorting out the meaning of the `truncate` prop

erinemay:
QA 👍 products with worksins like the ones brought up in the issue now have compatibility:
https://react-commerce-git-compatibility-variants-nullable-ifixit.vercel.app/products/nintendo-3ds-xl-replacement-battery
https://react-commerce-git-compatibility-variants-nullable-ifixit.vercel.app/products/sony-xbr-65x850d-65-inch-uhd-tv-motherboard

Other product pages with compatibility sections remain unchanged:
https://react-commerce-git-compatibility-variants-nullable-ifixit.vercel.app/products/500-gb-ssd-hybrid-2-5-hard-drive
https://react-commerce-git-compatibility-variants-nullable-ifixit.vercel.app/products/macbook-pro-14-2021-a2442-right-fan

New compatibility links are to the expected device pages.
image

erinemay:
> @erinemay I'm gonna dev_block ✌🏻 on sorting out the meaning of the `truncate` prop

Oops didn't see this. Well, it's working for now.

masonmcelvain:
un_dev_block ✌🏻

frontend/components/common/CompatibleDevice.tsx
danielcliu-ifixit: Ok so i think what happens is in this current code, if truncate = 5 and there are 5 variants it will show all 5. If there are 6 it will show 5 variants and a line with a link to the extra 1 in the later compatibility section for a total of 6 lines. In the old version the we would if there were 5 we would show 5, but if there were 6 we would show 4 variants and 1 link line for a total of 5 lines. I just think thats cleaner, but I would also agree that its borderline arbitrary.

frontend/components/common/CompatibleDevice.tsx
```diff
@@ -4,23 +4,27 @@ import { Product } from '@models/product';
...
    const variants = React.useMemo(
       () => device.variants.filter(Boolean),
       [device.variants]
    );
    const [visibleVariants, hiddenVariantCount] = React.useMemo(
       () =>
-         truncate > 0 && variants.length > truncate
-            ? [variants.slice(0, truncate), variants.length - truncate]
+         // If there are `max + 1` variants, show the last variant.
+         maxVisibleModels > 0 && variants.length > maxVisibleModels + 1
```
danielcliu-ifixit, line 21: but isn't the max visible numbers now technically `maxVisibleModels` + 1? So the parameter isn't always accurate

danielcliu-ifixit:
dev_block :v:

danielcliu-ifixit:
dev_block :v: on the max visible variants naming

frontend/components/common/CompatibleDevice.tsx
masonmcelvain: Do you have any ideas as to what we should call it? I was kind of at a loss for a clear name

masonmcelvain: We could do `maxModelLines` and restore the previous `-1` check.

danielcliu-ifixit: I honestly think `truncate` is _ok_ just not imo the best. I think `maxVariantsShown` could indicate, but basically i think theres two options here.

1.) We go to the old -1 way and the variable indicate the maximum lines under the specific devices section.
2.) We go with the current way and the variable indicates the maximum lines of variants (so the compatibility link will be +1 to the line count) under the specific device sections.

masonmcelvain: Yea I agree with you. See what you think about https://github.com/iFixit/react-commerce/pull/1183/commits/9ce8d42e2dbface16f5fbf209c42a67d81d46f94.  The +1, -1 logic is to handle the case where the # of variants === max lines to show.

frontend/components/common/CompatibleDevice.tsx
```diff
@@ -4,23 +4,26 @@ import { Product } from '@models/product';
...
       [device.variants]
    );
    const [visibleVariants, hiddenVariantCount] = React.useMemo(
       () =>
-         truncate > 0 && variants.length > truncate
-            ? [variants.slice(0, truncate), variants.length - truncate]
+         maxModelLines > 0 && variants.length > maxModelLines
+            ? [
+                 variants.slice(0, maxModelLines - 1),
+                 variants.length - maxModelLines + 1,
```
danielcliu-ifixit, line 23: yeha i think youre right there was a bug here

danielcliu-ifixit (approved):
CR :v: thanks

@QA should be a simple re-qa just make sure that the number of things in the device section is capped at 4.

masonmcelvain:
un_dev_block ✌🏻

masonmcelvain:
QA ✌🏻 

Checked the products that @erinemay checked earlier, and confirmed that the problem products now have compatibility.

Also checked that when there are 4 variants, 4 are shown

image


And when there are 5 or more, 3 are shown with the "and x more..." text

image

PR #1188, opened 2022-12-22
Increase CDN cache up to 1 day

### LINKED ISSUE #1187: Increase CDN stale while revalidate time
As part of #1156, we're taking a few measures to improve response time and thus our core web vitals score. While we're working on a server side Redis cache to mitigate slow responses on CDN cache misses, we can improve our current score by reducing the number of CDN cache misses in the first place. Right now our product pages are cached for 10 minutes, but we should experiment with longer cache times. Luckily thanks to SWR the CDN cache will be refreshed in background much more frequently, thus reducing the number of stale requests that we need to serve.

dhmacs:
closes #1187 

This PR increases stale time for product page as we aim to reduce cache misses.

Other changes:
- Added a small convenient duration library
- Standardized get server side props middleware conventions

## QA

Not sure how to QA this given that Vercel seems to override cache control on preview links. Got any idea @danielbeardsley @masonmcelvain @sterlinghirsh  ?
qa_req 0

### DIFF
--- a/frontend/helpers/cache-control-helpers.ts
+++ b/frontend/helpers/cache-control-helpers.ts
@@ -1,0 +1,30 @@
+import { GetServerSidePropsMiddleware } from '@lib/next-middleware';
+import { Duration } from '../lib/duration';
+
+interface CacheControlOptions {
+   sMaxAge: number;
+   staleWhileRevalidate: number;
+}
+
+export const withCacheControl =
+   (options: CacheControlOptions): GetServerSidePropsMiddleware =>
+   (next) =>
+   (context) => {
+      const sMaxAgeSeconds = options.sMaxAge / 1000;
+      const staleWhileRevalidateSeconds = options.staleWhileRevalidate / 1000;
+      context.res.setHeader(
+         'Cache-Control',
+         `public, s-maxage=${sMaxAgeSeconds}, stale-while-revalidate=${staleWhileRevalidateSeconds}`
+      );
+      return next(context);
+   };
+
+export const withCacheShort = withCacheControl({
+   sMaxAge: Duration(1).second,
+   staleWhileRevalidate: Duration(9).seconds,
+});
+
+export const withCacheLong = withCacheControl({
+   sMaxAge: Duration(1).minute,
+   staleWhileRevalidate: Duration(1).day,
+});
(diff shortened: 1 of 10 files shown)


frontend/helpers/cache-control-helpers.ts
```diff
@@ -0,0 +1,30 @@
...
+   };
+
+export const withCacheShort = withCacheControl({
+   sMaxAge: Duration(1).second,
+   staleWhileRevalidate: Duration(9).seconds,
+});
+
+export const withCacheLong = withCacheControl({
+   sMaxAge: Duration(1).minute,
+   staleWhileRevalidate: Duration(1).day,
```
danielbeardsley, line 29: I feel like this is a double-edged sword. Either it will:

* Significantly improve cache hit rates
   * If so, we'd be sending 1-day-old data to more people
   * Mostly this is a problem cause out-of-date quantities can effectively break the page
* Not improve cache hit rates
   * If so, then it doesn't really help

danielbeardsley:
I like the organisational changes, but I'm less sure about the 1-day TTL.

I vote we drop this back to 10-min and merge it cause we should keep the rest for sure.

frontend/helpers/cache-control-helpers.ts
dhmacs: > Mostly this is a problem cause out-of-date quantities can effectively break the page

What do you mean exactly by "break the page"? 

> I vote we drop this back to 10-min and merge it cause we should keep the rest for sure.

The main point of this PR is to increase the stale time. We want to see if core web vitals improves thanks to serving these pages faster. It might improve or it might not, but until we release this we won't know for sure. I hope to had this merged for a few days before we merge the Redis cache, so that we can see the performance impact of each change.

danielbeardsley: > What do you mean exactly by "break the page"?

I mean out-of-date quantities can mis-represent the truth and leave a product looking in-stock when it's not and vice-versa.

> The main point of this PR is to increase the stale time.

I'm aware of the intent. I'm suggesting let's merge the rest of this code change separately from the several-character change to the number. Outstanding code changes (especially those that refactor existing code) increase maintenance and may hinder development of other changes.

That said, I suppose we ought to let data guide us instead of intuition regarding out-of-stock or low-on-stock. I'm gonna try to write a query .....

Ok here's the results. This is the number of products each day who's in-stock status changed over the course of one day:
image

It's more than I expected. Though Iguess this change would mean for each product there'd probably be a max of a couple of page loads with out of date product info (cause of stale-while-revalidate).

Ok, let's just merge this and continue not worry that much about stock-changing problems until someone notices.

> before we merge the Redis cache,

I'm excited about this!

danielbeardsley (approved):
CR :+1:

danielbeardsley (approved):
CR :+1:

Issue #1181, opened 2022-12-21
Playwright: Fix `product/pro_user` test


ardelato:
## Description

This came up in https://github.com/iFixit/react-commerce/pull/1180. It seems this Playwright test is no longer working after the initial push in https://github.com/iFixit/react-commerce/pull/1149

### To Do:

- [ ] Figure out why we can no longer intercept and load the `pro user` correctly. 
- [ ] Unskip the test

PR #1189, opened 2022-12-22
Playwright: Fix broken pro_user test

mlahargou:
We forgot to add this param when we switched this to a Playwright test. It's needed so we can set pro user discounts.

QA:
--
The test should pass.

`pnpm playwright test --project=chromium --headed pro_user` 

Closes: https://github.com/ifixit/react-commerce/issues/1181

cc @ardelato

### DIFF
--- a/frontend/tests/playwright/utils.ts
+++ b/frontend/tests/playwright/utils.ts
@@ -6,6 +6,7 @@ export const interceptLogin = async (page: Page, userParams?: Object) => {
       algoliaApiKeyProduct: null,
       username: 'john',
       unique_username: 'john123',
+      privileges: [],
    };
 
    await page.route('**/api/2.0/user', (route) =>


frontend/tests/playwright/utils.ts
```diff
@@ -6,6 +6,7 @@ export const interceptLogin = async (page: Page, userParams?: Object) => {
       algoliaApiKeyProduct: null,
       username: 'john',
       unique_username: 'john123',
+      privileges: [],
```
ardelato, line 9: Wow that was all it took 😓.

Is there any way we can assign a matching type to this user object to keep it up-to-date? This property was not needed before to make the test pass.

ardelato (approved):
CR 👍  with question

frontend/tests/playwright/utils.ts
mlahargou: Let's make an issue for that. It's a small gotcha that took a bit to uncover.

mlahargou:
QA 📱 I ran the test and it passed.

PR #1193, opened 2022-12-22
Actions: Add workflow to run Playwright tests

### LINKED ISSUE #1114: Update our CI infrastructure to support running Playwright tests

batbattur:
## Summary
Adds workflow to run Playwright tests

## QA notes

- [x] CI should pass
- [x] Tracing should get uploaded as an artifact if the tests fail

**cc:** @mlahargou 
closes #1114


ardelato:
QA 👍 CI is working as expected and similar to `Cypress`

frontend/playwright.config.ts
```diff
@@ -114,8 +116,8 @@ const config: PlaywrightTestConfig = {
 
    /* Run your local dev server before starting the tests */
    webServer: {
-      cwd: '../',
-      command: 'pnpm run dev',
+      cwd: process.env.CI ? './' : '../',
```
mlahargou, line 119: What's this change about?

frontend/playwright.config.ts
```diff
@@ -114,8 +116,8 @@ const config: PlaywrightTestConfig = {
...
+      command: process.env.CI ? 'pnpm run build && pnpm start' : 'pnpm dev',
```
mlahargou, line 120: Should this be

`command: process.env.CI ? 'pnpm run build && pnpm start' : 'pnpm run dev',`

.github/workflows/E2Eplaywright.yml
```diff
@@ -21,7 +21,20 @@ jobs:
...
+         - name: Get installed Playwright version
+           id: playwright-version
+           run: echo "PLAYWRIGHT_VERSION=$(pnpm list -r "@playwright/test" | grep "@playwright/test" | tr -dc 1-9.)" >> $GITHUB_ENV
+
+         - name: Cache playwright binaries
+           uses: actions/cache@v3
+           id: playwright-cache
+           with:
+              path: ~/.cache/ms-playwright
+              key: ${{ runner.os }}-playwright-${{ env.PLAYWRIGHT_VERSION }}
```
mlahargou: Cool!

mlahargou (approved):
Looks good to me! 

Odd idea here, but should we have a job that depends on both of these jobs (cypress and playwright) that notifies us if one passes and the other doesn't? We could then easily tell if we are getting the same behavior between our cypress tests and playwright.

CR 📱

frontend/playwright.config.ts
batbattur: In CI, we want to run only the `frontend` (in the current `./` frontend folder) because the `backend` will be already running using docker.

However, when testing locally, we want both back and front ends to start so we have to be in the root folder (thus, we use `../`).

frontend/playwright.config.ts
batbattur: I think `pnpm run dev` vs `pnpm dev` are basically the same. From https://pnpm.io/cli/run#examples: 

> Let's say you have a watch script configured in your package.json, like so:
> ```
>"scripts": {
>    "watch": "webpack --watch"
>}
>```
> You can now run that script by using pnpm run watch! Simple, right? Another thing to note for those that like to save keystrokes and time is that all scripts get aliased in as pnpm commands, so ultimately pnpm watch is just shorthand for pnpm run watch (ONLY for scripts that do not share the same name as already existing pnpm commands).

ardelato: Ah, @batbattur we should make this consistent so there is no confusion. Let's remove the `run` part for `pnpm run build`


@mlahargou it is basically the same thing for `pnpm`

> Another thing to note for those that like to save keystrokes and time is that all scripts get aliased in as pnpm commands, so ultimately pnpm watch is just shorthand for pnpm run watch (ONLY for scripts that do not share the same name as already existing pnpm commands).

https://pnpm.io/cli/run

ardelato:
> Odd idea here, but should we have a job that depends on both of these jobs (cypress and playwright) that notifies us if one passes and the other doesn't? We could then easily tell if we are getting the same behavior between our cypress tests and playwright.

Can you elaborate on what you're thinking about the job? I am wondering what would be different than just checking the CI checks manually.

frontend/playwright.config.ts
batbattur: Updated in https://github.com/iFixit/react-commerce/pull/1193/commits/668c528b9ec44fa65d010d32203238d81bcc9476.

mlahargou:
> Can you elaborate on what you're thinking about the job? I am wondering what would be different than just checking the CI checks manually.

Sure!

Basically, the check would do `cypress_status XOR playwright_status`. Meaning it would only pass when one passed and the other failed. The benefit of a separate job would be that we could just look at a single status to determine if we are getting differing results between the two jobs. Otherwise, we would have to cross-reference manually. Anyways, just a thought, nothing to block on. Probably overkill as we are only going to have both running for a short amount of time.

mlahargou (approved):
CR 📱

PR #1194, opened 2022-12-23
Playwright Tests: Extract Search Query Response to Utils

ardelato:
## Description

This came up in https://github.com/iFixit/react-commerce/pull/1180#discussion_r1055888310.
We have this piece of code duplicated across some of our current tests, let's just extract it to a utils file and reuse it.

While extracting it, I also simplified the matching logic a bit.

## QA 

- [ ] CI should pass

Connects: https://github.com/iFixit/react-commerce/issues/1111

### DIFF
--- a/frontend/tests/playwright/utils.ts
+++ b/frontend/tests/playwright/utils.ts
@@ -22,6 +22,15 @@ export const interceptLogin = async (page: Page, userParams?: Object) => {
    );
 };
 
+export const waitForAlgoliaSearch = async (page: Page) => {
+   return page.waitForResponse(async (response) => {
+      return (
+         response.url().includes('1/indexes/*/queries') &&
+         response.status() === 200
+      );
+   });
+};
+
 /*
  * This function is used to resolve a path in an object using dot notation.
  * It is a way to mimic the `its` function in Cypress by using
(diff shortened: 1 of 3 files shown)


jordycosta:
QA 🟢 

CI is passing 
[View `E2E tests / playwright-run` report](https://github.com/iFixit/react-commerce/actions/runs/3767362736/jobs/6404817087)

Screenshot 2022-12-26 105304

frontend/tests/playwright/utils.ts
```diff
@@ -22,6 +22,15 @@ export const interceptLogin = async (page: Page, userParams?: Object) => {
    );
 };
 
+export const waitForAlgoliaSearch = async (page: Page) => {
+   return page.waitForResponse(async (response) => {
+      return (
+         response.url().includes('1/indexes/*/queries') &&
```
mlahargou, line 28: Not super familiar with the algolia requests, but why is there a `1` at the begining? Also, we did we drop the `algolia` check?

mlahargou (approved):
CR 📱

frontend/tests/playwright/utils.ts
ardelato: I am not sure about the `1` either. This is similar to what we do with Cypress 
https://github.com/iFixit/react-commerce/blob/main/frontend/tests/cypress/support/commands.ts#L28

PR #1195, opened 2022-12-29
Remove price facet from /Parts collections

### LINKED ISSUE #533: Results shown outside price range selected

### LINKED ISSUE #1190: Only show price range facet on tools collections
We've spent a bunch of time trying to make the price range facet work in a reasonable and predictable way since parts product groups typically have multiple variants with different prices. Shopping by price only really makes sense on tool collections, so let's hide the facet on parts and marketing pages. It should be shown on all-tools and any tools collection.

federicobadini:
closes #533
closes #1190 

### QA

1. Visit Vercel Preview
2. Verify that the price-range filter is not showing anymore in /Parts collections but is still visible in /Tools collections

### DIFF
--- a/frontend/templates/product-list/sections/FilterableProductsSection/facets/useFacets.tsx
+++ b/frontend/templates/product-list/sections/FilterableProductsSection/facets/useFacets.tsx
@@ -59,7 +59,10 @@ export function useFilteredFacets(productList: ProductList) {
       return facets.filter((facet) => !excludedToolsFacets.includes(facet));
    }
 
-   return usefulFacets;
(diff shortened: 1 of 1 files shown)


deltuh-vee:
QA 🎬 
* Price range facet has been removed from /Parts
* Price range facet is still visible and usable in /Tools

kwiens:
CR 👍

davidrans (approved):
CR :coffee:

Issue #1206, opened 2023-01-02
Featured Products on ProductDetailPage has no height limit


jeffsnyder:
Featured Products on ProductDetailPage has no height limit so the section become rather tall to accommodate the square five-across product grid.  And not all product images are sharp at this size (might be a separate issue/problem). 

![image](https://user-images.githubusercontent.com/7053450/210273174-479f429e-5415-4a1d-abe7-188fd1843640.png)

https://www.ifixit.com/products/fixmat

I'd propose we set a max height for this section and pull in more products to the section. It might make sense to convert this section to a carousel while we're at it.

davidrans:
Let's iterate on the design here, cause my hotfix isn't ideal @mmarcon91 @dhmacs

mmarcon91:
@jeffsnyder what resolution does your monitor have?

@davidrans  In my opinion you could solve it by have 6 products over 1536px

Base : 2x3 products
Above 768px : 3x2 products
Above at 1024px : 5 products
Above 1536px : 6 products

jeffsnyder:
I'll have to recheck when I'm in the office later this week. The monitor I'm on now is 1920 x 1080. The feature is different now and it all looks good to me, even when I zoom out. 👍

mmarcon91:
If we want to keep the product list inside a wrapper the design has to be rethought. Because the way it was designed it had to be (at least in most displays) all width.

sterlinghirsh:
@mmarcon91 I think you're right that we should take another look at the design. It's definitely too big with 5 products at the screen size we're looking at. 6 products or maybe even more could work. I'm ok with putting it in a wrapper if necessary, but I did like the look of the original full-width version on narrower windows. I will sometimes have a full screen browser window at 4k, and admittedly most websites don't look good, but we should at least try to make it look good at 1920 wide and try to make it look reasonable if we go past that.

mmarcon91:
> @mmarcon91 I think you're right that we should take another look at the design. It's definitely too big with 5 products at the screen size we're looking at. 6 products or maybe even more could work. I'm ok with putting it in a wrapper if necessary, but I did like the look of the original full-width version on narrower windows. I will sometimes have a full screen browser window at 4k, and admittedly most websites don't look good, but we should at least try to make it look good at 1920 wide and try to make it look reasonable if we go past that.

I suggest removing the wrapper and using these rules:

Base : 2x3 products
Above 768px : 3x2 products
Above at 1024px : 5 products
[...]

PR #1207, opened 2023-01-02
Featured Products: fix unbounded width

davidrans:
| Before | After |
| --------- | ------ |
| image | image |

CR
---
Ignore whitespace

Closes https://github.com/iFixit/react-commerce/issues/1206

### DIFF
--- a/frontend/templates/product/sections/FeaturedProductsSection/index.tsx
+++ b/frontend/templates/product/sections/FeaturedProductsSection/index.tsx
@@ -57,7 +57,7 @@ export function FeaturedProductsSection({
                }}
                spacing="1px"
                bg="gray.200"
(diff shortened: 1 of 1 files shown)


sumukhipandey (approved):
CR 🦋

deltuh-vee:
QA 🎬 
The width of the Featured Products section is now limited.

dhmacs:
@davidrans I think probably we should have solved this by increasing image resolution. And if a design change is really needed, imo this should have prompted a design iteration, because the design looks a bit odd now that the cards do not take the full width of the page

davidrans:
I've re-opened the issue :+1: https://github.com/iFixit/react-commerce/issues/1206

Issue #1201, opened 2023-01-02
Product Page: Add Playwright tests for Reviews


ardelato:
## Description

This is an ongoing effort in addressing https://github.com/iFixit/react-commerce/issues/452

### Requirements

- [ ] Check products with multiple reviews
	- [ ] Make sure `Sort by` and `Filter by` works
	- [ ] Click on `see more reviews` and make sure the rest of the reviews display properly

PR #1208, opened 2023-01-02
Product Page: Add reviews test

batbattur:
## Summary
Added a test for a product with multiple reviews.

The test compares the number of reviews before hitting the `see more reviews` button and after. 
Also expects that all of the reviews are visible after clicking the button.

## QA notes
Passing CI is enough (qa_req 0).

closes: #1201 Note: currently, we don't have the `filter by` or `sort by` buttons implemented for the reviews


davidrans (approved):
CR :coffee:

PR #1211, opened 2023-01-03
Hide aggregate rating and reviewCount if needed

### LINKED ISSUE #1140: Bug: Review hiding logic isn't working
Two problems with the Reviews section. 

1) Ratings with no text Review should not show up in the Reviews section 
2) If the average Rating is <4 and there are <10 total reviews, we should not show the Reviews section. (We might want to still show the "Write a review" banner.. this component is new since we defined the desired logic.)

Possibly related to https://github.com/iFixit/react-commerce/issues/215  ? Or maybe this issue didn't solve the problem? 

Issue from old template with links to related issues, logic, etc: https://github.com/iFixit/shopify-v2/issues/784

Example product: https://www.ifixit.com/products/anti-clamp

image

This should be fixed ASAP as it is affecting our new release products. 

@erinemay @dhmacs @sterlinghirsh

federicobadini:
closes #1140

This PR removes review data from the ld+json structured data unless `product.rating.value >= 4 || product.reviewsCount > 10`.

### QA

1. Visit Vercel preview ([here](https://react-commerce-prod-git-review-hiding-logic-remov-c801ba-ifixit.vercel.app/products/anti-clamp) is a sample product with hidden reviews)
2. Verify that the review section in the product ld+json is not present anymore if `product.rating.value < 4 && product.reviewsCount <= 10`

### DIFF
--- a/frontend/templates/product/MetaTags.tsx
+++ b/frontend/templates/product/MetaTags.tsx
@@ -135,7 +135,9 @@ export function MetaTags({ product, selectedVariant }: MetaTagsProps) {
                name: metaTitle || undefined,
                url: selectedVariantUrl,
                aggregateRating:
-                  product.rating && product.reviewsCount
+                  product.rating?.value &&
+                  product.reviewsCount &&
+                  (product.rating.value >= 4 || product.reviewsCount > 10)
                      ? {
                           '@type': 'AggregateRating',
                           ratingValue: product.rating.value,


frontend/templates/product/MetaTags.tsx
```diff
@@ -135,7 +135,9 @@ export function MetaTags({ product, selectedVariant }: MetaTagsProps) {
                name: metaTitle || undefined,
                url: selectedVariantUrl,
                aggregateRating:
-                  product.rating && product.reviewsCount
+                  product.rating?.value &&
+                  product.reviewsCount &&
+                  (product.rating.value >= 4 || product.reviewsCount > 10)
```
masonmcelvain, line 140: The spec says "If the average Rating is <4 and there are <10 total reviews, we should not show the Reviews section", so I would expect this to be

```suggestion
                  (product.rating.value >= 4 || product.reviewsCount >= 10)
```

But not sure if @jeffsnyder cares about that edge case.

masonmcelvain (approved):
CR ✌🏻 but deploy_block ✌🏻 on a minor detail

frontend/templates/product/MetaTags.tsx
federicobadini: Yep I know this is a slight deviation but there were already other references in code that enforced `> 10` without the equal.

I also checked the old repo and you can see [here](https://github.com/iFixit/ifixit/blob/5dc8b9d17fefc794a2e43b58746f0bd367922289/Cart/Objects/ProductReviews.php#L14-L15) that `10` is the max values of reviews **to hide**. 
That's why it was coded as above

masonmcelvain: Thanks for the explanation!

masonmcelvain:
un_deploy_block ✌🏻

deltuh-vee:
QA 🎬 
The review and aggregateRating fields are now gone for products that have a rating less than 4 and have less than 10 reviews.

PR #1213, opened 2023-01-04
Vulcan Proof-of-Concept: Create page

andyg0808:
This doesn't have anything in it yet, but it shows that I know how to create a new page which accepts a wiki name as a part of its URL. It also reserves this directory name for now.

## CR Notes
I'd particularly appreciate feedback on whether this is following appropriate conventions; I've tried to imitate some of the style I saw, but I'm not sure if I'm understanding the logic of it.

## QA Notes

qa_req 0 There's nothing interesting here, and CR can validate that this shouldn't affect any other page.

Epic https://github.com/iFixit/ifixit/issues/46024
Closes ifixit/ifixit#46052


frontend/templates/vulkan/server.tsx
```diff
@@ -0,0 +1 @@
+export { getServerSideProps } from '@templates/page/server';
```
dhmacs, line 1: Just to be sure, you're using the `getServerSideProps` from the `page` template just as a placeholder? Although `page` might seems like a generic concept, it's just a specific template to handle custom pages (such as the store home page, and any other one-off page made of sections). To be clear, it's not a template to be used for building up other templates

andyg0808: That's exactly the kind of thing I wanted to know; thanks. I assumed it was relatively generic. Is there a good place to get the props required by the layout? I didn't want to have to do a bunch of copy-paste to generate the correct data for it.

danielbeardsley:
Vulcan with a "K"? Is there some inside joke in that choice? I propose sticking with `Vulcan` otherwise, there may be constant misspellings. Also, I never know if we should use internal code names for project directories or not. Vulcan is definitely not descriptive and is effectively a random name and I haven't settled on whether that's better or worse than a more verbose but descriptive name.

andyg0808:
Thanks, @danielbeardsley. Totally missed that the codename is spelled with a "c"; I think the graphics API probably is the version of that word I've seen the most, so I just assumed that was the normal spelling.

andyg0808:
Comments addressed:
1. Renamed UI to `Vulcan` and renamed library files to `troubleshooting`
2. Replicated the setup being used in `page/server.tsx` in `troubleshooting/server.tsx`, adapted to work correctly there

andyg0808:
Sorry about the force-push; I'd messed up the test and didn't think it was worth leaving the mistake in history. Everything before [6cd4f8d](https://github.com/iFixit/react-commerce/pull/1213/commits/6cd4f8d0117be1cccdb5ef3884d10e0c687e84a3) should be the same as it was; I was just squashing a few recent commits.

frontend/templates/vulkan/server.tsx
dhmacs: `templates/vulkan/server.tsx` is the correct place where to define the `getServerSideProps` for the `vulkan` template. But you need to define the function yourself, not import it from another template, as each template would have its own data requirement. Looks like what you did with the troubleshooting template is the correct approach 👍 Some props will be the same for every template (e.g. the layouts data), but I'd expect every template to have also its own specific props based on the data dependencies of the template

frontend/tests/playwright/troubleshooting/loading.spec.ts
```diff
@@ -5,4 +5,12 @@ test.describe('Vulcan page', () => {
...
       await expect(page.getByText('iPhone')).toBeVisible();
    });
+
+   test('it should not be indexed', async ({ page }) => {
+      await page.goto('/Vulcan/iPhone');
+      // check that the meta robots tag is set to noindex
+      expect(
+         page.locator('meta[name="robots"][content="noindex"]')
+      ).toBeTruthy();
+   });
```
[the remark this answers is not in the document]
andyg0808: Fixed in [aa69e4d](https://github.com/iFixit/react-commerce/pull/1213/commits/aa69e4d935acf35805aafe77bfb08e62aa1bb86c).

danielbeardsley (approved):
CR :+1:

PR #1214, opened 2023-01-04
Show disabled product pages

### LINKED ISSUE #1153: Show Not For Sale page on Disabled product
## Problem
When a product is disabled, we currently display a 404, which was easier than designing an additional page state. However, it's a better UX to be able to show the product in a not-for-sale state.

## Do This
When all variants are disabled on a product, still show the product page, with some changes:
* Say "Not for Sale" instead of showing price
* No Variant details
* No Variant selector
* No add to cart
* No notify widget
* No shipping restrictions
* No International Buy Box
* No Value Propositions
* No cross sell (Featured products are OK)
* No stock details

Noindex these pages.

We will keep the images, description, warranty, video, etc.

There is no change to the existing behavior of a disabled variant on an otherwise enabled product.

Asana Ref: https://app.asana.com/0/1203495207405875/1203556380900043/f

masonmcelvain:
Shows disabled product pages, hides ecommerce components on those pages, and noindexes the pages.

## QA

Mark a product as disabled by setting the "Enabled 2" variant metafield to false for all variants on a product.

image

Then verify that the product page meets the spec in the parent issue.

CC @sterlinghirsh 

Closes #1153


masonmcelvain:
This is what the UI looks like. I am not wedded to it if we would prefer a different implementation, as I simply reused an existing component for this.

image

erinemay:
QA 🌴 Checked a [handful of product pages](https://docs.google.com/spreadsheets/d/1Jx0HSfwEeu3VGc8Ylv61LTlAmWoKHyhDF0NTc3sva8M/edit?usp=sharing) when none/some/all variants are disabled
Logic ✔️ 
- When at least one variant is enabled, only those enabled variants are shown or accessible via variant selection
- When all variants are disabled (either by setting `Enabled 2=false', or both `Disable When OOS=true` and available inventory = 0), the disabled page is shown.

Page details ✔️ 
- Displayed SKU and image are of the default variant
- No reviews
- All according to spec ✔️ 
   - No Variant details
   - No Variant selector
   - No add to cart
   - No notify widget
   - No shipping restrictions
   - No International Buy Box
   - No Value Propositions
   - No cross sell 
   - Yes featured products
   - No stock details 

deploy_block 🧷 on the appearance and a suggestion

>This is what the UI looks like. I am not wedded to it if we would prefer a different implementation, as I simply reused an existing component for this.

<details><summary>It looks a little busy on some pages because the description and kit contents sections are still automatically expanded.</summary>

https://react-commerce-git-not-for-sale-product-page-ifixit.vercel.app/products/galaxy-buds-plus-batteries

image

</details>

@masonmcelvain, what do you think about keeping all the detail dropdowns collapsed on disabled pages?

mmarcon91:
> Thanks @erinemay! I agree that it looks busy. Perhaps we could get @mmarcon91's opinion on the best UI for this situation?

On these pages as Erin says it would be better not to expand those sections. Giving a lot of information about something that you can't actually buy doesn't make a lot of sense.

At the same time I don't know how many such products there are and putting more conditions on the product page could slow down future implementations. 

1) UX side would be better to do.
2) Code side probably not.

Since it will not be a page that needs conversion optimization I would avoid "complicating" the code

masonmcelvain:
Thanks for your feedback @mmarcon91.

> Since it will not be a page that needs conversion optimization I would avoid "complicating" the code

This makes sense. However, this entire PR is already complicating the code for products that can be sold. I think that adding one more boolean check is ok if it provides a better UX. If we want to remove this feature for performance/simplicity reasons in the future, it shouldn't be too hard to revert.

masonmcelvain:
un_deploy_block ✌🏻 The accordians of disabled products are now collapsed by default.

image

sterlinghirsh (approved):
CR :zap:

k0rvusk0r4x:
- Products pages where all variants are enabled continue to load and function as on master.
- Product pages that have both active and inactive variants continue to behave as on master.
- Product pages where all variants are disabled:

   - Have the "Not for sale" message
   - Have the accordions collapsed when first loading
   - The accordions function (can be expanded and collapsed) if the user wishes to see the information despite the product not being available.

QA 😺

PR #1215, opened 2023-01-04
Actions: Add linter workflow

batbattur:
## Summary
1. Installed `eslint-plugin-jest` plugin and added `no-focused-tests` eslint rule to prevent `.only` in jest tests.
2. Fixed existing errors from the linter.
3. Added a new workflow to run the linter.
 
Note: Playwright tests already have `forbidOnly` (checks for `.only`) set correctly in the config so we don't have to worry about it.

## QA notes
1. Find a jest test and add `.only` like the following:
```diff
frontend/tests/jest/tests/CompatibleDevice.test.tsx

 describe('CompatibleDevice', () => {
-   it('renders and matches the snapshot', () => {
+   it.only('renders and matches the snapshot', () => {
       const product = getMockProduct();

 ```

2. Then run `cd frontend && pnpm lint`. 
4. Make sure the linter returns the following error: `Unexpected focused test  jest/no-focused-tests`

connects: https://github.com/iFixit/ifixit/issues/45873

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,46 @@
+name: ESlint
+
+on:
+   push:
+      branches-ignore:
+         - 'main'
+
+jobs:
+   linter:
+      runs-on: ubuntu-latest
+      steps:
+         - name: Checkout
+           uses: actions/checkout@v2
+           with:
+              token: ${{ secrets.ACTION_PAT }}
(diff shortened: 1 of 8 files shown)


deltuh-vee:
QA 🎬 
`.only` is caught by the `no-focused-tests` test.

mlahargou (approved):
CR 📱 Super cool!

PR #1218, opened 2023-01-04
Product Page: Add tests for images

### LINKED ISSUE #1200: Product Page: Add tests for Product Images
## Description

This is an ongoing effort in addressing https://github.com/iFixit/react-commerce/issues/452

### Requirements

- [ ] Check products with no image
- [ ] Check products with a single image
- [ ] Check products with multiple images
   - [ ] Select the other image and make sure the other image shows up

batbattur:
## Summary
Added tests that:
-  checks a product with no image and asserts `No photos available for this product` image placeholder is visible.
-  checks a product with a single image and asserts that the image is visible.
-  checks a product with multiple images and asserts that clicking on the next image button shows the other image by checking that they have different sources. 

## QA Notes
Passing CI is enough. 

qa_req 0
closes #1200

### DIFF
--- a/frontend/tests/playwright/product/product_image.spec.ts
+++ b/frontend/tests/playwright/product/product_image.spec.ts
@@ -1,4 +1,4 @@
-import { test, expect } from '@playwright/test';
+import { test, expect, Page } from '@playwright/test';
 
 test.describe('Product image test', () => {
    test('product with a single image ', async ({ page }) => {
@@ -9,4 +9,33 @@ test.describe('Product image test', () => {
          .first();
       expect(image).toBeVisible;
    });
+
+   test('product with multiple images ', async ({ page }) => {
+      await page.goto('/products/repair-business-toolkit');
+
+      const firstImageSrc = await getCurrentActiveImgSrc(page);
+      const firstImage = page.locator(`img[src="${firstImageSrc}"]`).first();
+      await expect(firstImage).toBeVisible();
+
+      // Click on the arrow for the next image
+      await page.getByTestId('swiper-next-image').first().click();
+
+      const secondImageSrc = await getCurrentActiveImgSrc(page);
+      const secondImage = page.locator(`img[src="${secondImageSrc}"]`).first();
+      await expect(secondImage).toBeVisible();
+
+      // Assert that currently active image has different src than the first one
+      expect(firstImageSrc).not.toEqual(secondImageSrc);
+   });
+
+   /*
+    * Returns the src of the currently displayed active image
+    * Applies only for product variants with multiple images
+    */
+   async function getCurrentActiveImgSrc(page: Page) {
+      const divElement = await page.$('.swiper-slide.swiper-slide-active');
+      const imgElement = await divElement!.$$('img');
+      const src = await imgElement[0].getAttribute('src');
+      return src;
+   }
 });
--- a/frontend/tests/playwright/product/product_image.spec.ts
+++ b/frontend/tests/playwright/product/product_image.spec.ts
@@ -1,0 +1,12 @@
+import { test, expect } from '@playwright/test';
+
+test.describe('Product image test', () => {
+   test('product with a single image ', async ({ page }) => {
+      await page.goto('/products/iflex-opening-tool');
+
+      const image = page
+         .getByRole('img', { name: 'iFlex Opening Tool' })
+         .first();
+      expect(image).toBeVisible;
+   });
+});
--- a/frontend/tests/playwright/product/product_image.spec.ts
+++ b/frontend/tests/playwright/product/product_image.spec.ts
@@ -1,4 +1,4 @@
-import { test, expect, Page } from '@playwright/test';
+import { test, expect } from '@playwright/test';
 
 test.describe('Product image test', () => {
    test('product with a single image ', async ({ page }) => {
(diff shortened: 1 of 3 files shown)


frontend/tests/playwright/product/product_image.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+
+   /*
+    * Returns the src of the currently displayed active image
+    * Applies only for product variants with multiple images
+    */
+   async function getCurrentActiveImgSrc(page: Page) {
+      const divElement = await page.$('.swiper-slide.swiper-slide-active');
+      const imgElement = await divElement!.$$('img');
+      const src = await imgElement[0].getAttribute('src');
+      return src;
```
mlahargou, line 39:
```suggestion
      return await imgElement[0].getAttribute('src');
```

frontend/tests/playwright/product/product_image.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+      expect(firstImageSrc).not.toEqual(secondImageSrc);
+   });
+
+   /*
+    * Returns the src of the currently displayed active image
+    * Applies only for product variants with multiple images
+    */
+   async function getCurrentActiveImgSrc(page: Page) {
+      const divElement = await page.$('.swiper-slide.swiper-slide-active');
+      const imgElement = await divElement!.$$('img');
```
mlahargou, line 37: I _think_ this could be something like this. 

```
return await page.locator('.swiper-slide-active').getByRole('img').getAttribute('src');
```

Might have to index into it still. Although I think locators automatically call `first()` if another locator is attached.

frontend/tests/playwright/product/product_image.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+         .getByRole('img', { name: 'iFlex Opening Tool' })
+         .first();
+      expect(image).toBeVisible;
+   });
+
+   test('product with multiple images ', async ({ page }) => {
+      await page.goto('/products/repair-business-toolkit');
+
+      const firstImageSrc = await getCurrentActiveImgSrc(page);
+      const firstImage = page.locator(`img[src="${firstImageSrc}"]`).first();
```
mlahargou, line 17: Are we sure this isn't grabbing the main active image? We want the image in the slider, right?

frontend/tests/playwright/product/product_image.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+      await expect(firstImage).toBeVisible();
+
+      // Click on the arrow for the next image
+      await page.getByTestId('swiper-next-image').first().click();
```
mlahargou, line 21: Do we need `first()`? I thought it automatically got the first one if another action is appended to it.

```suggestion
      await page.getByTestId('swiper-next-image').click();
```

frontend/tests/playwright/product/product_image.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+
+      const secondImageSrc = await getCurrentActiveImgSrc(page);
+      const secondImage = page.locator(`img[src="${secondImageSrc}"]`).first();
+      await expect(secondImage).toBeVisible();
+
+      // Assert that currently active image has different src than the first one
+      expect(firstImageSrc).not.toEqual(secondImageSrc);
```
mlahargou, line 28: Nice! I was wondering how we were going to differentiate the two.

frontend/tests/playwright/product/product_image.spec.ts
```diff
@@ -0,0 +1,41 @@
+import { test, expect, Page } from '@playwright/test';
```
mlahargou, line 1: Where do we use `Page`?

mlahargou (approved):
CR 📱 

Nice job on this! A couple of questions/comments. deploy_block 📱

frontend/tests/playwright/product/product_image.spec.ts
batbattur: Previously, we were using it in the helper function's parameter type in the older commit 
https://github.com/iFixit/react-commerce/blob/8d89f8f8b1e5fbcc984f4580de2d5d65038b92cb/frontend/tests/playwright/product/product_image.spec.ts#L35

Now, deleted the helper function in https://github.com/iFixit/react-commerce/pull/1218/commits/acbd5dea522fe74827e26b5626c071636a4869ac so we don't need to import it anymore.

frontend/tests/playwright/product/product_image.spec.ts
batbattur: I had to use `.first()` because it was throwing an error because it found 2 elements:

Screen Shot 2023-01-05 at 6 06 57 PM

> I thought it automatically got the first one if another action is appended to it.

This, I cannot confirm but it seems like it fails when it finds multiple elements.

frontend/tests/playwright/product/product_image.spec.ts
batbattur: The locator solution worked nicely and I had to use `first()`. It doesn't seem like it automatically calls `first()`.

Updated in https://github.com/iFixit/react-commerce/pull/1218/commits/acbd5dea522fe74827e26b5626c071636a4869ac to use the suggested changes above and got rid of the helper function.

frontend/tests/playwright/product/product_image.spec.ts
batbattur: Decided to get rid of this helper function and call it directly in the test (https://github.com/iFixit/react-commerce/commit/acbd5dea522fe74827e26b5626c071636a4869ac).

frontend/tests/playwright/product/product_image.spec.ts
batbattur: > Are we sure this isn't grabbing the main active image? 

I am not exactly sure what you meant by the `main active image`. 

> We want the image in the slider, right?

Yes! We get the `src` of the image in the slider by `const firstImageSrc = await getCurrentActiveImgSrc(page)`.

Then use `const firstImage = page.locator(img[src="${firstImageSrc}"]).first()` to get the element, and use 
`await expect(firstImage).toBeVisible()` to make sure it's visible. 

Then we click on the next image and repeat the same process.

Finally, we compare the sources of the first and second images shown in the slider, and make sure 
they're different (as noted in https://github.com/iFixit/react-commerce/pull/1218#discussion_r1063006118).

frontend/tests/playwright/product/product_image.spec.ts
mlahargou: TIL!

ardelato: To note, playwright will throw an error if more than one matching element is found https://playwright.dev/docs/locators#throws-an-error-if-more-than-one

ardelato (approved):
CR 👍 nice job @batbattur

batbattur:
un_deploy_block 🚛 addressed the questions and comment above

PR #1222, opened 2023-01-05
Sentry: upgrade to 7.29.0

davidrans:
We were quite a few (minor) versions behind here. The last release added some performance improvements for NextJS: https://github.com/getsentry/sentry-javascript/releases/tag/7.29.0


masonmcelvain (approved):
CR ✌🏻

erinemay:
I'm having local-dev troubles, even on main. I'm gonna remove my qaing label, but I assume we should be able to add and trigger an error and see it in Sentry still.

Are there any little gotchas like needing to bypass sampling ifixit/ifixit?

masonmcelvain:
I don't think we sample at all for react-commerce, meaning all errors should go to the react-commerce-dev project on Sentry.

deltuh-vee:
QA 🎬 
Still able to send errors to Sentry.

Issue #1203, opened 2023-01-02
Product Page: Add Tests for Breadcrumbs


ardelato:
## Description

This is an ongoing effort in addressing https://github.com/iFixit/react-commerce/issues/452

### Requirements

⚠️ It might be better to use Playwright for this.

- [ ] Make sure bread crumbs display properly on mobile and desktop.

PR #1227, opened 2023-01-06
Product Page: Add Tests for Breadcrumbs

batbattur:
## Summary
Added tests that check breadcrumbs on mobile/tablet and desktop.

On mobile and tablet (viewport width sizes below 1000px), we display only the last child breadcrumb
regardless of the number of breadcrumb links. Everything else is collapsed in the breadcrumb menu.
The test for mobile asserts that the last child breadcrumb and the mobile breadcrumb menu are visible.

On the desktop, we display 2 ancestors and the last child breadcrumb (in total 3 breadcrumbs). 
Everything else is collapsed in the breadcrumb menu. The tests for the desktop assert that the 
last child and ancestor breadcrumbs are visible, and the menu is visible/hidden when appropriate.  

## QA notes
Passing CI is enough.

qa_req 0
closes: #1203

### DIFF
--- a/frontend/tests/playwright/product/product_breadcrumb.spec.ts
+++ b/frontend/tests/playwright/product/product_breadcrumb.spec.ts
@@ -80,5 +80,40 @@ test.describe('Product breadcrumb test', () => {
          );
          await expect(mobileBreadcrumbMenu).not.toBeVisible();
       });
+
+      /*
+       * Same as the test above, but checks that the desktop breadcrumb menu is visible.
+       */
+      test('with more than 3 breacrumb links', async ({ page }) => {
+         await page.goto('/products/iphone-6s-plus-replacement-battery');
+
+         const lastChildBreadcrumb = page.getByTestId(
+            'breadcrumb-last-child-link'
+         );
+         await expect(lastChildBreadcrumb).toBeVisible();
+
+         const desktopBreadcrumbAncestorLinks = page.getByTestId(
+            'breadcrumb-ancestor-link-desktop'
+         );
+         const ancestorLinksCount =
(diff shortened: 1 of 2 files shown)


frontend/tests/playwright/product/product_breadcrumb.spec.ts
```diff
@@ -68,6 +68,17 @@ test.describe('Product breadcrumb test', () => {
             ).not.toBeNull();
          }
          expect(ancestorLinksCount).toBeLessThanOrEqual(3);
+
+         // If the product has less than or equal to 3 breadcrumb links, we don't display the breadcrumb menu
+         const desktopBreadcrumbMenu = page.getByTestId(
+            'breadcrumb-menu-desktop'
+         );
+         await expect(desktopBreadcrumbMenu).not.toBeVisible();
```
ardelato, line 76: If we are not using the set initialized variable later in the code why not pass the retrieval of the element directly to the assertion?

```suggestion
         await expect(page.getByTestId(
            'breadcrumb-menu-desktop'
         )).not.toBeVisible();
```

ardelato (approved):
CR 👍 with minor suggestion

frontend/tests/playwright/product/product_breadcrumb.spec.ts
batbattur: Thanks @ardelato for the suggestion! That makes it very cleaner! Updated to use the elements directly for the unused variables in https://github.com/iFixit/react-commerce/pull/1227/commits/7e5c2b4f96ac281a72e3117878516acd67cc5d31.

mlahargou (approved):
CR 📱 Tests look good!

PR #1229, opened 2023-01-07
Troubleshooting: Load wiki text on page

andyg0808:
This is using the new API created in https://github.com/iFixit/ifixit/pull/46057. It's built on top of the page from https://github.com/iFixit/react-commerce/pull/1213, now doing a fetch from the API and rendering separate components for each solution on the page.

The styling isn't pretty yet, but that's a problem for future us.
image

## CR Notes
The tests are expected to fail until the API from https://github.com/iFixit/ifixit/pull/46057 goes out.

## QA Notes

qa_req 0
This is still not really interesting to test, since it doesn't even look nice. The only real functionality is the basic loading of wikis, and I've not yet messed around with enough wikis to think that this should be considered particularly correct.

Closes ifixit/ifixit#46075
Closes ifixit/ifixit#46076


andyg0808:
deploy_block :+1: on #1213 going out.

frontend/templates/troubleshooting/index.tsx
```diff
@@ -2,19 +2,73 @@ import { useRouter } from 'next/router';
...
+   );
+   React.useEffect(() => {
+      client
+         .get(`Troubleshooting/${wikiname}`, {
+            credentials: 'include',
+         })
+         .then((res: any) => {
+            setWikiData(res);
+         });
+   }, [wikiname]);
```
dhmacs, line 46: Shouldn't data be fetched server side (`getServerSideProps`) so that the page is server side rendered?

andyg0808: Oh, good point. Done in 3d3ac9e0e126a7b9115eb97d57e7a2bf092c8386

andyg0808:
@ifixit/qae: I'm having trouble with this PR. The tests seem to pass locally, so I'm not sure why they're breaking here. I've tried merging `main`, and it doesn't seem to make them fail:

image

batbattur:
The page the test trying to load is returning 500 on the preview links: https://react-commerce-git-vulcan-render-from-troubleshoo-d7dc3c-ifixit.vercel.app/Vulcan/Dryer_Not_Spinning

It looks like it calls `Troubleshooting` endpoint from the react-commerce app? If it is, we might need to add it in the whitelist, similar to https://github.com/iFixit/ifixit/pull/45212. I think it's passing locally since you're ip in cominor whitelist.

mlahargou:
I wonder if we should allow all `GET` requests to cominor's api? It's a little non-intuitive to have a whitelist with only some of the routes enabled?

andyg0808:
@mlahargou One potentially-interesting option would be to make the API client in `react-commerce` always add some custom header or something, which we could then use to only authorize requests from our code. I think otherwise we're just exposing Cominor to whoever queries it. (I assume that there's not a consistent request IP or something for our GitHub Actions queries? Because, of course, it would be ideal if we could just whitelist only queries from our GitHub Actions code.)

mlahargou:
>  (I assume that there's not a consistent request IP or something for our GitHub Actions queries? Because, of course, it would be ideal if we could just whitelist only queries from our GitHub Actions code.)

Yeah, it kinda sucks. They have a huge list of ips and they aren't guaranteed to stay the same.

> Since there are so many IP address ranges for GitHub-hosted runners, we do not recommend that you use these as allow-lists for your internal resources. - https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#ip-addresses

mlahargou:
> One potentially-interesting option would be to make the API client in react-commerce always add some custom header or something, which we could then use to only authorize requests from our code.

I like this. Another benefit I potentially see if we do this is for Valkyrie testing. Currently, we can't access our valkyrie mysql dump in CI. We could potentially expose this if we allowed access based on a secure header.

andyg0808:
Yeah, the other reason I see that we wouldn't want to whitelist the GitHub Actions IPs is that potentially other users could write code in their GitHub Actions which would then make requests to our Cominor APIs. (Granted, our current solution is even less secure, but that's kinda the point of this discussion.)

andyg0808:
PR to allow requests to `Troubleshooting` API: https://github.com/iFixit/ifixit/pull/46167

andyg0808:
un_deploy_block :+1: 
#1213 has gone out!

jarstelfox (approved):
CR 🌵

Issue #1226, opened 2023-01-06
Upgrade Playwright to 1.29 (or higher?)


mlahargou:
I wanted to use [locator.all()](https://playwright.dev/docs/api/class-locator#locator-all) but we are on an older version. Let's see if we can upgrade to at least `1.29`.

cc @batbattur

PR #1231, opened 2023-01-10
Tests: Update playwright version

batbattur:
## Summary
1. Updated playwright to 1.29.1
2. Used `locator.all()` in `cross_sell` test to verify the feature from the new playwright version works properly. 

## QA notes
Passing CI is enough.

qa_req 0

cc: @mlahargou 
closes: #1226

### DIFF
--- a/frontend/tests/playwright/product/cross_sell.spec.ts
+++ b/frontend/tests/playwright/product/cross_sell.spec.ts
@@ -61,7 +61,6 @@ test.describe('Cross-sell test', () => {
       let expectedTotalPrice = 0;
 
       for (const product of await products.all()) {
-
          allProductTitles.push(
             await product.getByTestId('cross-sell-item-title').textContent()
          );
--- a/frontend/tests/playwright/product/cross_sell.spec.ts
+++ b/frontend/tests/playwright/product/cross_sell.spec.ts
@@ -9,15 +9,12 @@ test.describe('Cross-sell test', () => {
       page,
    }) => {
       const products = await page.getByTestId('cross-sell-item');
-      const count = await products.count();
 
       let currentProductTitle = null;
       let currentProductPrice = null;
       const otherProductTitles = [];
 
-      for (let i = 0; i < count; i++) {
-         const product = products.nth(i);
-
+      for (const product of await products.all()) {
          const isCurrent = await product
             .locator('span', { hasText: 'Current item' })
             .isVisible();
@@ -59,13 +56,11 @@ test.describe('Cross-sell test', () => {
 
    test('Cross-sell products can be added to cart', async ({ page }) => {
       const products = await page.getByTestId('cross-sell-item');
-      const count = await products.count();
 
       const allProductTitles = [];
       let expectedTotalPrice = 0;
 
-      for (let i = 0; i < count; i++) {
-         const product = products.nth(i);
+      for (const product of await products.all()) {
 
          allProductTitles.push(
             await product.getByTestId('cross-sell-item-title').textContent()
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -86,7 +86,7 @@
       "@graphql-codegen/typescript-operations": "2.5.4",
       "@ifixit/tsconfig": "workspace:*",
       "@next/bundle-analyzer": "12.1.6",
-      "@playwright/test": "1.28.1",
+      "@playwright/test": "1.29.1",
       "@svgr/cli": "5.5.0",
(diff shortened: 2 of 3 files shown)


frontend/tests/playwright/product/cross_sell.spec.ts
```diff
@@ -9,15 +9,12 @@ test.describe('Cross-sell test', () => {
...
-      const count = await products.count();
 
       let currentProductTitle = null;
       let currentProductPrice = null;
       const otherProductTitles = [];
 
-      for (let i = 0; i < count; i++) {
-         const product = products.nth(i);
-
+      for (const product of await products.all()) {
```
mlahargou, line 17: Can we use the `forEach` call on `locator.all()`?

```
products.all().forEach()
```

mlahargou (approved):
CR 📱

frontend/tests/playwright/product/cross_sell.spec.ts
batbattur: I think we can use `.forEach()` like the diff below, but now we're getting additional type errors, so we might need to cast them 😓  
Screen Shot 2023-01-09 at 4 45 12 PM

```diff
diff --git a/frontend/tests/playwright/product/cross_sell.spec.ts b/frontend/tests/playwright/product/cross_sell.spec.ts
index c9e52165..534d32d8 100644
--- a/frontend/tests/playwright/product/cross_sell.spec.ts
+++ b/frontend/tests/playwright/product/cross_sell.spec.ts
@@ -12,9 +12,9 @@ test.describe('Cross-sell test', () => {
 
       let currentProductTitle = null;
       let currentProductPrice = null;
-      const otherProductTitles = [];
+      const otherProductTitles: any = [];
 
-      for (const product of await products.all()) {
+      (await products.all()).forEach(async product => {
          const isCurrent = await product
             .locator('span', { hasText: 'Current item' })
             .isVisible();
@@ -26,7 +26,7 @@ test.describe('Cross-sell test', () => {
             currentProductTitle = await product
                .getByTestId('cross-sell-item-title')
                .textContent();
-            continue;
+            return;
          }
 
          // Deselect other cross-sell products
@@ -34,7 +34,7 @@ test.describe('Cross-sell test', () => {
             await product.getByTestId('cross-sell-item-title').textContent()
          );
          await product.getByTestId('cross-sell-item-select').click();
-      }
+      });
 
       // Assert total price matches price of product as it's the only one selected
       await expect(
@@ -48,7 +48,7 @@ test.describe('Cross-sell test', () => {
          .getByTestId('cart-drawer-body')
          .textContent();
 
-      otherProductTitles.forEach((otherProductTitle) => {
+      (otherProductTitles as any).forEach((otherProductTitle: any) => {
          expect(cartDrawerText).not.toContain(otherProductTitle);
       });
       expect(cartDrawerText).toContain(currentProductTitle);
```

ardelato: It is best to not use `forEach` with asynchronous tasks because the `async/await` is not respected.

https://gist.github.com/joeytwiddle/37d2085425c049629b80956d3c618971

frontend/tests/playwright/product/cross_sell.spec.ts
```diff
@@ -59,14 +56,11 @@ test.describe('Cross-sell test', () => {
...
       const products = await page.getByTestId('cross-sell-item');
-      const count = await products.count();
 
       const allProductTitles = [];
       let expectedTotalPrice = 0;
 
-      for (let i = 0; i < count; i++) {
-         const product = products.nth(i);
-
+      for (const product of await products.all()) {
```
ardelato, line 63: Hmm, will `product.all()` get executed once at just the start of the loop or every time we loop? 

I think we should set the result to a constant outside of the loop and then use evaluated constant to iterate on.

batbattur: I think there might be a misunderstanding with the `products` and `product` constants.

The `products` constant is defined on top of the test on line 58: 
 https://github.com/iFixit/react-commerce/blob/65ed8888f57297e94d2d2a8103763a002e043bda/frontend/tests/playwright/product/cross_sell.spec.ts#L58

We call `products.all()` to iterate for single `const product`.

I used the example in https://playwright.dev/docs/api/class-locator#locator-all.

ardelato: TIL!

Issue #1156, opened 2022-12-15
Improve server slow responses


dhmacs:
Some of the queries that we use to generate the pages are slow, so when there's a CDN cache miss the user gets a bad UX

## Things to do

- Increase stale time so we get less CDN cache misses
- Implement server cache so that whenever we deploy (and the CDN cache is nuked) we still serve pages fast
- Add client side queries for data that needs to be fresh (e.g. inventory quantities)

dhmacs:
This has been already partially addressed. Let's close this in favour of more granular issues

PR #1232, opened 2023-01-10
Add server swr cache library

dhmacs:
Related to #1156 

qa_req 0

This PR adds the library to implement the Redis swr cache. 
This PR does not cache any operation, I'll open separate PRs for that.

## Redis server

deploy_block 🛑 

We need a Redis instance for production and maybe one to use for preview. I've been using [Upstash](https://upstash.com/) while implementing this and it's great. Upstash is basically Redis as a Service and offers low latencies. 
I'd like to highlight that wether we choose to use Upstash or self host a Redis instance, the only thing that changes is the Redis connection URL, so even if we go with Upstash it'd be easy to switch to self hosting. Your call.

cc @sterlinghirsh @danielbeardsley


davidrans:
> if we go with Upstash it'd be easy to switch to self hosting

I'm fine with going with Upstash and evaluating pricing in a month. Do you have a ballpark of what you expect it to cost @dhmacs?

Uptime [looks okay](https://status.upstash.com/uptime/x20f3jcqg1vq?page=1) (one 16 minute incident in the last year). What kind of fault-tolerance have we implemented here for if Redis goes down?

frontend/lib/swr-cache/index.ts
```diff
@@ -0,0 +1,145 @@
...
+      fetch(`${APP_ORIGIN}/${endpoint}`, {
+         method: 'POST',
+         body: JSON.stringify(variables),
+      }).catch((error) => {
+         logger.error(
+            `[cache > start background revalidation]: unable to trigger revalidation\n${error}`
+         );
+      });
+      // We add a small delay to ensure the revalidation is triggered
+      await sleep(50);
```
masonmcelvain, line 81: I don't understand why revalidation wouldn't be triggered without this sleep. Could you explain it to me?

masonmcelvain (approved):
CR 🚀 This looks great

dhmacs:
> I'm fine with going with Upstash and evaluating pricing in a month. Do you have a ballpark of what you expect it to cost?

Let's do a quick calculation of the upper bound:

**Assumptions**:

1. We're caching product pages and product list pages
2. We have ~13500 products (according to Algolia)
3. We have ~3000 product list (according to Strapi)
4. At the CDN we consider entries stale after 1 hour
5. We consider Redis server stale every minute (this tells us that every time we'll have a CDN cache miss we'll also have a Redis revalidation request)
6. Every product list and every product is visited at least once every hour
7. Upstash pricing is $0.2 per 100K commands, up to $160 monthly max.

**Estimate**
- For every CDN cache miss we'd consume 3 Redis commands (1 to get the layout data, 1 to get the product / product list data and 1 for background revalidation of product / product list data)
- According to assumption 2) and 3) we'd have 13500 + 3000 = 16500 cache entries
- if each entry takes 3 command that result into 49500 commands/hour or ~36M monthly commands

That would translate into a monthly bill of 72$. It can probably be further improved with an in-memory cache layer.
This is an upper bound estimate though, I think it's unlikely that every page is visited every hour. With analytics data we could get closer to the real estimate.

This is a back of the envelope calculation that I just did, so take it with a grain of salt. I think the real numbers will be much lower, but much depends on actual traffic and on how we change CDN configs.

There are further improvements that we can make to the cache, for example in the future we might revalidate a product only when it changes on Shopify (with Shopify webhooks) so that also could reduce the bill, but this is just speculation.

> Uptime [looks okay](https://status.upstash.com/uptime/x20f3jcqg1vq?page=1) (one 16 minute incident in the last year). What kind of fault-tolerance have we implemented here for if Redis goes down?

Good point @davidrans ! Currently it will fail if Redis is down, but we can easily make it so Redis failures are considered cache misses (I think we can merge the PR since the library is not yet used, and then address this case separately if you agree). Redis outages will also be partially mitigated by the fact that we're also caching at the CDN level.

frontend/lib/swr-cache/index.ts
dhmacs: Because since we're not awaiting the fetch request, the response might be sent (hence the process terminated) before node has even the chance to start the network call. Adding a small delay helps make sure that the network request gets initiated. We can play with the delay a bit, but I fill like it's not necessary to optimize this for the time being since it should be sufficiently fast.

davidrans:
Thanks for doing that analysis!  I think, if the cost is anywhere close to that estimate, it's worth sticking with Upstash vs. maintaining our own instance.

I'm fine with deferring the error handling to a follow-up pull :+1:

frontend/lib/swr-cache/index.ts
masonmcelvain: Makes sense, thanks! I agree that we do not need to optimize this.

PR #1233, opened 2023-01-10
Sentry: Sample client side hydration errors

masonmcelvain:
Samples client side hydration errors.
https://github.com/getsentry/sentry-javascript/blob/master/packages/types/src/options.ts

qa_req 0

---

Slack thread: https://ifixit.slack.com/archives/C01GMQXJC95/p1673368942193239

CC @davidrans


masonmcelvain:
Added the env var to react-commerce-prod
image

davidrans:
> It would be ideal to sample only these hydration errors

A problem with sampling everything equally is that these hydration errors (which we are currently ignoring) are probably going to be the only thing that gets through since they're the majority of our errors at the moment.

masonmcelvain:
> > It would be ideal to sample only these hydration errors
> 
> A problem with sampling everything equally is that these hydration errors (which we are currently ignoring) are probably going to be the only thing that gets through since they're the majority of our errors at the moment.

Good point. I should ask, do events that are ignored via Sentry's UI still count against our quota? I would imagine they are.

I will change the code to only sample the hydration errors. dev_block ✌🏻

masonmcelvain:
un_dev_block ✌🏻 I dropped the now unused env var.

danielbeardsley (approved):
CR :+1:

PR #1237, opened 2023-01-10
Stats Timing: re-add logging

danielbeardsley:
In the pull that added statsd timing, we accidentally removed logging in a refactor that simplified some code in a response to a comment.

This adds both withLogging and withTiming helpers that wrap promise-Returning functions and changes the next logging middleware to use the withTiming helper (which wraps withLogging).

See [comment in #1162](https://github.com/iFixit/react-commerce/pull/1162#discussion_r1061912170)

### DIFF
--- a/frontend/helpers/next-helpers.ts
+++ b/frontend/helpers/next-helpers.ts
@@ -1,7 +1,7 @@
 import { PROD_USER_AGENT } from '@config/constants';
 import { CACHE_DISABLED } from '@config/env';
 import { setSentryPageContext } from '@ifixit/sentry';
-import { timeAsync } from '@ifixit/stats';
+import { withTiming } from '@ifixit/stats';
 import { clearCache } from '@lib/cache';
 import type { GetServerSidePropsMiddleware } from '@lib/next-middleware';
 import * as Sentry from '@sentry/nextjs';
@@ -18,8 +18,9 @@ interface WithLoggingMiddleware {
 export const withLogging: WithLoggingMiddleware =
    ({ pageName }) =>
    (next) => {
-      return async (context) => {
-         return timeAsync(`page.${pageName}.server_side_props`, async () => {
+      return withTiming(
+         `page.${pageName}.server_side_props`,
+         async (context) => {
             console.log('context.resolvedUrl', context.resolvedUrl);
             console.log('context.req.url', context.req.url);
             Sentry.setContext('Extra Info', {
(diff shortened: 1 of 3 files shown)


danielbeardsley:
QA :+1:  Logging is back:
image

frontend/helpers/next-helpers.ts
```diff
@@ -18,8 +18,9 @@ interface WithLoggingMiddleware {
 export const withLogging: WithLoggingMiddleware =
    ({ pageName }) =>
    (next) => {
-      return async (context) => {
-         return timeAsync(`page.${pageName}.server_side_props`, async () => {
+      return withTiming(
```
danielcliu-ifixit, line 21: It looks like `timeAsync` wraps `withTiming`, is there a reason to change the use of `timeAsync`?

danielcliu-ifixit (approved):
CR :v: with small question

frontend/helpers/next-helpers.ts
danielbeardsley: I created `withTiming` to match a more composable pattern, requiring one less closure. `timeAsync()` is still for single parameter-less function invocations if that's more convenient, but here using `withTiming` results in a cleaner callsite. Note that `timeAsync` doesn't *wrap* `withTiming`, it simply calls it an invokes the return value immediately.

dhmacs (commented):
Seems that `withLogging` (the middleware) is supposed to compose sentry logs and and timing logs. Can I suggest to create a `withSentry` and a `withTiming` middleware and then compose them like so:

```ts
import compose from 'lodash/flowRight';

export const withLogging: WithLoggingMiddleware = ({ pageName }) =>
   compose(withTiming({ pageName }), withSentry);
```

In this way it would be easy to add other kind of logging if we need, or remove the parts that we don't need anymore.

For completeness I add the back of the envelope code for `withSentry` and `withTiming`:

```ts
// Sentry logs
const withSentry: GetServerSidePropsMiddleware = (next) => {
   return (context) => {
      // Add current implementation for sentry logs here
   };
};

// Time logs
interface TimingMiddlewareOptions {
   pageName: string;
}

interface WithTimingMiddleware {
   (options: TimingMiddlewareOptions): GetServerSidePropsMiddleware;
}

const withTiming: WithTimingMiddleware =
   ({ pageName }) =>
   (next) => {
      return async (context) => {
         return timeAsync(`page.${pageName}.server_side_props`, async () => {
            return next(context);
         });
      };
   };
```

As a bonus this enables us also to use these middleware separately

dhmacs:
My comment was merely a refactor of the code, so I don't think we need to revert. If @danielbeardsley agrees with the proposal it can be done in another PR

danielbeardsley:
A follow up is fine. I'll add it to my list.

Issue #1224, opened 2023-01-05
Playwright: Fix CORS issue for Safari Testing


ardelato:
## Description

We cannot run Playwright tests using `webkit` (i.e. Safari) because all requests fail due to Safari's default `cross-site` protection being enabled.

## To Do:

- [ ] Get tests to properly run on `webkit` by disabling `cross-site` protection or something else.
- [ ] Add `webkit` to the list of projects Playwright should run 

**Reference**
https://ifixit.slack.com/archives/C01GMQXJC95/p1670959148403209?thread_ts=1670951372.555509&cid=C01GMQXJC95****

ardelato:
It doesn't look like there is a straightforward way to disable `CORS` for `webkit`. As opposed to Chromium there are no launch option argument flags that we can pass to WebKit to disable the security features. In addition, I've tried to intercept all requests and remove the `sec-fetch` headers as a way to fake a no-cores request but that didn't seem to work either.


Instead of investing more time into this problem, we decided to not proceed with running the Playwright tests for Safari. I will create a pull to remove the commented-out section in the playwright config.

PR #1240, opened 2023-01-10
Playwright: Remove mention of Safari project

ardelato:
## Description

It doesn't look like there is a straightforward way to disable CORS for WebKit. As opposed to Chromium, there are no launch option argument flags that we can pass to WebKit to disable the security features. I also tried intercepting all requests and removing the `sec-fetch` headers as a way to fake a no-cors request but that didn't seem to work either.

Therefore, we will not be able to run the tests in WebKit without some convoluted setup to disable CORS. I don't think it's worth the effort to do that. We can revisit this if there is a need to run the tests in WebKit, but for now, let's drop the mention of the Safari project in the config.

## QA Notes

qa_req 0, CI should pass

Closes: #1224

### DIFF
--- a/frontend/playwright.config.ts
+++ b/frontend/playwright.config.ts
@@ -88,14 +88,6 @@ const config: PlaywrightTestConfig = {
          },
       },
 
-      // We need to enable cross-site tracking for Safari to work locally.
-      // {
-      //    name: 'webkit',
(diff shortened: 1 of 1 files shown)


batbattur (approved):
CR 🚛

PR #1241, opened 2023-01-11
Vulcan: Format page

andyg0808:
This now looks decently like the mockup, less the features we're not planning to implement at this stage. I don't think it's particularly pretty, and I'd appreciate design feedback as well as standard code review.

image

## QA Notes
[Figma link](https://www.figma.com/file/jRG4EyQXRDMGfyLqut08KP/iFixit---Working-files?node-id=9318%3A156345&t=brQ6FiZj8ToHphwW-0)

The URLs that this responds to look like `/Vulcan/Samsung_Galaxy_S10_Will_Not_Turn_On`. Any wiki from the `WIKI` namespace can be used as the last component of the URL and will be rendered as these solution cards.

Related epic: https://github.com/iFixit/ifixit/issues/46024

Closes ifixit/ifixit#46092


andyg0808:
deploy_block :+1: on https://github.com/iFixit/react-commerce/pull/1229 going out.

k0rvusk0r4x:
Design suggestions:
- For the list of "Causes," bump the font for the causes (Solution 1, 2, etc.) up, and the word "Causes" up by one more than the list. Right now, at the same font size as the rest of the regular text, it feels lost.
- I'm not super keen on the wording of "Solution x: <cause>." It seems off to say we are talking about a solution, but list the cause instead. Not sure exactly how to change it at this point. Maybe something like "Solution 1: Charge Battery" with the same text? Something to think about as we go.
- I'm curious how it would look if we lined up the text of the Solutions with the list at the top and thus had the numbers off to the left a bit. Not sure it would be better, but might be worth checking if it feels more . . . lively.
- The current margins are pretty wide compared to other pages. Having the text cover so small an area makes the page feel kind of empty.
- Safe to assume we expect images, or at least a top image by the end? (I don't have access to that figma yet.)

I'm going to leave this without block or signoff for now so we can get some more eyes and more ideas.

sctice-ifixit:
Updated the Figma link to [the updated version that Mattia just added to Working Files](https://www.figma.com/file/jRG4EyQXRDMGfyLqut08KP/iFixit---Working-files?node-id=9318%3A156345&t=brQ6FiZj8ToHphwW-0). Some things changed out from under @andyg0808.

dev_block :cactus: on taking another pass to adjust the pull to the updated mock-up (sorry :disappointed:).

Once we're unblocked, maybe @andyg0808 can list [the features being targeted in this pull](https://docs.google.com/spreadsheets/d/1BjjKhp4GmGWT4pgaXVOHm77AjPY6gb-iOf-f03ENsOo/edit#gid=0), and we can keep feedback on the pull to places where those features diverge from the mock-up?

@k0rvusk0r4x Can you please move your suggestions to comments on [the new QA discussion](https://github.com/iFixit/ifixit/discussions/46185)? Maybe after you've had a chance to compare the updated implementation to the updated mock-up. I think a few of the things you suggested are already addressed by the mock-up.

andyg0808:
I'm gonna close this and create a new one, since the mockup is so different and now the API has changed as well. There just doesn't seem to be that much here worth grabbing.

PR #1245, opened 2023-01-11
Sentry: sample hydration errors with regex

masonmcelvain:
Fixes a bug where some hydration errors weren't matched because they were suffixed with 'See more info here: https://nextjs.org/docs/messages/react-hydration-error'

Followup to #1233

### DIFF
--- a/frontend/sentry.client.config.ts
+++ b/frontend/sentry.client.config.ts
@@ -43,7 +43,7 @@ Sentry.init({
       const ex = hint.originalException;
       if (ex && typeof ex == 'object' && ex.message) {
          // Sample hydration errors.
-         if (hydrationErrors.includes(ex.message)) {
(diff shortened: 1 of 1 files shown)


masonmcelvain:
QA ✌🏻 

I was able to reproduce all the hydration errors by adding this to `ProductListBreadcrumbs`
```tsx
{typeof window !== 'undefined' && (<BreadcrumbItem><Text>Client only</Text></BreadcrumbItem>)}
```

And verify that they are all sampled. Before, only "There was an error while hydrating. Because the error happened outside of a Suspense boundary, the entire root will switch to client rendering." was sampled.

danielcliu-ifixit:
CR :v:

davidrans (approved):
CR :coffee:

PR #1247, opened 2023-01-12
Handle Redis errors as cache misses in SWR cache library

dhmacs:
Related to https://github.com/iFixit/react-commerce/issues/1156

qa_req 0

This address concern expressed here https://github.com/iFixit/react-commerce/pull/1232#issuecomment-1377527739
> Uptime [looks okay](https://status.upstash.com/uptime/x20f3jcqg1vq?page=1) (one 16 minute incident in the last year). What kind of fault-tolerance have we implemented here for if Redis goes down?


masonmcelvain (approved):
CR ✌🏻

Issue #1077, opened 2022-11-29
Sentry frontend web vitals recording: We're missing most of the data


danielbeardsley:
Not 100% sure why, but likely this issue is describing the same problem:

https://github.com/getsentry/sentry-javascript/issues/3803 and
https://github.com/getsentry/sentry-javascript/issues/3833

Specifically, only ~2.5% of our [performance samples](https://sentry.io/organizations/ifixit/discover/results/?display=default&field=count%28%29&field=count_if%28measurements.lcp%2Cgreater%2C0%29&field=equation%7Ccount_if%28measurements.lcp%2Cgreater%2C0%29+%2Fcount%28%29&field=avg%28measurements.lcp%29&interval=1h&name=All+Events&project=6469069&query=event.type%3Atransaction+transaction.op%3Apageload++browser.name%3AChrome&sort=-count%28%29&statsPeriod=14d&user=669438&yAxis=equation%7Ccount_if%28measurements.lcp%2Cgreater%2C0%29+%2Fcount%28%29&yAxis=count%28%29) actually have FCP / LCP values: (Note, this is already limited to Chrome browsers)

![image](https://user-images.githubusercontent.com/26855/204607607-507a3fdc-229c-4c34-a3bf-9b369c3f8325.png)

### Do This
Possibly this can be fixed with this option: 
https://github.com/getsentry/sentry-javascript/issues/3833#issuecomment-909652059

CC @jarstelfox

danielbeardsley:
I've experimented (and failed) and read all the issues about LCP on sentry's github. It didn't seem like my concern (LCP is not being captured) wasn't represented so I made an issue:
https://github.com/getsentry/sentry-javascript/issues/6742

Note: they've deprecated and removed the option referenced in the description. This isn't a problem with next.js, nor this app, [the same thing is present in our php app too](https://sentry.io/organizations/ifixit/discover/results/?display=default&field=count%28%29&field=count_if%28measurements.lcp%2Cgreater%2C0%29&field=equation%7Ccount_if%28measurements.lcp%2Cgreater%2C0%29+%2Fcount%28%29&field=avg%28measurements.lcp%29&interval=1h&name=All+Events&project=1281569&query=event.type%3Atransaction+transaction.op%3Apageload&sort=-count%28%29&statsPeriod=14d&user=669438&yAxis=equation%7Ccount_if%28measurements.lcp%2Cgreater%2C0%29+%2Fcount%28%29&yAxis=count%28%29)

[...]

PR #1250, opened 2023-01-12
Sentry Tracing: increase sample rate significantly

danielbeardsley:
We're already paid for this much data collection in sentry, so let's use it.

Connect #1077


jarstelfox (approved):
CR 🌵

jarstelfox (approved):
CR 🌵

Issue #1239, opened 2023-01-10
Deprecate our use of Cypress


ardelato:
## Description

We have been running `Playwright` in parallel with `Cypress` now for a bit and have no issues. In addition, @iFixit/qae has been finding it more enjoyable to work with `Playwright`'s syntax then `Cypress`'. Therefore, let's deprecated our usage of `Cypress` and fully transition to using just `Playwright` for our `E2E` tests.

PR #1251, opened 2023-01-12
Tests: Remove Cypress

batbattur:
## Summary
Removed all of the Cypress-related files including:
- test-helpers
- cypress and cypress-related packages/libraries. Used `pnpm remove -D <package-name>`.
- all cypress tests and configs
- cypress workflow

## QA notes
1. Go to the collections page (either `/Parts` or `/Tools`) and click the pagination button for the next search result. 
Then make sure it scrolls smoothly (same as on master).
2. Run `git grep -i cypress` and make sure no actual code shows up in the result. 
Note: there should be only one cypress comment in `playwright/utils.ts`. 

closes: #1239 #1111


deltuh-vee:
QA 🎬 
* Autoscrolling is still smooth on collection pages
* No results appear from `git grep -i cypress` except for the one in `playwright/utils.ts`

package.json
```diff
@@ -21,8 +21,6 @@
       "strapi": "cd backend && pnpm run strapi",
       "test": "pnpm run -r test",
       "transform-svg": "cd frontend && pnpm run transform-svg",
-      "cypress:open": "cd frontend && pnpm run cypress:open",
```
ardelato, line 24: Oooh shoot I didn't know we had these scripts in the root `package.json`. We should probably add these scripts for `playwright`

ardelato (approved):
CR ⚡ There were a lot more references to `Cypress` then I thought. Nice job covering all bases @batbattur 🙌

package.json
batbattur: Yes! I think we should add them back and also update the `Test` section in read me.
Let's do it in a follow-up issue 👍

batbattur: Created a follow-up issue: https://github.com/iFixit/react-commerce/issues/1253.

PR #1252, opened 2023-01-12
Hide empty product list children (server-side version)

### LINKED ISSUE #822: ProductList: Do not include "empty" ProductLists in gallery of Children

masonmcelvain:
Hides empty product list children on parts product list pages.

Makes an extra request to Algolia on the server only to identify all child devices with products. This is used to filter the child devices collected from Strapi.

## QA

Examples of collections with no products:

https://www.ifixit.com/Parts/Computer_Hardware
https://www.ifixit.com/Parts/Samsung_Galaxy_S22

Also make sure that no collections that are non-empty end up being hidden by this change.

## References

https://www.algolia.com/doc/api-reference/api-methods/search/
https://www.algolia.com/doc/api-reference/api-parameters/filters/
https://support.algolia.com/hc/en-us/articles/5588550618897-Clarifying-confusion-with-facets-facet-values-maxFacetHits-and-maxValuesPerFacet

CC @dhmacs @sterlinghirsh 

Closes #822


jordycosta:
QA 🟢 

Confirmed that:

Non-empty collections are still being shown as before ✔️ 
Empty collections are now being hidden ✔️ 

---

### Testing Procedure:

1. Navigate to the Vercel [preview](https://react-commerce-git-hide-empty-child-collections-2-ifixit.vercel.app/Parts) _(non-prod)_
2. Open up https://www.cominor.com/Parts
3. Compare collection(s) pages
4. Ensure that empty collections are hidden

> also, confirm that no collections are being hidden that are non-empty

---

### Screen Capture:

Screenshot 2023-01-12 141409

[Empty collection example](https://www.cominor.com/Parts/iMac_M_Series)

frontend/lib/algolia/server/findDescendantDevicesWithProducts.ts
```diff
@@ -0,0 +1,39 @@
...
+
+async function fetchDevices(device: string) {
+   const client = algoliasearch(
+      ALGOLIA_APP_ID,
+      ALGOLIA_API_KEY,
+      CLIENT_OPTIONS
+   );
+   const index = client.initIndex(ALGOLIA_PRODUCT_INDEX_NAME);
+   const deviceSuffix = device
+      ? ` AND device:"${escapeFilterValue(device)}"`
```
danielbeardsley, line 30: It seems like these two things aren't necessarily exhaustively overlapping:

* list of children devices of X
* list of devices from all products that are compatible with a descendant of X

It's possible that the second definitely contains all of the first, but I'm a little leery of that being 100% true and it would suck if some were left out.

Here's the kind of things that this query returns for `iphone`:
image

I'm tempted to say we should do something more exact that involves our API but strapi product lists aren't tied directly to our device hierarchy...  so I'm not sure what to suggest. I'm nervous about depending on the list of facets being returned being completely exhaustive, especially when it's long. I feel like we should thoroughly review the differences here. Maybe just rendering them differently for now, or logging them or something.

Thoughts?

danielbeardsley:
deploy_block :+1:  on a conversation, but it looks like this is right.

frontend/lib/algolia/server/findDescendantDevicesWithProducts.ts
```diff
@@ -0,0 +1,39 @@
+import {
```
dhmacs, line 1: I suggest keeping this file within the product list model folder, since this is product list related business logic

frontend/lib/algolia/server/findDescendantDevicesWithProducts.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export async function findDescendantDevicesWithProducts(device: string) {
+   return cache(
+      `algolia.findDescendantDevicesWithProducts.${device}`,
+      () =>
+         timeAsync('algolia.findDescendantDevicesWithProducts', () =>
+            fetchDevices(device)
+         ),
+      60 * 10
+   );
```
dhmacs, line 19: It's probably not worth it to apply in-memory cache, since we're caching at most 100 items (across the board) so likely it wouldn't impact much. Beside, once I add the Redis cache this will be already covered by that

frontend/models/product-list/server.ts
```diff
@@ -42,15 +43,19 @@ export async function findProductList(
    deviceItemType: string | null = null
 ): Promise<ProductList | null> {
    const filterDeviceTitle = filters.deviceTitle?.eqi ?? '';
+   const isPartsList = filterDeviceTitle || filters.handle?.eq === 'Parts';
```
dhmacs, line 46: I suggest to use `productListType` enum to assert  this, instead of hard-coding the url

dhmacs (changes requested):
This is looking good 👍 

dev_block 🔢  on some suggestions

frontend/lib/algolia/server/findDescendantDevicesWithProducts.ts
masonmcelvain: Good call 👍🏻 https://github.com/iFixit/react-commerce/pull/1252/commits/2b57a1d0859dac53801995c7fe0c4509387ee93f

frontend/lib/algolia/server/findDescendantDevicesWithProducts.ts
masonmcelvain: https://github.com/iFixit/react-commerce/pull/1252/commits/bdcfbba626eb7f4a0840febdd71c0f0ad3131539

frontend/models/product-list/server.ts
masonmcelvain: Good idea https://github.com/iFixit/react-commerce/pull/1252/commits/2d8be9df791fd5789d6ff4b43ce63786858755b5

In order to make the requests in parallel, I chose to pass the product list type into this function instead of waiting for the product list to resolve to get the type.

This commit uses the product list type from the caller in the product list object, whereas before we used the product list type from Strapi. If you think this will cause problems, let me know, but it seemed unlikely that the two would be different unless the Strapi data was invalid.

danielbeardsley:
Sounds like this hides children with products in some cases, so dev_block :+1:  on that.

frontend/lib/algolia/server/findDescendantDevicesWithProducts.ts
masonmcelvain: One possible solution:

In Strapi the `deviceTitle` field only accepts one value.
image
If this field (or another field like `relatedDeviceTitles`) accepted a list of values (for example, `[Appliance, Vehicle, etc]`), then we could better match on devices in algolia that don't have a product list entry in Strapi, but exist in the iFixit device hierarchy.

image
image

danielbeardsley: I'm not sure I understand your suggestion. Strapi product lists currently represent one since device. It seems like you're suggesting having them represent more than one? It seems like that would create far more problems than it would solve (multiple sets of ancestors, multiple sets of children, ...). Maybe I'm misunderstanding your suggestion.

frontend/models/product-list/server.ts
```diff
@@ -139,6 +148,21 @@ export function getProductListType(
...
+   devicesWithProducts: string[] | null
+) {
+   if (devicesWithProducts == null) {
+      return productListChildren;
+   }
+   if (devicesWithProducts.length === 0) {
+      return [];
+   }
+   return productListChildren.filter((child) =>
+      devicesWithProducts.some((device) => device === child.deviceTitle)
```
danielbeardsley, line 162: Let's use a `Map` so we can do a faster check than a linear array search.

masonmcelvain: https://github.com/iFixit/react-commerce/pull/1252/commits/93f54047b62476cd6da28dedef209128678710bd

frontend/lib/algolia/server/findDescendantDevicesWithProducts.ts
masonmcelvain: Danny and I talked, and came up with the solution implemented in https://github.com/iFixit/react-commerce/pull/1252/commits/f8b30b6e93545bbdf0768a8f3ec10fdb0d461685

masonmcelvain:
un_dev_block ✌🏻 The implementation no longer depends on the iFixit device hierarchy matching that of Strapi. However, this came at the cost of running the algolia query in parallel with the Strapi query.

We could optimize this by running the algolia query in parallel with iFixit image query, but I'm going to hold off on that until we decide we like this solution.

frontend/helpers/algolia-helpers.ts
```diff
@@ -23,3 +23,10 @@ export function formatFacetName(algoliaName: string): string {
    }
    return FACETS_NAME_OVERRIDES[algoliaName];
 }
+
+/**
+ * https://www.algolia.com/doc/api-reference/api-parameters/filters/#usage-notes
+ */
+export function escapeFilterValue(value: string) {
+   return value.replaceAll("'", "\\'").replaceAll('"', '\\"');
```
danielbeardsley, line 31: Even though their docs are interactive and provide code examples, it's still unclear what the *spec* is. After interpreting their code examples and interacting with the validator I read the spec as follows.

> All single quotes and double quotes should be preceded with a backslash: `\`

In most languages, this requires two backslashes when used in a string literal which is the case here. I wrote this mostly to make sure my understanding is correct.

frontend/models/product-list/server.ts
```diff
@@ -139,6 +145,23 @@ export function getProductListType(
...
+   if (!isPartsList) {
+      return productListChildren;
+   }
+   const childDevices = productListChildren.reduce(
+      (acc, child) => (child.deviceTitle ? [...acc, child.deviceTitle] : acc),
+      [] as string[]
+   );
+   const devicesWithProducts = await findDevicesWithProducts(childDevices);
+   return productListChildren.filter(
+      (child) => child.deviceTitle && devicesWithProducts[child.deviceTitle] > 0
```
danielbeardsley, line 161: Are there some strapi product lists that have no `deviceTitle`?

frontend/models/product-list/server.ts
```diff
@@ -401,3 +424,27 @@ function createPublicAlgoliaKey(appId: string, apiKey: string): string {
...
+         ALGOLIA_APP_ID,
+         ALGOLIA_API_KEY,
+         CLIENT_OPTIONS
+      );
+      const index = client.initIndex(ALGOLIA_PRODUCT_INDEX_NAME);
+      const deviceDisjunctiveFilters = devices
+         .map((device) => `device:"${escapeFilterValue(device)}"`)
+         .join(' OR ');
+      const deviceFilterSuffix = deviceDisjunctiveFilters
+         ? ` AND ${deviceDisjunctiveFilters}`
```
danielbeardsley, line 440: I would expect this would need to be wrapped in parenthesis. Otherwise will end up with `public=1 AND device:X OR device:Y OR device:Z`

frontend/helpers/algolia-helpers.ts
masonmcelvain: I understand it the same way you do. I think Algolia's client wants to receive a string with with every `'` and `"` preceded by `\`.

frontend/models/product-list/server.ts
```diff
@@ -401,3 +424,27 @@ function createPublicAlgoliaKey(appId: string, apiKey: string): string {
...
+         : '';
+      const { facets } = await index.search('', {
+         facets: ['device'],
+         filters: `public = 1${deviceFilterSuffix}`,
+         maxValuesPerFacet: 1000,
+         hitsPerPage: 0,
```
danielbeardsley, line 446: Ask for 0 hits, sneaky!

frontend/models/product-list/server.ts
masonmcelvain: Good catch

frontend/models/product-list/server.ts
masonmcelvain: TS thinks it's nullable. Perhaps we can correct our type if it's never null in practice

frontend/models/product-list/server.ts
masonmcelvain: Yep! That way the only data we get back is the list of devices, instead of all the json included in the product hits

frontend/models/product-list/server.ts
```diff
@@ -139,6 +145,23 @@ export function getProductListType(
...
 
+async function filterDevicesWithNoProducts(
+   productListChildren: ProductListChild[],
+   isPartsList: boolean
+) {
+   if (!isPartsList) {
+      return productListChildren;
+   }
+   const childDevices = productListChildren.reduce(
+      (acc, child) => (child.deviceTitle ? [...acc, child.deviceTitle] : acc),
```
danielbeardsley, line 156: This is effectively an `O(n^2)` implementation.

How about `children.map(child => child.deviceTitle).filter(child => child.deviceTitle)`

masonmcelvain: Good eye! This motivated me to do some learning too: https://prateeksurana.me/blog/why-using-object-spread-with-reduce-bad-idea/

frontend/models/product-list/server.ts
```diff
@@ -95,12 +98,15 @@ export async function findProductList(
       // Strapi sort order is case sensitive, so we need to improve on it in memory
       children: await fillMissingImagesFromApi(
          sortProductListChildren(
-            filterNullableItems(
-               productList?.children?.data.map(
-                  createProductListChild({
-                     deviceWiki,
-                  })
-               )
+            await filterDevicesWithNoProducts(
```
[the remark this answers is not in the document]
masonmcelvain: I'm interested in making these requests in parallel if we like this solution and it works, so I'll play with it a little.

frontend/models/product-list/server.ts
masonmcelvain: https://github.com/iFixit/react-commerce/pull/1252/commits/7df5256696bc5596c6bcdc0a2abc1408854ad754

frontend/models/product-list/server.ts
masonmcelvain: https://github.com/iFixit/react-commerce/pull/1252/commits/b9e1765f1ad39c7e3a53eeedb8e3dde1163a11b9

frontend/models/product-list/server.ts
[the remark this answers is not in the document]
masonmcelvain: This is what I came up with, if we feel it makes the code to complex we can revert it. https://github.com/iFixit/react-commerce/pull/1252/commits/0894c1d5e39a94f3556cf3bbb8aa0b1c220558dd

masonmcelvain:
un_dev_block ✌🏻 Thanks @danielbeardsley!

frontend/models/product-list/server.ts
```diff
@@ -75,15 +76,17 @@ export async function findProductList(
    const algoliaApiKey = logSync('algolia:create key', () =>
       createPublicAlgoliaKey(ALGOLIA_APP_ID, ALGOLIA_API_KEY)
    );
-   const productListType = getProductListType(productList?.type);
 
    const ancestors = createProductListAncestors(parents);
+   const isPartsList =
+      productListType === ProductListType.AllParts ||
+      productListType === ProductListType.DeviceParts;
```
masonmcelvain, line 83: Now that we're making the Algolia request after we fetch the product list, we don't need to pass `productListType` into `findProductList`, we can just pull it off the productList directly.

masonmcelvain: https://github.com/iFixit/react-commerce/pull/1252/commits/f969a0e12764617e206205c42995d4d1156a8ae2

masonmcelvain:
un_deploy_block ✌🏻

danielbeardsley (approved):
CR :+1:   Well done.

k0rvusk0r4x:
- Empty lists are hidden.
- Non-empty lists are still visible and functional.
- Going directly to the page vial url (as a user might if it's bookmarked) does not break anything.

QA 😺

Issue #1249, opened 2023-01-12
Make upsell product title/image a link


sterlinghirsh:
It looks like a link. Make it act like a link to the product.

PR #1254, opened 2023-01-13
Add Links to Product to Upsell Section

danielcliu-ifixit:
## Overview

Lets have links to the product being upsold in the upsell section so that users can get more information about the product if they are interested.

## QA

Make sure that when you have something in your cart, the upsell section has links on the product name and on the product image. 

Now the entire upsell section is a link except the add to cart button basically.
~~Note: this links will be urls to the same domain that is serving the api responses, so if you are on localhost it will likely link you to some cominor domain.~~ 

Closes: #1249


deltuh-vee:
QA 🎬 
Product Upsells now have links to their Product Pages on the name and image.

packages/ui/cart/drawer/Upsell.tsx
```diff
@@ -42,12 +42,14 @@ export function Upsell({ item }: UpsellProps) {
...
-            </Text>
+            </Link>
             <Text color="brand.500">Add the product below to your order</Text>
             <HStack mt="3" align="center" spacing="3">
-               <CartLineItemImage src={item.imageSrc} alt={item.name} />
+               <a href={item.url}>
+                  <CartLineItemImage src={item.imageSrc} alt={item.name} />
+               </a>
                <Text fontSize="sm">{item.marketingBlurb}</Text>
             </HStack>
```
masonmcelvain, line 54: Opinion: I know the spec only requests wrapping the image and the title, but I personally would expect the entire upsell card to be a link, since this is what we do on product list pages, featured product section, and the cross sell section.

masonmcelvain (approved):
This looks right, but dev_block ✌🏻 on the NextLink change

packages/ui/cart/drawer/Upsell.tsx
```diff
@@ -42,12 +42,14 @@ export function Upsell({ item }: UpsellProps) {
...
          >
-            <Text color="brand.500" fontWeight="semibold">
+            <Link href={item.url} color="brand.500" fontWeight="semibold">
                {item.marketingTitle ?? item.name}
-            </Text>
+            </Link>
             <Text color="brand.500">Add the product below to your order</Text>
             <HStack mt="3" align="center" spacing="3">
-               <CartLineItemImage src={item.imageSrc} alt={item.name} />
+               <a href={item.url}>
```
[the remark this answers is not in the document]
masonmcelvain: I guess the manta and PTT links will be external (non-Next) links. If that causes problems with `<NextLink>`, maybe we could use `/products/${handle}` as the url instead.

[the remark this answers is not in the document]
danielcliu-ifixit: yeah rn our product urls are all absolute, is there a distinct advantage to using the `<NextLink>`?

packages/ui/cart/drawer/Upsell.tsx
danielcliu-ifixit: @sterlinghirsh what do you think of this design idea?

packages/ui/cart/drawer/Upsell.tsx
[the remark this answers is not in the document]
masonmcelvain: Anchor tags will cause a full page reload, while NextLink won't. NextLink's also prefetch the url so the transition is faster when the user clicks on the link.

deltuh-vee:
dev_block ✊
The image is no longer a product link

danielcliu-ifixit:
un_dev_block :v:

packages/ui/cart/drawer/Upsell.tsx
```diff
@@ -42,14 +44,19 @@ export function Upsell({ item }: UpsellProps) {
             borderColor="brand.200"
             borderStyle="solid"
          >
-            <Link href={item.url} color="brand.500" fontWeight="semibold">
-               {item.marketingTitle ?? item.name}
-            </Link>
-            <Text color="brand.500">Add the product below to your order</Text>
+            <NextLink href={`/products/${item.handle}`} passHref>
+               <LinkOverlay
+                  id={item.handle}
```
masonmcelvain, line 60: I'm curious why we're including an id here?

packages/ui/cart/drawer/Upsell.tsx
```diff
@@ -42,14 +44,19 @@ export function Upsell({ item }: UpsellProps) {
...
+                  fontWeight="semibold"
+                  color="brand.500"
+                  _groupHover={{ color: 'black' }}
```
masonmcelvain, line 63: What does this _groupHover achieve? I'm not that familiar with it.

If this is to style the underline on hover, I'd say we should just set `textDecoration="none"` because I don't think we show underline on hover in the rest of our product cards.

packages/ui/cart/drawer/Upsell.tsx
```diff
@@ -51,6 +53,7 @@ export function Upsell({ item }: UpsellProps) {
             borderWidth="1px"
             borderColor="brand.200"
             borderStyle="solid"
+            onClick={onClose}
```
masonmcelvain, line 56: I'm curious why this is necessary. Does the cart drawer remain open after navigation?

masonmcelvain (approved):
CR ✌🏻 Nice use of `LinkOverlay`! deploy_block ✌🏻 on a couple questions

deltuh-vee:
dev_block ✊ 
I don't think we want the cart drawer to close after clicking Add to Cart. Makes it harder to finish checking out.
https://user-images.githubusercontent.com/5289715/212774125-dad81882-898c-4d78-861c-ca49477af2cb.mp4

packages/ui/cart/drawer/Upsell.tsx
danielcliu-ifixit: Yes, you could click the link and the cart would stay open. This closes it.

packages/ui/cart/drawer/Upsell.tsx
danielcliu-ifixit: it changes the color of the text, which allows for a visual indication that there is a link to click basically.

packages/ui/cart/drawer/Upsell.tsx
danielcliu-ifixit: i copied pasted this and i just assumed that the id was a good thing to use.

danielcliu-ifixit:
un_dev_block :v:

packages/ui/cart/drawer/Upsell.tsx
masonmcelvain: Gotcha. I don't think it's wrong, but I also don't think it does anything here and we don't include id's on most components.

masonmcelvain (approved):
CR ✌🏻 Though now it seems like the drawer will only close if the image or the text is clicked. In other words, there are ways to click the link that will keep the cart drawer open.

danielcliu-ifixit:
> CR ✌🏻 Though now it seems like the drawer will only close if the image or the text is clicked. In other words, there are ways to click the link that will keep the cart drawer open.

I'm not sure im getting the same behavior. It seems to work fine for me.

masonmcelvain:
> > CR ✌🏻 Though now it seems like the drawer will only close if the image or the text is clicked. In other words, there are ways to click the link that will keep the cart drawer open.
> 
> 
> 
> I'm not sure im getting the same behavior. It seems to work fine for me. 

Great! I hadn't tested it, only read the code.

k0rvusk0r4x:
- Empty cart continues to look and behave the same.
- A cart with something in it has the upsell component.
- The image, title, and description of the upsell item is visible.
- The add to cart button is visible.
- The entire section with the item is a link. It works properly.
- The add to cart button is not a link to the item.

QA 😺

PR #1255, opened 2023-01-13
Chakra: update to latest version (fix for Safari 13)

danielbeardsley:
Previous version of chakra caused a syntax error in Safar v13. This is a browser that came out <4 years ago and should be still supported.

This is the exact issue:
https://github.com/chakra-ui/chakra-ui/issues/6828

It looks like the commit that fixes the issue is present in
> 2.4.5, but 2.4.8 just came out, so let's bump to that.

Closes Fixit/ifixit#45976

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -24,7 +24,7 @@
       "transform-svg": "npx @svgr/cli --typescript -d assets/svg assets/svg/files && prettier --write assets/svg"
    },
    "dependencies": {
-      "@chakra-ui/react": "2.3.6",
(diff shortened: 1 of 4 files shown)


danielbeardsley:
QA :+1:  -- Fixes the problem with Safari 13 (which works correctly now!)

I clicked around in chrome and everything I tried continued to work (add to cart, cart drawer, search, facets, ...)

masonmcelvain (approved):
CR ✌🏻

masonmcelvain:
QA ✌🏻 carry forward, only the tests were updated

PR #1256, opened 2023-01-13
Unused deps: remove them

danielbeardsley:
I've noticed some unmet peer dependencies and some unused dependencies. This change drops some that seem like they aren't needed.


frontend/package.json
```diff
@@ -110,7 +110,6 @@
       "eslint-plugin-react": "7.23.0",
       "eslint-plugin-react-hooks": "4.2.0",
       "graphql-tag": "2.12.6",
-      "graphqurl": "1.0.1",
```
batbattur, line 115: It looks like we use the cli `gq` version here:
https://github.com/iFixit/react-commerce/blob/d528e7820d58d4f6f1ea3fd6e1f14e40ebf42f2b/frontend/lib/shopify-storefront-sdk/generate-schema.js#L20

Found it by running `git grep "gq "`

batbattur:
dev_block 🚛 on double checking `graphqurl`, ref https://github.com/iFixit/react-commerce/pull/1256#discussion_r1070150062.

frontend/package.json
danielbeardsley: OMG, thank you... this is the the second time I've done this exact thing. Derp. Will fix.

danielbeardsley:
un_dev_block :+1: 

Thanks @batbattur !

frontend/package.json
```diff
@@ -129,5 +130,12 @@
    },
    "msw": {
       "workerDirectory": "public"
+   },
+   "pnpm": {
+      "peerDependencyRules": {
+         "allowedVersions": {
+            "graphql": "15"
+         }
+      }
```
masonmcelvain, line 139: I remember rush complains about unmet peer deps, as part of QA we should check that this doesn't break updating the submodule in our php app

masonmcelvain (approved):
CR ✌🏻 Nice!

frontend/package.json
danielbeardsley: Fair... but this pull *reduces* the number of unmet peer deps, not increases.

masonmcelvain: Oops I didn't read the docs, I thought this config was doing something different. Seems like there's no concern here.

> Unmet peer dependency warnings will not be printed for peer dependencies of the specified range.

https://pnpm.io/package_json#pnpmpeerdependencyrulesallowedversions

k0rvusk0r4x:
Smoke test shows everything continues to work the same as master.

QA 😺

PR #1257, opened 2023-01-13
Fontawesome: use deep imports for duo-tone icons

danielbeardsley:
This is one of the recommended ways to reduce bundle size in case tree-shaking isn't happening like it should:

https://fontawesome.com/v5/docs/apis/javascript/tree-shaking

Hard to know until bundle-analysis tells us.

Let's try this on the duo-tone imports cause it's way fewer changes.


danielbeardsley:
Nevermind. Looks like tree-shaking was already happening and this isn't needed.

danielbeardsley:
image

PR #1258, opened 2023-01-13
Playwright: Create custom fixtures to help interact with elements

ardelato:
## Description

By using Playwright custom fixtures[^1] we will be creating `iFixit/ifixit`'s extension-like classes to help with interacting with elements and making assertions.

## CR Notes

We will be adding more methods and fixtures in the future to fully address #1217, but for now this is a good start.

## QA

- [ ] CI Passes

Connects: https://github.com/ifixit/react-commerce/issues/1217

[^1]: https://playwright.dev/docs/test-fixtures#creating-a-fixture

qa_req 0

Issue #1217, opened 2023-01-04
Playwright: Add Utility Functions


ardelato:
## Description

As we began to add more `Playwright` tests for #452, we began to notice reoccurring actions throughout different test files. Now that we have a good number of tests, let's extract some of this duplicate code into utility functions similar to the `Extension` classes we have in the `iFixit/ifixit` repo. 

We can possibly extend `Playwright`'s built-in fixtures to include additional assertions or targeted selectors (https://playwright.dev/docs/test-fixtures).

## To Do:

- [x] Make it easier to grab the `SKU` off a product page
- [x] Make it easier to grab a product price as a float
- [x] Make it easier to select a variant option (_image based, not the dropdown selector_)
	- We might need to update the Components to include a `selected` or `active` tag to make it easier to figure out which variant option we currently have selected 
- [x] Add some utility functions around the mini cart
	- [x] Assert totals
	- [x] Asserting a Product is in the drawer
	- [x] Assert quantities in the drawer
	- [x] Decrease and increase quantities for a product
	- [x] Deleting a product from the cart

**Ref:**
https://ifixit.slack.com/archives/C02SMF0AZPF/p1672873076552309

**cc:** @mlahargou

PR #1235, opened 2023-01-10
CI: Add datadog workflow to send job metrics

batbattur:
## Summary
1. This is a copy of `datadog.yml` workflow from `iFixit/ifixit`. 
The only difference is that this runs on github hosted ubuntu-latest machine (free).

2. Deflakes a test that's flaky in https://github.com/iFixit/react-commerce/pull/1235/commits/74674f6e889da10c76e470d37d57cb793c45a45d

Note: I created a new API key on Datadog and added it to this repo's secrets

## QA notes
We can test this after merging it, same as in https://github.com/iFixit/ifixit/pull/44282.

After merging it, the data should be visible on https://app.datadoghq.com/dashboard/3xd-yax-e2h/github-actions. 
Cloned the existing graphs and added filtering for each repo. 

qa_req 0

connects #1234

PR #1261, opened 2023-01-16
Playwright: Add more utility functions for the Product Page

ardelato:
## Description

This is a continuation of https://github.com/iFixit/react-commerce/pull/1258

### Changes:

- Adds utility functions to target the prices of the product page we are on
- Adds utility functions to help select the different variants
- Adds additional attributes to help locate and select price elements and the variant options
- Integrates the utility functions into our current tests

### CR:

I did not extend the utility functions to interact with all types of variant selectors (i.e. the dropdown menu) as we primarily interact with the image type selector in our tests. If the need arises, we can extend the utility functions in the future.

### QA:

- [ ] CI should pass

Connects: #1217 
qa_req 0

### DIFF
--- a/frontend/tests/playwright/fixtures/product-page.ts
+++ b/frontend/tests/playwright/fixtures/product-page.ts
@@ -39,27 +39,27 @@ export class ProductPage {
    /**
     * @description Locates and returns the current product price as a string with the format $###.## or $###
     */
-   async getCurrentPrice(): Promise<string> {
+   async getCurrentPrice(): Promise<number> {
       const price = await this.page
          .getByTestId('product-price-section')
          .getByTestId('current-price')
          .textContent();
       expect(price).not.toEqual('');
       expect(price).toMatch(/\$[0-9]+(\.[0-9]{1,2})?$/);
-      return parseFloat(price!.slice(1)).toFixed(2);
+      return parseFloat(price!.slice(1));
    }
 
    /**
     * @description Locates and returns the original product price as a string with the format $###.## or $###
     */
-   async getDiscountedPrice(): Promise<string> {
+   async getDiscountedPrice(): Promise<number> {
       const price = await this.page
          .getByTestId('product-price-section')
          .getByTestId('compare-at-price')
          .textContent();
       expect(price).not.toEqual('');
       expect(price).toMatch(/\$[0-9]+(\.[0-9]{1,2})?$/);
-      return parseFloat(price!.slice(1)).toFixed(2);
+      return parseFloat(price!.slice(1));
    }
 
    /**
--- a/frontend/tests/playwright/fixtures/product-page.ts
+++ b/frontend/tests/playwright/fixtures/product-page.ts
@@ -35,4 +35,30 @@ export class ProductPage {
       expect(sku).toMatch(/IF\d{3}-\d{3}(-\d{1,2})?$/);
       return sku;
    }
+
+   /**
+    * @description Locates and returns the current product price as a string with the format $###.## or $###
+    */
+   async getCurrentPrice(): Promise<string> {
+      const price = await this.page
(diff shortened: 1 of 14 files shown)


frontend/tests/playwright/product/pro_user.spec.ts
```diff
@@ -21,13 +18,9 @@ test.describe('Pro user test', () => {
...
-         .getByTestId('current-price')
-         .first()
-         .innerText();
+      const proPrice = await productPage.getCurrentPrice();
 
-      expect(parseFloat(originalPriceString.replace('$', ''))).toBeGreaterThan(
-         parseFloat(proPriceString.replace('$', ''))
-      );
+      expect(proPrice).not.toEqual(originalPrice);
+      expect(parseFloat(proPrice)).toBeLessThan(parseFloat(originalPrice));
```
mlahargou, line 24: Why do we need to `parseFloat` again? Isn't the value returned from `getCurrentPrice` already a float?

mlahargou (approved):
CR 📱 

Looks great!

frontend/tests/playwright/product/pro_user.spec.ts
ardelato: `toFixed` returns a string so if we are trying to compare the numerical values between the prices then we need to convert it back.

I originally did this because I was influenced by this fix [`74674f6` (#1235)](https://github.com/iFixit/react-commerce/pull/1235/commits/74674f6e889da10c76e470d37d57cb793c45a45d), but seeing as it is really only used in one place. I think returning a `number` and having the Devs handle the `precision` would be better.

mlahargou: Ah gotcha! I'm good either way 🤷 

Maybe `parseFloat(parseFloat(val).toFixed(2))`?

frontend/tests/playwright/fixtures/product-page.ts
```diff
@@ -35,4 +35,58 @@ export class ProductPage {
...
+    * @description Locates and returns the original product price as a string with the format $###.## or $###
+    */
+   async getDiscountedPrice(): Promise<number> {
+      const price = await this.page
+         .getByTestId('product-price-section')
+         .getByTestId('compare-at-price')
+         .textContent();
+      expect(price).not.toEqual('');
+      expect(price).toMatch(/\$[0-9]+(\.[0-9]{1,2})?$/);
+      return parseFloat(price!.slice(1));
```
mlahargou, line 62: I'm almost inclined to have this return `parseFloat(parseFloat(price!.slice(1)).toFixed(2));`

mlahargou (approved):
CR 📱 but deploy_block 📱 on a minor suggestion. I worry that we might forget to add `toFixed` to another usage of this function down the line which will result in flaky tests.

frontend/tests/playwright/fixtures/product-page.ts
ardelato: If we are working with a price that has `.00` then we will need to repeat the `toFixed` call anyways since `parseFloat` will remove any decimal points that are `0`. Therefore, `parseFloat(parseFloat(price).toFixed(2))` is the same as `parseFloat(price)` where the price has `.00`.

zero points

ardelato: Hmm, I am not sure if it is good practice, but we could add an alternative function to return the `price` as a `string`.

```typescript

 async getCurrentPriceAsString(): Promise<string> {
      return (await this.getCurrentPrice()).toFixed(2);
   }
      
```

ardelato:
un_deploy_block ☁️ spoke with @mlahargou and we decided if the `toFixed` is forgotten a lot more than anticipated we can revisit changing the return type or adding an additional method.

PR #1262, opened 2023-01-16
Playwright: Add Cart Drawer Utility Functions

ardelato:
## Description

This is a continuation of https://github.com/iFixit/react-commerce/pull/1258 that builds on top of https://github.com/iFixit/react-commerce/pull/1261.

### Changes

- Adds utility functions for opening and closing the cart drawer
- Adds utility functions for incrementing/decrementing the quantities of cart items
- Adds a utility function to remove specific cart items
- Adds utility functions for asserting quantities and the presence of cart items

### QA 

- [ ] Confirm CI Passes

qa_req 0
Connects: #1217

### DIFF
--- a/frontend/tests/playwright/fixtures/cart-drawer.ts
+++ b/frontend/tests/playwright/fixtures/cart-drawer.ts
@@ -6,4 +6,12 @@ export class CartDrawer {
    constructor(page: Page) {
       this.page = page;
    }
+
+   async close() {
+      await this.page.getByTestId('cart-drawer-close').click();
+   }
+
+   async open() {
+      await this.page.getByTestId('cart-drawer-open').click();
+   }
 }
--- a/frontend/tests/playwright/fixtures/cart-drawer.ts
+++ b/frontend/tests/playwright/fixtures/cart-drawer.ts
@@ -1,4 +1,4 @@
-import { Page, Locator } from '@playwright/test';
+import { Page, Locator, expect } from '@playwright/test';
 
 export class CartDrawer {
    readonly page: Page;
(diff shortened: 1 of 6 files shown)


frontend/tests/playwright/fixtures/cart-drawer.ts
```diff
@@ -74,4 +74,12 @@ export class CartDrawer {
    async assertCartTotalQuantity(quantity: number): Promise<void> {
       expect(await this.getTotalQuantity()).toBe(quantity);
    }
+
+   async assertItemIsPresent(sku: string): Promise<void> {
```
mlahargou, line 78: In our iFixit/ifixit code, present and visible are two different things. Visible means it has a height/width and is not `display:none` (I think). Present just means it's on the dom.

Should we change this function to be `assertItemIsVisible`? Or `assertItemInDrawer`?

I'm just now realizing this function reads as "assert item is present in cart drawer" not "assert item is present on dom." Hmm, I think this is fine.

mlahargou (approved):
CR 📱 Looks great! So much easier to read :)

frontend/tests/playwright/fixtures/cart-drawer.ts
ardelato: > Should we change this function to be assertItemIsVisible? Or assertItemInDrawer?

Since this is a method of `CartDrawer` then I was thinking that the context in which the item is present, can be understood as within the `CartDrawer`. 

However, it is true that we are only asserting that the item is in the DOM and displayed within the `cart-drawer` element. 

Should we change it to `assertItemIsVisible`?

mlahargou: I think it's fine as is.

PR #1263, opened 2023-01-16
withLogging(): Fix sad bug!

danielbeardsley:
When that initialization was in the outermost function, the time() call was only happening once per callsite. This resulted in calling the *same* `done()` closure over and over, logging larger and larger time values.

Effectively, it *was* show time-since-boot each time a request finished.

Previously: 
image


danielbeardsley:
QA :+1: 

```
page.product.server_side_props: 511ms
```

andyg0808 (approved):
CR :+1: 
Looks reasonable to me.

PR #1264, opened 2023-01-17
Add Next.js api endpoint

dhmacs:
Related to #1156 

The goal of this PR is to define a prefix path for APIs defined in Next.js. This sets the stage for caching revalidation endpoints.

Following current iFixit convention (e.g. `/api/2.0`) I'm proposing to use `/api/3.0` and forward this path and all subpaths to the Next.js app.
I understand that you might want to use this for new versions of the iFixit PHP api, so I'm open for ideas. An alternative that I have in mind is `/api/rest`.
Thoughts?

👁️ Once we agree on the prefix, a forwarding rule must be set to serve all /api/3.0 subpaths through the Next.js app.

cc @sterlinghirsh @danielbeardsley 

qa_req 0

### DIFF
--- a/frontend/pages/api/nextjs/index.ts
+++ b/frontend/pages/api/nextjs/index.ts
@@ -1,0 +1,7 @@
+import type { NextApiHandler } from 'next';
+
+const handler: NextApiHandler = async (req, res) => {
+   res.status(200).json({ version: '3.0' });
+};
+
+export default handler;
--- a/frontend/pages/api/nextjs/cache/product.ts
+++ b/frontend/pages/api/nextjs/cache/product.ts
@@ -1,0 +1,7 @@
+import type { NextApiHandler } from 'next';
+
+const handler: NextApiHandler = async (req, res) => {
+   res.status(200).json({ model: 'product' });
+};
+
+export default handler;


danielbeardsley:
> but to me it sounds like a beta version of an api

I hear you. I think the thing I care about most is clear communication to our devs about where these API endpoints are hosted and perhaps their intent.

Sterling suggested `/api/2.0/next` , I'm kind of inclined to drop the version number since we likely won't be versioning this api (we've been on v2.0 for 10+ years and have just added to it). I think the important distinctions here will be:

* Which api endpoints are served by our next.js app
* Which api endpoints are meant for internal use
   * i.e. these cache endpoints will likely only be hit by our php app when product data changes, browsers won't be talking to them
   * The Internal idea could be communicated with another prefix: `/api/{whatever}/internal` so that's pretty easy to account for.

The problem stems from not having a great name for our next.js app. `react-commerce` will soon be even more of a misnomer. If we had a short code-name for this app, I think that would be a great prefix. Let me talk with the team.

danielbeardsley:
We spent a bit talking and the best arguments were given for `/api/nextjs`. It clearly communicates that these endpoints are served by a different app and *which app*.  `/api/next` may confuse for the reason you mentioned, that it seems like an upcoming version.

dhmacs:
Ok, I'm refactoring to `/api/nextjs` 👍

dhmacs:
I think the PR is now ready. Once merged we need to setup a Cloufront rule for this I think

cc @danielbeardsley @sterlinghirsh

frontend/pages/api/nextjs/index.ts
```diff
@@ -0,0 +1,7 @@
+import type { NextApiHandler } from 'next';
+
+const handler: NextApiHandler = async (req, res) => {
+   res.status(200).json({ version: '3.0' });
```
masonmcelvain, line 4: I'm curious why we're setting the initial version to `3.0` instead of `1.0`. @danielbeardsley's [argued](https://github.com/iFixit/react-commerce/pull/1264#issuecomment-1385939176) against versioning at all, but if we are going to use versions, I feel like it's ok to start from 1.0 since this api is distinct from iFixit's php api.

dhmacs: Don't pay attention to this, it's just a dummy endpoint that we'll use to test the path forwarding before we merge another PR with the real caching endpoints.

danielbeardsley (approved):
CR :+1:

dhmacs:
@danielbeardsley @sterlinghirsh let me know when the forwarding is done. Currently https://ifixit.com/api/nextjs redirects to the iFixit api

danielbeardsley:
> let me know when the forwarding is done.

It's deployed: https://www.ifixit.com/api/nextjs/product

PR #1272, opened 2023-01-19
Add indexId to correctly handle the Algolia state

### LINKED ISSUE #1191: Emptying search query doesn't properly update url
## Problem
When visiting https://www.ifixit.com/Parts?q=vive, the search query correctly starts pre-filled with vive. Editing the search query properly updates the url, but completely clearing the search query returns the url to the state it was in when the page loaded, i.e. ?q=vive.

## Do This
Make sure that clearing the query actually removes the query from the url.

federicobadini:
closes #1191 

Exploiting multiple times the same index on a page requires that all instances of `<Index />` have a unique `indexId` prop specified.
Otherwise the Algolia internal state is not updated correctly.

### QA

1. Visit Vercel preview
2. Navigate to a product-list page via a url with some pre-filled filters
3. Try to remove the filters and check that they are correctly removed from the url
4. Verify that filters continue to work also for interaction without pre-filled filters in the url

### DIFF
--- a/frontend/components/common/InstantSearchProvider.tsx
+++ b/frontend/components/common/InstantSearchProvider.tsx
@@ -64,7 +64,7 @@ export function InstantSearchProvider({
    const routing: RouterProps<UiState, RouteState> = {
       stateMapping: {
          stateToRoute(uiState) {
-            const indexUiState = uiState[indexName];
+            const indexUiState = uiState['main-product-list-index'];
             const routeState: RouteState = {};
             if (indexUiState.query) {
                routeState.q = indexUiState.query;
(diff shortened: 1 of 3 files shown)


masonmcelvain (approved):
CR ✌🏻 Good find. I wrote an issue to make a test for this, since it's kind of subtle https://github.com/iFixit/react-commerce/issues/1276

k0rvusk0r4x:
- After going directly to a product list page with a query via url, adding filters, then clearing the query, no longer returns the url back to it's original state.
- After going directly to a product list page with a query and filters via url, adding/subtracting filters, and changing/clearing/adding a new query changes the url and results as intended.
- Getting to list pages with queries and filters the "normal" way works. 
- After getting to list pages the "normal" way, adding/subtracting filters, and adding/changing/clearing queries works as intended.

QA 😺

PR #1273, opened 2023-01-19
Make tool categories selectable from other tool categories

### LINKED ISSUE #306: Make tool categories selectable from other tool categories
On https://www.ifixit.com/Store/Tools were presented with a list of tool categories on the left with the heading "BROWSE". Selecting "Gripping" will show the Tool Category page at https://www.ifixit.com/Store/Tools/Gripping . However, the list of other tool categories is still shown. Choosing one replaces the current selection, rather than refining the search more. This is nice for letting people browse our tool options.

On the new store, https://next.ifixit.com/Store/Tools shows the list of tool categories (only one is defined at time of writing this issue) but selecting a tool category doesn't continue to show the other options. There's still an Item Type facet, but it's just a list of refinements for tools that happen to exist in multiple categories.

It would be good to continue to show the list of tool categories to let someone browse our store instead of just refining the selection. In other words, we want to show sibling product lists (parent's children).

federicobadini:
🚧 - WIP - 🚧

closes #306


sterlinghirsh:
This is waiting on further discussion with @jeffsnyder and team to look into keeping the tool categories in sync with Strapi's tool product lists.

sterlinghirsh:
I looked at this again, and I think there's a problem with the spec that makes this tricky or impossible to solve. Right now, the top level children of the all-tools page are dictated by Strapi. Some of these match the Tool Category tag in Akeneo / Algolia (e.g. /Tools/Cleaning), but some are different (Algolia has Tool Category: 'ESD Prevention' but Strapi's url says /Tools/ESD_Safe). The list of sibling tool pages would want to come from Strapi instead of Algolia, but the more I think about it, the more I think it's just going to confuse things. I like having the filter that lets you refine the current tool category when tools have multiple tool category tags, so I'd rather not replace that. I also like having the child pages at the top, so I wouldn't want to confuse that by adding sibling pages.

Right now, my feeling is we bail on this. If we come up with a design that makes it clear, we could try implementing this by querying strapi for the all-tools children, but I don't think we should change our algolia logic or try to mimic what we have on Parts pages.

Sorry for the confusion and the waste of time. The problem wasn't clear to me until I saw this implementation and tried to work through resolving Strapi vs Algolia.

PR #1274, opened 2023-01-19
Footer Copyright: remove year

danielbeardsley:
We got 1000s of hydration errors around new-years when this value was different between server and client.

The year isn't particularly relevant here as copyrights last a long time. Technically, we don't even need a copyright message, but claims are more defensible if it was clear your content is copyrighted by you.

CC @sterlinghirsh @davidrans


davidrans (approved):
Ha!!

CR :coffee:

k0rvusk0r4x:
- The year is gone from the copyright.
- The text line(s) it was part of is still centered and looks good at various view port sizes.
- Otherwise, the footer continues to look and behave the same.

QA 😺

PR #1216, opened 2023-01-04
Product Page: Extend Add To Cart Tests with MSW handlers

ardelato:
## Description

This extends the `Playwright` `add_to_cart` tests by asserting the expected behavior for **low stocked** and **out of stocked** products. The list of assertions can be found in this issue https://github.com/iFixit/react-commerce/issues/1168.

### Changes

- Installed `MSW` and added a handler for the `shopify.findProduct` query
- Added an additional mocked product object for the find product query
- Updated the frontend `package.json` scripts for Playwright
- Added additional add to cart tests for low stocked and out of stocked products
- Changed the logic for how product inventory is determined in the `AddToCart` component

### CR Notes

There were a few _gotchas_ with setting up the environment to write these tests. Most of these _gotchas_ are explained in the commit messages.

Firstly, we needed to use `MSW` in order to mock the server-side requests. However, integrating `MSW` into our application was not very easy as our application is a lot more complex than the examples found online for integrating `MSW`, with `Playwright`, in a `Next.js` app. At the moment, we do not have a dynamic way of adding or removing handlers therefore we had to add conditionals for whether to intercept a request or let the request go through. In addition, it seems that `Playwright`'s `page.route` interceptor does not work when using `MSW`'s browser interceptor. As such, we choose to not enable `MSW`'s browser interceptor for now. 

There was another thing that should be noted. We use two different apis to reference the quantity of a product. When initially load a product page, we will use the quantity from the `shopify.findProduct` request. However, after adding the product to our cart, we will use the quantity from the `api/2.0/store/user/cart` endpoint, which uses our DB. In order to not mock the entire `cart` API, I changed the logic for which quantity we reference in the `AddToCart` component. We will now default to using `shopify`'s quantity first and fall back to the `cart`'s quantity as opposed to the reverse.  When the time comes, we should definitely use a single `api` for the `cart drawer` and the `product page`

### QA Notes

- [ ] Confirm CI should pass
- [ ] Confirm through the Vercel Preview app that visiting `/products/iphone-6s-plus-replacement-battery-low-stocked` or `/products/iphone-6s-plus-replacement-battery-out-of-stock` will return `404`
- [ ] Confirm running `NEXT_PUBLIC_MOCK_API=true pnpm dev` will cause the `msw` mocks to be used.
	- [ ] Visit  `/products/iphone-6s-plus-replacement-battery-low-stocked` and `/products/iphone-6s-plus-replacement-battery-out-of-stock` to confirm the pages are now rendered
[...]

PR #1275, opened 2023-01-20
Playwright: Enable the ability to set handlers dynamically

ardelato:
## Description

In #1216 we integrated **MSW** into our **Playwright** tests but we did not have the capability of mutating the responses of the server-side requests within the tests. We were only able to define static responses. In order to return different responses for the same request, we had to create conditional handlers.[^1] This is not ideal in the long run because it will lead to multiple conditionals to return different types of products for the same request.

Therefore, this pull request aims to add the capability of dynamically setting these server-side handlers within the tests.

<details>
<summary><strong>Details</strong></summary>

In order to dynamically set the handlers, there were quite a few changes that needed to be made that are not as straightforward as they seem. Hopefully, this section will help to explain the changes that were made.

#### Separate Processes

The inability to dynamically set the handlers within the tests is due to the **Playwright** tests running in a separate process from the **MSW** server. This means that the **Playwright** tests cannot directly access the **MSW** server to set the handlers because the **MSW** server is setup as we start up the **Nextjs** server (i.e. `pnpm run dev | pnpm run start`).

First MSW Integration

#### Custom Nextjs Server

To get around this, we need to spin the **Nextjs** server ourselves to return the instance to the **Playwright** tests. Using the **Nextjs** custom server documentation[^2] and a **Dev.to** article[^3] we were able to create and return the **Nextjs** server instance, with **MSW** integrated, to the **Playwright** tests.[^4].

It should be noted that our custom server is written in `Typescript` as opposed to `Javascript` like in the **Nextjs** documentation[^2]. In addition, we are using the built-in `http` package to create the server as opposed to the `Koa` package in the article[^3]. Aside from those significant differences, the logic in the code is the same.

#### Starting the Custom Server(s)

This was the most difficult part of this pull request mainly due to all the options that were available and having to choose the best approach.

The first approach as seen in the **Dev.to** article[^3] was to make use of the `beforeAll` and `afterAll` hooks in the **Playwright** tests. This approach would start a new **Nextjs** server for each test with the maximum number of servers set to the total number of workers. This would be a problem because we would need to start and listen for a custom server for each test in a file when there would be only a single test that would require the need to define a handler. This would be a waste of time and resources.

The second approach was to just start the custom server once and then use the `beforeEach` and `afterEach` hooks to define and reset the handlers. This approach would work but it would limit us to a single server. As such, this would cause a problem if tests mock the same endpoint but with different responses. We could run the tests serially to get around this but that would slow down the tests.

The last and best approach was inspired by this article [^5] that uses **Playwright's** powerful fixtures to create and start a custom server for each test worker. **Playwright** has both `test` and `worker` fixtures[^6], in which we can define the scope of the fixture. If we use the `worker` scope, then we can define a custom server for each worker instead of each test. This would allow us to define different handlers for the same endpoint in different tests. All while still running the tests in parallel. In addition, **Playwright** will only execute fixtures when a test or hook needs them, so we will only start a custom server when we need it. This is great in development where we are debugging or creating a single test and do not want to start a custom server. However, in CI, we will want to start a custom server for every worker before running the tests[^7] because of some _"**gotchas**"_. More on that later.

**Playwright's** fixture system was the best approach and made it easier to use the custom server in the tests without having to import it in every test file.

#### Final Integration

The final integration of the custom server was to optimize when and where we start the custom server. Using **Playwright**'s fixtures resolved the performance issue of starting the custom server for every test in a file. However, not every test requires a custom server. Therefore, to keep isolated environments, we will continue to start the regular server at port `3000` using `pnpm run start`. Regular tests will use the regular server and tests that require server-side mocking will use the custom server. This will allow us to keep the tests isolated and not have to worry about mocks affecting the requests of regular tests.

Final Integration

#### Gotchas

As mentioned above, there are some _"**gotchas**"_ when using the `worker` scope for the custom server fixture.

The first is that the custom server will not be started until a test or hook needs it. This means that if we have a test that does not require the custom server, then the custom server will not be started. However, if a test in the middle of a test suite requires it, then we will take longer to run the test because we have to wait until the server starts to be able to run that specific test. While not a problem in development, this could be a problem in CI if we time out the test because starting the server took too long. This is why we will want to start the custom server for every worker before running the tests and in addition increment the `test timeout` [^7][^8].

Similar to the first, if we start the custom server lazily, then it could cause a problem when running tests with multiple workers. To elaborate, if two workers running in parallel pick up tests that require the custom server, they will build and start the custom server around the same time. Both servers will be building and caching to the same directory (i.e. `.next/`). This means the build files will be overwritten and one server's webpack instance will throw a warning/error because it lost the link to some of the files it had built. As a result, we can get an error page and the test fails because it could not locate an expected element. To get around this in development, we will need to ensure that we run the tests with a single worker or skip the tests that require the custom server. In CI, we do not have to worry about this because we will build the app before running the tests and therefore the custom servers will use the same build files.[^9]

Cache Warning 2

</details>

### CR:

I tried my best to explain the changes in each commit message. I also tried to summarize the changes in the pull body. If you have any questions, please let me know.

### QA:

-  [x] Confirm CI passes
-  [x] Confirm that running the server with `NEXT_PUBLIC_MOCK_API=true pnpm dev` will enable **msw** and use the defined handlers in https://github.com/iFixit/react-commerce/blob/main/frontend/tests/playwright/msw/handlers.ts
   -  You can just copy what the file looked like before the changes and confirm visiting `localhost:3000/products/[handle]` will show the mock data based on the `handle` in the URL.

cr_req 2

[^1]: [Conditional Static Handlers](https://github.com/iFixit/react-commerce/blob/main/frontend/tests/playwright/msw/handlers.ts#L6-L46)
[^2]: [Nextjs Custom Server](https://nextjs.org/docs/advanced-features/custom-server)
[^3]: [Typescript Custom Nextjs Server](https://dev.to/kuroski/writing-integration-tests-for-nextjs-next-auth-prisma-using-playwright-and-msw-388m#:~:text=To%20make%20it%20work%2C%20first%2C%20we%20need%20to%20create%20a%20custom%20Next.js%20server)
[^4]: [Our Custom Server](https://github.com/iFixit/react-commerce/blob/feature--setup-ability-for-msw-dynamic-handlers/frontend/tests/playwright/fixtures/custom-nextjs-server.ts)
[^5]: [Nextjs Server as Playwright Fixtures](https://frontend-digest.com/using-playwright-to-test-next-js-applications-80a767540091#:~:text=Starting%20one%20instance%20per%20worker%20using%20Fixtures)
[^6]: [Playwright Worker Fixtures](https://playwright.dev/docs/test-fixtures#worker-scoped-fixtures)
[^7]: [Force Fixture Initialization in CI](https://github.com/iFixit/react-commerce/blob/1bb222b854de48cefbdd5670f813796a2f143803/frontend/tests/playwright/test-fixtures.ts#L38)
[^8]: [Test specific Timeout Override](https://github.com/iFixit/react-commerce/blob/1bb222b854de48cefbdd5670f813796a2f143803/frontend/tests/playwright/test-fixtures.ts#L39)
[^9]: [Production Build CI](https://github.com/iFixit/react-commerce/blob/feature--setup-ability-for-msw-dynamic-handlers/frontend/tests/playwright/fixtures/custom-nextjs-server.ts#L20-L26)


ardelato:
Going to close this PR for now, there are some additional changes that need to be made for the GitHub action in order to get this working in CI.

deltuh-vee:
QA 🎬
* CI is still passing
* Handlers can still be defined in /frontend/tests/playwright/msw/handlers.ts

frontend/tests/playwright/test-fixtures.ts
```diff
@@ -1,14 +1,65 @@
+/**
+ * Playwright test fixtures are very powerful and can be used to create a custom
+ * test environment.
+ *
+ * Fixtures can be isolated to a test or a worker depending on the scope defined.
+ *
+ * In addition, fixtures can be lazily loaded, which means that the fixture will only be created where it is used. If a test scoped fixture is not used in a test, it will not be created. If a worker scoped fixture is not used in a worker, it will not be created, and if a worker scoped fixture is used in a test, then it will remain instantiated for the duration of the worker. Overall this a very powerful feature that can help with expensive setup and teardown.
```
mlahargou, line 7: This is super cool!

mlahargou (commented):
Tooks a first look at this all and... damn! Nice job!

Still wrapping my head around a few things, but will continue CRing in a bit.

andyg0808 (commented):
CR :+1:
Looks pretty good to me. deploy_block :+1: on some questions.

This is neat!

frontend/tests/playwright/fixtures/custom-nextjs-server.ts
```diff
@@ -0,0 +1,94 @@
...
+         server.on('listening', async () => {
+            const port = (server.address() as any).port;
+            console.log(`> Ready on http://${hostname}:${port}`);
+
+            /*
+             * Although the server is ready, the Next.js app has not finished
+             * building yet.
+             * So we need to wait a bit before resolving the promise.
+             */
+            await new Promise((resolve) => setTimeout(resolve, 3000));
```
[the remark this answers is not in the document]
ardelato: While we can use the `port` to ping if the server has started, it does not necessarily mean that the server has finished building the app. That is why I set this delay, as I am not sure how we could tap into the `Next.js` server to know when the server has finished building the app. If we had direct access to the `webpack` server then we could curl for a file to know it was done building. This would be similar to how we did it in `iFixit/ifixit`.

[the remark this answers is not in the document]
ardelato: Do you have any suggestions I should try looking into?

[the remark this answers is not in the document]
andyg0808: Assuming that requests to the server in that state don't block, I'm not sure that I have any better ideas; seems like a reasonable approach, given the circumstances.

frontend/tests/playwright/product/add_to_cart.spec.ts
```diff
@@ -114,9 +116,22 @@ test.describe.serial('product page add to cart', () => {
...
+               const lowStockedProduct = cloneDeep(mockedProductQuery);
+               if (lowStockedProduct.product) {
+                  lowStockedProduct.product.variants.nodes[0].quantityAvailable = 3;
+               }
+               return res(ctx.data(lowStockedProduct));
+            })
+         );
+
+         await page.goto(
+            `http://localhost:${port}/products/iphone-6s-plus-replacement-battery-low-stocked`
```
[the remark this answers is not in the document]
ardelato: I made a follow-up issue about making it easier to interface https://github.com/ifixit/react-commerce/issues/1290.

> might it be worth always booting the MSW server

Hmm, if there is a way to dynamically modify the `productPage` fixture's dependency on the `MSW` server then that would be optimal, but if we can't then always booting the MSW server may slow down the Devs.

frontend/tests/playwright/msw/index.ts
```diff
@@ -15,7 +17,8 @@ export const setupMocks = async () => {
     * Therefore, we only set up MSW for the server, and not the browser.
```
mlahargou, line 17: I the comment above is out of date now.

frontend/tests/playwright/test-fixtures.ts
mlahargou: This is also a super long comment ;)

mlahargou (dismissed):
CR 📱 I think it looks great! Keeping deploy_blocked on Andrew's comments.

frontend/tests/playwright/fixtures/index.ts
```diff
@@ -1,9 +1,16 @@
...
 export type ProductFixtures = {
    productPage: ProductPage;
    cartDrawer: CartDrawer;
 };
 
-export { ProductPage, CartDrawer };
+export type CustomNextjsServer = {
+   serverRequestInterceptor: SetupServerApi;
+   port: number;
+};
```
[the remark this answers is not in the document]
ardelato: Addressed in [4798d8b](https://github.com/iFixit/react-commerce/pull/1275/commits/4798d8bc9268786f272032ee7902412d9add6a88)

frontend/tests/playwright/msw/index.ts
ardelato: Ah that latter half should probably get removed in #1289. I updated the issue to add that as a **To Do.**

ardelato: Addressed in [01dfc69](https://github.com/iFixit/react-commerce/pull/1275/commits/01dfc6974c1f134202aaacae4a5aaf5ed0be0f8a). We will be making some changes to this file with #1289 so I updated the comment.

frontend/tests/playwright/test-fixtures.ts
ardelato: Addressed in [8a3c12c](https://github.com/iFixit/react-commerce/pull/1275/commits/8a3c12cd936cbf92b837962b9708fc5ad8d32fcc)

andyg0808 (approved):
CR :+1:

deltuh-vee:
dev_block ✊ 
Handlers in `/frontend/tests/playwright/msw/handlers.ts` are not being rendered anymore.

ardelato:
un_dev_block ⚡ and QA 👍  

I confirmed adding handlers to `/frontend/tests/playwright/msw/handlers.ts` are still being rendered when running `NEXT_PUBLIC_MOCK_API=true pnpm dev`

handlers work

mlahargou:
CR 📱

PR #1280, opened 2023-01-20
Drop use of backwards-incompatible `Object.fromEntries()`

masonmcelvain:
Fixes a bug where `Object.fromEntries` is undefined on older browsers.

https://sentry.io/share/issue/d4e90fc3843348c599777eec6e1484d2/

We could polyfill this function instead, but we already follow this reduce pattern elsewhere in our code.

https://github.com/iFixit/react-commerce/blob/50ec98ff8596e4d505e466821525e9750ac252cd/packages/cart-sdk/hooks/use-cart.ts#L122 https://github.com/iFixit/react-commerce/blob/aedbde5560847162eb2373d61f268560fe312674/frontend/models/product.server.ts#L497

CanIUse: https://caniuse.com/mdn-javascript_builtins_object_fromentries

## QA

Visit a product list page on an older browser that doesn't have Object.fromEntries, and confirm there is no Sentry event reported and that you can view the price of the product.

### DIFF
--- a/frontend/templates/product-list/sections/FilterableProductsSection/useProductSearchHitPricing.ts
+++ b/frontend/templates/product-list/sections/FilterableProductsSection/useProductSearchHitPricing.ts
@@ -40,15 +40,13 @@ export function useProductSearchHitPricing(
 
    const proPricesByTier =
       product.price_tiers && !isEmpty(product.price_tiers)
-         ? Object.fromEntries(
-              Object.entries(product.price_tiers).map(([tier, price]) => [
-                 tier,
-                 {
-                    amount: price.default_variant_price,
-                    currencyCode: 'usd',
-                 },
-              ])
-           )
+         ? Object.entries(product.price_tiers).reduce((acc, [tier, price]) => {
+              acc[tier] = {
+                 amount: price.default_variant_price,
+                 currencyCode: 'usd',
+              };
+              return acc;
+           }, {} as Record<string, Money>)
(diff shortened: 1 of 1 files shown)


deltuh-vee:
dev_block ✊ 
I'm still seeing an "Object.fromEntries is not a function" error
https://sentry.io/organizations/ifixit/issues/3887715303/?project=6475315&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=1h

For reference, I was visiting https://react-commerce-git-object-from-entries-unsupported-ifixit.vercel.app/Parts on Chrome version 72.0.3626.109 through Browserstack.

masonmcelvain:
Hm actually it seems like these errors started happening the day we merged #1255 ([ref](https://github.com/iFixit/react-commerce/pulls?q=is%3Apr+is%3Aclosed+merged%3A2023-01-13)), which makes me think it's related to the chakra upgrade.

@sterlinghirsh @danielbeardsley do we care about these errors on older browsers? I had figured it was fixable on our end, but we may have to downgrade Chakra or open an issue on their end if we want to resolve them.

danielbeardsley:
Polyfilling seems easy enough and protects against usages in chakra and elsewhere. I vote for just updating our polyfill url to add: `Object.fromEntries` in the list of features.

masonmcelvain:
Good idea, I had even mentioned that in the PR comment but forgot it was an option!

Issue #1281, opened 2023-01-20
GitHub Actions: Cache Next Build


ardelato:
## Description

While testing for #1238, I noticed we do not cache the build for `Next` in CI. We probably should to speed up sequential CI runs

### To Do:

- [ ] Cache the `.next` directory to speed up builds in CI
https://nextjs.org/docs/advanced-features/ci-build-caching#github-actions

Issue #1279, opened 2023-01-20
GitHub Actions: Move starting Frontend Server as a Step


ardelato:
## Description

Playwright can start up the Frontend server if one is not already running. However, it is not a perfect feature as we sometimes hang or fail to boot up the server. As a result, the logs are hidden away by Playwright. 

![image](https://user-images.githubusercontent.com/22064420/213765834-c44823d1-cfad-40d7-b47d-15797289a2cf.png)

Let's move starting the Frontend server to a separate step in our actions workflow. This should ensure that Playwright will not time out nor hide the reason why the server did not boo up successfully.

### To Do:

- [ ] Add a step before running the Playwright tests in `main.yml` to start up the Frontend server with `pnpm build & pnpm start`
- [ ] Ensure Playwright does not start the webserver if we are in CI

ardelato:
Reference: https://github.com/iFixit/react-commerce/pull/1282

I chose to leave starting the Frontend server as a command Playwright runs, but extracted the building of the Nextjs app to an external step. It is quite messy to start the server as a different step and be able to know when the server has finished booting. If Playwright starts the server then it will automatically wait for it before running the tests. Hopefully, Playwright will not time out waiting on the server since we extracted the build step.

Issue #1238, opened 2023-01-10
Configure MSW for dynamic handlers


ardelato:
## Description

In #1216 we integrated `MSW` into our Playwright tests but we do not have dynamic handlers to mutate the response from the server-side requests. As a result, we had to conditionally set URL handles to return pre-set products.

This is not ideal in the long run because it will lead to multiple conditionals to return different types of products.

### To Do

- [x] Configure our integration of `MSW` to allow modifying the response for the handlers in each tests.

If we can get the dynamic handlers working we will add the `MSW` `browser worker` back in and use it to handle the `page.route` paths we have.

PR #1282, opened 2023-01-20
Playwright GitHub Actions: Move build step to separate step and cache it

ardelato:
## Description

We occasionally fail to run the Playwright tests because it times out waiting for the app to finish building and for the server to start up.

By moving the build step to a prior step, in the GitHub action workflow, we can ensure that Playwright just needs to start and wait for the server to be ready.

In addition, this build step will help when we create the custom Next.js servers needed for the dynamic `msw` handlers.[^1]

### Changes

- Moves the `pnpm build` step from the Playwright config into a separate GitHub action step
- Ensures the app will be built using the `.env.test` values
- Caches the `.next` directory to help speed up future CI runs.[^2]

### CR:

I could not extract the `Frontend Server` to an external GitHub action step as requested in #1279 because it is quite messy to start the server as a background process and be able to know when the server has finished booting. In addition, we set the requirement that Playwright will not reuse an existing server in CI. By keeping Playwright in control of starting the Frontend server, we can ensure the tests will only be run when the server finishes booting and will not be reused. Hopefully, Playwright will not time out waiting on the server since we extracted the build step. 

### QA:

qa_req 0 CI should pass and I confirmed the caching works in https://github.com/iFixit/react-commerce/actions/runs/3971918976/jobs/6809453451#step:9:15 

Closes: #1279 
Closes: #1281 
Connects: #1238 

[^1]: #1238 
[^2]: [nextjs.org/docs/advanced-features/ci-build-caching#github-actions](https://nextjs.org/docs/advanced-features/ci-build-caching#github-actions)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -55,6 +55,10 @@ jobs:
 
          - name: Build Next.js
            if: steps.nextjs-cache.outputs.cache-hit != 'true'
+           env:
+              NODE_ENV: test
+              NEXT_PUBLIC_STRAPI_ORIGIN: http://localhost:1337
+              ALGOLIA_API_KEY: ${{ secrets.ALGOLIA_API_KEY }}
            run: cd frontend && pnpm build
 
          - name: Run Playwright Tests
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,6 +46,17 @@ jobs:
            if: steps.playwright-cache.outputs.cache-hit != 'true'
            run: cd frontend && npx playwright install
 
+         - name: Cache Next.js build
+           uses: actions/cache@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -46,6 +46,17 @@ jobs:
            if: steps.playwright-cache.outputs.cache-hit != 'true'
            run: cd frontend && npx playwright install
 
+         - name: Cache Next.js build
+           uses: actions/cache@v3
+           id: nextjs-cache
+           with:
+              path: frontend/.next
+              key: ${{ runner.os }}-nextjs-${{ hashFiles('**/pnpm-lock.yaml') }}
```
mlahargou, line 54: Sweet!

PR #1283, opened 2023-01-23
Add SWR caching lib value validation

dhmacs:
Related to #1156 

In the previously merged implementation, the value schema was used only to infer the correct return type. This PR actually uses the schema to verify that the entry value is valid. This mechanism will guard from cache entries not matching a schema change, automatically treating those entries as cache misses

qa_req 0


masonmcelvain (approved):
CR ✌🏻

PR #1284, opened 2023-01-23
Algolia: Missing null check

masonmcelvain:
Fixes a bug where the code assumes the `indexUiState` is never undefined.

Sentry event: https://sentry.io/share/issue/ae902268ed2d4c2ebc4318ada29daeb4/

Bug introduced in #1272, where we started using the `indexId` instead of `indexName` to key the Algolia state.

## QA

Confirm that you can't reproduce the Sentry error on a product list page. Not sure on the exact reproduction steps, but it seems to be related to adding/removing facets or clicking on product list children links.


masonmcelvain:
QA ✌🏻 I can reproduce the error in the console by adding facets/clicking child links on www.ifixit.com, but can't reproduce on these previews

PR #1286, opened 2023-01-23
Polyfill `Object.fromEntries()`

masonmcelvain:
Fixes errors on older browsers that don't have Object.fromEntries(). These errors started happening when we upgraded Chakra in #1255.

See #1280 for background.

## QA

Visit a product list page on an older browser that doesn't have Object.fromEntries, and confirm there is no Sentry event reported and that you can view the price of the product.

### DIFF
--- a/frontend/pages/_app.tsx
+++ b/frontend/pages/_app.tsx
@@ -27,7 +27,7 @@ function MyApp({ Component, pageProps }: AppPropsWithLayout<any>) {
    return (
       <>
          <Script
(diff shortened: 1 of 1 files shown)


davidrans (approved):
CR :coffee:

deltuh-vee:
QA 🎬
I was able to visit a product list page on an older browser that doesn't have Object.fromEntries and not trigger a Sentry error. I was also able to view the price of the product.

Issue #1293, opened 2023-01-24
Hotfix: Create a timeout for the Playwright Action


ardelato:
## Description

It looks like `Playwright` tests are taking a long time because it can't communicate with the `backend` docker container

![Docker Request](https://user-images.githubusercontent.com/22064420/214434826-ab01364b-b8f4-40c0-9ac9-b0ec823b146b.png)

### To Do

- [ ] Apply a quick fix to timeout the action a lot sooner while we investigate why there is an issue communicating with the `backend`

**References:**
- https://ifixit.slack.com/archives/C02SMF0AZPF/p1674596480836209
- https://github.com/iFixit/react-commerce/actions/runs/4000511463/jobs/6865723423

PR #1294, opened 2023-01-24
Hotfix: Add timeout to Playwright Github Action

ardelato:
## Description

There is currently an issue with this action that requires further investigation. **Playwright** seems to be unable to communicate with the `backend` server (i.e. localhost:1337) which is causing the tests to be re-run multiple times and eventually fail with a total time of about 30 minutes.

For now, let's set a timeout of 15 minutes to prevent the action from running more than normal.

### QA

qa_req 0 CI should continue passing

Closes: #1293

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,6 +5,7 @@ env:
    npm_config_userconfig: './.npmrc'
 jobs:
    playwright-run:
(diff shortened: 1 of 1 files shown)


sterlinghirsh (approved):
CR :zap:

Issue #1296, opened 2023-01-24
Investigate why Playwright fails to talk to the Backend in CI


ardelato:
## Description

This a follow-up to https://github.com/iFixit/react-commerce/issues/1293, let's investigate why Playwright cannot talk to the `backend` server.

![image](https://user-images.githubusercontent.com/22064420/214446983-3b98bbda-c521-497f-8ec0-1e8cad19b4b3.png)

ardelato:
Closing this out. It looks like the changes from #1302 have resolved this issue and therefore we are no longer seeing the E2E actions reaching the 15-minute timeout https://github.com/iFixit/react-commerce/actions/workflows/main.yml.

PR #1301, opened 2023-01-26
Upgrade Next to 12.2.4 to possibly resolve socket hang up error

ardelato:
## Description

After some investigation, I found that the socket hang error might be resolved if we upgrade **Next** to `12.2.4` since the fetch error originates from the next server.

https://github.com/vercel/next.js/issues/36251#issuecomment-1199408517

```
at ClientRequest.<anonymous> (/home/runner/work/react-commerce/react-commerce/node_modules/.pnpm/next@12.2.3_t7ss3ubh4wigfvyfclbvqff3gm/node_modules/next/dist/compiled/node-fetch/index.js:1:65763)
```

## CR

The socket hang error is not easy to reproduce, but it seems after some retries I was able to trigger the error in https://github.com/iFixit/react-commerce/tree/testing--timeout

https://github.com/iFixit/react-commerce/actions/runs/4010961695/jobs/6888020732#step:11:23

That said, when updating **Next** to `12.2.4`, it seemed the **Playwright** tests were no longer prone to the error. However, we will only know if this was fixed after a couple more live CI runs. 

## QA: 

- [ ] CI passes
- [ ] A quick smoke test checking pages still look fine
- [ ] There should be no errors similar to what's shown in https://github.com/iFixit/react-commerce/issues/1296

Closes: #1296

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -55,7 +55,7 @@
       "ioredis": "5.2.4",
       "lodash": "4.17.21",
       "lru-cache": "7.10.1",
-      "next": "12.2.3",
+      "next": "12.2.4",
       "nextjs-progressbar": "0.0.14",
       "query-string": "7.0.1",
       "react": "18.2.0",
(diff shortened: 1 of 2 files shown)


ardelato:
The issue might be something related to the caching of the build. I will need to look into this a bit more

PR #1302, opened 2023-01-26
GitHub Actions: Change Nextjs Caching Logic

ardelato:
## Description

If we change some logic to the `getServerSideProps` function we need to bust the cache. We currently only rely on the packages installed to bust the cache which is not enough. 

This will implement the same cache-busting logic we use for Webpack in iFixit/ifixit.
https://github.com/iFixit/ifixit/blob/master/.github/workflows/E2E.yml#L35-L42

## QA: 

- [ ] CI should now pass
- [ ] Caching should still work for Nextjs

**cc:** @andyg0808

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,7 +52,10 @@ jobs:
            id: nextjs-cache
            with:
               path: frontend/.next
-              key: ${{ runner.os }}-nextjs-${{ hashFiles('**/pnpm-lock.yaml') }}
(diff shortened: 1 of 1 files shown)


andyg0808 (approved):
CR :+1:

andyg0808:
QA :+1: 
image

andyg0808:
CR :fish: 
Pulldasher doesn't seem to be picking up this pull.

Issue #1266, opened 2023-01-17
Analytics: Add Cart Drawer Cross-Sell event


erinemay:
Add 225 from our [event tracking spreadsheet](https://docs.google.com/spreadsheets/d/16Xas1pHf_K7FmfOX3yX2mYgeYr2DIENr8cyQp19FZkA/edit?pli=1#gid=1839136207&fvid=1843464925).

> Add to Cart Button for the upsell feature on the Mini Cart

![image](https://user-images.githubusercontent.com/10889879/213032295-2ae8ef7e-d0da-499c-9d72-89b0d5ce54ff.png)

![image](https://user-images.githubusercontent.com/72166715/213027347-13ce89f1-26ce-4197-af9d-2708665d9190.png)

cc: @addison-grant

addison-grant:
CC @jeffsnyder

jeffsnyder:
Sorry to be annoying. This is the "Cart Drawer" and it's a "Cross-sell" not an "upsell". 🤓 

@addison-grant will figure out the new event code we need.

addison-grant:
I'm contacting our consultant about updating the event to say "Add to Cart - Cart Drawer" instead of "Add to Cart - Mini Cart," and I am also asking if we can add the term "Cross-Sell" into the event schema somehow.

danielcliu-ifixit:
@jeffsnyder do we want both `Add to Cart - Cart Drawer` and `Add to Cart` events fired when we use the button in question, or just `Add to Cart - Cart Drawer`?

addison-grant:
Updated 
```
Category:	Add to Cart - Cart Drawer
Action:	Add to Cart - Cart Drawer - $productcode-$optionid
Name:	{{PageUrl}}
```
Let me confirm, but I don't think we need a separate `Add to Cart` event

jeffsnyder:
@addison-grant Correct, we should not fire two Add to Cart events for a single Add to Cart action.

danielcliu-ifixit:
yep that should be how it behaves now

PR #1308, opened 2023-01-27
Have Cart Drawer Cross Sell Use Specific Matamo Event

danielcliu-ifixit:
## Overview
Basically we want matomo events for adding to cart to be slightly different if they were added in unique ways (the two current unorthodox ways we want to track are `Frequently Bought Together` and the cart drawer's cross sell button). This pull's main point is to add the specific event to track cart drawer cross sells, but it also changes our `trackAddToCart()` to have its parameters better match the expected values from the [spreadsheet](https://docs.google.com/spreadsheets/d/16Xas1pHf_K7FmfOX3yX2mYgeYr2DIENr8cyQp19FZkA/edit?pli=1#gid=1839136207&fvid=1843464925).  

## QA

Make sure that the matamo events fired by adding to the cart normally, adding to the cart view the frequently bought together section, and adding by the cart drawer cross sell section all create the correct Matomo events as specified by the spreadsheet or by this issues comment (specifically in regards to the cart drawer). 

Closes: #1266

### DIFF
--- a/packages/analytics/index.tsx
+++ b/packages/analytics/index.tsx
@@ -30,21 +30,21 @@ export const trackInMatomoAndGA = (trackData: TrackEventMatomo) => {
 export function trackAddToCart(
    cart: CartLineItem[],
    addToCartInput: AddToCartInput,
-   message?: string
+   eventSpecification?: string
 ) {
    if (typeof window === 'undefined') {
       return;
    }
    trackMatomoCartChange(cart);
    trackGoogleAddToCart(addToCartInput);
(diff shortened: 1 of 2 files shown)


masonmcelvain (approved):
CR ✌🏻

erinemay:
QA 👍 
All 3 events are distinct. Events for `frequently bought together` and regular `add to cart` match the events sheet. 
Cart drawer adds match the new spec in the issue: https://github.com/iFixit/react-commerce/issues/1266#issuecomment-1405237957

All continue to have the correct itemcode, optionid, and url on which the action was made.

image

PR #1314, opened 2023-01-30
sentry sourcemaps: upload framework sourcemaps too

danielbeardsley:
We still see a fair number of errors (hydration errors) who's source maps are unavailable because the stack frames are in dist/framework.

Adding this option enables uploading those too. Unclear how bigger this makes the uploads, but we'll find out!

Connect #1209 

qa_req 0 - vercel does a `next build` which covers this.

See the docs: https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/#widen-the-upload-scope

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -140,11 +140,14 @@ const moduleExports = {
       return config;
    },
    sentry: {
+      autoInstrumentServerFunctions: false,
+      // Upload artifacts in dist/framework as well; this includes sourcemaps
+      // for react and other next.js code
+      widenClientFileUpload: true,
       ...(!SENTRY_AUTH_TOKEN && {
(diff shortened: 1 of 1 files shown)


masonmcelvain (dismissed):
CR ✌🏻 but there's conflicts

masonmcelvain (approved):
CR ✌🏻 carry forward, fixed a simple merge conflict

danielbeardsley:
This did upload more files (104 now, before it was 96). But it still doesn't include the ones of interest


Here's the contents of the deployment according to vercel:
image

sentry has the `698` chunk:
image

But not the `main` chunk nor several others:
image

danielbeardsley:
Likely we can still configure *more* options to the sentry-webpack plugin.

Issue #1289, opened 2023-01-24
Playwright: Integrate MSW into Browser


ardelato:
## Description

In #1216, we were unable to use `msw` for the client-side requests as it interfered with Playwright's interceptor `page.route`. In addition, we were unable to have dynamic handlers so opted to not integrate `msw` for the browser side until we had the ability for dynamic handlers.

We are soon going to merge this capability #1275 and therefore we can work on the browser side.

## To Do:

As we were testing dynamic handlers it seems that there is a package called `playwright-msw`[^1] that will make it easier to integrate `msw` into the browser as well as instantiate it as a Playwright fixture

- [ ] Install `playwright-msw`
- [ ] Replace `page.route` requests with `playwright-msw`
- [ ] Ensure static handlers can still be used for the `browser` when running `NEXT_PUBLIC_MOCK_API=true pnpm dev`
	- This is to help debug and test handlers without having to run Playwright

[^1]: https://www.npmjs.com/package/playwright-msw

Issue #1278, opened 2023-01-20
Cross sell integration tests fail when run from Europe


dhmacs:
Cross Sell section is initially hidden for European users until they explicitly select that they want to buy from the US store:

<img width="661" alt="Screenshot 2023-01-20 at 18 16 10" src="https://user-images.githubusercontent.com/4640135/213762482-d3506f79-db70-43ff-ac3b-8476ba236510.png">

This makes tests fail when they're being run from Europe.

## Do this

Force the US locale on the test so that tests behave predictably

cc @iFixit/qae

mlahargou:
Cool! We can get this patched up 👍

mlahargou:
We might be able to emulate the location: https://playwright.dev/docs/emulation

ardelato:
I am trying to reproduce this bug but I cannot do so.

The international buy box only renders if we change the store in www.cominor.com and continue using the same browser instance to retain the cookies.

https://user-images.githubusercontent.com/22064420/215567851-8b3952e8-b632-4aa7-847f-dc9b56b4b1a3.mp4

--- 

I also tried using Chrome's Locale tool and a VPN to change my location to Italy, but that did not cause the international buy box to render either.

![IP address does not change international buy](https://user-images.githubusercontent.com/22064420/215568254-efe0f05a-91a3-4a7e-93dd-839365ad0c49.png)

---

@dhmacs are you still able to reproduce this error, if so can you provide the steps?

dhmacs:
@ardelato Yes I'm able, and I'm not doing anything in particular, just start the dev server and run `pnpm playwright:run`

ardelato:
I found the issue and followed up with a solution in #1319.

Here are the reproduction steps:
1. Change IP location to Europe
2. Whitelist the IP for Cominor
3. Start up the Dev Server normally and validate the BuyBox is now rendered
4. Run tests that do not have the `api/2.0/internal/international_store_promotion/buybox` mocked. 
5. Tests should timeout because `Add to Cart` is not rendered and the `BuyBox` is rendered instead. 

[...]

PR #1319, opened 2023-01-31
Playwright: Integrate Browser Side Mock Service Worker

ardelato:
## Description

This adds the ability to dynamically mock the browser-side requests within our tests while keeping the mock API the same. This means we will be using `MSW` for both the server-side requests and the browser-side requests instead of using `Playwright` for the browser side and `MSW` for the server side.
 
### CR

To make it easier to integrate as a fixture, I am using a package called `playwright-msw` that I discovered on a discussion topic for the `MSW` repo. https://github.com/mswjs/msw/discussions/1322#discussioncomment-3799174. This package mainly wraps all the logic that would be needed to integrate the browser-side service worker into a Playwright fixture, thus saving us time from reimplementing this logic ourselves. 

https://github.com/valendres/playwright-msw/blob/main/packages/playwright-msw/src/router.ts

In addition, I discovered that one of our mocked requests `api/2.0/internal/international_store_promotion/buybox` helped to prevent the international buybox from rendering if we were located in Europe. As such, I moved this request handler to the default `handlers.ts` file. 
 
### QA

- [x] CI passes
- [ ] If IP address is in Europe, tests continue to pass
- [ ] If IP address is in Europe, `international buybox` does not render when running `NEXT_PUBLIC_MOCK_API=true pnpm dev`

cr_req 2

**cc:** @dhmacs 

Closes: #1289 
Closes: #1278


mlahargou (dismissed):
CR 📱 Looks great!

frontend/tests/playwright/product/add_to_cart.spec.ts
```diff
@@ -3,15 +3,14 @@ import { mockedProductQuery } from '@tests/jest/__mocks__/products';
...
-               status: 200,
-               body: '',
-            });
-         }
+   test.beforeEach(async ({ clientRequestHandler, rest }) => {
+      clientRequestHandler.use(
+         rest.post(
+            '/api/2.0/internal/international_store_promotion/buybox',
+            (req, res, ctx) => {
+               return res(ctx.status(200), ctx.json({}));
```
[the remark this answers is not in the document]
mlahargou: One question, why do we do `ctx.json({})` here but for the `notifyWhenSkuInStock` request below we don't add the body?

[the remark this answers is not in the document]
ardelato: Good question! 

I have no clue 😅 , I think I missed the return body when migrating it over.

deploy_block ☁️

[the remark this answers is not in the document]
ardelato: Good question! 

I have no clue 😅 , I think I missed the return body when migrating it over.

deploy_block ☁️

ardelato:
deploy_block ☁️  on CR comment

ardelato:
un_deploy_block ⚡ but dev_block ☁️ . It seems there is some cross over with the `international buybox` mock and the #1278 issue. In addition, there is a bug with the integration of the `browser side` worker when running `NEXT_PUBLIC_MOCK_API=true pnpm dev`

frontend/tests/playwright/product/add_to_cart.spec.ts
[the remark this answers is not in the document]
ardelato: So the `ctx.json({})` was actually a bug, and we should not have returned a body. The reason this continued to work was that it was trying to match against a `post` instead of a `get` and therefore the original request was still sent. 

I didn't think this was an issue until addressing #1278 where I discovered that this request was the reason the international buybox was rendering if my IP was located in Europe.

As such, I made additional changes to properly match against this request and fix #1278

ardelato:
un_dev_block ⚡

frontend/tests/playwright/msw/handlers.ts
```diff
@@ -1,8 +1,19 @@
...
  * rest and graphql requests.
  */
-export const handlers: RequestHandler[] = [];
+export const handlers: RequestHandler[] = [
+   /**
+    * Prepending the wildcard to the path would capture requests regardless of
+    * origin. Therefore, API requests to Cominor will also be matched against.
+    */
+   rest.get(
+      '*/api/2.0/internal/international_store_promotion/buybox',
```
mlahargou, line 14: A test could still override this mock easily if we wanted, right?

frontend/pages/_app.tsx
```diff
@@ -15,13 +16,24 @@ type AppPropsWithLayout<P> = AppProps<P> & {
...
-      setupMocks();
-   });
-}
-
 function MyApp({ Component, pageProps }: AppPropsWithLayout<any>) {
+   const [mswState, setMswState] = useState<'unused' | 'loading' | 'ready'>(
+      'unused'
+   );
+
+   if (process.env.NEXT_PUBLIC_MOCK_API === 'true' && mswState === 'unused') {
```
mlahargou, line 24: Not extremely familiar with this, but didn't you set the state right above? Won't `mswState === 'unused'` always be true here?

mlahargou (approved):
CR 📱 with a minor question deploy_block 📱

dhmacs:
@ardelato I've tested this on my computer and cross sell tests pass 👍 However there are still dozen of tests that are not passing (I don't know if its in the scope of this PR to solve them). I attach here a couple of screenshots for reference:

Screenshot 2023-02-01 at 09 52 08

Screenshot 2023-02-01 at 09 52 38

A ton of them seems to be hitting a timeout

This is running `NEXT_PUBLIC_MOCK_API=true pnpm dev` and `pnpm playwright:run`

dev_block 🔢

ardelato:
> I've tested this on my computer and cross sell tests pass

Awesome!

> However there are still dozen of tests that are not passing (I don't know if its in the scope of this PR to solve them). This is running `NEXT_PUBLIC_MOCK_API=true pnpm dev` and `pnpm playwright:run`

@dhmacs to confirm, did you run both commands at the same time? `NEXT_PUBLIC_MOCK_API=true pnpm dev` and `pnpm playwright:run`

ardelato:
@dhmacs It seems this is an issue on `main` as well. It is probably related to the custom next servers attached to each worker overriding the `.next` directory. 

Timeout Exception

In CI we build a production version of the app that all workers use, which can explain why these timeouts are not present in CI. If we narrow the scope of the tests and workers in Dev then the timeout exception is infrequent. 

Narrowed Testing

--- 

Overall, this issue is unrelated to this pull so I will follow-up on it in a separate pull.

frontend/tests/playwright/msw/handlers.ts
ardelato: Yes, we can either change it for an entire test runtime https://mswjs.io/docs/api/setup-worker/use#permanent-override or we can override the response for a single request using `res.once`
https://mswjs.io/docs/api/setup-worker/use#permanent-override

frontend/package.json
```diff
@@ -113,6 +113,7 @@
       "jest-watch-typeahead": "1.1.0",
       "msw": "0.49.2",
       "next-transpile-modules": "9.0.0",
+      "playwright-msw": "^2.1.0",
```
batbattur, line 116: What versioning should we use for `playwright-msw`? For `playwright-msw`, we have it set to the exact version (`"1.29.1"` without caret).

batbattur (dismissed):
CR 🚛 with minor question (not a blocker)

frontend/package.json
ardelato: Ahh yeah 😓 , I am wondering if we can create a linter to automate this for us or at least throw an error 🤔

ardelato: Addressed in [9bad1f1](https://github.com/iFixit/react-commerce/pull/1319/commits/9bad1f173891203d3a286cbf49a2cf7581013d7f)

batbattur (approved):
CR 🚛

mlahargou (approved):
CR 📱

ardelato:
QA ⚡ 

## Steps Taken:
1. Set a VPN to be located in Italy
2. Whitelisted the IP address for Cominor
3. Confirmed running just `pnpm dev` would render the international buybox
Italy No Mocking
4. Confirmed running `NEXT_PUBLIC_MOCK_API=true pnpm dev` enabled `MSW` and mocked the `buybox` API request
Italy Mocking Enabld
5. I confirmed that running the Playwright tests still passed

ardelato:
un_deploy_block 🦓 comments have been addressed https://github.com/iFixit/react-commerce/pull/1319#pullrequestreview-1278270872

Issue #1101, opened 2022-12-02
Images on list pages are squished.


k0rvusk0r4x:
The product images on product list pages (list and grid view) are slightly squished. Note that when I say "slightly" I mean I had to magnify the images quite a bit and compare to the image on the product page before I was able to see the subtle differences. 

Among the easier products to see the squish on is https://www.ifixit.com/products/anti-clamp, which is on this list, https://www.ifixit.com/Tools/Gripping?p=3. Again, note, easier, not easy. Looking at the logo, you can see it's not quite round on the list page.

This was brought to us by Carsten Frauenheim via the Slack bug report.

danielbeardsley:
Hmm.. I'm not seeing that. It's exactly square at normal browser widths:
![image](https://user-images.githubusercontent.com/26855/206781892-92c0fb84-6857-4871-9c77-a3a0cb97e69a.png)


A screenshot of the problem would be really helpful.

federicobadini:
I'm going to close this since I cannot find any wrong aspect-ratio on product-list
I spotted a 2px aspect-ratio error on the product page though. I'm going to open a PR for that

PR #1321, opened 2023-02-01
Correct product image aspect ratio

federicobadini:
connects #1101 

Product page images were not perfectly squared because the image height was computed as 100% + the 2px border.
This PR changes the way in which the gallery border is rendered to eliminate the problem.

Screenshot 2023-02-01 at 18 04 41

### QA

1. Visit Vercel preview
2. Verify that the gallery image aspect ratio is now exactly 1:1
3. No interaction (changing slide, zoom, etc) with the carousel should be compromised


masonmcelvain (approved):
CR ✌🏻

PR #1324, opened 2023-02-02
Add error message in cache handler

federicobadini:
This PR adds an error log in the catch of the `withCache` handler to have a clearer view of where [these errors](https://sentry.io/organizations/ifixit/issues/3908459644/?project=6469069&query=is%3Aunresolved&referrer=issue-stream) are originating from.


masonmcelvain (approved):
CR ✌🏻

masonmcelvain:
I don't think it's necessary to QA this, as this only adds a debugging log.

PR #1309, opened 2023-01-28
Revert "Display 4 rows devices on selected pages frontend"

jarstelfox:
Reverts iFixit/react-commerce#1158

See: https://ifixit.slack.com/archives/C0252FETV2S/p1674863795570009

Example log:

```
2023-01-28T00:09:49.116Z	20fecd52-a6c3-4ae6-83c8-2f18c1301edf	ERROR	GraphQL query failed with errors:
2023-01-28T00:09:49.116Z	20fecd52-a6c3-4ae6-83c8-2f18c1301edf	ERROR		[GRAPHQL_VALIDATION_FAILED] Cannot query field "defaultShowAllChildrenOnLgSizes" on type "ProductList".
strapi.getProductList: 56ms
findProductList: 58ms
server_side_props: 59ms
2023-01-28T00:09:49.261Z	20fecd52-a6c3-4ae6-83c8-2f18c1301edf	ERROR	Error: GraphQL query failed to execute: Bad Request
    at requester (/var/task/frontend/.next/server/chunks/786.js:1586:11)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Promise.all (index 0)
    at async findProductList (/var/task/frontend/.next/server/chunks/393.js:257:34)
    at async /var/task/frontend/.next/server/chunks/932.js:5277:35
    at async /var/task/node_modules/.pnpm/@sentry+nextjs@7.29.0_5cy3dema5qwip6n7exqk4sqtji/node_modules/@sentry/nextjs/cjs/config/wrappers/wrapperUtils.js:39:14
    at async Object.renderToHTML (/var/task/node_modules/.pnpm/next@12.2.3_t7ss3ubh4wigfvyfclbvqff3gm/node_modules/next/dist/server/render.js:573:20)
    at async doRender (/var/task/node_modules/.pnpm/next@12.2.3_t7ss3ubh4wigfvyfclbvqff3gm/node_modules/next/dist/server/base-server.js:915:38)
    at async cacheEntry.responseCache.get.isManualRevalidate.isManualRevalidate (/var/task/node_modules/.pnpm/next@12.2.3_t7ss3ubh4wigfvyfclbvqff3gm/node_modules/next/dist/server/base-server.js:1020:28)
    at async /var/task/node_modules/.pnpm/next@12.2.3_t7ss3ubh4wigfvyfclbvqff3gm/node_modules/next/dist/server/response-cache.js:69:36 {
  page: '/Parts/[...deviceHandleItemType]'
}
[...]

PR #1327, opened 2023-02-03
Revert "Revert "Display 4 rows devices on selected pages frontend""

sterlinghirsh:
Reverts iFixit/react-commerce#1309

We updated Strapi so this should be good to go.


sterlinghirsh:
This is failing the vercel deploy because of what I think is a bug in govinor. https://revert-1309-revert-1158-display-4-rows-devices-on-selected-pages-frontend.govinor.com/admin doesn't resolve. However, the prod preview works, so I think this is QA :zap:

andyg0808 (approved):
CR :+1:

PR #1329, opened 2023-02-03
Breadcrumbs: Copy from iFixit/ifixit

andyg0808:
This pull copies the breadcrumbs from the iFixit repo and modifies them to work here.

The iFixit/ifixit breadcrumbs potentially have a nicer small-screen behavior than the existing ones here.

Connects https://github.com/iFixit/ifixit/issues/46349

qa_req 0

### DIFF
--- a/packages/breadcrumbs/Breadcrumbs.tsx
+++ b/packages/breadcrumbs/Breadcrumbs.tsx
@@ -146,7 +146,7 @@ const IFixitBreadcrumb = memo(function IFixitBreadcrumb({
    );
 }, isEqual);
 
-const IFixitBreadcrumbItem = memo(function IFixitBreadcrumbItem({
+const IFixitBreadcrumbItem = function IFixitBreadcrumbItem({
    url,
    label: name,
    isLast,
@@ -172,10 +172,9 @@ const IFixitBreadcrumbItem = memo(function IFixitBreadcrumbItem({
          </Text>
       </BreadcrumbItem>
    );
-},
-isEqual);
+};
 
-const IFixitCollapsedBreadcrumb = memo(function IFixitCollapsedBreadcrumb({
+const IFixitCollapsedBreadcrumb = function IFixitCollapsedBreadcrumb({
    breadCrumbs,
    breadcrumbIcon,
 }: Pick<BreadCrumbsProps, 'breadCrumbs' | 'breadcrumbIcon'>) {
@@ -208,8 +207,7 @@ const IFixitCollapsedBreadcrumb = memo(function IFixitCollapsedBreadcrumb({
          {breadcrumbIcon}
       </>
    );
-},
-isEqual);
+};
 
 function IFixitCollapsedBreadcrumbItem({
    url,
--- a/packages/breadcrumbs/Breadcrumbs.tsx
+++ b/packages/breadcrumbs/Breadcrumbs.tsx
@@ -1,0 +1,211 @@
+import {
+   Flex,
+   Link,
+   Breadcrumb,
+   BreadcrumbItem,
+   Menu,
+   MenuButton,
+   MenuList,
+   MenuItem,
+   MenuItemProps,
+   IconButton,
+   Text,
+   BreadcrumbItemProps,
+   LinkProps,
+} from '@chakra-ui/react';
+import { FaIcon } from '@ifixit/icons';
+import { faEllipsis } from '@fortawesome/pro-regular-svg-icons';
+import { faChevronRight } from '@fortawesome/pro-solid-svg-icons';
+import { memo, useState, useCallback } from 'react';
+import isEqual from 'lodash/isEqual';
+import { HiddenWrapEffect, useHiddenWrap } from './FlexHiddenWrap';
+
+export interface BreadcrumbData {
+   name: string;
+   url: string;
+}
+
+export interface BreadCrumbsProps {
+   breadCrumbs: BreadcrumbData[];
+   breadcrumbsToShow?: number;
+   breadcrumbIcon?: JSX.Element;
+   usePlaceHolder?: boolean;
+}
+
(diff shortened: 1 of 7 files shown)


dhmacs:
dev_block 🛑  @andyg0808 can we put the component inside `@ifixit/ui` package instead of creating a specific package for breadcrumbs?

andyg0808:
@dhmacs I can if we really want to, but we're expecting to immediately start importing the breadcrumbs into iFixit, and I figured more packages were better, because it reduces the set of dependencies each one has, which makes it easier to avoid dependency issues when importing into iFixit/ifixit.

Thoughts?

dhmacs:
@andyg0808 I see your point, but this way will end up creating a ton of packages, it seems a lot of overhead 🤔 What needs to change on `@ifixit/ui` so that you're able to import it on `ifixit/ifixit` ?

I've also just noticed the bundle size impact. Do you why this is adding 7kb of JS to the product page, given that it's supposed to be a refactor?

andyg0808:
@dhmacs It's not so much that `@ifixit/ui` might need to change as that, once we import it, we're at risk of breaking from any other incompatible dependencies which might get added to `@ifixit/ui` down the road.

Re. the bundle size impact: I think this isn't a refactor as much as a new, different implementation. I'll look into why the bundle size has grown that much, or rather, what's making up that 7kb of JS it's adding.
deploy_block :+1: on doing that.

andyg0808:
Another blocker: I think these breadcrumbs have slightly different behavior than the previous ones; I think we should have a discussion about what we want.

packages/breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,232 @@
...
+   MenuList,
+   MenuItem,
+   MenuItemProps,
+   IconButton,
+   Text,
+   BreadcrumbItemProps,
+   LinkProps,
+} from '@chakra-ui/react';
+import { FaIcon } from '@ifixit/icons';
+import { faChevronRight, faEllipsis } from '@fortawesome/pro-solid-svg-icons';
```
jarstelfox, line 17: @andyg0808 I suspect the main bundle size growth is that these are not deep imports

jarstelfox (approved):
CR 🌵

packages/breadcrumbs/Breadcrumbs.tsx
andyg0808: Oooh, interesting point. Trying deep imports in 41cbaeff8d6857be14726947e1a92002b8d076c4.

packages/breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,233 @@
...
+   IconButton,
+   Text,
+   BreadcrumbItemProps,
+   LinkProps,
+} from '@chakra-ui/react';
+import { FaIcon } from '@ifixit/icons';
+import { faChevronRight } from '@fortawesome/pro-solid-svg-icons/faChevronRight';
+import { faEllipsis } from '@fortawesome/pro-solid-svg-icons/faEllipsis';
+import { memo, useState, useCallback } from 'react';
+import isEqual from 'lodash/isEqual';
```
jarstelfox, line 20: This could also be a culprit for the size increase. It's likely this can be dropped.

I was using it to deep equals check props for memoization. This is left over from when I was doing browser width rendering.

jarstelfox (approved):
CR 🌵

packages/breadcrumbs/Breadcrumbs.tsx
andyg0808: I checked it, and it didn't seem like it was costing much in size.

packages/breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,233 @@
...
+export type BreadcrumbItem = {
+   label: string;
+   url: string;
+};
+
+export interface BreadCrumbsProps {
+   breadCrumbs: BreadcrumbItem[];
+   breadcrumbsToShow?: number;
+   breadcrumbIcon?: JSX.Element;
+   usePlaceHolder?: boolean;
```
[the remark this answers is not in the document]
andyg0808: Fixed in 7e46cd3497ad2cbcd41df657762b05c6f8985483

andyg0808:
Sorry for the force push; I didn't think this through entirely. The main change should be that I stopped deleting the existing breadcrumbs implementation, which means we no longer have to get the new implementation compatible with the existing one before we can merge this.

jarstelfox (dismissed):
CR 🌵

packages/breadcrumbs/Breadcrumbs.tsx
sctice-ifixit: Did it make a difference? If so, I think we might have reached the point at which some anti-pattern tests are worth adding.

andyg0808: Not as far as I can tell. Interestingly, not using the new code at all doesn't seem to help either, so I'm not sure what's going on.

andyg0808:
un_dev_block :+1: 
I recognize that, if we continue doing this, we could make a lot of packages. I'm okay with that for now, because we can always merge it into `@ifixit/ui` later if we want and add a stub in its current position until we get the monorepo updated to reflect that.

jarstelfox (dismissed):
CR 🌵

andyg0808:
dev_block :+1: on copying a fix over from the source repo.

andyg0808:
un_dev_block :+1:

jarstelfox (approved):
CR 🌵

andyg0808:
un_deploy_block :+1: Dropped the use of the breadcrumbs, which also drops the extra 7kb of JS that was getting added.

Issue #1323, opened 2023-02-01
Playwright: Resolve Timeouts in Dev


ardelato:
## Description

If we run every test for Playwright we will get a bunch of `timeout` failures in Dev. This is probably due to our custom next servers using the same `.next` directory as the main dev server, which causes files to be overwritten and no longer referencable for our main dev server.

### To Do:

Resolve the `timeout` failures when running every test in Dev.       

We can possibly change the cache directory for each custom next server so there are no collisions or overwrites.
Alternatively, we could ensure all the custom next servers are booted up first before running tests to ensure the servers have all been built properly.          

_Originally posted by @dhmacs in https://github.com/iFixit/react-commerce/issues/1319#issuecomment-1411687291_

PR #1332, opened 2023-02-03
Playwright: Mitigate Timeouts

ardelato:
## Description

If we are running all the Playwright tests locally then the tests will often timeout. There are two main reasons this occurs.

Currently, we have no set limit to the number of workers we will use locally. Therefore, we could potentially have 4-8 workers running in parallel, all depending on the number of cores on the host machine. In CI, we use all available workers we can and since the GitHub action machine only has two cores, we really only use 2 workers in CI. To reduce the load on local runs, I've changed the number of workers to a quarter of available cores.

The other reason for the timeout is each worker will create its own custom Nextjs server but each server will use the same dist directory (i.e. `.next`). Therefore, while one instance is building files, another instance may be overwriting these files. As a result of this overlap, some files may not be present when trying to access a page which will cause a timeout. To resolve this, each worker will build to a different directory than the default `.next`.

### CR

In order to override the default dist folder for each custom server we need to set an environment variable because we cannot simply import the config file and override the `distDir` property. The `next.config.js` file is a regular Node.js module and not a JSON file.[^1] Then by using Playwright's `worker` context[^2] we can create a unique dist folder for each worker.

Lastly, this does not completely resolve timeouts. We do have some flaky tests that may fail so if we do want to run the entire Playwright test-suite then we should increment the number of retries. 

### QA

- [ ] Confirm running tests locally will not timeout
	```shell
	pnpm playwright:run --project="Desktop Chrome" --retries=2
	```

Closes: #1323 

[^1]: [next.config.js](https://nextjs.org/docs/api-reference/next.config.js/introduction#:~:text=next.config.js%20is%20a%20regular%20Node.js%20module%2C%20not%20a%20JSON%20file.%20It%20gets%20used%20by%20the%20Next.js%20server%20and%20build%20phases%2C%20and%20it%27s%20not%20included%20in%20the%20browser%20build.)
[^2]: [Playwright Worker Info](https://playwright.dev/docs/api/class-workerinfo#worker-info-worker-index)

### DIFF
--- a/frontend/tests/playwright/test-fixtures.ts
+++ b/frontend/tests/playwright/test-fixtures.ts
@@ -30,7 +30,9 @@ export const test = base.extend<
    customServer: [
       // eslint-disable-next-line no-empty-pattern
       async ({}, use, workerInfo) => {
-         process.env.NEXT_DIST_DIR = `./.next-worker-${workerInfo.workerIndex}`;
+         process.env.NEXT_DIST_DIR = process.env.CI
+            ? './.next'
+            : `./.next-worker-${workerInfo.workerIndex}`;
          const props: CustomNextjsServer = await Server();
          await use(props);
          // Clean up the dist directory after the worker is done.
--- a/frontend/tests/playwright/test-fixtures.ts
+++ b/frontend/tests/playwright/test-fixtures.ts
@@ -28,7 +28,8 @@ export const test = base.extend<
 >({
    customServer: [
       // eslint-disable-next-line no-empty-pattern
-      async ({}, use) => {
+      async ({}, use, workerInfo) => {
+         process.env.NEXT_DIST_DIR = `./.next-worker-${workerInfo.workerIndex}`;
          const props: CustomNextjsServer = await Server();
          await use(props);
       },
--- a/frontend/tests/playwright/test-fixtures.ts
+++ b/frontend/tests/playwright/test-fixtures.ts
@@ -20,6 +20,7 @@ import { graphql, rest } from 'msw';
 import type { ProductFixtures, CustomNextjsServer } from './fixtures';
 import { ProductPage, CartDrawer, Server } from './fixtures';
 import { format } from 'util';
+import { exec } from 'node:child_process';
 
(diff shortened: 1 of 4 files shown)


andyg0808 (approved):
CR :+1:

andyg0808 (approved):
CR :+1:

deltuh-vee:
The collections_scroll test is still timing out, but the fix_kit test is now passing.
CleanShot 2023-02-06 at 14 28 05

deltuh-vee:
QA 🎬 
.next folders are created for individual workers and are deleted after tests finish.

frontend/tests/playwright/test-fixtures.ts
```diff
@@ -35,9 +36,14 @@ export const test = base.extend<
...
       // eslint-disable-next-line no-empty-pattern
-      async ({}, use) => {
+      async ({}, use, workerInfo) => {
+         process.env.NEXT_DIST_DIR = process.env.CI
+            ? './.next'
+            : `./.next-worker-${workerInfo.workerIndex}`;
          const props: CustomNextjsServer = await Server();
          await use(props);
+         // Clean up the dist directory after the worker is done.
+         exec('rm -rf .next-worker-*');
```
danielbeardsley, line 46: I would think we'd want / need an await here as this is an async process.

ardelato: Although `exec` is an `async` process, we are not doing anything with the return value of `exec` nor are we running dependent logic after it. Therefore, the `await` would have no effect.

Await Not Needed

andyg0808: I'd double-check whether the test runner could potentially quit before this finishes running, or whether it'll block properly for this. Could change the command to a `sleep` possibly to just force it to wait.

danielbeardsley: Andrew captured what I was referring to but it seems like we're safe here as node will wait on outstanding invocations (of `exec` at least):
image

PR #1333, opened 2023-02-03
Add `bundle-stats` plugin

andyg0808:
This is the other plugin we run on iFixit/ifixit; I wanted to see if it would provide better information for figuring out what's going on with a build. It wasn't able to answer the question I had, but I figure it seems worth keeping the set up work I did.

qa_req 0
This is dev-only, and we're not that concerned if it doesn't work.

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -22,11 +22,15 @@ const withTM = require('next-transpile-modules')([
 
 const { withSentryConfig } = require('@sentry/nextjs');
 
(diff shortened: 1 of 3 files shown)


sterlinghirsh (approved):
CR :zap: Seems a little odd to have two bundle analysis / stats plugins, but I'm willing to try it out.

andyg0808:
I believe the bundle-stats report ends up in the `dist` directory, IIRC.

PR #1339, opened 2023-02-07
Cache: use null adapter while redis isn't ready

danielbeardsley:
We seem to be getting about [~250 errors per day](https://app.datadoghq.com/logs?query=source%3Avercel%20service%3Areact-commerce-prod.vercel.app%20%22Stream%20isn%27t%20writeable%22&agg_m=count&agg_q=%40path&agg_t=count&cols=host%2Cservice&index=&messageDisplay=inline&stream_sort=time%2Cdesc&top_n=5&top_o=top&viz=timeseries&from_ts=1675155609590&to_ts=1675760409590&live=true) that say the redis connection isn't a writable stream. These should be able to be caught and dealt with by returning as a miss, but they seem to be aborting the request.

Redis is communicated with via a persistent connection, so here we change caching to dynamically choose the backend based on the connection state. If it's anything other than "ready", we use the null adapter (everything looks like a miss).

Possible connection states from Typescript:
```typescript
type RedisStatus = "wait"
  | "reconnecting"
  | "connecting"
  | "connect"
  | "ready"
  | "close"
  | "end"
```

Looking at the client code, checking for `ready` seems like the right choice. I've also upped the connection timeout now that we have a fallback while connecting and that connecting won't delay any requests.

Connect #1322

### DIFF
--- a/frontend/lib/swr-cache/adapters/index.ts
+++ b/frontend/lib/swr-cache/adapters/index.ts
@@ -8,7 +8,8 @@ export const getCache = () => {
    if (REDIS_URL) {
       const redisClient = getRedisClient(REDIS_URL);
       const redisCacheAdapter = redisAdapter(redisClient);
-      return () => redisClient.status == "ready" ? redisCacheAdapter : nullCacheAdapter;
+      return () =>
+         redisClient.status == 'ready' ? redisCacheAdapter : nullCacheAdapter;
    } else {
       return () => nullCacheAdapter;
    }
--- a/frontend/lib/swr-cache/adapters/index.ts
+++ b/frontend/lib/swr-cache/adapters/index.ts
@@ -4,12 +4,19 @@ import { REDIS_URL } from '@config/env';
 import { nullAdapter } from './null-adapter';
 
 export const getCache = () => {
-   if (REDIS_URL === undefined) {
-      return nullAdapter();
+   const nullCacheAdapter = nullAdapter();
+   if (REDIS_URL) {
+      const redisClient = getRedisClient(REDIS_URL);
+      const redisCacheAdapter = redisAdapter(redisClient);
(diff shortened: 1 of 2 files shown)


dhmacs (changes requested):
It looks like they're all background revalidation requests to `api/nextjs/cache/product`, so they don't affect the user and it doesn't help to use the null adapter since the goal of these requests is to revalidate the Redis cache

deploy_block 🛑  let's take some time to inspect the error and find a fix

danielbeardsley:
> It looks like they're all background revalidation requests to `api/nextjs/cache/product`

Sorry, I had the query wrong, I've updated the link to show that it's happening on user requests too, but we catch those errors and treat them as misses correctly so they aren't 500s.

I've tried to look for a cause, but I haven't found anything obvious. I have noticed that they come in short 1-2 second bursts separated by many minutes of no errors. "Stream isn't writable" means the underlying TCP connection to redis is down for the moment, so attempting requests won't do anything. I do want this condition to be somewhat visible because I want to know if it gets worse or better. That said, I don't expect to be able to fix occasional connection interruptions with a 3rd party service.

I can try reducing our reconnection delay, perhaps we can recover more quickly, I'll open a pull.

Here's an example of the timestamps
image

erinemay:
@danielbeardsley thoughts on testing this?

danielbeardsley:
I think we should wait on the QA till I fix the conflicts.

Testing: run it locally with the REDIS_URL set to something that won't work: `redis://example.com` or some such. You shouldn't see any errors when accessing product pages. You can throw some log statements into the null adapter to ensure it's being used.

dev_block :+1: on fixing conflicts.

danielbeardsley:
This setting now just returns errors (effectively misses) when it can't connect to redis so this is no longer needed:
https://github.com/iFixit/react-commerce/blob/af589472e8106052f44f49f1d6907f8357062a1d/frontend/lib/redis/index.ts#LL18C12-L18C12

PR #1341, opened 2023-02-07
Redis: reconnect sooner

danielbeardsley:
We're seeing errors talking to redis on occasion; short several-second
bursts of errors, totalling ~600 per day.

It's a 3rd party service, interruptions happen, let's try reconnecting
more quickly. Delays look like this:

58ms
66ms
82ms
114ms
178ms
306ms
...

with a cap at 10 minutes (as before)

qa_req 0

Connect #1322

CC @dhmacs

### DIFF
--- a/frontend/lib/swr-cache/adapters/index.ts
+++ b/frontend/lib/swr-cache/adapters/index.ts
@@ -11,7 +11,10 @@ export const getCache = () => {
    const client = new Redis(REDIS_URL, {
       connectTimeout: 500,
       // Retry, connect every once in a while as cache misses / failures are OK
-      retryStrategy: (times) => Math.min(times * 5000, 10 * 60 * 1000),
+      retryStrategy: (times) => {
+         times = Math.min(20, times);
+         return Math.min(50 + 2 ^ (times + 3), 10 * 60 * 1000);
+      },
       maxRetriesPerRequest: 0,
       // Always try re-connecting since one instance is shared across the whole
       // process
--- a/frontend/lib/swr-cache/adapters/index.ts
+++ b/frontend/lib/swr-cache/adapters/index.ts
@@ -11,10 +11,7 @@ export const getCache = () => {
    const client = new Redis(REDIS_URL, {
       connectTimeout: 500,
       // Retry, connect every once in a while as cache misses / failures are OK
-      retryStrategy: (times) => {
-         times = Math.min(20, times);
-         return Math.min(50 + 2 ** (times + 3), 2 * 60 * 1000);
-      },
+      retryStrategy: (times) => Math.min(50 + 2 ** (times + 3), 2 * 60 * 1000),
       maxRetriesPerRequest: 0,
       // Always try re-connecting since one instance is shared across the whole
       // process
--- a/frontend/lib/swr-cache/adapters/index.ts
+++ b/frontend/lib/swr-cache/adapters/index.ts
@@ -13,7 +13,7 @@ export const getCache = () => {
       // Retry, connect every once in a while as cache misses / failures are OK
       retryStrategy: (times) => {
          times = Math.min(20, times);
-         return Math.min((50 + 2) ^ (times + 3), 10 * 60 * 1000);
+         return Math.min((50 + 2) ** (times + 3), 10 * 60 * 1000);
       },
       maxRetriesPerRequest: 0,
       // Always try re-connecting since one instance is shared across the whole
--- a/frontend/lib/swr-cache/adapters/index.ts
+++ b/frontend/lib/swr-cache/adapters/index.ts
@@ -13,7 +13,7 @@ export const getCache = () => {
       // Retry, connect every once in a while as cache misses / failures are OK
       retryStrategy: (times) => {
          times = Math.min(20, times);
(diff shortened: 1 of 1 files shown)


frontend/lib/swr-cache/adapters/index.ts
```diff
@@ -11,7 +11,10 @@ export const getCache = () => {
    const client = new Redis(REDIS_URL, {
       connectTimeout: 500,
       // Retry, connect every once in a while as cache misses / failures are OK
-      retryStrategy: (times) => Math.min(times * 5000, 10 * 60 * 1000),
+      retryStrategy: (times) => {
+         times = Math.min(20, times);
+         return Math.min((50 + 2) ^ (times + 3), 10 * 60 * 1000);
```
sterlinghirsh, line 16: ^ is the xor operator not exponentiation

sterlinghirsh:
CR :zap: dev_block :zap:

frontend/lib/swr-cache/adapters/index.ts
danielbeardsley: Well shoot.
```suggestion
         return Math.min((50 + 2) ** (times + 3), 10 * 60 * 1000);
```

danielbeardsley:
Derp, un_dev_block :+1:

frontend/lib/swr-cache/adapters/index.ts
```diff
@@ -11,7 +11,10 @@ export const getCache = () => {
    const client = new Redis(REDIS_URL, {
       connectTimeout: 500,
       // Retry, connect every once in a while as cache misses / failures are OK
-      retryStrategy: (times) => Math.min(times * 5000, 10 * 60 * 1000),
+      retryStrategy: (times) => {
+         times = Math.min(20, times);
+         return Math.min((50 + 2) ** (times + 3), 10 * 60 * 1000);
```
sterlinghirsh, line 16: What is this backoff supposed to be? 52**3 = 140608 which is over 2 mins if times == 0. If times == 1 it's 7311616 which is already over the 10 min cap. I think we should probably max out at 1 minute, maybe even 30 sec. I'm not really worried about spamming redis with connections if they're not going through. We should probably have this retry in under 1 s (maybe 100ms) for the first failure and then if we care, we can back it off from there.

sterlinghirsh:
CR :zap: dev_block :zap:

frontend/lib/swr-cache/adapters/index.ts
danielbeardsley: Well crap, I could have sworn that's not what I typed, or at least not what I typed at first...

Well what do you know... prettier screwed it up: https://github.com/iFixit/react-commerce/pull/1341/commits/9f50df6653c618e507478fe402222675159aba5e

Back when it was XOR, it helpful re-inforced the natural evaluation order for bitwise operators I guess... or something. Fixing. Sheesh.

danielbeardsley:
```suggestion
         return Math.min(50 + (2 ** (times + 3)), 2 * 60 * 1000);
```

danielbeardsley:
un_dev_block :+1:  :facepalm:

frontend/lib/swr-cache/adapters/index.ts
```diff
@@ -11,7 +11,10 @@ export const getCache = () => {
    const client = new Redis(REDIS_URL, {
       connectTimeout: 500,
       // Retry, connect every once in a while as cache misses / failures are OK
-      retryStrategy: (times) => Math.min(times * 5000, 10 * 60 * 1000),
+      retryStrategy: (times) => {
+         times = Math.min(20, times);
+         return Math.min(50 + 2 ** (times + 3), 2 * 60 * 1000);
+      },
```
sterlinghirsh, line 17:
```suggestion
      retryStrategy: (times) => Math.min(50 + 2 ** (times + 3), 2 * 60 * 1000),
```

sterlinghirsh (approved):
CR :zap: Simplified it a bit.

frontend/lib/swr-cache/adapters/index.ts
```diff
@@ -11,10 +11,7 @@ export const getCache = () => {
    const client = new Redis(REDIS_URL, {
       connectTimeout: 500,
       // Retry, connect every once in a while as cache misses / failures are OK
-      retryStrategy: (times) => {
-         times = Math.min(20, times);
-         return Math.min(50 + 2 ** (times + 3), 2 * 60 * 1000);
-      },
+      retryStrategy: (times) => Math.min(50 + 2 ** (times + 3), 2 * 60 * 1000),
```
danielbeardsley, line 14: I did the double-min because I figured if `times` got into the hundreds / thousands it might crash.

But I was wrong... looks like it just results in `Infinity`:

image

Which seems to work fine with `Math.min()`

sterlinghirsh: I had that "oh THAT'S why he did that!" thought immediately after committing it and checked it, but luckily it worked anyway.

danielbeardsley:
This addressed the vast majority of failures:
image

Note: these failures only result in slower responses (effectively cache misses), not 500s. So the remaining trickle is under the threshold of caring

PR #1342, opened 2023-02-07
React-commerce: Footer updates

aburke07:
## Overview
This pull updates the footer to match [the figma](https://www.figma.com/file/sfj6r7Bn6eREMvuqo0f7Du/iFixit---Components?node-id=401%3A18120&t=4WQlzhHuGSzzoNe5-0). Most of the components exactly match the iFixit footer components, so they are copied over from iFixit/ifixit. In another pull, we can go back and import the duplicate components (Link lists, social media section, and partners) into iFixit/ifixit and remove the duplicate code there. 

## CR
There are a lot of overlapping commits, so it might be easier to just look at the file diff.

## QA
Make sure the footer matches [the figma](https://www.figma.com/file/sfj6r7Bn6eREMvuqo0f7Du/iFixit---Components?node-id=401%3A18120&t=4WQlzhHuGSzzoNe5-0) and behaves as expected.

Connects to: https://github.com/iFixit/ifixit/issues/45964


danielbeardsley:
I think we should look into the bundle size changes. +10% more javascript seems like too much to not look into why.

dev_block :+1:  on the 10% across the board increase in JS.

frontend/layouts/default/Newsletter.tsx
```diff
@@ -1,86 +1,92 @@
...
-   InputProps,
    Stack,
-   StackProps,
    Text,
-   TextProps,
+   Flex,
+   Link,
 } from '@chakra-ui/react';
 import { faCircleCheck } from '@fortawesome/pro-solid-svg-icons';
+import { faEye } from '@fortawesome/pro-solid-svg-icons/faEye';
```
danielbeardsley, line 13: Hmm.. is the large JS increase because of these two different import styles? Hard to imagine... but it's an easy experiment to combine them and see (perhaps going with whatever is more common in this repo)

aburke07: I combined these in [0cc257f](https://github.com/iFixit/react-commerce/pull/1342/commits/0cc257fd80d9c7742356d33f63dffa7d2178cff7), and the bundle stats report increased from `+11 B` to `+307 B` on the global bundle size, so it seems like importing these separately is better, even though we have combined imports everywhere else in this repo. 

Correct me if I'm wrong, but isn't `+11 B` (bytes) extremely tiny in comparison to the whole build? Where are you getting the +10% increase from?

danielbeardsley: Yes, +300b is tiny, however, right below that in the same comment we have stats for each page:
image

aburke07: I ran the webpack analysis on my branch and main and they ended up being very similar:
**On my branch:**
Screen Shot 2023-02-17 at 10 59 15 AM

**On Main:**
Screen Shot 2023-02-17 at 10 56 39 AM


The only difference I can really see is the increase in sdk.ts in 647.js. That seems like something that could touch all of the above pages.

Is there a way to re-run the bundle analysis from CI? It didn't seem to update after [d09523d](https://github.com/iFixit/react-commerce/pull/1342/commits/d09523df726209f846f29f9d7f814836cccdbe76).

danielbeardsley: > Is there a way to re-run the bundle analysis from CI? It didn't seem to update after

Looks like it updated now:
image

It looks like your analysis is comparing the *server bundles* and not the client bundles (I think).

deltuh-vee: @aburke07 I reran the Next.js Bundle Analysis. Here's where I found the job to rerun it on this branch: https://github.com/iFixit/react-commerce/actions/workflows/nextjs_bundle_analysis.yml

danielbeardsley: Hmm...  I wonder if it has to do with there being conflicts on this branch.

packages/footer/components/LinkData.tsx
```diff
@@ -0,0 +1,191 @@
+import { FooterType } from './Shared';
```
danielbeardsley, line 1: I'm not 100% sure, but if these are types, we should be doing `import type { FooterType } ...`

This applies to all imports. I don't think it matters functionally, but it may be what's inflating the bundle size on this branch?

packages/footer/components/Navigation.tsx
```diff
@@ -10,9 +10,90 @@ import {
    ListProps,
    SimpleGrid,
    SimpleGridProps,
+   Text,
 } from '@chakra-ui/react';
+import { NewsletterComponent, NewsletterFormProps } from './Newsletter';
+import { Menu as MenuType, MenuItem } from '../../ui/menu/menu';
```
danielbeardsley, line 16: Oh... this is possibly a problem. I think we should be using the `@ifixit/ui` alias here instead of a relative path that goes *outside* this package directory.

packages/footer/components/Newsletter.tsx
```diff
@@ -0,0 +1,239 @@
...
+   Link,
+} from '@chakra-ui/react';
+import { faCircleCheck, faEye } from '@fortawesome/pro-solid-svg-icons';
+import {
+   Subscription,
+   SubscriptionStatus,
+   useSubscribeToNewsletter,
+} from '../../newsletter-sdk';
+import React from 'react';
+import { FaIcon } from '../../icons';
```
danielbeardsley, line 20: Same here... I think we should be using the package name: `@ifixit/icons` instead of a relative path that reaches *outside* the package directory.

aburke07:
> I looked at the timing of the changes that produced the positive changes in bundle size and it was the commit that used `import type` in a bunch of places.
> 
> [7157e77](https://github.com/iFixit/react-commerce/commit/7157e775c203f00340699b9a4254d4614a6b4d6c)
> 
> So I bet there's another couple of those that's still inflating the bundle size. Also could be the import problems that I mentioned above.

Yep. I'm doing the changes separately so we can see what actually fixed the bundle size increase. We are headed in the correct direction so far!

federicobadini:
dev_block 🙋‍♂️ on a couple of observations

I’ve looked at the PR and I think that it would be easier if we structure the content like in [this codesandbox](https://q699yx.csb.app/). This should allow for instance to avoid using negative margins and the structure looks simpler to me. Let me know what you think!

I’ve also noticed that there are differences in font sizes: the mockup seems to use smaller fonts

frontend/tests/jest/__mocks__/products.ts
```diff
@@ -1,5 +1,5 @@
 import { CurrencyCode, FindProductQuery } from '@lib/shopify-storefront-sdk';
-import { MenuItemType } from '@models/menu';
+import { MenuItemType } from '@ifixit/ui';
```
danielbeardsley, line 2:
```suggestion
import type { MenuItemType } from '@ifixit/ui';
```

frontend/models/store.ts
```diff
@@ -9,7 +9,7 @@ import {
    MenuItemType,
    ProductListLinkMenuItem,
    SubmenuMenuItem,
-} from './menu';
+} from '@ifixit/ui';
```
danielbeardsley, line 12: I think these are all types now... `import type ...`

packages/ui/menu/index.ts
```diff
@@ -0,0 +1 @@
+export * from './menu';
```
danielbeardsley, line 1: I don't think we need this file, I don't see anyone importing from `@ifixit/ui/menu`

frontend/tests/jest/__mocks__/products.ts
aburke07: Surprisingly,  `MenuItemType` is not a type. It is an enum. So, the compiler complains about importing this as `type`.

frontend/models/store.ts
aburke07: Stated above, MenuItemType is not a type. It is an enum. So, the compiler complains about importing this as type. Should I break out the types and import those separately from the same path?

packages/ui/menu/index.ts
aburke07: I'll test it without this file, but I believe we do need it. `/packages/ui/index.tsx` has `export * from './menu';` which points to `packages/ui/menu/index.ts`. So I think it is a necessary intermediary export.

frontend/models/store.ts
danielbeardsley: Yes, I'd try splitting it up into two statements (types and non types).

packages/footer/components/Newsletter.tsx
```diff
@@ -0,0 +1,239 @@
...
+         flexDirection="row"
+         alignItems="center"
+         padding="12px"
+         gap="8px"
+         w="100%"
+         backgroundColor="#EDF6FF"
+         border="1px"
+         borderColor="#77B5FF"
+         borderStyle="solid"
+         borderRadius="4px"
```
danielbeardsley, line 151: I'm a little surprised by all the explicit numbers and colors here, I would have expected to depend more on our chakra theme (and it's colors and spacing numbers).

packages/footer/components/Newsletter.tsx
```diff
@@ -0,0 +1,239 @@
...
+   newsletterForm,
+}: {
+   newsletterForm: NewsletterFormProps;
+}) {
+   const ref = React.useRef(null);
+   const inputRef = React.useRef<HTMLInputElement>(null);
+   const [subscription, subscribe] = useSubscribeToNewsletter();
+
+   const onSubscribe = React.useCallback(
+      async (event: React.FormEvent<HTMLElement>) => {
```
danielbeardsley, line 189: This doesn't use `await` and thus doesn't need to be declared as `async`

danielbeardsley: I actually think it'd be cleaner to move the Ref stuff and this event handler down into the `NewsletterForm` and have it just take in `onSubscribe: (email: string) => void`. Mostly because I think the handler should be in the component that is responsible for the form. Then this component can just do `onSubscribe={subscribe}`.

danielbeardsley:
This is really close!

I still think something extra is getting imported and inflating the build size, but I'm not 100% sure.

packages/footer/components/Newsletter.tsx
aburke07: Good catch. This code was copied from iFixit/ifixit where they are not defined. I can switch them over.

packages/footer/components/Newsletter.tsx
aburke07: Nice idea! Moved in [0389595](https://github.com/iFixit/react-commerce/pull/1342/commits/0389595395eb69a7a005f9a73c7bd123951b66c8)

packages/footer/components/Newsletter.tsx
aburke07: Colors: [2374ee0](https://github.com/iFixit/react-commerce/pull/1342/commits/2374ee0f899c5d43db0dd674a61c3e5e13dfce0d)
Spacing: [66f37ec](https://github.com/iFixit/react-commerce/pull/1342/commits/66f37ec5fbba6888204cf8593f6651ed5c7cabc8)

frontend/models/store.ts
aburke07: Split imports in [53ba854](https://github.com/iFixit/react-commerce/pull/1342/commits/53ba8541d62db5b4ea2f9e0f415f7bc2167169e8)

packages/footer/components/LinkData.tsx
aburke07: [893cac9](https://github.com/iFixit/react-commerce/pull/1342/commits/893cac98ed4ef1328248455754490c41ab0108f7)

packages/footer/components/Navigation.tsx
aburke07: [32deab5](https://github.com/iFixit/react-commerce/pull/1342/commits/32deab58142a96fc9145cf6932e6a8d825ab2574)

packages/footer/components/Newsletter.tsx
aburke07: [32deab5](https://github.com/iFixit/react-commerce/pull/1342/commits/32deab58142a96fc9145cf6932e6a8d825ab2574)

aburke07:
Whew... we are finally passing tests and I've implemented all of @danielbeardsley's ideas about bundle size, but it still is inflated [~+12kb per page](https://github.com/iFixit/react-commerce/pull/1342#issuecomment-1428443632). Is there anything else we could do? 

I'll review the file diffs and see if I missed any `type` imports, because that seemed to be the only change so far that has helped reduce the size.

frontend/tests/jest/__mocks__/products.ts
```diff
@@ -1,13 +1,14 @@
-import { CurrencyCode, FindProductQuery } from '@lib/shopify-storefront-sdk';
-import { MenuItemType } from '@models/menu';
-import { ProductSearchHit } from '@models/product-list';
-import { ProductReview } from '@models/product/reviews';
+import { CurrencyCode } from '@lib/shopify-storefront-sdk';
+import type { FindProductQuery } from '@lib/shopify-storefront-sdk';
+import { MenuItemType } from '@ifixit/ui/menu';
```
danielbeardsley, line 3:
```suggestion
import { MenuItemType } from '@ifixit/ui';
```

Don't know if it makes a difference.

danielbeardsley:
I can't think of anything else to change. It'd sure be nice to know which packages got bigger and why, but it's down to a 2.6% increase.

CR :+1: 

un_dev_block :+1:

aburke07:
It looks like the additional import type uses didn't have much of an effect on the bundle size. I ran the webpack analysis again, but this time compared the server bundles. The only significant difference was between the two packages is detailed below. Everything else was either smaller on my branch or within 1kb of change, so I think the increase is confined to the difference below:

Main: 
main 441

My Branch: 
my branch 413

frontend/tests/jest/__mocks__/products.ts
aburke07: We can't do this yet ([issue](https://github.com/iFixit/react-commerce/issues/1400#issue-1595575920)):
> The cause of this issue was the `index` file of the package we were trying to import, `@ifixit/ui`, in the mocked products file[^2], also included Typescript XML, `tsx`, files. This became a problem with Playwright since we use this `products.ts` file in one of our tests[^3]. Playwright tries to compile all the imported files defined in the `index` file of `@ifixit/ui` which included the `tsx` files[^4]. However, Playwright does not know how to parse `tsx` files so it throws an error. 

> The workaround for this was to explicitly import the `ts` file needed in the `products.ts` file - i.e. `@ifixit/ui/menu` instead of `@ifixit/ui` [^5]

danielbeardsley:
> but this time compared the server bundles

Did you mean "client bundles"? The client bundle is the one that's shipped to the browser, the one who's bytes matter a lot more.

aburke07:
> > but this time compared the server bundles
> 
> Did you mean "client bundles"? The client bundle is the one that's shipped to the browser, the one who's bytes matter a lot more.

Yes! Good catch. These are the client bundles. Last time I was incorrectly looking at the server bundles.

k0rvusk0r4x:
- The language is missing.
- The icon for "Help Translate" is white on figma, but gray here.
- The social media icons are a darker gray here than on figma.

Here: 
image

Figma: 
image

dev_block 😿

aburke07:
Un_dev_block :alien: 

> The language is missing.

This is intentional: react-commerce is not localized.

> The icon for "Help Translate" is white on figma, but gray here.
> The social media icons are a darker gray here than on figma.

Fixed in [b9d9927](https://github.com/iFixit/react-commerce/pull/1342/commits/b9d9927156b2f4c35fbedacd378e6d06b52aa6ea)

I also noticed that the partner logos were not grayscale, as they are in the figma. This is fixed in [8e6a36a](https://github.com/iFixit/react-commerce/pull/1342/commits/8e6a36a6ff88afc6ecdad5602de5b9010220df38).

Lastly, we still need to remove the `- [Privacy] - [Legal] - [Accessibility]` part of the bottom menu, but this can be done in Strapi after deploy.

jordycosta:
dev_block 🟥 

It seems like the font sizes are much larger, we should probably be consistent with the main homepage footer

Screenshot 2023-02-27 145548

dhmacs:
@aburke07 did we figure out the reason for the bundle increase?

aburke07:
Un_dev_block :alien: Updated fontSizes and spacing to match figma.

federicobadini:
I've not completed a full review but here is my first two feedbacks

- [ ] The markup structure seems to be still on 4 columns. Is there any reason to prefer this to the markup structure proposed [here](https://github.com/iFixit/react-commerce/pull/1342#issuecomment-1438229129)?
- [ ] I see that fields like `menu1` and `menu2` are not retrieved from Strapi anymore. It is not clear to me why we are proposing to remove this feature 

dev_block ⌛

packages/footer/components/Navigation.tsx
```diff
@@ -1,18 +1,102 @@
...
+
+const NavigationColumn = ({ menu }: { menu: MenuType | null }) => {
+   console.log('Menu: ', menu);
+
+   if (!menu) {
+      return null;
+   }
+   return (
+      <>
+         <Text fontSize="14px" fontWeight="bold" color="white" my={2}>
```
jordycosta, line 59:
```suggestion
         <Text fontSize="16px" fontWeight="bold" color="white" my={2}>
```

This would match the footer on the homepage

**Preview**
preview

**Homepage**
prod

federicobadini:
Regarding the additional KBs, this can be explained by the fact that there are [static data](https://github.com/iFixit/react-commerce/blob/footer-update-newsletter/packages/footer/components/LinkData.tsx) that have been included to populate the menu. Moreover [this line](https://github.com/iFixit/react-commerce/blob/6d673103f46aa7a54acc8d93b07805dc2ab62b4a/frontend/layouts/default/Footer.tsx#L30) imports a long list of country codes

packages/footer/components/Navigation.tsx
aburke07: The homepage footer will soon be replaced with an import of this footer, and the figma is meant to be the canonical for both of these. I'm fine with this change, and I personally think the larger text looks nicer for the header.

Should we stick with the figma design or make this change?

aburke07: Actually, I think it might be more difficult to use this component now that we are going to use strapi to populate it, so I'll match the homepage size to keep it consistent.

aburke07:
> * The markup structure seems to be still on 4 columns. Is there any reason to prefer this to the markup structure proposed [here](https://github.com/iFixit/react-commerce/pull/1342#issuecomment-1438229129)?

The current implementation works fine, so I don't see a huge reason to refactor this code. It was copied from iFixit/ifixit.

> * I see that fields like `menu1` and `menu2` are not retrieved from Strapi anymore. It is not clear to me why we are proposing to remove this feature

This was also copied from iFixit/ifixit where we use LinkData.tsx to populate the menus. @sctice-ifixit confirmed that we actually want to be using Strapi instead, so I will switch it back.

packages/footer/components/Navigation.tsx
```diff
@@ -1,18 +1,102 @@
...
+         <FooterNavigationList pl={-5}>
+            <NavigationColumn menu={menu3} />
+         </FooterNavigationList>
+         <NewsletterComponent newsletterForm={newsletterForm} />
+      </FooterNavigationSection>
+   );
+};
+
+const NavigationColumn = ({ menu }: { menu: MenuType | null }) => {
+   console.log('Menu: ', menu);
```
ardelato, line 52: Is this `console.log` supposed to be here?

packages/footer/components/Navigation.tsx
```diff
@@ -1,18 +1,102 @@
...
+            {menu.title}
+         </Text>
+         <FooterNavigationListItems menu={menu} />
+      </>
+   );
+};
+
+const FooterNavigationListItems = ({ menu }: { menu: MenuType }) => {
+   const listItems = menu.items.map((item: MenuItem, index: number) => {
+      console.log('Item: ', item);
```
ardelato, line 69: Likewise, was this `console.log` purposely added?

packages/footer/components/Navigation.tsx
aburke07: Nope! Great find. Removed in [136721d](https://github.com/iFixit/react-commerce/pull/1342/commits/136721df9799b9857110515151d1b26bfb7f3ab4).

packages/footer/components/Navigation.tsx
aburke07: Nope! Great find. Removed in [136721d](https://github.com/iFixit/react-commerce/pull/1342/commits/136721df9799b9857110515151d1b26bfb7f3ab4).

federicobadini:
> The current implementation works fine, so I don't see a huge reason to refactor this code. It was copied from iFixit/ifixit.

If the component was simply ported from `iFixit/iFixit`, I would consider this as an opportunity to improve the code we have in both repos.
The footer was originally ported from `react-commerce`, so I suspect it was iteratively updated inside `iFixit/iFixit` until the current situation, in which the markup is including several hacks to mimic the design. 

The original `react-commerce` footer featured a 4 column layout so it appears reasonable that originally the content was organised inside a `SimpleGrid`.

Screenshot 2023-03-01 at 09 55 47

The new mockup though, is structured differently.
We have a clear separation between the left side (three columns with equal span) and a separated component on the right.  

Screenshot 2023-03-01 at 09 57 28

Inspecting `iFixit/iFixit` we can see that  negative margins and absolute positioning are used to adapt the old 4 columns grid to the new design:

Screenshot 2023-03-01 at 10 50 58

The width of each `li` is currently 25% of the footer width instead of 33% of the area dedicated to lists. 
It is not the case right now but if we are asked for instance to make the whole `li` clickable we will have to deal with overlapping content. 
I feel like this markup cleaning will be something we will have to do on the long term so I though it was better to do it now.

Sorry for the very long message, but I wanted to clarify the elements that prompted me to propose the refactor 😊
I'm of course open to discuss this so feel free to ping me

danielbeardsley:
Hey!  Page sizes are back down!

aburke07:
Un_dev_block :alien: Reverted to Strapi to populate the footer links in [91fbd40](https://github.com/iFixit/react-commerce/pull/1342/commits/91fbd408a3d10053992b37ff3a47dafc1e19b000) and re-formatted the links/newsletter component in [2405663](https://github.com/iFixit/react-commerce/pull/1342/commits/24056639d8cac31714eacf1f404d425301e1e2fd)

deltuh-vee:
Is the language switcher going to be left for another pull?
image

aburke07:
> Is the language switcher going to be left for another pull?

Yes. Mentioned in the original issue here: https://github.com/iFixit/ifixit/issues/45964#issuecomment-1410069964

deltuh-vee:
Hmm the mockup seems to lack dividers on mobile view 
image

aburke07:
> the mockup seems to lack dividers on mobile view

Fixed in [edaaba4](https://github.com/iFixit/react-commerce/pull/1342/commits/edaaba46644aec7a56854d4992d16b186c1225b5)

deltuh-vee:
dev_block ✊ 
Are the differences in title links under Resource intentional?
image

I also noticed hovering over this is a little off
image
The highlight just flashes instead of staying bright when hovered over. The link also extends beyond the text to the right.

aburke07:
> Are the differences in title links under Resource intentional?

Nope! I didn't see that they changed. I updated the values in strapi for the branch. The caching might take 10-15 mins to turn over before you see the updated links.

> The highlight just flashes instead of staying bright when hovered over. The link also extends beyond the text to the right.

Limited the width in [a941870](https://github.com/iFixit/react-commerce/pull/1342/commits/a9418701355b3b25cb48f5142d60eef19b510d26). I couldn't get the link hover to flash in chrome or safari. Can you check again?

Thanks for sticking with this QA!

Un_dev_block :alien:

deltuh-vee:
> I couldn't get the link hover to flash in chrome or safari. Can you check again?
Hmm I think one of my extensions is messing with the page. It looks fine on Firefox in a private window.

deltuh-vee:
QA 🎬 
Footer now matches the mockup.

frontend/layouts/default/Footer.tsx
```diff
@@ -36,155 +23,41 @@ import {
...
 interface FooterProps {
    stores: StoreListItem[];
-   menu1: Store['footer']['menu1'];
-   menu2: Store['footer']['menu2'];
    partners: Store['footer']['partners'];
    bottomMenu: Store['footer']['bottomMenu'];
    socialMediaAccounts: Store['socialMediaAccounts'];
+   menu1: Store['footer']['menu1'];
+   menu2: Store['footer']['menu2'];
+   menu3: Store['footer']['menu3'];
```
jyee27, line 37: Can we make these more generic (e.g. make a `NavigationList` type, replace menu1-3 with a `menus: NavigationList[]` prop)? If it's difficult to change we can save it for later.

jyee27 (approved):
CR ☕

frontend/layouts/default/Footer.tsx
aburke07: I don't think we can change this prop type because each of the strapi menu types here are different. I don't think there's a way to generalize these three types together.

Issue #1290, opened 2023-01-24
Playwright: Make it easier to interact with MSW fixture


ardelato:
## Description

Once #1275, goes out we will have the ability to set dynamic handlers. However, we did not prioritize abstracting the API for setting the dynamic handlers which means the setup is a bit convoluted. 

1. Declare the following fixtures within the test (`serverRequestInterceptor`, `port`, and either `graphql` or `rest`)

	https://github.com/iFixit/react-commerce/blob/917b1ca6269d38d2430eeded9ef423a42c667514/frontend/tests/playwright/product/add_to_cart.spec.ts#L115-L122

 	- The `serverRequestInterceptor` is the actual `msw` service worker [^1]. 
 	- The `port` is needed to navigate to the custom server instead of the regular server. 
 	- Lastly, the `graphql` or `rest` fixture are `msw`'s request handlers for respectively capturing either `GraphQL API` requests or `Rest API` requests.

2. Declare the new handlers to be used by the `msw` service worker
	https://github.com/iFixit/react-commerce/blob/917b1ca6269d38d2430eeded9ef423a42c667514/frontend/tests/playwright/product/add_to_cart.spec.ts#L123-L131

3. Then use `port` to go to the custom server the `msw` service worker is running on.
	https://github.com/iFixit/react-commerce/blob/917b1ca6269d38d2430eeded9ef423a42c667514/frontend/tests/playwright/product/add_to_cart.spec.ts#L133-L135

## To Do:

- [ ] See if we can change the `baseURL` a test uses if we use the `msw` service worker. That way we do not need to directly import and call the `port`
- [ ] Create a utility function to make it easier to interface with the `graphql` and `rest` request handlers. 
	> Maybe a wrapper function that accepts an `API` path to match against, a request type (`graphql` | `rest`), and the return type/body for the request.
	> This way we do not need to call either request handler fixture, `rest` or `graphql`. In addition, it hides a lot of the boilerplate for defining a request handler.
	> We will need to consider the different types of operations for each request type (i.e. `mutation` and `query` for `graphql`) 
	>
	> Another idea might be to create a factory class to return either a `Graphql` or `Rest` request handler class. 
 

[^1]: https://mswjs.io/docs/api/setup-server

PR #1348, opened 2023-02-08
Playwright: Abstract Handler Creation for MSW

ardelato:
## Description

In #1275 we created the ability to set dynamic handlers for `MSW`. However, the API for creating a handler is rather bloated and a lot of it can be extracted to reduce the boilerplate.

As such, we created a new static helper function to create these handlers.

## CR:

Commit by commit would be the easiest way to understand the code as some obscure context is needed in some cases.

Overall, the end goal was to abstract creating these dynamic handlers just enough to reduce the boilerplate and still be simple to use. Most of this was accomplished by making use of Typescript's features. 

## QA:

Besides checking CI still passing, I am not sure if it will be worth validating if the error handling works, as that might be exhaustive. 

I already confirmed that the `international buybox` handler still works using similar steps as https://github.com/iFixit/react-commerce/pull/1319#issuecomment-1416451999.

I also confirmed that the type checking is working as expected for the overloaded signatures:

|Before|After|
|:---:|:---:|
|  GraphQL body type is allowed | Graphql only accepts an object not a string|
<p align=center> <strong>In the scenario above, a GraphQLHandler requires a body that is only of type <code>Object</code></strong></p>

cr_req 2
qa_req 0

Connects: #1290

### DIFF
--- a/frontend/tests/playwright/msw/request-handler.ts
+++ b/frontend/tests/playwright/msw/request-handler.ts
@@ -1,0 +1,1 @@
+export default class Handler {}
--- a/frontend/tests/playwright/msw/request-handler.ts
+++ b/frontend/tests/playwright/msw/request-handler.ts
@@ -1,1 +1,12 @@
-export default class Handler {}
+import { RequestHandler } from 'msw';
+
+export type RequestInfo = {
+   endpoint: string;
+   method: string;
+};
+
+export default class Handler {
+   static create(request: RequestInfo): RequestHandler {
+      const { endpoint, method } = request;
+   }
+}
--- a/frontend/tests/playwright/msw/request-handler.ts
+++ b/frontend/tests/playwright/msw/request-handler.ts
@@ -1,12 +1,65 @@
 import { RequestHandler } from 'msw';
 
+/**
+ * These methods are a subset of the GraphQL operational methods that are
+ * supported by MSW's `graphql` handler. There are of course other methods
+ * `MSW` supports, but we are only interested in these two for now.
+ *
+ * @see https://mswjs.io/docs/api/graphql#operation-kind
+ */
+export type GraphQLMethods = 'mutation' | 'query';
+
+/**
+ * These methods are a subset of the REST methods that are
+ * supported by MSW's `rest` handler - _with the exception of
+ * `options`_.
+ *
+ * @see https://mswjs.io/docs/api/rest#methods
+ */
+export type RestMethods = 'all' | 'get' | 'post' | 'put' | 'patch' | 'delete';
+
 export type RequestInfo = {
    endpoint: string;
-   method: string;
+   method: GraphQLMethods | RestMethods;
 };
 
 export default class Handler {
    static create(request: RequestInfo): RequestHandler {
       const { endpoint, method } = request;
+
+      if (isGraphQLMethod(method)) {
+         // return GraphQLHandler
+      } else if (isRestMethod(method)) {
+         // return RestHandler
+      } else {
+         throw new Error(
+            `Invalid method ${method}. Must be 'mutation', 'query', 'all', 'get', 'post', 'put', 'patch', or 'delete'.`
+         );
+      }
    }
+}
+
+/**
+ * This is a type guard to ensure that the `method` is a valid GraphQLMethod
+ * by using Typescript Predicate Return Types.
+ * @see https://www.typescriptlang.org/docs/handbook/2/classes.html#this-based-type-guards
+ */
+function isGraphQLMethod(method: string): method is GraphQLMethods {
+   return method === 'mutation' || method === 'query';
+}
+
+/**
+ * This is a type guard to ensure that the `method` is a valid RestMethod
+ * by using Typescript Predicate Return Types.
+ * @see https://www.typescriptlang.org/docs/handbook/2/classes.html#this-based-type-guards
+ */
+function isRestMethod(method: string): method is RestMethods {
+   return (
+      method === 'all' ||
+      method === 'get' ||
+      method === 'post' ||
+      method === 'put' ||
+      method === 'patch' ||
+      method === 'delete'
+   );
 }
--- a/frontend/tests/playwright/msw/request-handler.ts
+++ b/frontend/tests/playwright/msw/request-handler.ts
@@ -1,3 +1,4 @@
+import { isRecord } from '@ifixit/helpers';
 import { GraphQLHandler, RestHandler, RequestHandler } from 'msw';
 
 /**
@@ -53,6 +54,10 @@ export default class Handler {
       const { status, body, responseType } = response;
 
       if (isGraphQLMethod(method)) {
+         if (!isRecord(body)) {
+            throw new Error(`Invalid GraphQL body ${body}. Must be an object`);
+         }
+
          return new GraphQLHandler(
             method,
             endpoint,
--- a/frontend/tests/playwright/msw/request-handler.ts
+++ b/frontend/tests/playwright/msw/request-handler.ts
@@ -61,6 +61,7 @@ export type RequestInfo = {
  * to configure what the response should return if the request is matched.
  *
  * @property `status` - The HTTP status code to return
+ * @property `headers` - The headers to return with the response
  * @property `body` - The body of the response.
  * @property `responseType` - This the kind of transformation to apply to the
  * body when it will be returned. This is only applicable to REST requests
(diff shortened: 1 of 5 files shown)


frontend/tests/playwright/msw/request-handler.ts
```diff
@@ -82,37 +115,45 @@ export default class Handler {
              * regardless of location the request is made from.
              */
             '*',
-            (req, res, ctx) => {
-               if (passthrough) return req.passthrough();
-               if (once) return res.once(ctx.status(status), ctx.data(body));
+            customResolver ??
```
mlahargou, line 159: When do you imagine we would use a custom resolver?

mlahargou (approved):
CR 📱 

Nice job on this! I think it makes mocking these requests much simpler.

frontend/tests/playwright/msw/request-handler.ts
ardelato: I could see it being used when we want to use the values of the `request` to modify the `response` or reduce the number of handler creations needed.

For example, if wanted to validate each user type in a single test, we would need to create each handler that would override the same path.

Instead, we could bake each case into a single custom resolver:

```typescript
server.use(Handler.create({endpoint: '/user', method: 'get'}, pro-test-user-response))
// Run test

server.resetHandlers()
Handler.create({endpoint: '/user', method: 'get'}, ifixit-empl-user-response)
// Run test

server.resetHandlers()
...

// vs.

const customR = (req, res, ctx) => {
	const user = req.body.username

	if(username = 'pro-test-user') {
		return res(
			ctx.json({...defaultUserProps, discountTier: ''})	
	}
	else if (username = 'ifixt-empl-test-user') {}
	else if(username = 'reg-test-user') {}
	else 
	...
}

Handler.create({endpoint: '/user', method: 'get'}, customR)
```

ardelato: Obviously, we could just separate each `user` test into its own `test case` but I am just trying to give an example of how the `customResovler` option could be used.

mlahargou: oh cool! That makes sense 👍

frontend/tests/playwright/msw/request-handler.ts
```diff
@@ -0,0 +1,233 @@
...
+ * @see https://mswjs.io/docs/api/context
+ * @see https://mswjs.io/docs/api/response
+ *
+ */
+export type MockedResponseInfo = {
+   status: number;
+   headers?: Record<string, string | string[]>;
+   body?: Record<string, any> | string | number | boolean | null;
+   responseType?: 'json' | 'text' | 'xml' | 'raw';
+};
```
andyg0808, line 80: I'd suggest splitting this into two types: one for GraphQL requests and one for REST requests. I think trying to keep them together and then filter the appropriate keys out below is probably not simplifying things. If you want to avoid duplication on `status`, you can always create a `type BaseResponseInfo = { status: number }` and then intersect it into the field set for each subtype.

frontend/tests/playwright/msw/request-handler.ts
```diff
@@ -0,0 +1,233 @@
...
+export type RestResolver = ResponseResolver<
+   RestRequest<DefaultBodyType, PathParams<string>>,
+   RestContext,
+   DefaultBodyType
+>;
+
+export type RequestInfo = {
+   endpoint: string;
+   method: GraphQLMethods | RestMethods;
+};
```
andyg0808, line 57: I think it'll be easier to work with to have two copies of this: one for GraphQL and one for REST. The fact that `endpoint` is the same type and would be duplicated seems relatively unimportant to me. If you disagree, you can always define a `BaseRequestInfo` with just `{ endpoint: string }` in it.

frontend/tests/playwright/msw/request-handler.ts
```diff
@@ -0,0 +1,233 @@
...
+      options?: HandlerOptions;
+   }): RequestHandler {
+      const { endpoint, method } = request;
+      const { status, headers, body, responseType } = response;
+      const { once, passthrough, customResolver } = options;
+
+      if (isGraphQLMethod(method)) {
+         if (!isRecord(body)) {
+            throw new Error(`Invalid GraphQL body ${body}. Must be an object`);
+         }
```
andyg0808, line 148: I'm guessing that this check can be made redundant with enough care in typing. I'm not opposed to leaving it even in that case, in order to provide safety against untyped calls, but if TypeScript forced this check, it's probably indicative of a deeper problem.

frontend/tests/playwright/msw/request-handler.ts
```diff
@@ -0,0 +1,233 @@
...
+            /**
+             * This refers to the Path,or location, of the endpoint. By using a
+             * wildcard, we are saying that we want to match any endpoint
+             * regardless of location the request is made from.
+             */
+            '*',
+            customResolver ??
+               ((req, res, ctx) => {
+                  if (passthrough) return req.passthrough();
+                  if (once) return res.once(ctx.status(status), ctx.data(body));
```
andyg0808, line 162: Since `passthrough` and `once` don't do anything when `customResolver` is present, I might try to make the `HandlerOptions` type reflect that:
`type GraphQLHandlerOptions = { customResolver: GraphQLResolver } | { passthrough?: boolean; once: boolean; }`

andyg0808 (approved):
Lots of typey suggestions, but looks good overall.
deploy_block :+1: on comments.

frontend/tests/playwright/msw/request-handler.ts
```diff
@@ -0,0 +1,233 @@
...
+   GraphQLContext<Record<string, any>>
+>;
+/**
+ * These methods are a subset of the REST methods that are
+ * supported by MSW's `rest` handler - _with the exception of
+ * `options`_.
+ *
+ * @see https://mswjs.io/docs/api/rest#methods
+ */
+export type RestMethods = 'all' | 'get' | 'post' | 'put' | 'patch' | 'delete';
```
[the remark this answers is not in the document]
ardelato: Unfortunately, there is no `MSWRestMethods` type but I can still accomplish this by doing

```typescript
Omit<keyof (typeof rest), 'options`>
```

That is definitely better then the static typing so I'll make the change 👍

frontend/tests/playwright/msw/request-handler.ts
ardelato: If we make two respective request info types for each API type, should we remove the overloading for `create()`?

frontend/tests/playwright/msw/request-handler.ts
ardelato: It became redundant in [`cd6fcf4` (#1348)](https://github.com/iFixit/react-commerce/pull/1348/commits/cd6fcf42ebeae5708a4edfa0958a3c48987fe9dd). Prior to that, the `body` type for a GraphQL Handler could have been anything acceptable under a Rest `body` type. That is why the check was there but is no longer needed.

ardelato:
un_dev_block ⚡ un_deploy_block ⚡ I addressed the comments by exporting two different functions instead of a single class for creating each handler type.

frontend/tests/playwright/msw/request-handler.ts
```diff
@@ -86,148 +110,121 @@ export type MockedResponseInfo = {
...
+   response: GraphQLResponseInfo;
+   options?: GraphQLHandlerOptions;
+}): GraphQLHandler => {
+   const { endpoint, method } = request;
+   const { status, body } = response;
+
+   const customResolver =
+      typeof options === 'function' ? (options as GraphQLResolver) : null;
+   const { once = false, passthrough = false } =
+      typeof options === 'object' ? (options as HandlerOptions) : {};
```
andyg0808, line 157: Not a big deal, but have you tried leaving these casts out? I'd have expected TypeScript to potentially be able to infer the types here, and doing that would probably be safer for the long-term.

andyg0808 (approved):
CR :+1:

frontend/tests/playwright/msw/request-handler.ts
ardelato: Unfortunately, Typescript does not implicitly understand that `once` and `passthrough` exists in `options`

image

And `customResovler` would resolve to `null` instead of either `RestResolver` or `null`

image

mlahargou (approved):
CR 📱 Sweet!

PR #1354, opened 2023-02-09
Enable zoom only when image is larger than gallery container by 1.5 factor

### LINKED ISSUE #1097: Don't zoom low-res images
On Product Detail Page, disable hover zoom when product images are not high resolution enough to make zooming useful. I'm not sure at what resolution zooming is useful but I feel like it would be an easy question for dev to answer.

image

Example product: https://www.ifixit.com/products/lenovo-ideapad-c340-15iil-c340-15iwl-laptop-battery

Not a high priority, but this should be done at some point.

federicobadini:
closes #1097 

### QA

1. Visit Vercel preview
2. Verify that images with `inner_width * inner_height` only slightly larger than the gallery container are not zoomed
3. Only images with an `inner_width * inner_height >= container_width * container_height * 1.5` should have zoom enabled

### DIFF
--- a/frontend/templates/product/sections/ProductSection/ProductGallery.tsx
+++ b/frontend/templates/product/sections/ProductSection/ProductGallery.tsx
@@ -59,7 +59,7 @@ export function ProductGallery({
                ({ width, height }) =>
                   !width ||
                   !height ||
-                  width * height < containerWidth * containerHeight
+                  width * height < containerWidth * containerHeight * 1.5
             )
          );
       };


danielbeardsley:
A lot of screens these days have high DPI and a lot of people don't have above-average vision. I think there's a lot of value in being able to make a picture larger even if we don't have the resolution to show more pixels. In other words, even though zooming in on picture that's already shown at 1:1 doesn't show more pixel information, people may still appreciate being able to make that picture larger.

Thoughts?

sterlinghirsh (approved):
CR :zap:

sterlinghirsh:
@danielbeardsley As I mentioned on the issue, I agree there's some value in showing a zoomed blurry image. We may decide to abandon the limitation on zooming, but I'm willing to try it at 1.5x and see how it feels since it's a tiny change that's already done.

k0rvusk0r4x:
1. In the preview, went to the example of a product that should not have zoom according to the issue.
2. Noted it still has zoom.
3. Inspected the image and was not able to find `inner_width` or `inner_height` to compare to container size, so I can't be sure if the issue is this not doing what it's supposed to or the numbers being wrong.

dev_block 😿 

> Thoughts?

Most often, when I zoom in on tech products, I'm looking at small things on the edges, like fasteners, connectors, or the tips of screwdrivers. Blurry is not my favorite, but most of the time I'd prefer to zoom in on blurry than not at all.

danielbeardsley:
> but I'm willing to try it at 1.5x and see how it feels since it's a tiny change that's already done.

But it's not about *feel* it's about us deciding *which* product images allow zooming. This change is easy to make, but that doesn't mean it's *better* cause it's easy. From a users perspective, they don't get to know if there are more image pixels or not, we just enable zooming for that image or we don't.

As far as I understand, this is about deciding which images to enable the zoom feature on, not about changing how zooming works or feels. If we consider making the zoomed image *bigger* than it's natural size, that would change the feel of the zoom.

davidrans (approved):
CR :coffee:

k0rvusk0r4x:
If the size of the image is < 1.5x the container, zoom is disabled.

QA 😺 

Note in case further testing is needed: Easiest way to ensure the image is smaller is to make the browser full screen on a monitor where full screen is huge and silly.

PR #1357, opened 2023-02-10
Browser Caching: enable it!

danielbeardsley:
Problem: we have been using `s-maxage` and vercel listens to that, but strips it off so both downstream caches (Cloudfront and browsers) see the response as uncacheable.

Since we're telling vercel to cache our responses (via s-maxage), we might as well let browsers and cloudfront cache them too (via maxage).

The only reason this isn't the default recommended for vercel is because they auto-invalidate your caches on deploy and they can't do that for downstream caches. Given that our cache times are 1 minute, I'm OK with CF and browsers storing these responses for that amount of time.

Note: max-age is computed relative to the when the response was generated, not when it was received. So, if vercel sends a 40s-old page to CF and CF caches it, it'll only consider it fresh for another 20s, not 40s + 1min.

qa_req 0

### DIFF
--- a/frontend/helpers/cache-control-helpers.ts
+++ b/frontend/helpers/cache-control-helpers.ts
@@ -10,11 +10,11 @@ export const withCacheControl =
    (options: CacheControlOptions): GetServerSidePropsMiddleware =>
    (next) =>
    (context) => {
-      const sMaxAgeSeconds = options.sMaxAge / 1000;
+      const maxAgeSeconds = options.sMaxAge / 1000;
       const staleWhileRevalidateSeconds = options.staleWhileRevalidate / 1000;
       context.res.setHeader(
          'Cache-Control',
-         `public, s-maxage=${sMaxAgeSeconds}, stale-while-revalidate=${staleWhileRevalidateSeconds}`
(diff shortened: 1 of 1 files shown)


deltuh-vee:
QA 🎬 
max-age is now set for product page responses and Vercel's cache is getting hits.
image

deltuh-vee:
I see that max-age is now set, but are we expecting to see a cache hit from cloudfront? 
image

danielbeardsley:
> expecting to see a cache hit from cloudfront?

No. The vercel previews are not behind cloudfront. The only thing behind cloudfront is `www.ifixit.com`

deltuh-vee:
> No. The vercel previews are not behind cloudfront. The only thing behind cloudfront is www.ifixit.com
Does that mean we'll have to check this in production?

danielbeardsley:
I don't really think there is anything to QA here. It adds `max-age` to the cache-control header. CR about if this is a good idea is more important. I'll drop the QA requirement.

danielcliu-ifixit (approved):
CR :v:

danielbeardsley:
> Maybe we should have an api to specify `max-age` separate from `s-maxage` to make the difference between the two clear.

If needs for these two diverge I think it'd make sense to set them separately, but as of now the only difference is that browser caches are not cleared on deploy. Given that current times are short, deploys infrequent, and cached times are dwarfed by how long someone may keep a tab open (hours to days), I don't feel strongly.

Issue #1346, opened 2023-02-07
CI Feature: Add a package.json linter to ensure static versions are used


ardelato:
## Description

There are times when we install a new package but the semantic version syntax may default to pre-pending `^` or `~`. The rest of our packages all use a static semantic version.

It would be nice if we could automate this or at least catch it so the semantic version syntax is consistent across our `package.json`.

**References:**
- https://github.com/iFixit/react-commerce/pull/1216#discussion_r1062787286
- https://github.com/iFixit/react-commerce/pull/1319#discussion_r1095063250

mlahargou:
Taking this off of QAE backlog as it looks like IoP is going to take care of this.

federicobadini:
Probably the best option we have is using `npm-package-json-lint` adding it to the build & lint step in order for it to error out if some rule does not pass validation.
It just requires a base config like this:

```json
{
  "rules": {
    "prefer-absolute-version-dependencies": ["error", {}]
  }
}

```

We should also make sure to skip packages that have wildcard - like our own `@ifixit/*` packages - from validation, but it should be feasible too.

There are more lightweight alternatives like adding `"save-exact": "true"` to package.json or .npmrc.
However these solutions only reduce the likelyhood of non-static package version making their way into package.json.
So if the requirement is strict, I'd go with the first one.

Let me know what you think

ardelato:
I tested the `save-exact` option which is useful for adding new packages as the developers won't need to specify a version every time they want to install a package. https://github.com/iFixit/react-commerce/compare/testing-package-json-linter

However, to be more strict I tested the `npm-package-json-lint` package and found out the following.
1. The `exception` setting does not take in `regex` and therefore we need to explicitly list all modules we want to exclude. This means we need to list each of the `@ifixit` packages.
2. If we make use of the `lint-staged` `pre-commit` hook we already have, then we can run this linter every time we commit. 

---

I think we should include both the `save-exact` rule as well as the `npm-package-json-lint` package. Although, how we integrate the `npm-package-json-lint` package is still up for discussion.

If we use the `pre-commit` hook then it would be included as part of the normal workflow of a Dev. 

If we use it in CI, then we would need to add an additional job as the current [`linter` action ](https://github.com/iFixit/react-commerce/blob/main/.github/workflows/lint.yml#L29) only runs `eslint` for the `frontend` directory and will automatically commit the `eslint` fixes.

PR #1358, opened 2023-02-10
Breadcrumbs: Add missing dependencies

andyg0808:
Seems I missed some dependencies when I set this up previously. I'm assuming this is because we don't have PNPM set to particularly strict settings in this repo, so I didn't notice that they were missing. It could also be because the code wasn't actually being used yet, so there wasn't any build to break.

qa_req 0
This is dependency-only, and this code is currently unused. We'll test it when we integrate it on the monorepo side.

### DIFF
--- a/packages/breadcrumbs/Breadcrumbs.tsx
+++ b/packages/breadcrumbs/Breadcrumbs.tsx
@@ -12,6 +12,7 @@ import {
    Text,
    BreadcrumbItemProps,
    LinkProps,
+   useBreakpointValue,
 } from '@chakra-ui/react';
 import { FaIcon } from '@ifixit/icons';
 import { faChevronRight } from '@fortawesome/pro-solid-svg-icons/faChevronRight';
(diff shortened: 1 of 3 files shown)


dhmacs (commented):
@andyg0808 What do you think about specifying strict versions? Imo it'd be easier to know which versions we're relying on, and it would be more predictable to recreate lock files

related #1346

andyg0808:
@dhmacs Yeah, I'll do that. I initially was thinking that I'd want it to just use whatever peer version was available,but as I think about it, it'll throw errors if there's a version mismatch, so we shouldn't have problems with different-version copies of libraries floating around.

danielbeardsley:
I have a guess about the failures. Maybe this package was getting a different version of chakra previously and adding these explicit dependencies changed which version it ended up being resolved to?

The failure is the same across seemingly all the failures which makes me thing it's not a one-off temporary thing. That said, I'm having a hard time imagining what could be wrong about this...

danielbeardsley (approved):
CR :+1: 

That looks better. How did we previously arrive at the versions such that they didn't match the ones in other packages?

andyg0808:
@danielbeardsley The first time, I don't remember how I chose the version ranges; I might have referred to another `package.json`. The second time, when I made them fixed versions, I think I used whatever the ranges had resolved to in the `pnpm-lock.yaml`. The third time, I tried to find the existing version numbers in the `pnpm-lock.yaml` which weren't being added by my changes. And then this time, I mostly just made sure to get _all_ the versions to match existing entries in the `pnpm-lock.yaml`.

PR #1359, opened 2023-02-11
Playwright: Extend functionality of the `Product Page` fixture

ardelato:
## Description

This removes the need to require the `page` and `port` fixtures when using the `productPage` fixture. Thus, it will clean up the test case signatures a bit.

### CR:

We were calling both the `productPage` fixture and the `page` fixture in some tests. This made the test fixture usage signature a bit bloated. 

Test Case Sig

The only reason we were doing this was to use the `page` methods that are not directly available on the `productPage` fixture. 

However, the `productPage` fixture has a `page` property that is the same
instance as the `page` fixture. Therefore, we can just use this property to call any of the `page` methods.

Lastly, we had to import the `port` fixture because we cannot use the `productPage.gotoProduct` method as it is using the `baseURL` that is set in our configs. We want to navigate to the port our custom server is running on, which is why we were directly using `page.goto`. However, we can configure the `baseURL` that the `productPage` fixture uses if we set it as a property and update it when we define the `serverRequestInterceptor` fixture. This will allow us to navigate to both a product page on the default server and custom server without additional overrides by the user.

### QA:

qa_req 0 
- [ ] CI should continue to pass

Closes: https://github.com/ifixit/react-commerce/issues/1290


frontend/tests/playwright/product/add_to_cart.spec.ts
```diff
@@ -80,28 +78,28 @@ test.describe('product page add to cart', () => {
       await cartDrawer.open();
       await cartDrawer.assertCartTotalQuantity(0);
       await cartDrawer.assertItemIsNotPresent(sku);
-      await page.getByTestId('back-to-shopping').click();
+      await productPage.page.getByTestId('back-to-shopping').click();
```
mlahargou, line 81: Not too sure about this one, but does typescript have some sort of magic methods like php? I'm totally not sure if this is even possible in ts.

In php, we could do something like this on the product page:

```php
public function __call($name, $args) {
   if (<$name isn't a function in this class>) {
     return $this->page->$name($args);
   }
   return $this->$name($args);
}
```

Then this could just be 
```suggestion
      await productPage.getByTestId('back-to-shopping').click();
```

mlahargou (dismissed):
CR 📱 Looks great as is! Just a small question that might make things a bit more simple.

frontend/tests/playwright/product/add_to_cart.spec.ts
ardelato: I initially was looking into something like for `TS` but I could not find anything.

I'll do some more digging now that I learned they are called `magic methods`.

ardelato: I don't think there is an easy way to implement something like `__call` magic method in Typescript. I saw there is something called a `Proxy` object (https://stackoverflow.com/a/42244695) that could allow us to redirect all implicit methods to `Page` but because of Typescript, there are a lot of type checks we would need to add. In addition, it would convert the methods of the classes into properties.

https://codesandbox.io/s/typescript-playground-export-forked-k3iewl?file=/index.ts

mlahargou: Sounds good. I'm good with how this is 👍

ardelato:
deploy_block ⚡ on https://github.com/iFixit/react-commerce/pull/1348 going out first

ardelato:
dev_block ☁️ I need to resolve the conflicts since there were major changes in the base branch

ardelato:
un_dev_block ⚡ resolved the conflicts

ardelato:
dev_block ☁️ the test failures are relevant

ardelato:
un_dev_block ⚡ Failures were actually unrelated to code changes

ardelato:
un_deploy_block ⚡ https://github.com/iFixit/react-commerce/pull/1348 has gone out

mlahargou (approved):
CR 📱 Just a lot of merging in master 👍

PR #1368, opened 2023-02-15
Increase product list s-maxage and stale-while-revalidate duration

### LINKED ISSUE #1367: Increase `s-maxage` and `stale-while-revalidate` duration for product list pages
Currently we have `s-maxage` set to 10s and `stale-while-revalidate` set to 10min, this is a very short timespan, which mean that for most product list pages caching would not be used at all.
Since product list pages do not need realtime updates, let's increase the duration of cache control

dhmacs:
closes #1367 

This also increases to 5 minutes the default `s-maxage` of product pages

qa_req 0


danielbeardsley (approved):
CR :+1: 

Looks good!

Issue #1326, opened 2023-02-02
Make tests more resilient


ardelato:
## Description

In #1317 there was a small grammatical change that caused the tests to fail. https://github.com/iFixit/react-commerce/commit/4031350862efa7ef883f2a133e2af4ebbae51aa4

### To Do:

- [ ] Make the tests more resilient to changes like grammatical changes.

_Originally posted by @dhmacs in https://github.com/iFixit/react-commerce/issues/1317#issuecomment-1414465975_

dhmacs:
In my opinion, we should prioritize testing areas that, if broken, could affect revenue. I don't believe testing copy is valuable and can result in additional maintenance costs.
For instance, tests for the "notify back in stock" form should validate two key functionalities:

- The ability for the user to interact with the form
- The ability for the user to submit the form successfully.

mlahargou:
I agree, comparing strings here is not the best idea. I do, however, think it is important to verify the user's workflow. We should still ensure the div is visible, but we should do that based on the selector/class of the element, rather than the text on the element.

PR #1317, opened 2023-01-31
Add Out of Stock alert

dhmacs:
closes #951 

This PR adds an out of stock alert box as requested in #951 and contains also a couple of color corrections.

## QA

1. Open Vercel preview
2. Verify that the new design for Out of Stock has been implemented
    Screenshot 2023-01-31 at 14 36 05

PR #1371, opened 2023-02-16
Playwright: Make tests a bit more resilient

ardelato:
## Description

In https://github.com/iFixit/react-commerce/pull/1317 there was a small grammatical change that caused the tests to fail. https://github.com/iFixit/react-commerce/commit/4031350862efa7ef883f2a133e2af4ebbae51aa4

The main two reasons why this occurred are:
1. We use `toHaveText` instead of `toContainText`[^1] for asserting the text in an element. 
	- `toHaveText` will do an explicit match while `toContainText` will match if the string is included. Therefore, we should use the latter for validating that some text is within the element. For explicit assertions, we can still use `toContainText` but pass in a `RegExp` instead. 
2. We search for elements based on text content
	- By searching for elements based on text, we might end up nabbing a different element than expected if our text is too vague. On the other hand, if our text is too specific, then we risk our assertions failing if we add a minor change-- like a `period`.   

This pull tries to address these issues to make our tests more resilient to text changes.

### CR Notes:

In order to make our tests more resilient I made a few changes to our tests.

- I replaced all uses of `toHaveText` with `toContainText`
- I added `data-testid`s to certain components in order to replace some usages of `getByText`.
- To avoid adding multiple `data-testid`s I narrowed down the searching for an element by scoping it down from the `page` to a closer `parent` element. This allows us to ensure we do not reference a different element with similar text.

### QA Notes:

qa_req 0 CI continues to pass

Closes: #1326 

[^1]: https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-contain-text

### DIFF
--- a/frontend/templates/product-list/sections/FilterableProductsSection/index.tsx
+++ b/frontend/templates/product-list/sections/FilterableProductsSection/index.tsx
@@ -189,6 +189,7 @@ const ProductListEmptyState = forwardRef<EmptyStateProps, 'div'>(
                px="2"
                textAlign="center"
                {...otherProps}
+               data-testid="product-list-no-results"
             >
                <Icon
                   as={SearchEmptyStateIllustration}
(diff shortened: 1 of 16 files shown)


batbattur (approved):
CR 🏎️ Nice job on adding `data-testid`s!

PR #1375, opened 2023-02-16
Refactor product list facets

### LINKED ISSUE #1362: Refactor product list facets

federicobadini:
closes #1362


ardelato:
The Playwright `cross-sell` test failure is unrelated to the PR. I am going to make a new PR to fix it

masonmcelvain:
@ardelato It looks like @federicobadini already opened a test to fix it here: https://github.com/iFixit/react-commerce/pull/1385

ardelato:
Ah nice! I did not see that. I'll go ahead and close my PR

frontend/templates/product-list/sections/FilterableProductsSection/Toolbar.tsx
```diff
@@ -92,15 +96,27 @@ export function Toolbar(props: ToolbarProps) {
...
                </ProductViewSwitch>
             </Flex>
+
+            {currentRefinements.items.length > 0 && (
+               <Flex
+                  wrap="wrap"
+                  align="flex-start"
+                  display={{ base: 'flex', md: 'none' }}
+               >
+                  <CurrentRefinements />
```
masonmcelvain, line 117: I noticed that the design displayed the current refinements in a single row with overflow-x hidden, whereas the implementation wrapped them onto multiple rows. It looks nice as implemented, but thought I'd point out the discrepancy.

| Spec | Implementation |
|-|-|
| image | image |

frontend/tests/playwright/product-list/filters.spec.ts
```diff
@@ -5,12 +5,21 @@ import { waitForAlgoliaSearch, resolvePath } from '../utils';
...
-   await expect(page.getByTestId(`current-refinement-${value}`)).toBeVisible();
+   if (layout === 'desktop') {
+      await expect(
+         page.getByTestId(`current-refinement-${value}`).first()
+      ).toBeVisible();
+   } else {
+      await expect(
+         page.getByTestId(`current-refinement-${value}`).last()
+      ).toBeVisible();
+   }
```
masonmcelvain, line 22: It's not clear to me why we're making the distinction between mobile and desktop in this test. When I inspect the `<CurrentRefinements>` component, it appears that all current refinements are rendered on both mobile and desktop, so I'm confused why we can't make the same assertion on both.

masonmcelvain (approved):
CR ✌🏻 Nice work! I have a couple non-blocking questions but looks good to me

frontend/templates/product-list/sections/FilterableProductsSection/Toolbar.tsx
federicobadini: Yep, you are right. During the development together with @mmarcon91 we discussed this.
Having chips on a single row make more room for the product list but make the current selection harder to read.
So we decided to keep the multiple row approach for a while and eventually to move to single row only if needed

frontend/tests/playwright/product-list/filters.spec.ts
federicobadini: We have multiple `<CurrentRefinements>` components in dom. However their visibility depends on the layout.
If we are on a mobile layout `<CurrentRefinements>` is visible only in the toolbar component.
If we are on a desktop layout `<CurrentRefinements>` is located in the upper part of the side panel.

To avoid altering the test, I made the above distinction.

masonmcelvain: Thanks for the explanation! I missed the fact that there are 2 components.

federicobadini:
Aligned with main to include the test fix

masonmcelvain (approved):
CR ✌🏻

k0rvusk0r4x:
The color is off: 
Branch: 
image
Mockup: 
image

Is this deliberate for accessibility or similar?

Mobile only:
Branch has the chosen filters wrapping: 
image
On the mockup, they scroll: 
image

I think wrapping is better here, but it is a difference we want to be sure we are doing deliberately.

dev_block 😿 to assess these differences and decide if we want to change the speck or change the branch.

federicobadini:
> Is this deliberate for accessibility or similar?

It is for uniformity with the other chips that we have onto the website - we can update the mockup but that was discussed with @mmarcon91 

> I think wrapping is better here, but it is a difference we want to be sure we are doing deliberately.

Discussed [here](https://github.com/iFixit/react-commerce/pull/1375#discussion_r1112390541) :)

federicobadini:
un_dev_block 👍

masonmcelvain (approved):
CR ✌🏻 I'm uncertain why the tests are failing. I'm also noticing that builds on `main` are failing. I'm looking into it

masonmcelvain (dismissed):
CR ✌🏻

erinemay:
In general, it looks good. dev_block 🧷 on some questions and a couple minor things I noticed. 

<details><summary>1. Desktop: Show more/Show less not working</summary>

On a collection like https://react-commerce-git-refactor-product-list-facets-ifixit.vercel.app/Parts/iPhone, I see the item-type list is fully expanded, but show more/show less is still present and doesn't have any effect. I'm seeing it on Chrome, FF, and Safari.

https://user-images.githubusercontent.com/72166715/221687689-9a17226d-4a96-41e5-82d4-dc29436f981e.mp4

</details>


<details><summary>2. Desktop: Should the facet container resize dynamically?</summary>

This might have been discussed elsewhere, but the mockup appears to have the facet container dynamically sized (like the current implementation). This branch has it as the same static size, filling the vertical viewport space.

| Mockup/current | this branch |
| - | - |
| image | image |

</details>

<details><summary>5. Tablet: Facet container extends beyond the screen</summary>

This is the iPad 9th with Safari 15. It looks like the container keeps going until the end of the productlist.

image
 
</details>

<details><summary>4. Mobile: Should the filter list have some room to peek at the product list?</summary>

Or is this just a demonstration that it should open as a drawer?

| Mockup | this branch |
| - | - |
| image | image |




</details>

federicobadini:
> Desktop: Show more/Show less not working

Fixed

> Desktop: Should the facet container resize dynamically?

Fixed

> This is the iPad 9th with Safari 15. It looks like the container keeps going until the end of the productlist

Without the fix above, the facet container maxHeight was 100vh minus some margins. However on safari 100vh is not taking into account the space of the address bar, which reduces the available viewport.
So 100vh - when the bar is visible - is effectively more than the visible viewport height.
The fact that the container was sticky also created this "illusion" that its height was effectively equal to the product list height. Now that the fix above is in place we should have the same behaviour we currently have in production.

> Mobile: Should the filter list have some room to peek at the product list?

It would be nice, however it was excluded from this PR since we have several other drawers that would need to be styled too (cart, navigation etc). So it will be discussed in a separate PR

un_dev_block 👍

masonmcelvain (approved):
CR ✌🏻

masonmcelvain (dismissed):
CR ✌🏻

ardelato:
My bad, I accidentally pushed to this branch while trying to deflake the `parts_page_search` test 😓 . I undid the commit and had to force push.

ardelato:
CR 🚗 carry-over

erinemay:
The facets are working as expected and the above noted things are fixed. qa 🦖 

deploy_block 🧷 @federicobadini, I've got a weird one for you. 

I'm seeing this gap to the right side of some mobile device/browser combinations, but I'm not sure if it's an artifact of previewing and browserstack, or an actual bug. Are you seeing this? I haven't confirmed it's present on real devices. I'm asking around now for someone with an iPhone. I'm not seeing it on mobile Android + Chrome.

Here's an example problematic URL: https://react-commerce-git-refactor-product-list-facets-ifixit.vercel.app/Parts/Teenage_Engineering_OP-Z

And the main version: https://react-commerce.vercel.app/Parts/Teenage_Engineering_OP-Z

<details><summary>Screenshots and details</summary>
Browserstack - iPhone 14 + Safari 16

| this branch | main |
| - | - |
| image | image |

FF dev tools site preview as iPhone 11 Pro (happening on MacOS and Windows)

| this branch | main |
| - | - |
| image | image |



There isn't any obvious element that's expanding beyond the page limit. 

https://user-images.githubusercontent.com/72166715/221998127-06e01ed4-d8c6-46a3-8202-21be8f5eb05c.mp4

</details>

deltuh-vee:
I see it on an iPhone SE
IMG_CF493CF1808D-1

federicobadini:
> I've got a weird one for you.

The list and grid items were updated for the new layout in #1376 
Since it was merged, I am pretty confident that just updating from main will resolve that issue

federicobadini:
un_dev_block 👍

erinemay:
Awesome. Yep. That did it. 

QA 👍 

- On desktop, tablet, and mobile, the above issues have been resolved or addressed. 
- The filter facets look like the mockup and behave in an intuitive manner. 
- Adding selections works
- Removing selections (individually and all together) works
- The container responds dynamically
- Available facets refresh immediately when one is selected that changes availability of others
- In general, I poked around and added/removed/paginated, moved up and down the hierarchy and didn't find anything else of note.

deltuh-vee:
Carry Over QA 🎬

masonmcelvain (approved):
CR ✌🏻

PR #1378, opened 2023-02-16
Troubleshooting: Add nav bar and edit link

andyg0808:
This adds a nav bar and the `Edit` button/link on it. It's not adding anything else just yet, as this is about the easiest part of the bar to build first.

## QA Notes
Link: https://react-commerce-git-vulcan-add-nav-bar-ifixit.vercel.app/Vulcan/troubleshooting-test

Please check that the nav bar looks decent and that the edit link works. You may have to click the link and then edit the URL to point at Cominor, because it only uses a relative URL. Deployed, that shouldn't be a problem, because we'll be on the `ifixit.com` domain.

Closes ifixit/ifixit#46529

### DIFF
--- a/frontend/templates/troubleshooting/index.tsx
+++ b/frontend/templates/troubleshooting/index.tsx
@@ -22,36 +22,37 @@ const Wiki: NextPageWithLayout<{
    layoutProps: DefaultLayoutProps;
 }> = ({ wikiData }) => {
    return (
+      <Flex direction="column" alignItems="center" width="100%" fontSize="16px">
          <NavBar />
-      <Flex
-         direction="row"
-         justifyContent="center"
-         width="100%"
-         fontSize="16px"
-      >
-         <Flex padding="0px 32px 32px" gap="48px" maxW="1280px" id="wrapper">
-            <Flex gap="16px" direction="column" id="main">
-               <Head>
-                  <meta name="robots" content="noindex" />
-               </Head>
-               <Heading as="h1">{wikiData.title}</Heading>
-               {wikiData.introduction.map((intro) => (
-                  <IntroductionSection key={intro.heading} intro={intro} />
-               ))}
-               {wikiData.solutions.map((solution, index) => (
-                  <SolutionCard
-                     key={solution.heading}
-                     index={index + 1}
-                     solution={solution}
-                  />
-               ))}
-               {wikiData.conclusion.map((conclusion) => (
-                  <ConclusionSection
-                     key={conclusion.heading}
-                     conclusion={conclusion}
-                  />
-               ))}
-            </Flex>
+         <Flex
+            padding="0px 32px 32px"
+            paddingLeft="48px"
+            paddingRight="48px"
+            gap="16px"
+            maxW="1280px"
+            direction="column"
+            id="main"
+         >
+            <Head>
+               <meta name="robots" content="noindex" />
+            </Head>
+            <Heading as="h1">{wikiData.title}</Heading>
+            {wikiData.introduction.map((intro) => (
+               <IntroductionSection key={intro.heading} intro={intro} />
+            ))}
+            {wikiData.solutions.map((solution, index) => (
+               <SolutionCard
+                  key={solution.heading}
+                  index={index + 1}
+                  solution={solution}
+               />
+            ))}
+            {wikiData.conclusion.map((conclusion) => (
+               <ConclusionSection
+                  key={conclusion.heading}
+                  conclusion={conclusion}
+               />
+            ))}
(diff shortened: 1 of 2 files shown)


frontend/templates/troubleshooting/index.tsx
```diff
@@ -46,6 +57,51 @@ const Wiki: NextPageWithLayout<{
    );
 };
 
+function NavBar() {
+   return (
+      <Flex
+         w="100%"
```
ardelato, line 56: I am not sure how much styling tweaking you are planning on doing for the early stages of Vulcan, but just to note, without some padding on the children elements. The breadcrumbs and area link text will ride the borders.

image

frontend/templates/troubleshooting/index.tsx
```diff
@@ -46,6 +57,51 @@ const Wiki: NextPageWithLayout<{
...
+         variant="link"
+         bgColor="transparent"
+         textColor="brand"
+         borderLeftColor="gray.200"
+         borderLeftWidth="1px"
+         borderRadius="0px"
+         padding="9px, 16px, 9px, 16px"
+      >
+         <Text
+            fontFamily="SF Pro"
```
ardelato: Do we want to specify a font family? I am completely out of the loop about our usage of fonts, but I thought we were going to use system fonts?

frontend/templates/troubleshooting/index.tsx
```diff
@@ -22,36 +22,37 @@ const Wiki: NextPageWithLayout<{
...
-               ))}
-               {wikiData.conclusion.map((conclusion) => (
-                  <ConclusionSection
-                     key={conclusion.heading}
-                     conclusion={conclusion}
-                  />
-               ))}
-            </Flex>
+         <Flex
+            padding="0px 32px 32px"
```
ardelato, line 20: I am little bit confused as to why you are setting the padding for right and left here but then overriding it directly with `paddingLeft` and `paddingRight`?

Why not set them here `padding="0px 48px 32px"`

frontend/templates/troubleshooting/index.tsx
```diff
@@ -70,14 +61,22 @@ function NavBar({ editUrl }: { editUrl: string }) {
          justify="center"
       >
          <Flex maxW="1280px" flexGrow="1">
-            <Box flexGrow="1">Breadcrumbs</Box>
-            <Box>Parts/Guide/Answers</Box>
+            <Breadcrumbs />
+            <AreaLinks />
```
ardelato, line 65: Are we going to insert the text content in `Breadcrumbs` and `AreaLinks` in the future?

ardelato (dismissed):
CR 👍 but deploy_block ☁️ on some questions.

frontend/templates/troubleshooting/index.tsx
andyg0808: It probably won't be just text content, but yes, we plan to have both those components eventually.

frontend/templates/troubleshooting/index.tsx
andyg0808: I'm fine with that right now, because they're empty. When I go to add stuff to them, I'll either need to add a `gap` on this flexbox or add padding to them.

frontend/templates/troubleshooting/index.tsx
andyg0808: Neither do I! I must have missed that there was already padding on this. I'll combine them and push a commit.

andyg0808: Should be fixed in d110a7a1ddf9bebac1a09429479772f626db6dbf.

frontend/templates/troubleshooting/index.tsx
andyg0808: I think this got fixed in a later commit.

andyg0808:
un_deploy_block :+1: 
Comments addressed.

frontend/templates/troubleshooting/index.tsx
ardelato: To confirm, we do want `32px` for left and right padding instead of `48px`?

ardelato (approved):
CR 👍

frontend/templates/troubleshooting/index.tsx
andyg0808: Yes: looking back at the mockup, I'm pretty sure that's the correct value.

k0rvusk0r4x:
- Header is present.
- Edit button is present.
- Button goes to the correct place.
- Checked on various browsers.

QA 😺

PR #1380, opened 2023-02-17
@ifixit/breadcrumbs: Require types as dev dependency

andyg0808:
This adds the lodash types as a dev dependency to `@ifixit/breadcrumbs`

qa_req 0
This is dev-only.


jarstelfox (approved):
CR 🌵

PR #1381, opened 2023-02-18
Hot Fix: Make Runners Explicitly use Node Version 16

ardelato:
## Description

The default node version for the runners was changed to 18, which is not compatible with our current setup. Unfortunately, this change was only tracked as an issue and was not announced on GitHub's blog (https://github.blog/changelog/label/actions/), which is why we did not notice it. 

In order to fix this, we need to explicitly set the node version to 16 for all our workflows that use node/pnpm.

## QA Notes:

qa_req 0 CI should pass

References:
https://github.com/actions/runner-images/issues/7002
https://ifixit.slack.com/archives/C02SMF0AZPF/p1676677777249139

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -13,6 +13,12 @@ jobs:
            uses: actions/checkout@v2
            with:
               token: ${{ secrets.ACTION_PAT }}
+
+         - name: Install Node.js
+           uses: actions/setup-node@v3
+           with:
+              node-version: 16
+
          - name: Setup and cache pnpm
            uses: ./.github/actions/pnpm-setup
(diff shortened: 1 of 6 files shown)


mlahargou (approved):
CR 📱 

Thanks for hoping on this!

danielbeardsley:
Some of our tests are failing because part of the next-js install process downloads a binary from `sharp` (an image processing module) and that service only payed for so much egress data:

`sharp: Installation error: Status 503 Egress is over the account limit.`

At least that's how it seems. 

That isn't the fault of this pull, though it'd be nice to depend on them less so we aren't downloading very much or anything during CI.

CR :+1:

Issue #1386, opened 2023-02-20
Playwright: Fix Cross Sell Test


ardelato:
## Description

The cross-sell test validates whether some cross-sell products have, or have not, been added to the cart drawer by seeing if the product title is contained in an element of the cart drawer. The problem is the element we target includes the `upsell` component. As such, if we validate that a cross-sell product is not included in our cart, we may mistakenly see the product title in the `upsell` component, thus unexpectedly failing the test.

![image](https://user-images.githubusercontent.com/22064420/220195839-a1f426fb-bc3c-4d7c-9171-e664a99cbe8b.png)

![image](https://user-images.githubusercontent.com/22064420/220195773-38eaa619-8724-40b1-b124-55536fc5793f.png)

### To Do:

- [ ] Ensure we only validate against the cart list items and not the entire cart body.
- [ ] Update the `cart-drawer` extension to allow passing a `product title` to assert if an item is in the list instead of only by `SKU`s

ardelato:
closing in favor of https://github.com/iFixit/react-commerce/issues/1384

PR #1385, opened 2023-02-20
Correct cross sell test

federicobadini:
closes #1384 

This PR introduces a new `data-testid` that targets the line items container.
The upsell component was also moved outside the line-items `ul`.

### QA

1. Visit Vercel preview
2. Verify that there is no visual changes to the upsell banner inside the cart drawer
3. Verify that the cross sell tests are now running correctly

PR #1387, opened 2023-02-20
Playwright: Ignore Upsell Product for Cross Sell Test

ardelato:
## Description

We were grabbing the text of the entire cart drawer body which includes the text of the upsell product component. This became a problem when the upsell product was one of the products of the cross-sells. As such, when we were validating that the cross-sell product title was not in the cart drawer body we would find it as part of the upsell product and fail the test.

By narrowing down the selector to only grab the `li` elements we will avoid grabbing the text of the upsell product component. In doing so, we need to update the `cartDrawerText` variable name to `cartDrawerItems` to better reflect what it is. Furthermore, we need to update the assertion to  `toContainText` instead of `toContain` as the latter only works with a single string and the former can work with an array of strings.

### QA

qa_req 0
- [ ] CI should now pass

**cc:** @mlahargou 

Connects: #1386

### DIFF
--- a/frontend/tests/playwright/product/cross_sell.spec.ts
+++ b/frontend/tests/playwright/product/cross_sell.spec.ts
@@ -30,9 +30,11 @@ test.describe('Cross-sell test', () => {
          }
 
          // Deselect other cross-sell products
-         otherProductTitles.push(
-            await product.getByTestId('cross-sell-item-title').textContent()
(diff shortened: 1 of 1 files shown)


ardelato:
Closing in favor of https://github.com/iFixit/react-commerce/pull/1385.

PR #1398, opened 2023-02-22
Correct zoom container size

### LINKED ISSUE #1397: Fix zoom container width regression
Zoom container width was calculated by computing a difference between the page width and the right column width.
The new wrapper caused a regression since the zoom container is slightly larger than the page itself.

zoom-container

federicobadini:
close #1397 

### QA

1. Visit [Vercel preview](https://react-commerce-git-fix-zoom-wrapper-regression-ifixit.vercel.app/products/iphone-7-plus-screen)
2. Verify that the zoom container is now sized correctly and is not overflowing anymore outside of the viewport

### DIFF
--- a/frontend/templates/product/sections/ProductSection/index.tsx
+++ b/frontend/templates/product/sections/ProductSection/index.tsx
@@ -95,8 +95,8 @@ export function ProductSection({
                   top="0"
                   w={{
                      md: '320px',
-                     lg: 'calc(400px + 16px + ((100vw - 960px) / 2 - 24px))',
-                     xl: 'calc(400px + 16px + ((100vw - 1100px) / 2 - 24px))',
+                     lg: 'calc(16px + 400px + 32px - 24px)',
+                     xl: 'calc(16px + 400px + 32px + (100vw - 1280px) / 2 - 24px)',
                   }}
                   left="calc(100% + 24px)"
                   boxShadow="md"
--- a/frontend/templates/product/sections/ProductSection/index.tsx
+++ b/frontend/templates/product/sections/ProductSection/index.tsx
@@ -93,6 +93,7 @@ export function ProductSection({
                   id="zoom-container"
                   position="absolute"
                   top="0"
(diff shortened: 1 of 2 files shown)


masonmcelvain (dismissed):
CR ✌🏻 Seems reasonable but it's hard for me to know where the values for this calculation come from. From the context, it seems like `400px` comes from the product info section box, and `1280px` (aka `80rem`) comes from the `<Wrapper>` component maxW. I assume the other values come from margin or padding across various components. I trust that it works, but if I had to refactor this code myself in the future, I would have a hard time tracking down the origin of each value.

masonmcelvain:
@federicobadini messaged me a great explanation of the CSS width calculations in this PR, reposting it here for others to see:

<details>

> For completeness: the zoom needed to appear as an overlay spaced 24px right from the gallery container until 24px to the right end of the screen.
> Since the element is placed inside the gallery we start by placing it left="calc(100% + 24px)"
> The gallery is separated from the right information column by a 40px margin and since we start from 24px we start adding the remaining 16px to the width
> The right information column is 400px wide on lg and xl  so we sum that too.
> The page wrapper element has a 32px padding so we sum that too.
> On lg screen (below 1280px) that's it, we just go back 24px from the right margin
> On xl screen we have to take into account that the wrapper has a max-width=1280 so we have also to sum the delta between 100vw and 1280px and again then we subtract 24px for the right margin

Screenshot 2023-02-22 at 17 42 27

</details>

k0rvusk0r4x:
The zoom container no longer overflows the viewport.

QA 😸 

> @federicobadini messaged me a great explanation of the CSS width calculations in this PR

Would it be possible to put this, or a more concise version if possible, into a comment with the code change. I'm imagining five years down the line needing to remember the calculation and having to find that comment to explain it.

deploy_block 😿 on trying to get the explanation into a comment.

masonmcelvain (approved):
CR ✌🏻 QA ✌🏻 un_deploy_block ✌🏻 Good idea @k0rvusk0r4x

PR #1407, opened 2023-02-23
Memory Cache: don't clear on vercel_no_cache

danielbeardsley:
We are using Alertra to monitor uptime and we add a _vercel_no_cache=1 query param to ensure we actually run app code. We don't want this to constantly clear the in-memory cache because we have alerta checking twice a minute.

The in-memory cache only stores a few global settings and it expires every hour.

See comment here: https://github.com/iFixit/ifixit/issues/46312#issuecomment-1442379352

qa_req 0

### DIFF
--- a/frontend/helpers/next-helpers.ts
+++ b/frontend/helpers/next-helpers.ts
@@ -2,7 +2,6 @@ import { PROD_USER_AGENT } from '@config/constants';
 import { CACHE_DISABLED } from '@config/env';
 import { setSentryPageContext } from '@ifixit/sentry';
 import { withTiming } from '@ifixit/helpers';
-import { clearCache } from '@lib/cache';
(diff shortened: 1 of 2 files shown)


frontend/helpers/next-helpers.ts
```diff
@@ -2,7 +2,6 @@ import { PROD_USER_AGENT } from '@config/constants';
 import { CACHE_DISABLED } from '@config/env';
 import { setSentryPageContext } from '@ifixit/sentry';
 import { withTiming } from '@ifixit/helpers';
-import { clearCache } from '@lib/cache';
```
masonmcelvain, line 5: It looks like this was the only usage of this `clearCache` function, so we could drop the definition too.

masonmcelvain (approved):
CR ✌🏻

Issue #1392, opened 2023-02-21
Playwright Strapi: Investigate Flakiness 


ardelato:
## Description

There are times when running the `Playwright` tests because the tests fail to talk to the `Strapi` backend: https://github.com/iFixit/react-commerce/actions/runs/4230450839/jobs/7350631355#step:12:852

We are not sure if this is a problem with spinning up multiple `strapi` servers as part of a matrix or some other factors.

### To Do

- [ ] Investigate if GH runners in a matrix share any resources
- [ ] Export the `docker logs` of the `strapi` server when playwright tests fail so we can gain insight what was going on in the case where the `strapi` server was having problems

mlahargou:
So I figured out some interesting things. 

Firstly, we only see these strapi errors on job reruns. 
- First run (successful): https://github.com/iFixit/react-commerce/actions/runs/4257656032/jobs/7408032866
- Second run (failure): https://github.com/iFixit/react-commerce/actions/runs/4257656032/jobs/7408197758

That lead me to think that there was some state that gets transferred from the first job that causes the second to fail. After digging into the workflow, I noticed that we are caching the entire `frontend/.next` directory. The second run drops the entire `frontend/.next` directory in from the first run and **does not** run `pnpm build`.

I did a little research and noticed that most other workflows [only cache the `.next/cache` directory](https://nextjs.org/docs/advanced-features/ci-build-caching#github-actions) but will still build the app. I did that and things behaved like normal on both runs. We do get a bit of a slowdown because we call `pnpm build` every time, but I think that's worth it.

For reference, here is my test branch: https://github.com/iFixit/react-commerce/compare/playwright--spit-out-docker-logs-for-strapi

The caching isn't quite working b/c it's not respecting the js/ts hashes, but I think that should be an easy fix?
![image](https://user-images.githubusercontent.com/29667789/221070763-8462b823-00fc-4bbd-8c34-a455f0f9303e.png)
![image](https://user-images.githubusercontent.com/29667789/221070983-fd1b4cd0-f798-43ce-9659-3b526e720fcf.png)

---

@ardelato I'm leaning towards building every time and just caching the `.next/cache` directory. Thoughts?

PR #1410, opened 2023-02-24
Playwright: Only cache the nextjs cache dir

mlahargou:
All reruns were failing b/c we were not building the nextjs app (`pnpm build`) on reruns:
- First run (successful): https://github.com/iFixit/react-commerce/actions/runs/4257656032/jobs/7408032866
- Second run (failure): https://github.com/iFixit/react-commerce/actions/runs/4257656032/jobs/7408197758

That lead me to think that there was some state that gets transferred from the first job that causes the second to fail. After digging into the workflow, I noticed that we are caching the entire `frontend/.nex`t directory. The second run drops the entire `frontend/.next` directory in from the first run and does not run `pnpm build`.

I did a little research and noticed that most other workflows [only cache the .next/cache directory](https://nextjs.org/docs/advanced-features/ci-build-caching#github-actions) but will still build the app. I did that and things behaved like normal on both runs. We do get a bit of a slowdown because we call `pnpm build` every time, but I think that's worth it.

QA:
--
Once CI finishes, rerun a playwright test. We shouldn't see any of the `FetchError: request to http://localhost:1337/graphql failed` errors that we saw in the issue.

cc @ardelato 

Closes: #1392

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -53,18 +53,17 @@ jobs:
            if: steps.playwright-cache.outputs.cache-hit != 'true'
            run: cd frontend && npx playwright install
 
-         - name: Cache Next.js build
+         - name: Cache Next.js cache directory
            uses: actions/cache@v3
-           id: nextjs-cache
            with:
-              path: frontend/.next
-              key: ${{ github.ref_name }}-nextjs-${{github.sha}}
+              path: frontend/.next/cache
+              # Generate a new cache whenever packages or source files change.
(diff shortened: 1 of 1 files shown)


dhmacs:
@mlahargou CI fails due to the bug being fixed here https://github.com/iFixit/react-commerce/pull/1412

mlahargou:
QA 📱 

The rerun is passing tests: https://github.com/iFixit/react-commerce/actions/runs/4284062397/jobs/7461434521

ardelato (approved):
CR 💾

Issue #1416, opened 2023-02-27
Playwright: Deflake Parts Page Search Test


ardelato:
## Description

The `parts_page_search.spec.ts` test has been flaky and may be a result of the response getting captured but the Frontend state not getting updated in time when we reach the assertion.

https://github.com/iFixit/react-commerce/blob/main/frontend/tests/playwright/product-list/parts_page_search.spec.ts#L19-L22

### To Do:

- [ ] Possibly add a `waitForURL` before the assertion so we give the Frontend a chance to update.
	- We should probably test while throttling the network and CPU for the Playwright test to confirm it has been resolved.

**cc:** @federicobadini

PR #1419, opened 2023-02-28
Playwright: Deflake Parts Page Search Test

ardelato:
## Description

This test is flaky as we wait for the algolia query to finish but this event does not mean that the page has fully been updated. As a result, we sometimes fail since we have not updated the url after the query has finished.

This flakiness became more of an issue in https://github.com/iFixit/react-commerce/pull/1375. Locally the test passes for the branch. However, when we run the entire test-suite the flakiness of the test increases. This means when the machine is under higher loads the test begins to fail more often. 

To fix this, we will instead wait for the url to be updated with a query parameter after we enter the search text but before the assertion is made. This will give the page enough time to update the URL. 

### QA

qa_req 0 CI should continue to pass.

I confirmed this helped to deflake the test by running the following command to overload the playwright workers:

```
pnpm playwright:run parts_page_search --repeat-each=5 --workers=5
```

|Before|After|
|:---:|:---:|
|  image  | image |

**cc:** @federicobadini 

Closes: #1416

### DIFF
--- a/frontend/tests/playwright/product-list/parts_page_search.spec.ts
+++ b/frontend/tests/playwright/product-list/parts_page_search.spec.ts
@@ -1,4 +1,3 @@
-import { waitForAlgoliaSearch } from './../utils';
 import { test, expect } from '../test-fixtures';
 
 test.describe('parts page search', () => {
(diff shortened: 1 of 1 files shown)


federicobadini (approved):
CR 👍

PR #1423, opened 2023-03-01
Breadcrumbs: Allow mobile to be dynamic

andyg0808:
This is ported from our main monorepo.

Original issue: https://github.com/iFixit/ifixit/pull/46667

qa_req 0
We'll test this when we integrate the breadcrumbs from here in the monorepo.


aburke07 (approved):
CR :alien:

hcutrone (approved):
CR :burrito:

PR #1444, opened 2023-03-07
Add menu3 to strapi store

aburke07:
1. This copies over the autogenerated changes from https://github.com/iFixit/react-commerce/pull/1342.
2. We need to add menu3 to strapi before trying to access it in the frontend.

## QA
Qa_req 0 - Tests should cover any regressions.


hcutrone (approved):
CR :burrito:

Issue #1415, opened 2023-02-27
Product page: setEcommerceView event needs to come before trackPageView


addison-grant:
This is the iFixit/react-commerce corresponding issue to https://github.com/iFixit/shopify-v2/issues/995 and https://github.com/iFixit/ifixit/issues/46724.

We need to check if this is relevant to this repository. 

The `setEcommerceView` method needs to happen before the `trackPageView` method.

QA notes
===

At the beginning of viewing a web page, two requests go out to Matomo:
1. to fetch the tracking JS code
2. to track a page view

The page view tracking request on product pages currently do not include the SKU, (full) product name, and the price (converted to USD). This should amend that.

addison-grant:
Example where it is working (fixed in https://github.com/iFixit/shopify-v2/pull/994): https://store.ifixit.de/products/irobot-roomba-s9-s9plus-battery 

<img width="1031" alt="image" src="https://user-images.githubusercontent.com/10889879/223490295-ee0fa142-e5c4-48c0-b180-b66d24110455.png">

addison-grant:
Example where it is not working: https://www.ifixit.com/products/steam-deck-fan

Scrolling through the payload, one does not see the necessary product parameters.

<img width="1046" alt="image" src="https://user-images.githubusercontent.com/10889879/223491665-47fe3955-b6e0-4100-80cb-beeb9fc1e767.png">

addison-grant:
I wasn't sure how to tell if GA is working, but I think not. I am trying out https://chrome.google.com/webstore/detail/google-analytics-debugger/jiokjpcfnipnicblkgdklknjjeohnfln/related?hl=en-US, and comparing the shopify-v2 product page to the react-commerce product page:

Shopify-v2 | react-commerce
--- | ---
<img width="1387" alt="image" src="https://user-images.githubusercontent.com/10889879/223496153-9aac2c75-0c8c-4a63-a3df-68479dd0ed97.png"> | <img width="1389" alt="image" src="https://user-images.githubusercontent.com/10889879/223496288-55d10d43-ad85-4d96-8bd4-7657884b513b.png">

Note the lack of porduct view event on react-commerce

addison-grant:
Relevant files/directories:

- [frontend/templates/product/index.tsx](https://github.com/iFixit/react-commerce/blob/8e291cc4054071f8bede0470d9d171c237502798/frontend/templates/product/index.tsx#L7)
[...]

PR #1447, opened 2023-03-07
Define _paq on window if it isn't yet defined: fixes setEcommerceView issue

addison-grant:
...and add `_paq` to the window interface. (Not sure if the latter is necessary, but I was getting messages from my editor about the `window` object not having the property or something like that.)

Do this inside an exported function that encapsulates the `_paq = []` definition and `_paq.push` functionality.

> `Property '_paq' does not exist on type 'Window & typeof globalThis'.`

The `setEcommerceView` functionality didn't seem to be working properly before because `_paq` wasn't being created on the global `window` object if `_paq` wasn't yet defined on it.

https://github.com/iFixit/react-commerce/blob/matomo--fix-set-ecommerce-view/packages/analytics/matomo/index.tsx#L34-L37

# QA notes
At the beginning of viewing a web page, two requests go out to Matomo:

- to fetch the tracking JS code
- to track a page view

The page view tracking request on **product pages currently do not include the SKU, (full) product name, and the price** (converted to USD). This should amend that.

Closes https://github.com/iFixit/react-commerce/issues/1415


packages/analytics/matomo/usePaq.ts
```diff
@@ -1,6 +1,14 @@
+declare global {
+   interface Window {
+      _paq: any;
```
danielbeardsley, line 3: I think it should be `{push: (args: array) => void}` or however it's expressed in typescript because we use it like `_paq.push([...])`. Does this make sense?

packages/analytics/matomo/usePaq.ts
```diff
@@ -1,6 +1,14 @@
...
+   }
+}
+
 export function usePaq(): any[] | undefined {
    if (typeof window === 'undefined') {
       return undefined;
    }
-   return (window as any)._paq;
+
+   var _paq = (window._paq = window._paq || []);
```
jarstelfox, line 12: Although this works, it seems to spy a larger bug to me.

We are trying to use matomo *before* we have even included matomo on the page.

packages/analytics/matomo/usePaq.ts
addison-grant: I think so. Because either it's an array with `push` or because it's the initialized object from Matomo that has `push` defined on it.

addison-grant: Are you saying I can put `{push: (args: array) => void}` for the type?

packages/analytics/matomo/usePaq.ts
danielbeardsley, line 12: Oh wait... this may be troublesome cause callers may capture this value in a closure (it's really easy to do accidentally).

I'm tempted to say we should return a `pushToPaq(args: array): void` function that will do this dynamic lookup each time it's called. Thoughts?

addison-grant: I think it's okay. The Matomo tracking object takes whatever you've added the to the `_paq` array and adopts it and the tracking object has a `push` method that sort of masks over it.

danielbeardsley: That's always been the way that analytics platforms work (GA, matomo, pwik, ...). They use an array interface so `window._paq` starts as an array and matomo then (when it loads) replaces that with something that exposes a `push()` method just like array does and at the same time consumes the already pushed calls.

danielbeardsley:
deploy_block :+1:  on considering my suggestion.

packages/analytics/matomo/usePaq.ts
jarstelfox: Agreed, but I think this indicates matomo needs to be included higher on in the layout template IMO

packages/analytics/matomo/usePaq.ts
```diff
@@ -1,6 +1,14 @@
+declare global {
```
danielbeardsley, line 1: Forgot to delete this file.

addison-grant: https://github.com/iFixit/react-commerce/pull/1447/commits/580e167ca687a70c98ee0b35b8c3688cf6c78177

danielbeardsley (approved):
CR :+1:

packages/analytics/matomo/usePaq.ts
masonmcelvain: Next.js [recommends](https://nextjs.org/docs/api-reference/next/script#afterinteractive) that analytics scripts be loaded after the page becomes interactive. We're currently following this suggestion, but we could experiment with running the script `beforeInteractive` if it's important to us.

https://github.com/iFixit/react-commerce/blob/34257205922f58e7628a0f9a559e7dfd6d1febd2/frontend/components/analytics/Matomo.tsx#L19

packages/analytics/matomo/matomoPush.ts
```diff
@@ -0,0 +1,14 @@
+declare global {
+   interface Window {
+      _paq?: { push: (args: any[]) => void };
+   }
+}
```
masonmcelvain, line 5: Might be better long-term to put this in a `global.d.ts` file, but definitely not a blocker for this. IIRC you can't `declare global` more than once in a package but I could be wrong

masonmcelvain (approved):
CR ✌🏻

packages/analytics/matomo/usePaq.ts
jarstelfox: Oh man, I wasn't aware of the nuance of the Next.js script loading here.

If I (briefly) ignore how next.sj works. I totally agree the rendered script should be something like this: `<script async="" src="...matomo.js"></script>`

This script should be de-prioritized for page loading, but high enough in the dom that the parsing (and background downloading) happens as soon as possible.

My apologies here, but I am not well versed in how the script component exactly ends up rendered in this case.

For clarity here:

It is totally fine for us to push the array before we have included the matomo script tag on the page. The fact that we hit a case where the global array was not yet defined seems to indicate that we have code that ran before the matomo script in the dom.

This pull will address the problem of the array being undefined, which is great. My claim is we probably can move the matomo script rendering higher up in the dom tree.

addison-grant:
@iFixit/qa-all let me know if you have any questions.

jordycosta:
@addison-grant would this be in the web-console network requests upon loading a product page?

deltuh-vee:
QA 🎬 
The page tracking request now includes:
* the price in USD
* the sku
* and the full product name


image

PR #1449, opened 2023-03-08
Restructure product page next data

### LINKED ISSUE #1443: Improve __NEXT_DATA__ for product pages
This is an analysis of html node sizes for a page with just one product variant.

Screenshot 2023-03-07 at 09 29 59

As the number of variant increases the `__NEXT_DATA__` json becomes the predominant element of the page

Screenshot 2023-03-07 at 09 30 20

We can note two things here: 

1. We are almost duplicating data to provide `variants` and `allVariants` as well as `images` and `allImages`
2. By looking inside either `variants` or `allVariants` we can notice that each element contains its related crossSellVariants with all their details. This means that when a high number of variants shares the same cross sell recommendation we are repeating the same data over and over.

Screenshot 2023-03-07 at 10 13 42

The need to have both the `variants` and the associated `allVariants` property may be substituted via a flag that defines which variant are active and which ones are not.
The relation between a variant and its associated crossSell recommendations could be done via id, allowing to keep the list of cross sell products in a separate place without any duplicate

connects https://github.com/iFixit/react-commerce/pull/1092#issuecomment-1455270179

federicobadini:
connects https://github.com/iFixit/react-commerce/pull/1092#issuecomment-1455270179
closes #1443 

Cross sell variants are now grouped at product level and referenced by id.
The duplicated structures were removed - according to how many active variants we have we pass down the data and the images associated to the active elements (`> 0` active variants) or the one associated with the first variant (`= 0` active variants)

### QA

1. Visit [Vercel preview](https://react-commerce-git-restructure-product-page-next-data-ifixit.vercel.app/products/iphone-6s-screen)
2. Check that there is no change in how product pages or cross sell section behave wrt production
3. Check also that products with disabled variants and products with no variants available for sale are still working as they do in production

### DIFF
--- a/frontend/components/common/CompatibleDevice.tsx
+++ b/frontend/components/common/CompatibleDevice.tsx
@@ -11,7 +11,7 @@ export function CompatibleDevice({
    device,
    maxModelLines = 0,
 }: CompatibleDeviceProps) {
-   const variants = React.useMemo(() => device.variants, [device.variants]);
+   const variants = device.variants;
    const [visibleVariants, hiddenVariantCount] = React.useMemo(
       () =>
          maxModelLines > 0 && variants.length > maxModelLines
--- a/frontend/components/common/CompatibleDevice.tsx
+++ b/frontend/components/common/CompatibleDevice.tsx
@@ -11,10 +11,7 @@ export function CompatibleDevice({
    device,
    maxModelLines = 0,
 }: CompatibleDeviceProps) {
-   const variants = React.useMemo(
-      () => device.variants.filter(Boolean),
-      [device.variants]
-   );
+   const variants = React.useMemo(() => device.variants, [device.variants]);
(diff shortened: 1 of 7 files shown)


masonmcelvain (dismissed):
CR ✌🏻 Great optimization!

federicobadini:
dev_block 🛑  to verify [this](https://github.com/iFixit/react-commerce/pull/1449#discussion_r1131248888)

masonmcelvain (dismissed):
CR ✌🏻 Thanks for the change! un_dev_block ✌🏻

deltuh-vee:
QA 🎬 
* The product pages and cross sell sections are still the same as on production
*  products with disabled variants and products with no variants available for sale are still working

danielbeardsley:
Did we measure the affect on the JSON size for some example pages?

I'm excited to see how much of a change this produced.

PR #1451, opened 2023-03-08
Troubleshooting: Add some frontend tests

andyg0808:
I realized there were some new features which were easy enough to add tests for.

qa_req 0
This only adds tests.

Connects https://github.com/iFixit/ifixit/issues/46529
Connects https://github.com/iFixit/ifixit/issues/46560


djmetzle (approved):
CR :link:

Issue #1414, opened 2023-02-27
Update "Frequently Bought Together" Cross Sell Event


addison-grant:
Parent issue https://github.com/iFixit/ifixit/issues/46547

<details><summary> Quote </summary>

> ## Problem
> We need to fix the way our Cross Sell events are titled. This will require an update to our Events sheet. The old legacy cross sell events were titled by the listing first, then the additional cross sells after the slash:
> 
> ![cross sell2](https://user-images.githubusercontent.com/66840433/219199705-45a8bffb-4e63-427d-926e-15828c438a1c.png)
> 
> The new version of the Cross Sell event title has the SKUs out of order, so it is difficult to tell which listing these cross sells are on:
> 
> ![cross sell1](https://user-images.githubusercontent.com/66840433/219199878-3f8f0a31-90bc-4275-870e-ed7859a81418.png)
> 
> ## Do This
> From John B:
> 
> > * So, this is the current definition of the Cross Sell event:
> > * Add to Cart - Frequently Bought Together - $productcode-$optionid
> > * And we should edited this way:
> > * Add to Cart - Frequently Bought Together - $currentProductSku/$relatedSku,$relatedSku...
> 
> Update the "Frequently Bought Together" Cross Sell Events titling format to match John B's spec in the [Events Spreadsheet](https://docs.google.com/spreadsheets/d/16Xas1pHf_K7FmfOX3yX2mYgeYr2DIENr8cyQp19FZkA/edit?pli=1#gid=1839136207&fvid=1843464925).
> 
 [ ... ]
> 
> ## Background
> cc: @addison-grant @jeffsnyder

</details>

addison-grant:
CC @sterlinghirsh

PR #1452, opened 2023-03-08
Include current itemcode in Frequently Bought Together event

addison-grant:
This allows the E-Commerce team to see what variant "Frequently Bought Together" items are associated with. Currently in Matomo, you can see the URL, but that doesn't include the variant.

See row 223 of [iFixit EVENTS Spreadsheet (updated 03-08-2023)
](https://docs.google.com/spreadsheets/d/16Xas1pHf_K7FmfOX3yX2mYgeYr2DIENr8cyQp19FZkA/edit?pli=1#gid=1839136207&fvid=1843464925) for spec details.

A little discussion in the comments of this PR around the spec: https://github.com/iFixit/react-commerce/pull/1452#discussion_r1130094635

Closes https://github.com/iFixit/react-commerce/issues/1414

### DIFF
--- a/packages/analytics/index.tsx
+++ b/packages/analytics/index.tsx
@@ -42,8 +42,10 @@ export function trackAddToCart(
    const itemcodes =
       addToCartInput.type === 'product'
          ? addToCartInput.product.itemcode
-         : `${addToCartInput.bundle.currentItemCode}/`
-         + `${addToCartInput.bundle.items.map((item) => item.itemcode).join(', ')}`;
+         : `${addToCartInput.bundle.currentItemCode}/` +
+           `${addToCartInput.bundle.items
+              .map((item) => item.itemcode)
+              .join(', ')}`;
    trackInMatomoAndGA({
       eventCategory: event,
       eventAction: `${event} - ${itemcodes}`,
--- a/packages/analytics/index.tsx
+++ b/packages/analytics/index.tsx
@@ -42,7 +42,8 @@ export function trackAddToCart(
    const itemcodes =
       addToCartInput.type === 'product'
          ? addToCartInput.product.itemcode
-         : addToCartInput.bundle.items.map((item) => item.itemcode).join(', ');
+         : `${addToCartInput.bundle.currentItemCode}/`
+         + `${addToCartInput.bundle.items.map((item) => item.itemcode).join(', ')}`;
    trackInMatomoAndGA({
       eventCategory: event,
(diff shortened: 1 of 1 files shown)


packages/analytics/index.tsx
```diff
@@ -42,7 +42,10 @@ export function trackAddToCart(
    const itemcodes =
       addToCartInput.type === 'product'
          ? addToCartInput.product.itemcode
-         : addToCartInput.bundle.items.map((item) => item.itemcode).join(', ');
+         : `${addToCartInput.bundle.currentItemCode}/` +
+           `${addToCartInput.bundle.items
+              .map((item) => item.itemcode)
+              .join(', ')}`;
```
masonmcelvain, line 48: The spec has the format `$currentProductSku/$currentProductSku,$relatedSku1,$relatedSku2,...`, and this implementation looks slightly different. Specifically, the current sku might not be the first sku after the `/`, and the delimiter here is `, ` instead of `,`.

masonmcelvain (approved):
CR ✌🏻 My comment isn't a blocker, I'm just trying to understand the expected behavior

packages/analytics/index.tsx
addison-grant: I think the spec can be a little confusing because there's an explanation for this all the way to the right in the Notes:

> "$currentProductSku in the first position indicates the product page.
> $currentProductSku in the second position should be included in the Event Action only if the product is added to cart.
> 
> SITUATION 1
> Let's assume the event in the report is in the format below:
> Add to Cart - Frequently Bought Together - $currentProductSku/$currentProductSku,$relatedSku1,$relatedSku2
> 
> The meaning:
> Both the current product and the related products were added to cart.
> - currrent product
> - related product 1
> - related product 2
> 
> 
> SITUATION 2
> Now let's assume the event in the report is in this format (without $currentProductSku after the slash):
> Add to Cart - Frequently Bought Together - $currentProductSku/$relatedSku1,$relatedSku2
> 
> The meaning:
> Only the related products of the current product were added to cart. 
> - related product 1
> - related product 2"

I think that should answer the question, but let me know if not!

masonmcelvain: Thank you, I totally missed that! This implementation seems on point then 👍🏻

jordycosta:
QA 🟢 

The itemcode is shown for the frequently bought together items after clicking on `Add to cart` in that section ✅ 

FBT

matomo

PR #1448, opened 2023-03-08
Refactor product page featured products section

dhmacs:
closes #1434 
closes #1206 

## QA

1. Open [product page preview](https://react-commerce-git-refactor-product-page-featured-dc6043-ifixit.vercel.app/products/repair-business-toolkit)
2. Verify that the featured products section adheres to [mockup](https://www.figma.com/file/jRG4EyQXRDMGfyLqut08KP/iFixit---Working-files?node-id=6650-80646&t=lm5sZ8asm7PGe46K-4) (Note: The title is the same of production, and the description in the default scenario is not shown. This is something that will be configurable when the section config will come from Strapi)

    Screenshot 2023-03-08 at 17 31 51

   > 👁️ Note that on wider screens the section displays 6 columns to satisfy #1206 requirements

3. Verify that the section is responsive

PR #1455, opened 2023-03-09
Readd menu3 to strapi

aburke07:
I'm not sure how this was removed in the last few days, but we still need this for #1342 to pass CI.

After this deploys, I think we need to redeploy strapi again.

Qa_req 0

### DIFF
--- a/frontend/lib/strapi-sdk/generated/sdk.ts
+++ b/frontend/lib/strapi-sdk/generated/sdk.ts
@@ -2667,6 +2667,118 @@ export type FindStoreQuery = {
                      } | null;
                   } | null;
                } | null;
+               menu3?: {
+                  __typename?: 'MenuEntityResponse';
+                  data?: {
+                     __typename?: 'MenuEntity';
+                     attributes?: {
+                        __typename?: 'Menu';
+                        title: string;
+                        items: Array<
+                           | {
+                                __typename: 'ComponentMenuLink';
+                                name: string;
+                                url: string;
+                                description?: string | null;
+                             }
+                           | {
+                                __typename: 'ComponentMenuLinkWithImage';
+                                name: string;
+                                url: string;
+                                image: {
+                                   __typename?: 'UploadFileEntityResponse';
+                                   data?: {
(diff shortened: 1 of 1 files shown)


hcutrone (commented):
CR :burrito: but deploy_block :burrito: - should we figure out why this was removed? Will it happen again?

aburke07:
> should we figure out why this was removed? Will it happen again?

@danielbeardsley Found it here: https://github.com/iFixit/react-commerce/pull/1448#discussion_r1131383164

It was just accidentally removed.

Un_deploy_block :alien:

masonmcelvain:
Doesn't this cause a diff every time you run a dev server? deploy_block ✌🏻

aburke07:
> Doesn't this cause a diff every time you run a dev server? deploy_block ✌🏻

The diff I'm getting is adding back this menu3 type. Wouldn't adding this to main remove the diff? I'm not sure why this is adding a diff in the first place either.

aburke07:
> I added to #1342 where we query for it, so lets see if that passes CI. If so, we can close this pull.

Vercel deployed successfully there, so adding menu3 on that pull seems like the way to go.

danielbeardsley:
> The diff I'm getting is adding back this menu3 type. 

The SDK is built off whatever is in your strapi DB, is it not? Strapi itself will I think adjust it's database (does it remove columns too??) to whatever is defined in the schema files in the backend each time it's booted.

masonmcelvain:
> Hmm, I thought the strapi type needed to be updated before we query for it?

Ah, I meant that we needed to add the field to strapi in `backend/`, which you did in #1444 in `backend/src/components/store/footer.json`. The types in `frontend/lib/strapi-sdk/generated/sdk.ts` get autogenerated based on what we're querying for in `frontend/lib/strapi-sdk/operations` I think. So, when we start querying for `menu3` in `findStore.graphql`, that's when we'd expect the type of `menu3` to be added to the generated type schema.

masonmcelvain:
This [file](https://github.com/iFixit/react-commerce/blob/main/frontend/codegen.yml) controls the schema generation, for reference

Issue #1276, opened 2023-01-20
Write a test for empty search query


masonmcelvain:
We recently fixed a bug where emptying the search query didn't update the url properly.

Let's write a test to prevent regression.

Followup to https://github.com/iFixit/react-commerce/issues/1191

PR #1458, opened 2023-03-09
Tests: Add search for empty query test

batbattur:
## Summary
Added a test that checks if the URL is updated after clearing the search text.
It makes sure the URL is updated and has no query parameters such as `?q=...`.

## QA notes
Passing CI is enough.

closes https://github.com/iFixit/react-commerce/issues/1276
qa_req 0

### DIFF
--- a/frontend/tests/playwright/product-list/parts_page_search.spec.ts
+++ b/frontend/tests/playwright/product-list/parts_page_search.spec.ts
@@ -45,24 +45,22 @@ test.describe('parts page search', () => {
       await expect(page.getByTestId('product-list-no-results')).toBeVisible();
    });
 
-   test.only("Should update url on empty search query", async ({
-    page,
- }) => {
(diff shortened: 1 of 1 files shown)


frontend/tests/playwright/product-list/parts_page_search.spec.ts
```diff
@@ -44,4 +44,23 @@ test.describe('parts page search', () => {
       await expect(page.getByTestId('list-view-products')).not.toBeVisible();
       await expect(page.getByTestId('product-list-no-results')).toBeVisible();
    });
+
+   test.only('Should update url on empty search query', async ({ page }) => {
```
batbattur, line 48:
```suggestion
   test('Should update url on empty search query', async ({ page }) => {
```

ardelato (approved):
CR 👍

PR #1459, opened 2023-03-10
Github Actions: Add stress-test workflow

### LINKED ISSUE #1431: Stress Test the Tests
## Description

We are now in a more stable place for our testing infrastructure.

We have a good set of tests that cover a majority of our business logic[^1]. We now have an easy-to-use interface to quickly write more tests[^2]. Lastly, we are slowly making our tests more robust[^3].

As such, this is a good time to create a workflow like in `iFixit/ifixit` where we stress test new or changes tests.[^4].

By doing this, we can ensure we don't increment the number of flaky tests.

### To Do:

- [ ] Mimic the `stress test` workflow from `iFixit/ifixit` 

[^1]: https://github.com/iFixit/react-commerce/issues/452
[^2]: https://github.com/iFixit/react-commerce/issues/1290 and https://github.com/iFixit/react-commerce/issues/1217
[^3]: https://github.com/iFixit/react-commerce/issues/1326, https://github.com/iFixit/react-commerce/issues/1416, https://github.com/iFixit/react-commerce/issues/1392
[^4]: https://github.com/iFixit/ifixit/blob/master/.github/workflows/E2E.yml#L212-L242

batbattur:
## Summary
Added stress-test ci for playwright and jest tests.

The new workflow has 3 jobs:
- The first job checks for the test changes and the logic for skipping the stress-test check.
  - It looks for modified jest test files in `frontend/tests/playwright/` folder with `*.spec.ts` file extension and jest tests in `frontend/tests/jest/tests/` with `*.test.tsx`. If changes were found, it will trigger the corresponding jobs to run the stress-test.
   - Also, if the last commit message has [skip stress-test] commit message, then it will not run stress tests (copy from `ifixit`).
- The second job runs `playwright` stress-test if there were test changes. It runs stress-test for `Desktop Chrome` and `Mobile Chrome` projects (we can add `Firefox` and `Tablet` but seems a bit too many). It uses `--repeat-each` option to run each test method `5` times.
- The third job runs `jest` stress-test and uses regular bash for loop to run jest test-suite `5` times (unfortunately `--repeat-each` doesn't exist for jest).

The only difference from the `iFixit/ifixit`'s stress-test workflow is that in `react-commerce`, we can't easily get the modified individual testMethod so we end up running the whole test class (which I think is fine). 

## QA
Check the following cases:
1. Make sure it doesn't run the stress-test when no test changes were made. 
2. Make sure it does run the stress-test when test changes were made.  

closes #1431

### DIFF
--- a/.github/workflows/stress_test.yml
+++ b/.github/workflows/stress_test.yml
@@ -114,3 +114,30 @@ jobs:
               name: playwright-artifacts
               path: frontend/tests/playwright/test-results/artifacts
               retention-days: 7
+
+   jest:
+      name: jest-stress-test
+      runs-on: ubuntu-latest
+      needs: check-diff
+      if: needs.check-diff.outputs.modified-jest-tests != ''
+      steps:
+         - name: Checkout
+           uses: actions/checkout@v2
+
+         - uses: actions/setup-node@v3
+           with:
+              node-version: 16
+
+         - name: Setup and cache pnpm
+           uses: ./.github/actions/pnpm-setup
+
+         - name: Install workspaces
+           env:
+              FONT_AWESOME_NPM_TOKEN: ${{ secrets.FONT_AWESOME_NPM_TOKEN }}
+           run: pnpm install:all
+
+         - name: Run Jest tests
+           run: for i in {1..5}; do jest -- ${{ needs.check-diff.outputs.modified-jest-tests }}; done
+           env:
+              NEXT_PUBLIC_STRAPI_ORIGIN: http://localhost:1337
+              ALGOLIA_API_KEY: ${{ secrets.ALGOLIA_API_KEY }}
--- a/.github/workflows/stress_test.yml
+++ b/.github/workflows/stress_test.yml
@@ -70,20 +70,10 @@ jobs:
          - name: Setup and cache playwright dependencies
            uses: ./.github/actions/playwright-setup
 
-         - name: Cache Next.js cache directory
-           uses: actions/cache@v3
+         - name: Build and cache frontend for with test setup
+           uses: ./.github/actions/build-test-frontend
            with:
-              path: frontend/.next/cache
-              key: ${{ runner.os }}-nextjs-${{ hashFiles('**/package-lock.json') }}-${{ hashFiles('**.[jt]s', '**.[jt]sx') }}
-              restore-keys: |
-                 ${{ runner.os }}-nextjs-${{ hashFiles('**/package-lock.json') }}-
-
-         - name: Build Next.js
-           env:
-              NODE_ENV: test
-              NEXT_PUBLIC_STRAPI_ORIGIN: http://localhost:1337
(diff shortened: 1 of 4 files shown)


batbattur:
dev_block 🚛 in progress

batbattur:
Steps used to test:
1. Made test changes for both jest and playwright multiple tests, ref: https://github.com/iFixit/react-commerce/pull/1459/commits/a794ee33860beb73b6ad692b18995a7209ef013f. 
2. Watched the stress logs https://github.com/iFixit/react-commerce/actions/runs/4380606193
   - Made sure it ran only the modified jest tests 5 times: [logs](https://github.com/iFixit/react-commerce/actions/runs/4380606193/jobs/7667855655#step:6:22).
   - Same for playwright tests, made sure each test method ran 5 times, [logs](https://github.com/iFixit/react-commerce/actions/runs/4380606193/jobs/7667855495#step:12:110).

batbattur:
un_dev_block 🚛

.github/workflows/stress_test.yml
```diff
@@ -0,0 +1,36 @@
+name: Stress-test
+
+on: [pull_request]
```
ardelato, line 3: Is the `concurrency` configurations we have in our `iFixit/ifixit` `E2E.yml` only needed if we are using a `self-hosted` runner?

batbattur: > Is the concurrency configuration we have in our iFixit/ifixit E2E.yml only needed if we are using a self-hosted runner?

I believe we can add in react-commerce as well but since they're free I think we don't care too much. The reason why we have it on `iFixit/ifixit` is because we have a limited amount of runners.

.github/workflows/stress_test.yml
```diff
@@ -34,3 +34,83 @@ jobs:
...
+              NEXT_PUBLIC_STRAPI_ORIGIN: http://localhost:1337
+              ALGOLIA_API_KEY: ${{ secrets.ALGOLIA_API_KEY }}
+           run: cd frontend && pnpm build
+
+         - name: Run Playwright Tests
+           env:
+              NODE_ENV: test
+              NEXT_PUBLIC_STRAPI_ORIGIN: http://localhost:1337
+              ALGOLIA_API_KEY: ${{ secrets.ALGOLIA_API_KEY }}
+           run: cd frontend && pnpm playwright:run --project="${{ matrix.project }}" ${{ needs.check-diff.outputs.modified-playwright-tests }} --repeat-each=5
```
ardelato, line 105: Hmm, what are your thoughts on extracting most of these steps into a composite action since there are a lot of steps that are similar between this workflow and the `main.yml` workflow?

ardelato (approved):
CR 👍 but deploy_block ☁️ on possible reducing duplicate code with composite actions.

ardelato (approved):
CR 👍 and un_deploy_block ⚡

batbattur:
QA 🚛 carry-over 

The new composite actions seem to be working fine.

PR #1465, opened 2023-03-14
Update footer to the new PageWrapperLayout

federicobadini:
This PR moves from the work done by @aburke07 to on the footer to include the new Wrapper component.
Together with @mmarcon91 we also added also some additional styles to avoid situations in which portions of the ui ended up being to close to each other.

### QA

1. Visit [Vercel preview](https://react-commerce-git-fix-footer-details-ifixit.vercel.app/Parts)
2. Check that - tanks to the new wrapper - the footer is now wide as much as the main content of the page


packages/footer/components/Partners.tsx
```diff
@@ -7,12 +7,11 @@ export const FooterPartners = forwardRef<BoxProps, 'div'>(
...
          <Flex
             ref={ref}
-            mt="7"
+            mt="8"
             alignItems="center"
             justifyContent="center"
             flexWrap="wrap"
-            paddingLeft="4"
-            paddingRight="4"
+            gap="3"
```
aburke07, line 14: I don't think we want to be using gap any more: https://github.com/iFixit/ifixit/pull/46613

aburke07 (approved):
CR :alien: Nice work!

packages/footer/components/Partners.tsx
aburke07: See: https://ifixit.slack.com/archives/C01GVQAAGDU/p1678818687071049

@jarstelfox mentioned we want to be using chakra `spacing` prop instead: https://chakra-ui.com/docs/components/stack/usage#usage

deltuh-vee:
QA 🎬 
Footer content is now as wide as the main content on the page.

aburke07:
Flex gap doesn't work in older versions of Safari, so we still shouldn't be using chakra "gap" if we can come up with an alternative solution. We should either revert to using a Stack with an alternative dom structure, use a Grid, or handle the margins ourselves.

PR #1479, opened 2023-03-16
Footer Package: don't rely on @ifixit/ui

aburke07:
We are running into a ton of rush errors in https://github.com/iFixit/ifixit/pull/46933 where rush doesn't have an updated rush.json with all of the packages from @ifixit/ui. Instead of adding a ton of dependencies to this, lets just extract Menu from
@ifixit/ui into its own package and not rely on @ifixit/ui at all in the footer package.

## QA
Make sure the footer looks and functions the same as before

Connects to https://github.com/iFixit/ifixit/pull/46933 and https://github.com/iFixit/ifixit/issues/45964


k0rvusk0r4x:
The prod and dev environments here are pretty different, so it's hard to know exactly which/how much needs to be fixed.

Main https://me.cominor.com/: 
image

Dev version https://react-commerce-6dg5p22yk-ifixit.vercel.app/ (capitalization and lists are the same, but spacing (vertical and horizontal) is wrong: 
image

Prod version https://react-commerce-prod-ls4hmuftn-ifixit.vercel.app/ (capitalization, lists, and spacing (vertical and horizontal) are wrong): 
image

dev_block 😿 to fix the spacing, and possibly the capitalization and lists.

Usually the primary difference between the dev and prod version is how accurate/up-to-date product information is, so I'm not sure which one we should focus on as "this is what it will look like in the end." Can you please clarify so we don't accidentally block when it's not necessary?

aburke07:
Cominor.com (home page, not store pages) use the iFixit footer, which has slightly different spacing than the react-commerce footer. This will be addressed in a future pull to clean up the code. 

Please only compare the react-commerce dev/prod footers, as those should match exactly, sans the list content. I will update the lists in strapi to match, but don't worry about the matching lists or capitalization for now.

erinemay:
qa 🦖 
- No unexpected differences between the dev preview on [this branch](https://react-commerce-git-fix-ui-package-ifixit.vercel.app/Parts) and [main](https://react-commerce.vercel.app/Parts).
    - The items in the Resources column are different, but those are data issues ([branch govinor Resources column](https://fix-ui-package.govinor.com/admin/content-manager/collectionType/api::menu.menu/10?plugins[i18n][locale]=en), [maingovinor Resources column](https://main.govinor.com/admin/content-manager/collectionType/api::menu.menu/10?plugins[i18n][locale]=en)).
- No differences between prod preview on [this branch](https://react-commerce-prod-git-fix-ui-package-ifixit.vercel.app/Parts) and [main](https://react-commerce-prod.vercel.app/Parts).

Aside from the data issue noted above:
- Footers look identical
- Links still work
- Newsletter signup ui still works
- Emails are getting saved in `ifixit_guide.newsletter`
- Store selector works
- No new console errors that weren't also on main

packages/menu/package.json
```diff
@@ -0,0 +1,20 @@
...
+      "type-check": "tsc --pretty --noEmit"
+   },
+   "devDependencies": {
+      "@ifixit/tsconfig": "workspace:*",
+      "@types/react": "18.0.24",
+      "@types/react-dom": "18.0.8",
+      "react": "18.2.0",
+      "react-dom": "18.2.0",
+      "typescript": "4.8.4"
+   }
```
danielbeardsley, line 19: I don't know if it makes much different in the end, but I would call react + react-dom "peer" dependencies. I'd look at several other packages in this repo to see what they do.

aburke07: These were actually not needed. Removed in [caf0fc1](https://github.com/iFixit/react-commerce/pull/1479/commits/caf0fc1ff162d21da4e83821afb14e9eb7a2f5f9).

Other packages do list react/react-dom as devDependencies, not peer dependencies.

erinemay:
QA 👍 Both dev and prod are now identical on this branch and main. No new errors that aren't also on main. The same things are working ➡️ https://github.com/iFixit/react-commerce/pull/1479#issuecomment-1474415409

packages/menu/package.json
```diff
@@ -0,0 +1,19 @@
...
+      ".": "./index.ts"
+   },
+   "scripts": {
+      "build": "",
+      "clean": "",
+      "type-check": "tsc --pretty --noEmit"
+   },
+   "devDependencies": {
+      "@ifixit/tsconfig": "workspace:*",
+      "typescript": "4.8.4"
```
danielbeardsley, line 17: I'm almost *certain* this package still wants react and chakra as some kind of dependencies (they are assumed and used in the code I think). So pnpm needs to make a symlink in order to access them. 

It's possible that it's still works because we are unintentionally hoisting everything: https://github.com/iFixit/ifixit/blob/master/common/config/rush/.npmrc#L48

Likely we only meant to hoist the listed things.

aburke07: I still don't believe we use react or chakra in this package. The only actual code is in menu.ts, which only consists of 50 lines of typescript types and an enum. Nothing is imported into that file either. 
The old UI package that menu.ts was in definitely had react and chakra, but the whole reason for extracting menu is so we don't need to rely on the extra dependencies of the UI package.

danielbeardsley: Oh wow, I didn't notice how little was in it. You're right. Carry on then!

danielbeardsley (approved):
CR :+1: 

Though I worry a little that it'll be easy to re-add dependencies on the ifixit/ui package to the footer without realizing the hardship it causes elsewhere.

PR #1488, opened 2023-03-21
Remove featured product list section from product list page - frontend changes

### LINKED ISSUE #1483: Facets disappear briefly on page load
I'm also seeing it on MacOS + Safari and FF
This recording is on Windows 10+Chrome with speed throttled to fast 3G. It's easier to see with speed throttled. 

https://user-images.githubusercontent.com/72166715/225971260-30b8979e-27f6-46fb-b5a5-3c449b3653cc.mp4

federicobadini:
closes #1483

This PR removes the featured-product-list section since the new mockups do not include that component anymore.

Since the featured-product-list section was relying on Algolia, this forced us to put in place a ["multi-index" Algolia configuration](https://www.algolia.com/doc/guides/building-search-ui/ui-and-ux-patterns/multi-index-search/react-hooks/#searching-the-same-index-with-different-search-parameters) for the page, and this seems to bring some unwanted side effects. So this is also an opportunity to get rid of that. 

In this PR we remove **all the frontend code** linked to the featured-product-list component.
Another PR will follow for the Strapi cleanup.

Screenshot 2023-03-20 at 17 48 16

### Detailed explanation

Having multiple indexes on the page means that the internal Algolia state needs to take into account both a facet-filtered search (the one associated to the main content) and a search without any facet (the one associated to a featured-product-list section).
So we effectively perform two queries towards Algolia.

By inspecting how the Algolia state updates, there is a short time window in which the internal Algolia state clears the list of facets. This is what causes the flicker described in #1483.
This was present also before we chose to hide the facets panel if empty: it was only less visible because the panel remained in place but the facets were briefly removed and then re-added.
Please note that this is not a transition from the previous facets to the new facets since the new facets were already visible before the facets were temporarily emptied.

You can see the two queries, the old behaviour of facets and all the details described above in the following video:

https://user-images.githubusercontent.com/7805759/226594651-1d1badc8-a2cd-4da2-b8ae-6fccf6e5cd92.mov

### QA

1. Visit Vercel preview
2. Verify that product list pages are still displaying correctly
3. Verify that the issue described in #1483 is not present anymore
4. Verify that we are now performing a single query towards Algolia (while performing client side navigation)


frontend/models/product-list/types.ts
```diff
@@ -157,12 +155,6 @@ export interface ProductListRelatedPostsSection {
    tags: string | null;
 }
 
-export interface ProductListFeaturedProductListSection {
-   type: ProductListSectionType.FeaturedProductList;
-   id: string;
-   productList: FeaturedProductList;
-}
-
 export interface FeaturedProductList {
```
masonmcelvain, line 158: Should we get rid of this interface too?

frontend/templates/product-list/ProductListView.tsx
```diff
@@ -32,19 +31,17 @@ export function ProductListView({
          <SecondaryNavigation productList={productList} />
          <Wrapper py={{ base: 4, md: 6 }}>
             <VStack align="stretch" spacing={{ base: 4, md: 6 }}>
-               <Index indexName={indexName} indexId="main-product-list-index">
```
masonmcelvain, line 35: Very minor, but I think we can drop the `Index` import now.

masonmcelvain (approved):
CR ✌🏻 with some non-blocking comments. Nice job!

frontend/models/product-list/types.ts
federicobadini: Nice catch! Removed :)

masonmcelvain (approved):
CR ✌🏻 Thanks!

deltuh-vee:
QA 🎬 
* Product list pages are still displaying correctly
* https://github.com/iFixit/react-commerce/issues/1483 is no longer occurring
* Only a single call is made towards the Algolia API when performing client side navigation

Issue #1492, opened 2023-03-21
Bad page number in url results in 500


danielbeardsley:
Problem
===

We had several thousand of these errors over the weekend (there's a separate issue for figuring out why they weren't sent to sentry). Bad input (page numbers in this case) should not result in a 500 response, nor a notice to us.

```
START RequestId: 30b245bf-3743-46ee-85fa-8e1369d37e5f Version: $LATEST
2023-03-21T09:31:16.944Z	undefined	INFO	info  - Loaded env from /var/task/frontend/.env.production
2023-03-21T09:31:21.518Z	30b245bf-3743-46ee-85fa-8e1369d37e5f	INFO	context.resolvedUrl /Parts/MacBook_Pro/Display_Components?p=-1%20OR%203%2B947-947-1%3D0%2B0%2B0%2B1
2023-03-21T09:31:21.518Z	30b245bf-3743-46ee-85fa-8e1369d37e5f	INFO	context.req.url /Parts/MacBook_Pro/Display_Components?p=-1%20OR%203%2B947-947-1%3D0%2B0%2B0%2B1
2023-03-21T09:31:21.566Z	30b245bf-3743-46ee-85fa-8e1369d37e5f	INFO	iFixit API GET:cart/part_collections/devices/MacBook_Pro => 26.58ms
2023-03-21T09:31:21.576Z	30b245bf-3743-46ee-85fa-8e1369d37e5f	INFO	Cache miss for 'storeList' => 50.20ms
2023-03-21T09:31:21.579Z	30b245bf-3743-46ee-85fa-8e1369d37e5f	INFO	Cache miss for 'globalSettings' => 60.11ms
2023-03-21T09:31:21.604Z	30b245bf-3743-46ee-85fa-8e1369d37e5f	INFO	strapi.getProductList => 66.13ms
2023-03-21T09:31:21.622Z	30b245bf-3743-46ee-85fa-8e1369d37e5f	INFO	algolia.findDescendantDevicesWithProducts => 14.38ms
2023-03-21T09:31:21.622Z	30b245bf-3743-46ee-85fa-8e1369d37e5f	INFO	findProductList => 84.63ms
2023-03-21T09:31:22.217Z	30b245bf-3743-46ee-85fa-8e1369d37e5f	INFO	Cache miss for 'store-us' => 689.23ms
2023-03-21T09:31:22.217Z	30b245bf-3743-46ee-85fa-8e1369d37e5f	INFO	layoutProps => 698.39ms
[...]

PR #1503, opened 2023-03-24
Add Parameter Check for Negative Page Number

TomiFixit:
Closes [#1492](https://github.com/iFixit/react-commerce/issues/1492)

**QA**

If the page number is invalid (<0), the parameter is ignored.

### DIFF
--- a/frontend/components/common/InstantSearchProvider.tsx
+++ b/frontend/components/common/InstantSearchProvider.tsx
@@ -230,7 +230,7 @@ export function InstantSearchProvider({
 
             return {
                q: String(q || ''),
-               p: typeof p === 'string' ? parseInt(p) : undefined,
+               p: typeof p === 'string' && parseInt(p) >= 0 ? parseInt(p) : undefined,
                filter: filterObject,
             };
          },
--- a/frontend/components/common/InstantSearchProvider.tsx
+++ b/frontend/components/common/InstantSearchProvider.tsx
@@ -230,7 +230,10 @@ export function InstantSearchProvider({
 
             return {
                q: String(q || ''),
-               p: typeof p === 'string' && parseInt(p) >= 0 ? parseInt(p) : undefined,
+               p:
+                  typeof p === 'string' && parseInt(p) >= 0
+                     ? parseInt(p)
+                     : undefined,
                filter: filterObject,
             };
          },


frontend/components/common/InstantSearchProvider.tsx
```diff
@@ -230,7 +230,10 @@ export function InstantSearchProvider({
 
             return {
                q: String(q || ''),
-               p: typeof p === 'string' ? parseInt(p) : undefined,
+               p:
+                  typeof p === 'string' && parseInt(p) >= 0
+                     ? parseInt(p)
+                     : undefined,
```
ardelato, line 236: Looking at the code it seems that `p` can by of type `any`.

However, `parseInt()` will return `NaN` if it can't parse the integer https://developer.mozilla.org/en-US/docs/web/javascript/reference/global_objects/parseint

So we can extract the parsing to a line above and then simply check the negative value in this line. This will make the line less bloated and not have to run `parseInt` twice.

So something like 

```

const pageParam = parseInt(p)

return {
   q: String(q || ''),

   p: pageParam >= 0 ? pageParam : undefined

```

[Typescript Example](https://www.typescriptlang.org/play?#code/DYUwLgBADgXBCGA7AngbgFCkleAnAziACYD6UEAvNHoQJKJgAUUAlOgMYD2i+noAdME4BzRgAMAgimrCQEAAp54AWwgASAN44CIek1YBfMW0zgZcqgCIAjAAZLp7DWJl4sytR17mbkGy48fCCCIuLynPgAlmCRAG5yOO6KuCrqWs7eiX5GJlgQiCDC8DHxru5WALQALA552oSkBUUlIGUWnnQMjE3Fca1Z-ty8AkKiYgByhb3x5gpKqpr1ul09LW0sOegcQ0EhY1LIswBmkYjwwGlLpOQAfFS2EAD8HS7kcACuiEQgJwVEmwFhsFRmEItE+sdTudLs5rr4IHcIA9nlc2hAPl8fqdiACdiNQhMpi1IWcLotYSRVn00YjkS9GkTqVl0RBPt9fjjjNtAvjRON4OMEfcICwgA)

ardelato (approved):
CR 👍 but deploy_block ☁️ on minor suggestion.

Also is `Add` supposed to be the full title of this PR?

ardelato:
un_deploy_block ⚡ spoke with @TomiFixit and `p` was not just `any` so we would still need to `type` check the value. Therefore, the current way works.

PR #1515, opened 2023-03-30
Use product data from Strapi when available

### LINKED ISSUE #1509: Add Strapi as a source for featured product page sections
Now that we have created the Strapi production collection type and that we've refactored the product page template we can finally use Strapi as a data source to build custom featured product pages.
In this first version we'll mostly have the standard sections, but we'll add one after the other all the sections that have been designed specifically for featured products

dhmacs:
closes #1509


deltuh-vee:
QA 🎬 
* With no overrides or saved drafts, product pages are still normally rendered.
* With a published override, only the added sections are shown.

masonmcelvain (approved):
CR ✌🏻

PR #1516, opened 2023-03-30
Add products anchor

### LINKED ISSUE #1510: Add an id to the products section in product list pages
@ifixitmaxb proposed to add the ability to link to the products section in product list pages:

> I shared this on private channel as well but I imagine y'all would have a good answer too:
> If I'm on a collection page in our US store and I click to go to page 2, page 2 loads and lands me at the product list (not the top of the page). Is there a parameter we can add to a collection page link that would land the user at the product list instead of at the top of the page? I was on mobile this am and I clicked the Marlin banner from a guide page and it took me to the top of the filtered collection - the UX wasn't ideal imo - I thought I was on the wrong page tbh. It'd be ideal imo if the banner link, and any shareable filtered collection link, landed the user at the product list (maybe by default?) instead of at the top of the page
> Fwiw I was on this Guide and the Marlin banner is below Step 1 https://www.ifixit.com/Guide/OnePlus+7+Pro+Battery+replacement/147766

This can be achieved by assigning an html id to the section (e.g. `products`) and compose the link like this: `/Tools?q=Marlin#products`

dhmacs:
closes #1510 

## QA

Open a product list preview from mobile and verify that by appending `#products` the browser scrolls down to the list of products.
For example check with this link [/Tools?q=Marlin#products](https://react-commerce-git-add-products-anchor-to-product-list-ifixit.vercel.app/Tools?q=Marlin)

### DIFF
--- a/frontend/templates/product-list/sections/FilterableProductsSection/index.tsx
+++ b/frontend/templates/product-list/sections/FilterableProductsSection/index.tsx
@@ -75,6 +75,7 @@ export function FilterableProductsSection({ productList }: SectionProps) {
    return (
       <Flex
          ref={productsContainerScrollRef}
+         id="products"
          as="section"
          direction="column"
          align="stretch"


sterlinghirsh (approved):
CR :zap:

dhmacs:
@ifixitmaxb you should be able to update the link to `/Tools?q=Marlin#products` now

PR #1517, opened 2023-03-30
Enable pre-commit format hoook

masonmcelvain:
Re-enables the pre-commit format hook. This uses a new tool called [lint-staged](https://github.com/okonet/lint-staged) to only format staged files, making the hook much faster than formatting all files in the repo.

Connects https://github.com/iFixit/react-commerce/issues/1026

## QA

Make an intentional formatting error, stage it, and try to make a commit. The commit should fail if the formatting error is the only change. If there are other changes, then your commit should be successful with all formatting errors fixed.

CC @danielbeardsley @sterlinghirsh

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,2 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
+
+pnpm lint-staged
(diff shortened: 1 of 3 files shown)


danielbeardsley (approved):
CR :+1: 

Thanks!

deltuh-vee:
QA 🎬 
* Single formatting errors prevent committing.
* Formatting errors are automatically fixed when committing if accompanied by other changes.

Issue #1400, opened 2023-02-22
Tests: Clean up Mock Files


ardelato:
## Description

In this pull https://github.com/iFixit/react-commerce/pull/1342, there was a compiling error that was occurring when trying to run the Playwright tests[^1]

![image](https://user-images.githubusercontent.com/22064420/220714199-903668e7-50b0-409a-b964-de305359aa7c.png)

The cause of this issue was the `index` file of the package we were trying to import, `@ifixit/ui`, in the mocked products file[^2], also included Typescript XML, `tsx`, files. This became a problem with Playwright since we use this `products.ts` file in one of our tests[^3]. Playwright tries to compile all the imported files defined in the `index` file of `@ifixit/ui` which included the `tsx` files[^4]. However, Playwright does not know how to parse `tsx` files so it throws an error. 

The workaround for this was to explicitly import the `ts` file needed in the `products.ts` file - i.e. `@ifixit/ui/menu` instead of `@ifixit/ui` [^5]

### To Do

We now have quite a bit of coupling between `Playwright` and the mocks that live in the `jest` directory. We should extract the mocks from `jest` and make it its own independent directory so that both testing frameworks can use the mocks but not be tied to the other's compiling rules.

- [x] Also investigate whether we can get `Playwright` to compile `tsx` files
- [ ] Update Playwright README with information

[^1]: https://github.com/iFixit/react-commerce/actions/runs/4234773967/jobs/7357533031#step:12:22
[^2]: https://github.com/iFixit/react-commerce/blob/main/frontend/tests/jest/__mocks__/products.ts
[^3]: https://github.com/iFixit/react-commerce/blob/main/frontend/tests/playwright/product/add_to_cart.spec.ts#L2
[^4]: https://github.com/iFixit/react-commerce/blob/footer-update-newsletter/packages/ui/index.tsx
[^5]: [`721c290` (#1342)](https://github.com/iFixit/react-commerce/pull/1342/commits/721c2904c71a6174fc1c061d30bf54bbac8ee9b1)

andyg0808:
Is there not a way to make Playwright handle tsx files appropriately? Is it just that the tests themselves don't need the tsx, so splitting it out makes the most sense?

ardelato:
At the moment I can't find an immediate answer. I'll update the issue to also look if we can get `Playwright` to compile `tsx` files.

Aside from that, I think overall the `tests` directory could use a cleanup since there is some overlap between `jest` and `Playwright` now.

[...]

PR #1518, opened 2023-03-30
Extract Playwright Specific Mock from Jest

ardelato:
## Description

This product snapshot is only used in Playwright tests, so it makes more sense to move it into the Playwright fixtures directory instead of having the Playwright tests import it from the Jest mocks directory.

This will also help to avoid similar problems as described in the original issue #1400 

### CR

I will be cleaning up the rest of the `__mocks__` folder in a later pull since it will be related to just `jest`.

### QA

Connects: #1400


batbattur (approved):
CR 🚛

PR #1519, opened 2023-03-30
Cleanup Jest Mocks

ardelato:
## Description

The `jest/__mocks__/products.ts` file originally only housed a single `Product` mock but as we continued to write more tests and extend the types, the file became muddled with other mock types.

In order to clean up this file, I've extracted each mock type into its own file with the exception of the multiple `product` mocks. I created a new file for each type of `product` and moved them into their own directory.

I also moved the utility functions for retrieving and modifying these mocks into the associated mock file to retain the context of what they use.

### CR

Although this does not resolve the issue of manually updating the mocks when there is a `schema` change, it should help with the top-level organization of the mocks used for `jest` tests.

### QA

Connects: #1400


ardelato:
un_deploy_block ⚡ blocking pull got merged

mlahargou (approved):
CR 📱 Checks out to me! Nice :)

batbattur (approved):
CR 🚛 looks very nice and easy to find the mocks

PR #1522, opened 2023-03-31
Sentry: Enable `autoInstrumentServerFunctions`

masonmcelvain:
Enables error monitoring on api routes and data fetching methods (like getServerSideProps).

https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/#configure-server-side-auto-instrumentation


masonmcelvain:
@dhmacs I see that we added `autoInstrumentServerFunctions: false` in https://github.com/iFixit/react-commerce/commit/f13399d4093859a5626d5537c60c73523661549a. If you remember, could you describe the motivation for adding that option? It's unclear to me based on the context of the PR. It might be preventing us from sending certain kinds of errors to Sentry (see https://github.com/getsentry/sentry-javascript/issues/7602#issuecomment-1485036262), though there are workarounds.

PR #1524, opened 2023-04-01
Troubleshooting: Link primary author's name to their profile

andyg0808:
We want to link the primary author to their profile and the "contributors" message (if there is one) to the history page.

## QA Notes
Please confirm that it matches the description in the issue.
https://react-commerce-git-troubleshooting-use-separate-l-020767-ifixit.vercel.app/Vulcan/Dryer_Not_Spinning should have multiple contributors; note that the links will point to the Vercel page, not the iFixit page, so they won't work directly when clicked. You'll need to edit the URL you end up on to switch from the Vercel hostname to the `cominor.com` hostname.

Closes ifixit/ifixit#47194


aburke07 (approved):
CR :alien:

k0rvusk0r4x:
- The links are there.
- The links work.

However:

- The issue shows the "x contributors" link to be blue.
- The issue also shoes a line under "and" on hover, but that isn't the most intuitive when we are dealing with two separate links that go to different places. I would leave that out as it is now.

dev_block 😿 for color change

andyg0808:
un_dev_block :+1: 
Color changed.

erinemay:
QA 👍 
Single: https://react-commerce-git-troubleshooting-use-separate-l-020767-ifixit.vercel.app/Vulcan/troubleshooting-test
image

Two: https://react-commerce-git-troubleshooting-use-separate-l-020767-ifixit.vercel.app/Vulcan/erins_troubleshooting_test
image

Greater than two: https://react-commerce-git-troubleshooting-use-separate-l-020767-ifixit.vercel.app/Vulcan/Dryer_Not_Spinning
image

- Author link goes to author profile.
- Contributors link goes to wiki history.
- Links are blue.
- Looks like the mockup.

PR #1531, opened 2023-04-04
Fix footer dependencies

aburke07:
This should _hopefully_ fix the rush errors that we are getting when trying to update iFixit/ifixit code. We were relying on some packages that required a ton of other dependencies that we don't want in iFixit, so this pull extracts the needed code into their own packages without tons of dependencies.

Qa_req 0

Connects to: [iFixit/ifixit#47214](https://github.com/iFixit/ifixit/pull/47214)


frontend/components/common/AppProviders.tsx
```diff
@@ -64,17 +65,19 @@ export function AppProviders({
 
    return (
       <AppProvider ifixitOrigin={ifixitOrigin ?? IFIXIT_ORIGIN}>
```
danielbeardsley, line 67: Since we have a new context for the ifixit origin, I think we should remove this value from the AppProvider. Thoughts?

danielbeardsley:
Looking good!

dev_block :+1:  on one suggestion.

frontend/components/common/AppProviders.tsx
aburke07: Hmm, AppProvider only contains ifixitOrigin, so now I'm questioning why we split it out into its own package. The original goal is to remove packages with tons of unneeded dependencies from newsletter-sdk, but app doesn't seem to have a bunch of extra dependencies. Let me ping Andrew to clarify.

aburke07: Force pushed them out :+1:

danielbeardsley (dismissed):
CR :+1: 

Nice work!

danielbeardsley (approved):
CR :+1:

aburke07:
Un_dev_block :alien:

PR #1538, opened 2023-04-05
Add fortawesome/solid-icons to footer package

aburke07:
We are running into typechecker errors on the iFixit side when trying to update the react-commerce submodule because this package was not added to the footer package. This should fix it.

Qa_req 0 - The icons still render in the footer.

Connects to https://github.com/iFixit/ifixit/pull/47233


danielbeardsley:
CR :+1:

danielbeardsley (approved):
CR :+1:

PR #1546, opened 2023-04-06
Add react as dependency to new packages

aburke07:
These packages use react, so we need to add the dependency. According to: https://stackoverflow.com/questions/30451556/what-is-the-correct-way-of-adding-a-dependency-to-react-in-your-package-json-for we should add it as a peer dependency and a dev dependency.

We are failing the build-assets on https://github.com/iFixit/ifixit/pull/47270, so this should fix that.

Qa_req 0

Connects to https://github.com/iFixit/ifixit/pull/47270


danielbeardsley:
CR :+1:

danielbeardsley (approved):
CR :+1:

PR #1549, opened 2023-04-07
Troubleshooting: Move solution title to same line as number

jyee27:
Adjust solution titles based on the updated mockup. This also includes some other text spacing changes, but we should do another pass after this pull for things like header sizes and spacing.

<details><summary>Image</summary>

image

</details> 

Closes: https://github.com/iFixit/ifixit/issues/47301


deltuh-vee:
QA 🎬 
Solution title is now on the same line as the number.

djmetzle (approved):
CR :medal_sports:

PR #1551, opened 2023-04-10
Add strapi image loader

### LINKED ISSUE #1550: Strapi images not loading on cominor
https://www.cominor.com/products/repair-business-toolkit is correctly picking up strapi sections, but the uploaded images don't display. The urls are rendering like https://www.cominor.com/uploads/newproducts_v1_89907dd0d2.jpg but should be https://main.govinor.com/uploads/newproducts_v1_89907dd0d2.jpg

masonmcelvain:
Closes #1550 

Adds a loader to correctly resolve images hosted on Strapi. We get urls like "/uploads/newproducts_v1_89907dd0d2.jpg" from Strapi's API, but the images are hosted on Strapi's origin, not the same origin as the page.

## QA

Make sure the section images on `/products/repair-business-toolkit` load correctly.

CC @sterlinghirsh

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -134,6 +134,7 @@ const moduleExports = {
          'www.cominor.com',
          'guide-images.cdn.ifixit.com',
          process.env.STRAPI_IMAGE_DOMAIN,
+         new URL(strapiOrigin).hostname,
       ].filter((domain) => domain),
    },
    i18n: {


masonmcelvain:
dev_block ✌🏻 the loader isn't working

frontend/components/sections/SocialGallerySection.tsx
```diff
@@ -80,6 +81,7 @@ function SocialPost({ post: { image, author } }: SocialPostGridItemProps) {
             <Box w="full" position="relative">
                <AspectRatio ratio={1}>
                   <ResponsiveImage
+                     loader={strapiLoader}
```
danielbeardsley, line 84: Is there a reason we didn't utilize the same technique we've done with other image urls:
https://github.com/iFixit/react-commerce/blob/main/packages/ui/misc/ResponsiveImage/index.tsx#L19-L27

masonmcelvain: It's trickier to unambiguously identify these urls ("/uploads/newproducts_v1_89907dd0d2.jpg") since they don't have an origin (we could match on `/uploads/` if we feel certain we won't ever encounter other relative urls that start with that segment). We also don't have STRAPI_ORIGIN available in the context of the packages, it's only available to `frontend/` currently.

danielbeardsley: Hmm... actually we already did this:
https://github.com/iFixit/react-commerce/blob/main/packages/ui/misc/ResponsiveImage/index.tsx#L25

Which looks like we're expecting strapi to upload to S3:
https://github.com/iFixit/react-commerce/blob/main/packages/ui/misc/ResponsiveImage/iFixitUtils.ts#L47-L52

Oh perhaps on govinor strapi isn't set up to upload to S3 (but in production it is?)

masonmcelvain: Thanks for the context.

You might be right that dev strapi doesn't upload to s3, the govinor links resolve but I can't figure out an aws link that works.
https://main.govinor.com/uploads/newproducts_v1_89907dd0d2.jpg
https://ifixit-dev-strapi-uploads.s3.amazonaws.com/uploads/newproducts_v1_89907dd0d2.jpg

Even if dev strapi stores images locally though, I'd expect the rewrite below to work for next links with relative `/upload/blah` urls to work even without the custom loader.
https://github.com/iFixit/react-commerce/blob/764d538991a6fc39bcf68fb6f19b430943f6c7c6/frontend/next.config.js#L60-L67

masonmcelvain: The [docs](https://nextjs.org/docs/api-reference/next.config.js/rewrites) seem to imply that rewrites only apply to client side routing, maybe this strategy doesn't work for images?

> Rewrites are applied to client-side routing, a <Link href="/about"> will have the rewrite applied in the above example.

danielbeardsley: > instead of using next.js

Totally. Here's the [list of forwarded paths](https://github.com/iFixit/ifixit/blob/master/Conf/haproxy/haproxy.development.cfg.php#L111) and it doesn't include those.

dhmacs: @danielbeardsley images in prod are uploaded to s3, so they have the s3 url contrary to dev/staging images

danielbeardsley: >  they have the s3 url contrary to dev/staging images

My question was that if S3 upload is not configured, I'd expect strapi image urls to have a strapi hostname. i.e. `{branch}.govinor.com/uploads/...`. Yet here we are expecting them to have a next.js hostname? Maybe some code is building urls incorrectly?

dhmacs: @danielbeardsley we use relative url combined with the rewrite because we are using the Next image component, and that component requires to whitelist URLs. Now since we can't whitelist every possible staging strapi URL, we just use the rewrite rule and make it look like it's a Next.js path.. This is a cosmetic trick that we only need in dev/staging so that images work seamlessly.. In prod we'll have full s3 url, so we don't need to worry about any of this

danielbeardsley: > the rewrite

Hmm... Does this mean next.js is acting as a proxy (to strapi) in dev? I wouldn't have guessed that altering the config in next would have it run a proxy. Also, enjoy your holiday, I wasn't expecting any replies today! :-)

masonmcelvain:
https://github.com/iFixit/ifixit/pull/47331 solved the problem, nothing is wrong with this repo's configuration.

frontend/components/sections/SocialGallerySection.tsx
dhmacs: > Does this mean next.js is acting as a proxy (to strapi) in dev?

Exactly 👍 

> enjoy your holiday, I wasn't expecting any replies today

Thanks, I'm going to bed now 😁

PR #1561, opened 2023-04-11
Sentry: Turn on auto instrument server funcs

masonmcelvain:
https://docs.sentry.io/platforms/javascript/guides/nextjs/manual-setup/#configure-server-side-auto-instrumentation

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -149,7 +149,6 @@ const moduleExports = {
       return config;
    },
    sentry: {
-      autoInstrumentServerFunctions: false,
       // Upload artifacts in dist/framework as well; this includes sourcemaps
       // for react and other next.js code
       widenClientFileUpload: true,


danielbeardsley:
CR :+1: 

I think this is the right thing, but it's hard for me to believe we aren't instrumenting with sentry at all except for the places where we do it explicitly.

deploy_block :+1:  on looking at those explicit places and ensuring we won't be double-reporting.

masonmcelvain:
Great suggestion! I did a quick audit of the places we explicitly use sentry, and I think these were the only potential double-captures. Our [`applySentryFetchMiddleware()`](https://github.com/iFixit/react-commerce/blob/83d0f5661ff0b02c434f3b2dd0adbd90f8b5d1d4/packages/sentry/index.tsx#L46) middleware doesn't look like it will double report.

un_deploy_block ✌🏻

danielbeardsley (approved):
CR :+1: 

Nice work!

Issue #1496, opened 2023-03-21
GitHub Actions: Generate urls to view trace files for Playwright


ardelato:
## Description

It would be nice to view `trace.zip` files in https://trace.playwright.dev/ without having to download the artifacts first.

From the Playwright documentation, it looks like there is a way to do so if we host the zip files somewhere.

https://playwright.dev/docs/trace-viewer#viewing-remote-traces

### To Do

- [ ] See about creating links that will automatically load https://trace.playwright.dev/ with the `trace.zip` file of the test that failed in CI.

[ifixit.slack.com/archives/C01DBJK0H7U/p1678813084862249](https://ifixit.slack.com/archives/C01DBJK0H7U/p1678813084862249)

PR #1563, opened 2023-04-11
Reorganize the Directory Structure of the Trace files

ardelato:
## Description

This pull request adds a new formatting fixture for trace file output paths in Playwright tests. It modifies the `testInfo.outputDir` property to create a nested directory structure that makes it easier to locate trace files for specific tests.

The new directory structure for trace files will look like this:
```bash
[subtest-directory]/test-file-name/test-name/device-name/
```

For example:
```bash
product/add_to_cart/back-to-shopping-button-works/desktop-chrome/
```

This nested directory structure will allow us to easily find the trace files for a given test as the original trace files were underneath a single directory where the directory name was a flattened version of the new directory structure.

|Before|After|
|:---:|:---:|
|  image  | image     |

### QA Notes

- [ ] Verify that a trace file is created in the expected folder path for a test file in the playwright directory.
	- Check if a `.spec.ts` file lives in the `playwright` folder so that the `[subtest-directory]` path will be omitted and the trace file will still successfully be outputted
- [ ] Test the proper formatting of directory names when special characters are used in test names.
	- Basically check `'` and other special characters that would be used in a test case name will be properly formatted for the use of a directory name

Connects: #1496

### DIFF
--- a/frontend/tests/playwright/fixtures/trace-override.ts
+++ b/frontend/tests/playwright/fixtures/trace-override.ts
@@ -19,10 +19,7 @@ export function traceOutputDirTemplate(testInfo: TestInfo) {
 }
 
 function formatTestNameForDirectory(testName: string) {
-   // Replace any special characters and spaces with a hyphen
    const formattedName = testName.replace(/[^a-zA-Z0-9]+/g, '-');
-
-   // Ensure the directory name is in lowercase
    const directoryName = formattedName.toLowerCase();
 
    return directoryName;
--- a/frontend/tests/playwright/fixtures/trace-override.ts
+++ b/frontend/tests/playwright/fixtures/trace-override.ts
@@ -1,0 +1,29 @@
+import { TestInfo } from '@playwright/test';
+import path from 'node:path';
+
+export function traceOutputDirTemplate(testInfo: TestInfo) {
+   // The path to where all the trace files will be outputted as defined
+   // in the Playwright config
+   const globalOutputDir = testInfo.project.outputDir;
+
+   // titlePath houses the path to the test file relevant to the test directory
+   // e.g. 'product/add_to_cart.spec.ts' -> 'product/add_to_cart'
+   const testFile = testInfo.titlePath[0].replace(/\.spec\.ts$/, '');
+
+   const testName = formatTestNameForDirectory(testInfo.title);
+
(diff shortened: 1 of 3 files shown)


frontend/tests/playwright/fixtures/trace-override.ts
```diff
@@ -0,0 +1,29 @@
...
+import path from 'node:path';
+
+export function traceOutputDirTemplate(testInfo: TestInfo) {
+   // The path to where all the trace files will be outputted as defined
+   // in the Playwright config
+   const globalOutputDir = testInfo.project.outputDir;
+
+   // titlePath houses the path to the test file relevant to the test directory
+   // e.g. 'product/add_to_cart.spec.ts' -> 'product/add_to_cart'
+   const testFile = testInfo.titlePath[0].replace(/\.spec\.ts$/, '');
```
mlahargou, line 11: I know we only have `spec.ts` files right now. Do you think that will change? Maybe we add `spec.tsx` at some point?

mlahargou (approved):
CR 📱 with a minor comment. Fixtures are so powerful!

frontend/tests/playwright/fixtures/trace-override.ts
ardelato: > I know we only have spec.ts files right now. Do you think that will change?

I don't think it will change, there might be a chance someone who adds a `.test.ts` file but that would assume they didn't notice the common usage of `.spec.ts` and we can comment on in CR.

As for the `spec.tsx`, the usage of `tsx` would imply that there are some `react` elements used in the test file which `playwright` is not capable of parsing. This is what caused this issue https://github.com/iFixit/react-commerce/issues/1400

mlahargou: Sounds good :)

deltuh-vee:
`Clicking + and - Buttons Changes Item Quantity in Cart` gets processed as clicking-and-buttons-changes-item-quantity-in-cart. Would we want to not use "+" and "-" characters?

mlahargou:
Yeah, let's just change the test name or forget about it. I don't think it's a huge deal.

deltuh-vee:
QA 🎬
* Traces are created in the expected folder path
* Directory names are not heavily affected by test case names

PR #1565, opened 2023-04-12
ResponsiveImage: let next.js handle strapi images

### LINKED ISSUE #1560: Strapi image auto-resize or use next
We are serving full-size images from strapi. We should figure out whether we want to keep doing that, auto-resize images on strapi upload, or route through next. We had some gigantic images on the repair business toolkit.

danielbeardsley:
We want folks who upload images to strapi to not have to care about making their images web-ready. They should be able to upload high quality, high-res images and we can then lean on next/image to downsample and re-compress them.

Yes, vercel likely clears the image cache more than we'd like, but we have cloudfront sitting in front of them, so I'm not too worried.

I've left this check in here so there's an explicit callout (strapi images should be processed by next) rather than just an absence.

Closes #1560

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -135,6 +135,7 @@ const moduleExports = {
          'guide-images.cdn.ifixit.com',
          process.env.STRAPI_IMAGE_DOMAIN,
       ].filter((domain) => domain),
+      minimumCacheTTL: 3600,
    },
    i18n: {
       locales: ['en-US'],
--- a/packages/ui/misc/ResponsiveImage/index.tsx
+++ b/packages/ui/misc/ResponsiveImage/index.tsx
@@ -23,7 +23,7 @@ export function ResponsiveImage(props: ImageProps) {
       } else if (isGuideImage(props.src)) {
          loader = getIFixitImageLoader(guideImageSizeMap, 'huge', props.width);
       } else if (isStrapiImage(props.src)) {
-         unoptimized = true;
+         unoptimized = false; // let next handle optimization cache resizing
       }
    }
 


masonmcelvain (approved):
CR ✌🏻 though I think the comment is a little confusing

We could also globally bump the next.js [image cache TTL](https://nextjs.org/docs/api-reference/next/image#minimum-cache-ttl) in next.config.js if we wanted to (defaults to 60 seconds).

danielbeardsley:
> We could also globally bump the next.js [image cache TTL](https://nextjs.org/docs/api-reference/next/image#minimum-cache-ttl) in next.config.js if we wanted to (defaults to 60 seconds).

Totally, I'll bump it to 1 hour (arbitrary), but the docs say:

> The expiration (or rather Max Age) of the optimized image is defined by either the minimumCacheTTL or the upstream image Cache-Control header, whichever is larger.

And in the case of strapi images, it looks like they serve them with 1-month expiration times:

image

masonmcelvain (approved):
CR ✌🏻

danielbeardsley:
Firstly let's look at the image requests on the RBT page.

main branch:
image

this branch:
image


Now let's look at an individual image:
550kb - https://ifixit-strapi-uploads.s3.amazonaws.com/rbt_banner_bc373aecec.jpg
vs
321kb - https://react-commerce-prod-j3oz962ug-ifixit.vercel.app/_next/image?url=https%3A%2F%2Fifixit-strapi-uploads.s3.amazonaws.com%2Frbt_banner_bc373aecec.jpg&w=1920&q=75

That said... there's plenty of opportunity to improve the props we send to the image component so it sizes these correctly (no need for a 2,000px image on a mobile device) but that can be in a follow up pull.

QA :+1:

danielbeardsley:
Also notice that the images are still cacheable from next (thus will be cached by vercel and CF):
image

Issue #1436, opened 2023-03-03
Product Page Compatibility: Limit displayed model numbers per device by default


sterlinghirsh:
## Problem
On highly compatible products like https://www.ifixit.com/products/frigidaire-241798231-ice-maker-assembly#compatibility, the list of model numbers per device becomes ridiculously long (thousands of lines). We still want to have all these numbers visible to Google, but it's a bad UX to have such a long list.

## Do This
Use css or a disclosure element like `<details>` or something to limit the display to the first 10 or so variants. When the list is truncated, include a link like "Show all 4528 models..." that expands the list for a single device. Still not the best, but it's an easy step in the right direction.

dhmacs:
cc @mmarcon91 for a quick design iteration

PR #1570, opened 2023-04-12
Product Page: Add compatibility disclosure

masonmcelvain:
Collapses the product page compatibility component when there are more than 10 model number lines.

## CR

The diff is easier to view with whitespace ignored.

It also seems like the Jest test that was changed was broken before. The product being tested only has 3 variants on the compatible device, so the previous test implementation should have produced `other model...`, however the test was asserting for `other models...`.

## QA
- Find a product with a compatible device that has >10 model numbers.
- Make sure the implementation matches the disclosure [mockup](https://www.figma.com/file/BolmFORO6ZfHXePOSyEGSS/iFixit---Page-sections?node-id=159-5560) (disregard the search bar for this PR).
- Make sure _both_ compatibility components otherwise match their `main` counterparts.

Closes https://github.com/iFixit/react-commerce/issues/1436

### DIFF
--- a/frontend/components/common/CompatibleDevice.tsx
+++ b/frontend/components/common/CompatibleDevice.tsx
@@ -1,71 +1,127 @@
-import { Flex, Img, Text } from '@chakra-ui/react';
+import {
+   chakra,
+   Flex,
+   HTMLChakraProps,
+   Img,
+   Text,
+   useDisclosure,
+} from '@chakra-ui/react';
 import type { Product } from '@pages/api/nextjs/cache/product';
-import * as React from 'react';
+import NextLink from 'next/link';
 
-export type CompatibleDeviceProps = {
+export interface CompatibleDeviceProps extends HTMLChakraProps<'a'> {
    device: NonNullable<Product['compatibility']>['devices'][number];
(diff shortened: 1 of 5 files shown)


deltuh-vee:
Is showing 1 extra model for compatible models intentional?
image
image

masonmcelvain:
Ah man, I continue to misunderstand the meaning of the `maxModelLines` prop. To me, it means "how many model lines should be shown", but to previous implementors it meant something different. Fix incoming.

masonmcelvain:
Should be fixed when that commit finishes deploying

deltuh-vee:
QA 🎬 
* Implementation matches the mockup
* Both compatibility components match their counterparts on main

danielcliu-ifixit:
> Ah man, I continue to misunderstand the meaning of the maxModelLines prop. To me, it means "how many model lines should be shown", but to previous implementors it meant something different. Fix incoming.

Yeah i remember having this discussion about it originally. If its a headache for developers we can just change it so its more universally intuitive.

danielbeardsley (approved):
CR :+1:

PR #1571, opened 2023-04-12
Sentry: Drop invalid strapi filter reporting

masonmcelvain:
Drops a sentry report about invalid strapi filters. We already have a UI component to alert Strapi admin users that the filter field is invalid, so reporting to Sentry is just extra noise.

qa_req 0


danielcliu-ifixit:
CR :v:

davidrans (approved):
CR :coffee:

PR #1572, opened 2023-04-13
Product Page: Use Sorted Variants

danielcliu-ifixit:
## Overview

The variant selection on our product pages isn't always in a sensible order, and Shopify doesn't help you much in terms of making product data easy to return in the order you want it to. This uses the api introduced in https://github.com/iFixit/ifixit/pull/47291 to get back the products variants sorted correctly and use that data to have the variant dropdown options display in a sensible order.

## QA
Do product page variants work correctly and have a sensible order? you can check the preset order by looking at `ifixit_cart.attribute_options` `sort_order` column.

### DIFF
--- a/frontend/lib/ifixit-api/devices.ts
+++ b/frontend/lib/ifixit-api/devices.ts
@@ -5,6 +5,16 @@ import { z } from 'zod';
 
 export type DeviceWiki = Record<string, any>;
 
+export type DeviceWikiApiResponse = {
+   variantOptions: {
+      [o_code: string]: string[];
+   };
+};
+
+export type VariantOptions = {
+   [o_code: string]: string[];
+};
+
 export async function fetchDeviceWiki(
    client: IFixitAPIClient,
    deviceTitle: string
@@ -16,6 +26,22 @@ export async function fetchDeviceWiki(
          'deviceHandle cannot be a blank string'
       );
       return await client.get(`cart/part_collections/devices/${deviceHandle}`);
+   } catch (error: any) {
+      return null;
+   }
+}
+
+export async function fetchProductData(
+   client: IFixitAPIClient,
+   handle: string
+): Promise<DeviceWikiApiResponse | null> {
+   const productHandle = encodeURIComponent(handle);
+   try {
+      invariant(
+         productHandle.length > 0,
+         'productHandle cannot be a blank string'
+      );
+      return await client.get(`store/product/${productHandle}`);
    } catch (error: any) {
       return null;
    }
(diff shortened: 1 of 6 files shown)


frontend/lib/ifixit-api/devices.ts
```diff
@@ -5,6 +5,16 @@ import { z } from 'zod';
 
 export type DeviceWiki = Record<string, any>;
 
+export type DeviceWikiApiResponse = {
```
masonmcelvain, line 8: Naming: Doesn't this type represent a `ProductApiResponse`, not a device wiki api response? I also feel like this type and the fetcher don't belong in the `devices.ts` file, but maybe a new `product.ts` file

frontend/templates/product/server.tsx
```diff
@@ -23,9 +23,11 @@ export const getServerSideProps: GetServerSideProps<ProductTemplateProps> =
       const { stores, ...otherLayoutProps } = await getLayoutServerSideProps({
          storeCode: DEFAULT_STORE_CODE,
       });
+      const ifixitOrigin = ifixitOriginFromHost(context);
```
masonmcelvain, line 26: Minor: Now that we're storing the result of `ifixitOriginFromHost(context)` in a variable, we can use this variable in `pageProps.appProps` below instead of calling the function again

masonmcelvain (approved):
CR ✌🏻 Nice work, glad it was an easy integration!

deltuh-vee:
QA 🎬 
The sort order is the same as what is returned from the live API.

Looking at the [Fix the World T-Shirt](https://react-commerce-prod-git-product-page-sorted-variants-ifixit.vercel.app/products/fix-the-world-t-shirt), the apparel size seems to be have a different order on cominor.
image
Here are the Women's sizes in sort order from cominor.
image

Curiously enough, there's no sort order populated using the same query on Metabase.
image

frontend/lib/ifixit-api/devices.ts
danielcliu-ifixit: this got merged so fast after qa that i didn't get aroudn to addressing these.I guess ill staple these changes on to the inevitable compatibility notes PR

masonmcelvain:
Why are women's large and medium before women's small? That seems like a bug

masonmcelvain:
They are ordered correctly in or db and on akeneo
```
select * from attribute_options where attribute = "apparel_size" and locale = "en_US" order by sort_order
```

image
https://akeneo.cominor.com/#/configuration/attribute/apparel_size/edit

masonmcelvain:
Seems like a problem with the api, and not the ui so no need to revert this pull or anything.
https://www.ifixit.com/api/2.0/store/product/fix-the-world-t-shirt

danielcliu-ifixit:
@masonmcelvain its a problem with the live data, theres no sort_order data in the live db.

Issue #1578, opened 2023-04-13
Bug: Next.js Bundle Analysis Comment Duplication


ardelato:
## Description

It looks like the `nextjs_bundle_analysis.yml` is continually creating new bundle stat comments. The root cause of the problem is twofold:

1. The `nextjs-bundle-analysi`s package was updated, resulting in a change to the comment text that the existing search logic was trying to find. The relevant change can be found in the following commit: https://github.com/hashicorp/nextjs-bundle-analysis/commit/067cb5de682d51ad30b8aac2f69437303825c3b8
2. The use of `npx` to execute the package causes it to download the latest version of `nextjs-bundle-analysis`, which may introduce breaking changes in CI due to inconsistencies between the versions being used and the expectations in CI.

### To Do:

- [ ] Update the comment finding logic similar to https://github.com/iFixit/ifixit/blob/master/.github/actions/issue-comment/action.yml
- [ ] Use a specific version of `nextjs-bundle-analysis` in CI. This will prevent breaking changes from being inadvertently introduced into the CI process.

PR #1579, opened 2023-04-13
GitHub Actions: Fix Duplicate Nextjs Bundle Analysis Comments

ardelato:
## Description 

The `nextjs-bundle-analysis package` has been updated, which caused a change in the comment text that the search logic was previously relying on. The relevant change can be found in the following commit: hashicorp/nextjs-bundle-analysis@067cb5d

Due to this update, a new `Nextjs Bundle Analysis` comment is created on every new run instead of updating the existing comment in the PR.

### CR Notes:

This updates the comment finding logic to use a different substring and to ensure it is from the `github-actions[bot]` user. 

To avoid this issue in the future I added a specific version to use for `npx -p nextjs-bundle-analysis`. 

### QA Notes:

- [ ] Verify the `Next.js Bundle Analysis` comment does not get duplicated on new pull workflow runs.

Closes: #1578

### DIFF
--- a/.github/workflows/nextjs_bundle_analysis.yml
+++ b/.github/workflows/nextjs_bundle_analysis.yml
@@ -54,7 +54,7 @@ jobs:
          # Here's the first place where next-bundle-analysis' own script is used
          # This step pulls the raw bundle stats for the current bundle
          - name: Analyze bundle
-           run: npx -p nextjs-bundle-analysis report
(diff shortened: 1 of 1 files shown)


deltuh-vee:
QA 🎬 
Duplicate Next.js Bundle Analysis comments are not created.

batbattur (approved):
CR 🚛

PR #1581, opened 2023-04-13
Product Page: Use Sorted Variants (with no bugs!)

danielcliu-ifixit:
## Overview
Alt Text

A redux of https://github.com/iFixit/react-commerce/pull/1572
Same as last time but this time empty option fields fall back to shopify option values. This should solve the issues we were seeing on pages like [repair business toolkit](https://ifixit.slack.com/archives/C0252FETV2S/p1681408439081469)

## QA
Same as last time but make sure it works on the repair business toolkit as well.


masonmcelvain (approved):
CR ✌🏻

PR #1585, opened 2023-04-14
Product Page Cache API Optimization

### LINKED ISSUE #1523: Caching api optimizations
We might be able to make some slight optimizations to the caching system. Right now, we have a baseline 50ms delay on any page that triggers a recache. This is because we make an http request to https://www.ifixit.com/api/nextjs/cache/product, but if the serverless function completes before the request is made, the cache is never refreshed.

50ms isn't too bad, but we might still be able to do better. There might be a library for this, but if not, we could replace our use of `fetch` with a lower level `http` call. https://nodejs.org/api/http.html#httprequesturl-options-callback

```js
new Promise((resolve, reject) => {
   http.request(...).end(resolve).on('error', reject);
})
```

I haven't tried it, but I think this should be able to make the request and finish posting without waiting for a response, though we'll have to try it to be sure. Might have to put a call to destroy() in there somewhere.

The second idea was instead of requesting from https://www.ifixit.com which will go through CloudFront, we might be able to request from the actual vercel domain to avoid the extra round-trips to CloudFront. It's probably a small improvement, but since it will be on so many pages, it might be good to do anyway.

masonmcelvain:
Closes #1523

- Replaces `fetch(); sleep();` with `request().end()` so we can resume the serverless function as soon as the request completes.
- Makes revalidation requests against the vercel domain instead of www.ifixit.com to avoid going through cloudfront.

## QA
Make sure the product page still loads normally. Verify that revalidation time is generally faster than it was before. Verify that we still get cache hits after revalidation. (Note: product page redis cache entries become stale after 1 minute)

CC @sterlinghirsh @dhmacs


masonmcelvain:
Preliminary results:

`main` (https://www.cominor.com/products/iphone-5c-replacement-battery):
```
api/nextjs/cache/product.key: "api/nextjs/cache/product:{"handle":"iphone-5c-replacement-battery","ifixitOrigin":"https://www.cominor.com","storeCode":"test"}"
event cache.findProduct.stale -> 1
cache.findProduct.stale => 61.91ms
---
api/nextjs/cache/product.key: "api/nextjs/cache/product:{"handle":"iphone-5c-replacement-battery","ifixitOrigin":"https://www.cominor.com","storeCode":"test"}"
event cache.findProduct.stale -> 1
cache.findProduct.stale => 131.58ms
---
api/nextjs/cache/product.key: "api/nextjs/cache/product:{"handle":"iphone-5c-replacement-battery","ifixitOrigin":"https://www.cominor.com","storeCode":"test"}"
event cache.findProduct.stale -> 1
cache.findProduct.stale => 55.44ms
```

Feature (https://react-commerce-git-cache-optimization-ifixit.cominor.com/products/iphone-5c-replacement-battery):
```
api/nextjs/cache/product.key: "api/nextjs/cache/product:{"handle":"iphone-5c-replacement-battery","storeCode":"test"}"
event cache.findProduct.stale -> 1
cache.findProduct.stale => 16.72ms
---
api/nextjs/cache/product.key: "api/nextjs/cache/product:{"handle":"iphone-5c-replacement-battery","storeCode":"test"}"
event cache.findProduct.stale -> 1
cache.findProduct.stale => 6.50ms
---
api/nextjs/cache/product.key: "api/nextjs/cache/product:{"handle":"iphone-5c-replacement-battery","storeCode":"test"}"
event cache.findProduct.stale -> 1
cache.findProduct.stale => 46.85ms
```

deltuh-vee:
dev_block ✊ 
I'm seeing nothing but cache misses when reloading [product pages](https://react-commerce-git-cache-optimization-ifixit.vercel.app/products/iphone-6s-screen).
image

frontend/lib/swr-cache/index.ts
```diff
@@ -15,8 +15,9 @@
  *   e.g. `const productList = await ProductList.get({ ... })`
  */
 
-import { APP_ORIGIN } from '@config/env';
+import { APP_ORIGIN, VERCEL_URL } from '@config/env';
 import { log as defaultLog, Logger } from '@ifixit/helpers';
+import { request } from 'https';
```
dhmacs, line 20: Does this work on localhost? We're not using `https` during localhost development, so I think this should break 🤔

frontend/lib/swr-cache/index.ts
```diff
@@ -72,18 +72,29 @@ export const withCache = <
...
+         // Don't delay the lambda beyond 50ms for revalidation
+         setTimeout(resolve, 50);
+         const postData = JSON.stringify(variables);
+         const req = request({
+            hostname: VERCEL_URL || new URL(APP_ORIGIN).hostname,
+            path: `/${endpoint}`,
+            method: 'POST',
+            headers: {
+               'Content-Type': 'application/json',
+               'Content-Length': Buffer.byteLength(postData),
```
dhmacs, line 85: Out of curiosity, why did you specify this header?

dhmacs (commented):
We discussed this last week, and I remember @sterlinghirsh told us that this solution was not really improving perfs due to the request taking longer on average than the `50ms` pause to be sent. Did you and @sterlinghirsh discuss this offline afterward?

deploy_block ⏱️

frontend/lib/swr-cache/index.ts
```diff
@@ -72,18 +72,29 @@ export const withCache = <
...
-         body: JSON.stringify(variables),
-      }).catch((error) => {
-         logger.error(
-            `${endpoint}.error: failed to trigger revalidation\n${printError(
-               error
-            )}`
-         );
+      await new Promise((resolve, reject) => {
+         // Don't delay the lambda beyond 50ms for revalidation
+         setTimeout(resolve, 50);
```
danielbeardsley, line 77: Do we need to cancel this if the request finishes first? i.e. Does calling `resolve()` twice cause any problems?

frontend/lib/swr-cache/index.ts
```diff
@@ -72,18 +72,29 @@ export const withCache = <
...
+         }).on('error', (error) => {
+            logger.error(
+               `${endpoint}.error: failed to trigger revalidation\n${printError(
+                  error
+               )}`
+            );
+            reject();
+         });
+         req.write(postData);
+         req.end(resolve);
```
danielbeardsley, line 96: @sterlinghirsh I remember experimenting with you and we ended up getting it quite low (with localhost I think) but I don't remember if this was the technique we ended with...

masonmcelvain:
@dhmacs
> We discussed this last week, and I remember @sterlinghirsh told us that this solution was not really improving perfs due to the request taking longer on average than the 50ms pause to be sent. Did you and @sterlinghirsh discuss this offline afterward?

Thanks for cluing me in, I didn't realize Sterling and Danny had done some testing already. I spoke to Sterling this morning, and we are going to add some more detailed performance logging to understand how long it typically takes to send the request in the Vercel environment.

frontend/lib/swr-cache/index.ts
masonmcelvain: I haven't tried testing on localhost, so it's very possible that this does not work as intended on localhost. Looking at it again, we can probably use the `http` module and specify the `protocol` (http or http) based on the presence of the VERCEL_URL.

frontend/lib/swr-cache/index.ts
masonmcelvain: No good reason, I was working off the example implementation in Node.js [docs](https://nodejs.org/api/http.html#httprequesturl-options-callback), and copied it from there. I will try dropping it if we can, to keep things simple.

frontend/lib/swr-cache/index.ts
masonmcelvain: @sterlinghirsh had the same question! I don't think so, this always produces `'done'` when I run it with node

```js
(async () => {
   const p = new Promise((resolve, reject) => {
      resolve('done');
      resolve('second resolution');
      throw new Error('oh no!');
      reject('rejection');
   });
   const res = await p;
   console.log(res);
})();
```

MDN also [suggests](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/Promise#effect_of_calling_resolvefunc) that calling it twice has no effect.

> Calling resolveFunc causes the promise to become resolved, so that calling resolveFunc or rejectFunc again has no effect.

masonmcelvain: That being said, I'm happy to clean up the timeout if we don't like this pattern.

danielbeardsley: It's fine by me to leave it.

frontend/lib/swr-cache/index.ts
dhmacs: I think of have to use both, and pick the right one based on the env

danielbeardsley: I think the modules need to be switched based on the url. We can either do an in code switch or use a 3rd party http module which provides a combined interface.

```
import http from "node:http";
import https from "node:https";
const httpModule = url.startsWith("https://") ? https : http;
```

frontend/lib/swr-cache/index.ts
```diff
@@ -72,18 +75,41 @@ export const withCache = <
...
+            logger.success.event(`${endpoint}.revalidation.end`);
+            logger.success.timing(`${endpoint}.revalidation.end`, elapsed);
+            clearTimeout(timeout);
+            resolve();
+         });
       });
-      // We add a small delay to ensure the revalidation is triggered
-      await sleep(50);
+      const elapsed = performance.now() - start;
+      logger.info.event(`${endpoint}.revalidation.total`);
```
danielbeardsley, line 111:
```suggestion
      logger.info.event(`${endpoint}.revalidation.count`);
```

I might also move this to *before* the await. Just in case the promise is rejected.

frontend/lib/swr-cache/index.ts
```diff
@@ -72,18 +75,41 @@ export const withCache = <
...
+               `${endpoint}.error: failed to trigger revalidation\n${printError(
+                  error
+               )}`
+            );
+            reject();
+         });
+         req.write(postData);
+         req.end(() => {
+            const elapsed = performance.now() - start;
+            logger.success.event(`${endpoint}.revalidation.end`);
```
danielbeardsley, line 104:
```suggestion
            logger.success.event(`${endpoint}.revalidation.request_end`);
```

frontend/lib/swr-cache/index.ts
```diff
@@ -72,18 +75,41 @@ export const withCache = <
...
+            logger.success.timing(`${endpoint}.revalidation.end`, elapsed);
```
danielbeardsley, line 105:
```suggestion
            logger.success.timing(`${endpoint}.revalidation.request_end`, elapsed);
```

frontend/lib/swr-cache/index.ts
```diff
@@ -72,18 +75,41 @@ export const withCache = <
...
+            ...(!VERCEL_URL && { port: new URL(APP_ORIGIN).port }),
+            path: `/${endpoint}`,
+            method: 'POST',
+         }).on('error', (error) => {
+            logger.error(
+               `${endpoint}.error: failed to trigger revalidation\n${printError(
+                  error
+               )}`
+            );
+            reject();
```
danielbeardsley, line 99: Might want to record a stat here too... not sure how common it'll be, but it might be nice to know.

frontend/lib/swr-cache/index.ts
```diff
@@ -72,18 +75,41 @@ export const withCache = <
...
+      await new Promise<void>((resolve, reject) => {
+         // Don't delay the lambda beyond 50ms for revalidation
+         const timeout = setTimeout(() => {
+            const elapsed = performance.now() - start;
+            logger.warning.event(`${endpoint}.revalidation.timeout`);
+            logger.warning.timing(`${endpoint}.revalidation.timeout`, elapsed);
+            resolve();
+         }, 50);
+         const postData = JSON.stringify(variables);
+         const req = request({
```
danielbeardsley, line 88: I agree that we should keep things simple, but the docs indicate that setting the `Content-Length` header does have a change in behavior:

image

I don't know if it'll impact performance, but it seems reasonable to test it in preview.

danielbeardsley (dismissed):
CR :+1:

masonmcelvain:
Thanks for the feedback!

We have some more logging, and it appears to work on localhost and Vercel.

Todo:
- Still not sure how to measure whether this change has any effect on the overall lambda response time on a stale cache.
- Play with the Content-Length header to see the effect.
- Maybe drop the post body type checking in the handler, if we're confident on the type. It changed as I was tweaking things, so I added that for consistent results.

danielbeardsley:
> Still not sure how to measure whether this change has any effect on the overall lambda response time on a stale cache.

We have these numbers in datadog, so we should be able to see changes on deploy (if there are any).

danielbeardsley:
CR :+1:

danielbeardsley (approved):
CR :+1:

masonmcelvain:
QA ✌🏻 

✔️ Revalidation works on localhost with upstash
✔️ Revalidation works on Vercel, with subsequent hits after a stale cache
image
✔️ We have logging
image

masonmcelvain:
un_deploy_block ✌🏻 We will analyze the performance on Datadog once this deploys to production

danielbeardsley:
> We have logging

I'm curious... what are those screenshots from? That doesn't look like vercel.

masonmcelvain:
Yea that is the Vercel dashboard. Here is the [link](https://vercel.com/ifixit/react-commerce/4ZRyDqvvjSo36KiVHKETKsobTAsS/logs?page=1&timeline=past30Minutes&startDate=1681925550403&endDate=1681927350403) to the logs page. I think they re-did the logs UI recently or something

masonmcelvain:
Feel free to un_deploy_block and merge if you don't have any more questions

danielbeardsley:
un_deploy_block :+1:

PR #1588, opened 2023-04-14
Strapi: add deploy script

danielbeardsley:
Yes, the use of two repos and a hard-coded path here isn't a great idea. But since we'll be merging these two repos soon, I'm not too worried about it.

Connect iFixit/ifixit#42822


danielbeardsley:
QA :+1: 

Deploy works!
image

masonmcelvain (approved):
CR ✌🏻

PR #1573, opened 2023-04-13
Product pages: Add Reviews Edit Button for Admin

sumukhipandey:
The new react product pages didn't have the `Edit` option for reviews. This PR adds the edit button for admins.

## QA
- If you go to this [link](https://react-commerce-git-product-pages-add-admin-reviews-edit-ifixit.vercel.app/products/mako-driver-kit-64-precision-bits), for example, should now have an edit option if logged in as Admin.
- The redirect from the Edit button should be similar to how it is [here](https://www.ifixit.com/Store/Tools/Pro-Tech-Toolkit/IF145-307#)

Closes #1500

PR #1589, opened 2023-04-15
Reviews Edit : Use sku for url

### LINKED ISSUE #1500: Make Product Reviews Editable on Product Pages
**Problem:**

Product Pages used to have a feature where iFixit staff could edit product reviews, but it is missing on the new react pages.

Mako example: https://www.ifixit.com/products/mako-driver-kit-64-precision-bits

Mako Driver Kit_ Precision Screwdriver Bit Set for Phone, Computer

**Do this:**

Make reviews editable on React product pages. This involves adding an edit link to each review when the logged in user is an admin. The link should follow the same pattern as on the PTT page, i.e. to `https://www.ifixit.com/User/Reviews/IF145-307-4?userid=<userid of reviewer>`

As a reference for how the react pages should look, the special Pro Tech Toolkit and Manta Driver Kit pages still have editable reviews:

Pro Tech Toolkit example: https://www.ifixit.com/Store/Tools/Pro-Tech-Toolkit/IF145-307

Pro Tech Toolkit_ Precision Tools for Phone, Computer Repair (1)

cc: @jeffsnyder @sterlinghirsh

sumukhipandey:
@c00perdanny pointed out in https://github.com/iFixit/react-commerce/issues/1500#issuecomment-1507555803 that the url that the Edit button points to is incorrect. 

In https://github.com/iFixit/ifixit/pull/47399, we add the sku field to the API. This allows us to access the sku for the product that corresponds to a review.

This PR uses this new field to build the correct url. 

Closes #1500 

CC: @danielbeardsley

### DIFF
--- a/frontend/templates/product/sections/ProductReviewsSection/index.tsx
+++ b/frontend/templates/product/sections/ProductReviewsSection/index.tsx
@@ -230,7 +230,7 @@ function ProductReviewLineItem({ review }: ProductReviewLineItemProps) {
          borderColor="gray.200"
          data-testid="product-review-line-item"
       >
-         {isAdminUser && variantSku && (
+         {isAdminUser && (
             <Button
                variant="link"
                as={Link}
--- a/frontend/templates/product/sections/ProductReviewsSection/index.tsx
+++ b/frontend/templates/product/sections/ProductReviewsSection/index.tsx
@@ -122,7 +122,6 @@ export function ProductReviewsSection({
                            <ProductReviewLineItem
                               key={review.reviewid}
                               review={review}
-                              variantSku={selectedVariant.sku}
                            />
                         );
                      })}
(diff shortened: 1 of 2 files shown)


sumukhipandey:
dev_block 🦋 

Waiting for https://github.com/iFixit/ifixit/pull/47399 to be merged

frontend/templates/product/sections/ProductReviewsSection/index.tsx
```diff
@@ -122,7 +122,6 @@ export function ProductReviewsSection({
                            <ProductReviewLineItem
                               key={review.reviewid}
                               review={review}
-                              variantSku={selectedVariant.sku}
```
danielbeardsley, line 125: This prop is still used, though I'm not sure exactly *why* it was in the conditional, we should probably just drop it.

sumukhipandey: f6e90815cb01adbb52c0ec655209951c9e0e5e2d. 
Thanks Danny! Previously, we were using the `variantSKu = selectedVariant.sku` in the url so if the variantSku somehow ended up as undefined or null the link would break which is why we had it in the conditional. 
https://github.com/iFixit/react-commerce/pull/1573#discussion_r1165722871

deltuh-vee:
QA 🎬 
Reviews are now editable for the specific variant on react-commerce.

PR #1591, opened 2023-04-17
Strapi: Update script only deploy on backend/ changes

danielbeardsley:
Specifically, we use a last-strapi-deploy tag and compute the diff from that. If it includes changes to the backend/ dir, we deploy, otherwise, we bail early.

Connect https://github.com/iFixit/ifixit/issues/42822


backend/strapi-deploy
```diff
@@ -1,14 +1,37 @@
...
+
+# Short circuit if the build branch is not main.
+DEPLOY_BRANCH="main"
+if [[ $BUILD_BRANCH != $DEPLOY_BRANCH ]]; then
+   echo "Skipping $BUILD_BRANCH. We only deploy $DEPLOY_BRANCH.";
+   exit 0
+fi
+
+echo "Previous deploy was $(git rev-parse "$DEPLOY_TAG"), listing backend/ changes:"
+if git diff "$DEPLOY_TAG...HEAD" --exit-code --name-only -- backend/; then
```
masonmcelvain, line 18: This logic seems backwards. If there are _are_ changes to `backend/`, then `git diff` will exit with `1`, and this if statement will exit the script early.
```suggestion
if ! git diff "$DEPLOY_TAG...HEAD" --exit-code --name-only -- backend/; then
```

masonmcelvain (approved):
CR ✌🏻 deploy_block ✌🏻 on double checking a potential logic bug

backend/strapi-deploy
danielbeardsley: We get an exit 0 when there is no diff and 1 when there *is* a diff. As backward as it seems, that's the way diff (and git diff) works:

image

backend/strapi-deploy
```diff
@@ -1,14 +1,37 @@
...
 export STRAPI_COMMIT=$(git rev-parse HEAD)
+DEPLOY_TAG=last-strapi-deploy
 
-dir=$(git rev-parse --show-toplevel)
-cd "$dir/backend"
+git fetch --tags --quiet
+
+# Short circuit if the build branch is not main.
+DEPLOY_BRANCH="main"
+if [[ $BUILD_BRANCH != $DEPLOY_BRANCH ]]; then
```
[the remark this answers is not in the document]
danielbeardsley: It's a variable [provided by cimpler](https://github.com/danielbeardsley/cimpler/blob/master/plugins/git-build.js#L34) .

danielbeardsley:
un_deploy_block :+1:

danielbeardsley:
QA :+1: 

image

backend/strapi-deploy
masonmcelvain: Derp, I was confused that a `0` exit code is truthy in bash and an exit code of `1` is falsy. Reading it this way, it now looks right 👍🏻

PR #1602, opened 2023-04-19
Cache timing logs: use statName

danielbeardsley:
endpoint is "/api/nextjs/cache/product" and doesn't get parsed by datadog without changing our log parsing filters.

Use statName like the rest of our timing and event logs in this function.

qa_req 0  (CI + typescript is enough)


masonmcelvain (approved):
CR ✌🏻

PR #1603, opened 2023-04-19
Cache invalidation: make request to stable host

danielbeardsley:
In prod, VERCEL_URL looks like: react-commerce-prod-48uflsij4-ifixit.vercel.app

While this does correct to the correct service, it's reported to the
in the logs (and thus to datadog) as a different service who's name
keeps changing. This makes filtering logs for just react-commerce logs
really hard.

Let's allow overriding this in an env var.

qa_req 0 (CI + typescript is enough)


masonmcelvain (approved):
CR ✌🏻

danielbeardsley:
One unrelated failure. Merging anyway.

PR #1604, opened 2023-04-19
Product Page: Update compatibility linking

masonmcelvain:
Closes [#47433](https://github.com/iFixit/ifixit/issues/47433)

- Uses a `<details>` element instead of a chakra disclosure for compatibility section dropdowns. This gives us the find-in-page feature in chromium browsers.
- Updates compatibility linking to be more intuitive.

## CR

Best viewed with [whitespace hidden and split](https://github.com/iFixit/react-commerce/pull/1604/files?diff=split&w=1).

## QA

- Visit a product with more than 10 model numbers on one of it's devices. For example, `/products/refrigerator-ice-maker-assembly-wr30x10093` in the production preview.
- Make sure both compatibility components meet the spec as described in the [issue](https://github.com/iFixit/ifixit/issues/47433) and on [slack](https://ifixit.slack.com/archives/C01K32B4FPX/p1681837185634129).
- Make sure that on chromium browsers, using find-in-page (Ctrl+f) for a model number in a collapsed compatibility section dropdown opens the dropdown and jumps to the model number.

CC @sterlinghirsh


frontend/components/common/CompatibleDevice.tsx
```diff
@@ -26,102 +28,134 @@ export function CompatibleDevice({
...
+         display="flex"
+         alignItems="start"
+         transition="all 300m"
+         {...(!truncateModels && {
+            onClick: (event) => {
+               toggle();
+               event.preventDefault();
+               event.stopPropagation();
+            },
+            onKeyDown: (event) => {
```
sterlinghirsh, line 43: Do you know why we're handling onKeyDown here?

frontend/components/common/CompatibleDevice.tsx
```diff
@@ -26,102 +28,134 @@ export function CompatibleDevice({
...
-               src={device.imageUrl}
-               alt={device.deviceName ?? ''}
-               w="12"
-               mr="3"
-               borderWidth="1px"
-               borderStyle="solid"
-               borderColor="gray.300"
-               borderRadius="base"
-            />
+            <NextLink href="#compatibility" passHref>
```
sterlinghirsh, line 73: I wonder if there's any advantage to using NextLink for an in-page anchor like this.

frontend/components/common/CompatibleDevice.tsx
masonmcelvain: My intent was to toggle the dropdown if the div was focused and the user pressed 'Enter', but it looks like you can't focus the div. Do you think it should be focusable, or just omit the onKeyDown?

frontend/components/common/CompatibleDevice.tsx
```diff
@@ -26,102 +28,134 @@ export function CompatibleDevice({
...
+               <chakra.a>
+                  <Text
+                     my="2px"
+                     fontWeight="medium"
+                     _groupHover={{ color: 'brand.500' }}
+                  >
+                     {device.deviceName}
+                  </Text>
+               </chakra.a>
```
sterlinghirsh, line 82: If we're going to wrap with chakra.a, is there a reason to also have the Text element? I'd expect chakra.a takes the same style params.

frontend/components/common/CompatibleDevice.tsx
masonmcelvain: None that I'm aware of, but you caught a typo! That's supposed to be `device.deviceUrl` I think

frontend/components/common/CompatibleDevice.tsx
masonmcelvain: I'll try dropping Text, hadn't thought of that.

frontend/templates/product/sections/ProductOverviewSection/CompatibleDevices.tsx
```diff
@@ -21,14 +21,16 @@ export function CompatibleDevices({ product }: CompatibleDevicesProps) {
...
-            <CompatibleDevice
-               key={index}
-               device={device}
-               maxModelLines={MAX_MODEL_LINES_PER_DEVICE}
-               role="group"
-               mb="6px"
-            />
+         {devices.slice(0, MAX_VISIBLE_DEVICES).map((device, index) => (
+            <LinkBox key={index}>
+               <LinkOverlay href="#compatibility" />
```
sterlinghirsh, line 26: Why the LinkOverlay here?

masonmcelvain: Makes the whole CompatibleDevice card a link. Wrapping it in an anchor causes hydration errors, so this is a workaround.

packages/ui/hooks/useDetails.tsx
```diff
@@ -0,0 +1,22 @@
+import { useBoolean } from '@chakra-ui/react';
+import { useCallback, useEffect, useRef } from 'react';
+
+export function useDetails(): readonly [
```
sterlinghirsh, line 4: This looks like it will work though I'm not sure I understand why it's necessary. The places where we use isOpen are pretty limited, just showing different text for different open/closed states, which I would be okay with dropping or using a pure css solution like:

```html
<details>
  <summary>
    <span class="whenClosed">Show More</span>
    <span class="whenOpen">Show Fewer</span>
   </summary
   Stuff goes here
</details>
```
```css
details .whenOpen {
   display: none;
}
details[open] .whenOpen {
   display: initial;
}
details[open] .whenClosed {
   display: none;
}
```

For activating the toggle behavior when clicking the outer component, if we don't need to track the open/closed state, we might be able to get away without a ref or useDetails if the click handler is like:

```js
onClick: (e) => {
   e.currentTarget.getElementsByTagName('details')[0]?.toggleProperty('open');
}
```

packages/ui/components/Summary.tsx
```diff
@@ -0,0 +1,3 @@
+import { ChakraComponent, chakra } from '@chakra-ui/react';
+
+export const Summary: ChakraComponent<'summary'> = chakra('summary');
```
sterlinghirsh, line 3: Many things inside packages/ui are components. I think this could go inside packages/ui/misc rather than creating a new subdirectory. If we end up wanting to keep useDetails

sterlinghirsh (commented):
CR :zap: dev_block :zap: I left a few comments here. I think we should see if we can get away with relying more on the browser defaults for details.

frontend/templates/product/sections/ProductOverviewSection/CompatibleDevices.tsx
sterlinghirsh: I'd like to see what you mean about wrapping the whole thing in an anchor. Is that with a NextLink or also with a normal a tag?

packages/ui/hooks/useDetails.tsx
masonmcelvain: You're right that we don't necessarily need a custom hook for details and chakra wrapper for Summary. However, it's a departure from the pattern of using hooks and Chakra components (instead of native DOM accessors and pure CSS) that we regularly use in this repo. Your solution might be preferable, but it seems more like a conversation of "Should we use native features (potentially  more performant styling without JS) instead of convenience Chakra wrappers?".

When I wrote the useDetails hook, I was trying to mimic Chakra's [useDisclosure](https://chakra-ui.com/docs/hooks/use-disclosure#return-value) hook, for the record.

packages/ui/components/Summary.tsx
masonmcelvain: Ah forgot about ui/misc, thanks

frontend/components/common/CompatibleDevice.tsx
sterlinghirsh: Looks like the Summary element itself will be focusable, so let's omit the onKeyDown.

packages/ui/hooks/useDetails.tsx
sterlinghirsh: I agree it's a departure and 99% of the time I'd be against it, since it's so counter to the react way of doing things. On the other hand, if we can get away without worrying about the open state of the details element in js, then I'd rather keep it simple.

frontend/templates/product/sections/ProductOverviewSection/CompatibleDevices.tsx
masonmcelvain: I don't think it matters which link component we use. I think react is complaining that we are wrapping anchor tags in other anchor tags (since CompatibleDevice has anchor tags in it).

```ts
{/* causes hydration errors */}
<a>
   <CompatibleDevice />
</a>
```

https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element

> [Transparent](https://html.spec.whatwg.org/multipage/dom.html#transparent), but there must be no [interactive content](https://html.spec.whatwg.org/multipage/dom.html#interactive-content-2) descend[a](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element)nt, a element descendant, or descendant with the [tabindex](https://html.spec.whatwg.org/multipage/interaction.html#attr-tabindex) attribute specified.

masonmcelvain: https://chakra-ui.com/docs/components/link-overlay/usage

Chakra's linkbox works around this specification with zindex it seems

> The LinkBox lifts any nested links to the top using z-index to ensure proper keyboard navigation between links.

packages/ui/hooks/useDetails.tsx
sterlinghirsh: BTW with regard to being more performant vs more convenient, I think we've made the decision to be more convenient a lot of times. However, I'm not convinced that in this case the convenience tradeoff is there, especially if you use your new chakra summary element. I think we should see how this looks without the hook and then decide.

frontend/components/common/CompatibleDevice.tsx
masonmcelvain: Kept the next link, because sometimes it's a /Device/... link https://github.com/iFixit/react-commerce/pull/1604/commits/b02d16c0a1446e996e4dab46802b6888c1ad0325

packages/ui/hooks/useDetails.tsx
masonmcelvain: 👍🏻 

I put the CSS as `sx` in a container component to keep it closer to the html. If you think it's better to put it in a dedicated css file just lmk

https://github.com/iFixit/react-commerce/pull/1604/commits/f9ae8642b031243af7163955453d7340813796ab

masonmcelvain:
un_dev_block ✌🏻 Thanks for the review, I think I took a stab at all the comments

frontend/components/common/CompatibleDevice.tsx
```diff
@@ -88,13 +79,18 @@ export function CompatibleDevice({
...
                      event.stopPropagation();
                   },
-                  onKeyDown: (event) => {
-                     if ('Enter' === event.code) {
-                        event.preventDefault();
-                        event.stopPropagation();
-                     }
-                  },
                })}
+               sx={{
```
sterlinghirsh, line 83: Minor point but this seems like it could be on the summary element closer to where it's relevant.

frontend/components/common/CompatibleDevice.tsx
```diff
@@ -114,32 +110,29 @@ export function CompatibleDevice({
...
-               <details ref={detailsRef}>
-                  <Summary
-                     mt="2px"
-                     lineHeight="short"
-                     fontSize="xs"
-                     fontWeight="medium"
-                     color="brand.500"
-                     cursor="pointer"
+               <details>
+                  <summary
```
sterlinghirsh, line 114: Could we keep the style props by using `<chakra.summary>` here?

frontend/components/common/CompatibleDevice.tsx
```diff
@@ -26,102 +27,128 @@ export function CompatibleDevice({
...
-                     onClick={(event) => {
-                        onToggle();
-                        event.preventDefault();
-                        event.stopPropagation();
+               ))}
+            </Flex>
+            {truncateModels && hiddenVariants.length > 0 && (
+               <Text mb="2px" lineHeight="short" fontSize="xs" color="gray.600">
+                  And {hiddenVariants.length} other model
+                  {hiddenVariants.length > 1 ? 's' : ''}...
```
sterlinghirsh, line 109: Do you think this will be good enough?
```suggestion
                  And {hiddenVariants.length} more...
```

frontend/components/common/CompatibleDevice.tsx
```diff
@@ -26,102 +27,128 @@ export function CompatibleDevice({
...
+                  >
+                     <span className="whenClosed">
+                        Show {hiddenVariants.length} more
+                     </span>
+                     <span className="whenOpen">Show fewer</span>
+                  </summary>
+                  <Flex
+                     flexDir="column"
+                     {...(!truncateModels && {
+                        onClick: (event) => {
```
sterlinghirsh, line 132: This click handler isn't doing anything anymore I think.

frontend/components/common/CompatibleDevice.tsx
```diff
@@ -26,102 +27,128 @@ export function CompatibleDevice({
...
+                        cursor: 'pointer',
                      }}
-                     onKeyDown={(event) => {
-                        if ('Enter' === event.code) {
-                           onToggle();
+                  >
+                     <span className="whenClosed">
+                        Show {hiddenVariants.length} more
+                     </span>
+                     <span className="whenOpen">Show fewer</span>
```
sterlinghirsh, line 127: I do like this better than toggling the strings based on the hook, though honestly I would be fine if it just said "Show X more" in both cases. I'm okay to leave it like this if you want though.

frontend/templates/product/sections/ProductOverviewSection/CompatibleDevices.tsx
sterlinghirsh: Ah fair point, though it does raise the question of why we're including a link in CompatibleDevice if we're trying to wrap the whole box in a different link.

frontend/components/common/CompatibleDevice.tsx
masonmcelvain: Yes we could. How is that different than using the summary element as you described here though? https://github.com/iFixit/react-commerce/pull/1604#discussion_r1171940067 Besides the fact that we aren't defining it in a separate file.

frontend/components/common/CompatibleDevice.tsx
masonmcelvain: Wasn't sure how to do a css parent selector for the `[open]` attribute

frontend/components/common/CompatibleDevice.tsx
masonmcelvain: Yea, probably good to be consistent in the wording anyway

frontend/components/common/CompatibleDevice.tsx
masonmcelvain: It prevents the dropdown from being toggled when the model numbers are clicked. I though this was important so users could copy/paste the numbers, but I'm not wedded to it

frontend/components/common/CompatibleDevice.tsx
masonmcelvain: Works for me, and makes things simpler!

deltuh-vee:
dev_block ✊ 
Browser search isn't opening the Compatibility dropdown even though there is a valid search result within.

deltuh-vee:
QA 🎬 
* Compatibility links in the main product section link to the Compatibility section
* Extra compatibility models are hidden by a details section
* The title and image in the Compatibility cards link to the device page.

masonmcelvain:
dev_block ✌🏻 on using the SmartLink from https://github.com/iFixit/react-commerce/pull/1605

frontend/templates/product/sections/ProductOverviewSection/CompatibleDevices.tsx
masonmcelvain: Good point. I chose to conditionally include the links in e6575c214325d052ea834a5e47d069c95afd6fef, which means we no longer need to use the linkbox too.

masonmcelvain:
un_dev_block ✌🏻

deltuh-vee:
dev_block ✊ 
Clicking the image or title in the Compatbility cards just toggles the details, instead of linking to the Device page.

masonmcelvain:
Thanks @deltuh-vee! Should be fixed now, un_dev_block ✌🏻

erinemay:
dev_block 🦑 Right now, clicking on the white space of the card expands the list. Is it possible to make clicking on the text also expand the list?

https://user-images.githubusercontent.com/72166715/235212224-5f338dfa-150c-4591-8e1a-df1b2f486ed6.mp4

Otherwise:

- Product overview compatiblity cards link to #compatibility instead of the device pages.
- `See all compatible` devices links to #compatibility
- Compatibility section cards collapse/expand on click (except for text noted above🛑) 
- Only the title and image of the card in the #compatibility section link to the device page.
- Searching the page for a device number finds the device number. If the number is below the show more button, the toggle is expanded on search and all matches are shown without issue.

masonmcelvain:
> Right now, clicking on the white space of the card expands the list. Is it possible to make clicking on the text also expand the list?

@erinemay I'd love your opinion on this actually. I intentionally prevented the list from expanding/collapsing when the text is clicked, so that users can select model numbers to copy/paste them without the UI changing. However, I'm not a UX expert. Is it more intuitive for the clicking the text to toggle the dropdown, at the expense of users being unable to copy/paste the numbers?

erinemay:
> > Right now, clicking on the white space of the card expands the list. Is it possible to make clicking on the text also expand the list?
> 
> @erinemay I'd love your opinion on this actually. I intentionally prevented the list from expanding/collapsing when the text is clicked, so that users can select model numbers to copy/paste them without the UI changing. However, I'm not a UX expert. Is it more intuitive for clicking the text to toggle the dropdown, at the expense of users being unable to copy/paste the numbers?

Oh that's a really good point. I'd muuuuch rather have that. It's super frustrating to try to copy text that doesn't want to be copied. I think the show more toggle is pretty dang clear enough that people shouldn't have trouble expanding the list. For sure un_dev_block 🦑 and QA 💯 everything else is working.

masonmcelvain:
Thanks for your feedback!

> I don't think it's expected that the card also expands the list

Yea, I'm open to dropping this feature if we don't think it's worth the complexity. @ifixitmaxb requested it, but I think it's something that could be debated. It seems to work as-is though, so we can go ahead with this for now.

sterlinghirsh (approved):
CR :zap:

Issue #1607, opened 2023-04-21
Strapi: Investigate Slowness


danielbeardsley:
### Problem
Some of our strapi graphql queries (specifically for `getProductList`) take a long time (100-200ms is common). It's unclear if it's because of what we are requesting, the size of the response, or random variability.

### Do this
* [X] #1608 - Add `timeAsync()` to all other strapi queries
* Experiment with strapi queries on localhost to see if there's anything that contributes to our response times
   * High variability in strapi time on [iPhone `getProductList` query](https://app.datadoghq.com/logs?query=service%3Areact-commerce-prod.vercel.app%20%40path%3A%22en-US%2FParts%2F%5B...deviceHandleItemType%5D%22%20%40proxy.path%3A%22%2FParts%2FiPhone%22&agg_m=%40timing.strapi.getProductList&agg_t=pc95&cols=host%2Cservice%2C%40timing.strapi.getProductList&index=%2A&messageDisplay=inline&stream_sort=time%2Cdesc&viz=stream&from_ts=1682104404701&to_ts=1682445076359&live=false)

### Supporting data
Here's a distribution for response times from a single query (`/Parts/iPhone_5c`):
![image](https://user-images.githubusercontent.com/26855/233738383-561465f6-9d70-48ab-a44c-5e90e615867f.png)
The same for a less popular page:
![image](https://user-images.githubusercontent.com/26855/233738746-2bb6f07f-0073-427d-8f73-68c0e0681247.png)

Note: These measurements come from vercel and thus include the network overhead from vercel to strapi (ifixit load balancer).

sterlinghirsh:
This may be somewhat mitigated because we have 2 strapi machines and we have improved caching. Reevaluate whether this is important. We could take a brief look into it, but it's not causing an issue right now.

danielbeardsley:
We looked into the variability and it seems unaffected by the one machine -> two machine upgrade. It's also impossible now to see strapi requests for individual pages because most of all the strapi querying happens behind the `api/*` request which has no url params.

Since strapi is a node app, we *might* be able to get some insight into why some requests take a long time by using node's debug tooling (allowing chrome dev tools to connect to it and profile it).

sterlinghirsh:
Haven't had Strapi issues in a while. Closing this.

PR #1608, opened 2023-04-21
Strapi: time all calls to strapi

danielbeardsley:
Let's get `timeAsync()` around every call so we can graph them in datadog. Previously, we were only timing `getProductList`, now we time them all.

Connect #1607


sterlinghirsh (approved):
CR :zap:

danielbeardsley:
QA :+1: 

image


Hmm.. got a 502 (bad Gateway) during one of the vercel deployments. Seems entirely unrelated.

PR #1609, opened 2023-04-24
Strapi: tag with strapi:latest as well

danielbeardsley:
We're moving to using the standard best-practice of strapi:latest for deploys over only strapi:{commit hash}. We now tag images with both.

This makes auto-deploy easier and booting new machines easier.

Connect https://github.com/iFixit/ifixit/issues/47518


danielbeardsley:
QA :+1: 

The script tagged and pushed `strapi:latest`!

sterlinghirsh (approved):
CR :zap:

PR #1610, opened 2023-04-24
Footer Code Cleanup: export some things

aburke07:
We want to reuse these components in the iFixit/ifixit repo, so we need to be able to import them. We also want the styling to remain the same. Some of the default styles are slightly different in ifixit, so we need to set these here. Lastly, we can add an otherProps argument to components that we want to pass different styles into on the ifixit side.

## QA
The footer should look the exact same.

### DIFF
--- a/packages/footer/components/Navigation.tsx
+++ b/packages/footer/components/Navigation.tsx
@@ -21,7 +21,7 @@ import type { NewsletterFormProps } from './Newsletter';
 import type { Menu as MenuType, MenuItem } from '@ifixit/menu';
 import { useTrackedOnClick } from '@ifixit/tracking-hooks';
 
-type NavSectionProps = {
+export type NavSectionProps = {
    menu1: MenuType | null;
    menu2: MenuType | null;
    menu3: MenuType | null;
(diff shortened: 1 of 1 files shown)


jordycosta:
QA 🟢 

The footer looks good to me!
No visual changes when comparing [prod store footer](https://www.ifixit.com/Parts) to the footer on [this preview](https://react-commerce-prod-git-footer-export-cleanup-ifixit.vercel.app/) ✔️

packages/footer/components/Navigation.tsx
```diff
@@ -52,21 +52,33 @@ export const NavigationSection = ({
    );
 };
 
-const NavigationColumn = ({ menu }: { menu: MenuType | null }) => {
+export const NavigationColumn = ({
+   menu,
+   ...otherProps
+}: {
+   menu: MenuType | null;
+}) => {
```
hcutrone, line 60: We can import chakra's `TextProps` so we can state what we are expecting to be in `otherProps`
```suggestion
}: {
   menu: MenuType | null;
} & TextProps) => {
```

hcutrone (approved):
dev_block :burrito: Minor suggestion to be more type safe

aburke07:
Un_dev_block :alien:

jordycosta:
QA 🟢 

Footer is unchanged 👍

hcutrone (approved):
CR :burrito:

Issue #1612, opened 2023-04-25
Product Lists: add to redis cache


danielbeardsley:
### Problem
The [iPhone `getProductList` strapi query](https://app.datadoghq.com/logs?query=service%3Areact-commerce-prod.vercel.app%20%40path%3A%22en-US%2FParts%2F%5B...deviceHandleItemType%5D%22%20%40proxy.path%3A%22%2FParts%2FiPhone%22&agg_m=%40timing.strapi.getProductList&agg_t=pc95&cols=host%2Cservice%2C%40timing.strapi.getProductList&index=%2A&messageDisplay=inline&stream_sort=time%2Cdesc&viz=stream&from_ts=1682104404701&to_ts=1682445076359&live=false) is qutie slow and highly variable. It's pretty easy for a lot of traffic to swamp next.js and we end up overloading strapi.

### Do This
* Add redis caching to `getProductList` using our `swr-cache` library
   * Probably cache *just* the strapi querying portion as it's values change infrequently (maybe a multi-hour ttl? and a several-day stale time?)

Connect https://github.com/iFixit/ifixit/issues/47517

PR #1613, opened 2023-04-25
Product list swr cache

masonmcelvain:
- Adds a [dev dependency](https://the-guild.dev/graphql/codegen/plugins/typescript/typescript-validation-schema) ([GitHub](https://github.com/Code-Hex/graphql-codegen-typescript-validation-schema/tree/main)) to generate zod schemas for all inputs in Strapi's graphql schema.
- Adds a swr-cache endpoint to cache the product list model, with a 1 hour TTL and a 1 day SWR.

Closes https://github.com/iFixit/react-commerce/issues/1612

## QA

- Smoke test the product list pages to make sure they match main and don't produce any new errors in the logs or Sentry.
- While visiting the product list pages, open the Vercel logs and confirm that we get cache misses, stales, and hits when appropriate based on the timings listed above.

### DIFF
--- a/frontend/pages/api/nextjs/cache/product-list.ts
+++ b/frontend/pages/api/nextjs/cache/product-list.ts
@@ -1,0 +1,21 @@
+import { Duration } from '@lib/duration';
+import { ProductListFiltersInputSchema } from '@lib/strapi-sdk';
+import { withCache } from '@lib/swr-cache';
+import { findProductList } from '@models/product-list/server';
+import { ProductListSchema } from '@models/product-list/types';
+import { z } from 'zod';
+
+export default withCache({
+   endpoint: 'api/nextjs/cache/product-list',
+   statName: 'cache.findProductList',
+   variablesSchema: z.object({
+      filters: ProductListFiltersInputSchema(),
+      ifixitOrigin: z.string(),
+   }),
+   valueSchema: ProductListSchema.nullable(),
+   async getFreshValue({ filters, ifixitOrigin }) {
+      return findProductList(filters, ifixitOrigin);
+   },
+   ttl: Duration(1).hour,
+   staleWhileRevalidate: Duration(1).day,
+});
--- a/frontend/templates/product-list/server.tsx
+++ b/frontend/templates/product-list/server.tsx
@@ -51,6 +51,7 @@ export const getProductListServerSideProps = ({
       let shouldRedirectToCanonical = false;
       let canonicalPath: string | null = null;
       const ifixitOrigin = ifixitOriginFromHost(context);
+      const cacheOptions = { forceMiss: hasDisableCacheGets(context) };
 
       switch (productListType) {
          case ProductListType.AllParts: {
@@ -60,7 +61,7 @@ export const getProductListServerSideProps = ({
                      filters: { handle: { eq: 'Parts' } },
                      ifixitOrigin,
                   },
-                  hasDisableCacheGets(context)
+                  cacheOptions
                )
             );
             break;
@@ -96,7 +97,7 @@ export const getProductListServerSideProps = ({
                      },
                      ifixitOrigin,
                   },
-                  hasDisableCacheGets(context)
+                  cacheOptions
                )
             );
 
@@ -115,7 +116,7 @@ export const getProductListServerSideProps = ({
             productList = await timeAsync('findProductList', () =>
                ProductListCache.get(
                   { filters: { handle: { eq: 'Tools' } }, ifixitOrigin },
-                  hasDisableCacheGets(context)
+                  cacheOptions
                )
             );
             break;
@@ -130,7 +131,7 @@ export const getProductListServerSideProps = ({
             productList = await timeAsync('findProductList', () =>
                ProductListCache.get(
                   { filters: { handle: { eqi: handle } }, ifixitOrigin },
-                  hasDisableCacheGets(context)
+                  cacheOptions
                )
             );
 
@@ -164,7 +165,7 @@ export const getProductListServerSideProps = ({
                      },
                      ifixitOrigin,
                   },
-                  hasDisableCacheGets(context)
+                  cacheOptions
                )
             );
             shouldRedirectToCanonical =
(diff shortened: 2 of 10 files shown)


dhmacs (commented):
@masonmcelvain this seems to be taking a different approach from the existing caching strategy, as we're only caching the **Strapi request**, and not the entire **product list model**. The pattern for caching endpoints up to now is to cache models, not individual queries. Are we intentionally taking a more granular approach from now on?

In my opinion, this has pros & cons.

On the **pros side**:
- This allows setting different expirations for different data sources.

On the **cons side**:
- We will increase the number of Redis requests (potentially slower hit times).
- We will have a less straightforward mental model (For example, if we're only caching the Strapi response, I wouldn't call the cache endpoint `product-list`, but rather I'll come up with some other naming convention, like maybe `strapi-product-list`).

masonmcelvain:
Thanks for your feedback @dhmacs!

@danielbeardsley and I reviewed the timing of the other async requests used to build the product list model (Strapi was the initial motivation), and we agree that caching the whole model will be beneficial for end users. 👍🏻 

We are concerned about changes to other data sources not being reflected in the UI in a timely manner (i.e. iFixit's device wiki and image api's, and Algolia for devices with products). Solutions we came up with:

- Setting a 1 hour TTL and a 1-2 day SWR (up for debate/tuning)
  - With a TTL/SWR that's too high, it may take longer for Google to index new products because they won't appear on product list pages until the cache is revalidated.
  - With a TTL/SWR too low, we risk continuing to overload Strapi with requests during revalidation.
- Extending the `disableCacheGets` query parameter to coerce a swr cache miss will give admin users a mechanism to refresh the cache if they are actively making changes to products/devices.

frontend/pages/api/nextjs/cache/product-list.ts
```diff
@@ -0,0 +1,21 @@
...
+import { z } from 'zod';
+
+export default withCache({
+   endpoint: 'api/nextjs/cache/product-list',
+   statName: 'cache.findProductList',
+   variablesSchema: z.object({
+      filters: ProductListFiltersInputSchema(),
+      ifixitOrigin: z.string(),
+   }),
+   valueSchema: ProductListSchema.nullable(),
```
danielbeardsley, line 15: Does that mean the model sometimes returns null too?

frontend/templates/product-list/server.tsx
```diff
@@ -100,7 +113,10 @@ export const getProductListServerSideProps = ({
          }
          case ProductListType.AllTools: {
             productList = await timeAsync('findProductList', () =>
-               findProductList({ handle: { eq: 'Tools' } }, ifixitOrigin)
+               ProductListCache.get(
+                  { filters: { handle: { eq: 'Tools' } }, ifixitOrigin },
+                  hasDisableCacheGets(context)
```
danielbeardsley, line 118: I wonder if we should change `swr-cache` to take in `context` instead of just a bool so we don't have to remember to repeat this function call in every callsite. Thoughts?

dhmacs: Imo it would be better to pass an `options` object that includes flags specific to the model instead. This would help maintain a separation of concerns, as the model shouldn't be aware of the HTTP request context. By keeping this distinction, we can adhere to the MVC, where the model remains independent from the controller (Next.js request). This approach ensures that the model can be used without an HTTP context if needed.

frontend/pages/api/nextjs/cache/product-list.ts
masonmcelvain: `findProductList` can return null, if that's what you're asking. I might not be understanding the question though.

https://github.com/iFixit/react-commerce/blob/09d390a43ab91e18fd3961efd1b2351f3efc9ff7/frontend/models/product-list/server.ts#L38-L41

frontend/templates/product-list/server.tsx
masonmcelvain: I implemented the options idea, let me know what you think https://github.com/iFixit/react-commerce/pull/1613/commits/5b873869a9040430f21a8b6a4ad86730778297a8

dhmacs: @masonmcelvain LGTM

danielbeardsley (approved):
CR :+1:

deltuh-vee:
dev_block ✊ 
I'm getting this console error that isn't present on main when loading /Parts.

image

masonmcelvain:
@deltuh-vee this isn't unique to this branch as far as I can tell. For example, https://react-commerce-git-eslint-no-unused-vars-ifixit.vercel.app/Parts from https://github.com/iFixit/react-commerce/pull/1626 has the same console error.

un_dev_block ✌🏻

erinemay:
>Adds a swr-cache endpoint to cache the product list model, with a 1 hour TTL and a 1 day SWR.
While visiting the product list pages, open the Vercel logs and confirm that we get cache misses, stales, and hits when appropriate based on the timings listed above.

@masonmcelvain, I'm not quite sure I understand how to see what should be happening here. 

To test this, do we need to wait an hour and a day, or is there some way to spoof that change after visiting a page? 

When I go to the same page repeatedly, all I'm seeing are misses. 
https://vercel.com/ifixit/react-commerce/A5wjtz5nJxbTRqrRJioK1CGcv3TX/logs?timeline=absolute&domains=react-commerce-git-product-list-swr-cache-ifixit.vercel.app&startDate=1682718000000&endDate=1682718240000&page=1&selectedLogId=168271820807319275263097870&selectedLogTimestamp=1682718208073&forceEndDate=1682718223516

Is this the right place to check that? In the sidebar details on a specific log event?
image

erinemay:
Ok. Talked to Mason and got the low-down. Plan here for posterity. 

1. Validate hit - Check a common page.
2. Miss - Find an obscure page.
3. Stale - Check back on that obscure page after an hour with logs open.
4. Smoke test collections.

I was looking at the vercel cache. We log our own caching in that same sidebar. Here's an example of a hit.
`cache.findProductList.hit -> 1`
image

erinemay:
Following the steps above, I validated a hit, miss, and stale happened as expected based on my previous comment. After the first stale and miss, reloading or returning to the same page within the hour were hits. 

Did a smoke test and found nothing amiss. Compared this branch with the main preview:
- /Parts and /Tools
- Nested collections
- Navigating the hierarchy via the breadcrumbs
- Using facets, including item-type facets and navigating hierarchy 
- Using the back button

QA 👍

PR #1615, opened 2023-04-26
Upgrade strapi v4.9.2

### LINKED ISSUE #1606: Upgrade Strapi to v4.9.2

dhmacs:
closes #1606 
qa_req 0

This PR upgrades Strapi to enable the new data transfer feature which we aim to use to improve the dev/staging db seeding process. It also patches a [security vulnerability](https://github.com/strapi/strapi/releases/tag/v4.8.0).

❗ ❗  This updates requires a `TRANSFER_TOKEN_SALT` environment variable to be set (you can generate the token with `crypto.randomBytes(16).toString('base64')` in Node.js). 
We also need to set `STRAPI_ADMIN_BACKEND_URL` to an empty string (in Govinor instances having it not being set resulted in an admin client error)

deploy_block 🛑  until we've confirmed that the 2 env vars are set in production

cc @masonmcelvain @sterlinghirsh @danielbeardsley

### DIFF
--- a/backend/config/admin.ts
+++ b/backend/config/admin.ts
@@ -5,4 +5,9 @@ export default ({ env }) => ({
    apiToken: {
       salt: env('API_TOKEN_SALT'),
    },
+   transfer: {
+      token: {
+         salt: env('TRANSFER_TOKEN_SALT'),
+      },
+   },
 });
--- a/backend/docker-compose.yaml
+++ b/backend/docker-compose.yaml
@@ -4,7 +4,7 @@ services:
       build: .
       environment:
          NODE_ENV: production
-         STRAPI_VERSION: 4.6.2
+         STRAPI_VERSION: 4.9.2
          SEED_DB: 'true'
          STRAPI_ADMIN_ENABLE_ADDONS_DANGEROUS_ACTIONS: '${STRAPI_ADMIN_ENABLE_ADDONS_DANGEROUS_ACTIONS}'
          DATABASE_CLIENT: postgres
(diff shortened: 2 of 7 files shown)


masonmcelvain:
> deploy_block :stop_sign: until we've confirmed that the 2 env vars are set in production

Doesn't look like pulldasher picked it up from the pr comment

masonmcelvain:
> > We also need to set STRAPI_ADMIN_BACKEND_URL to an empty string
> 
> I assume this should be set for the strapi deploy and doesn't matter for vercel, right?

Yes that's how I read it. Doesn't look like that env var is referenced in frontend/, only backend/

danielbeardsley:
> This updates requires a `TRANSFER_TOKEN_SALT` environment variable to be set

I set it to random values that differ between Prod and Preview:
image

masonmcelvain:
> I set it to random values that differ between Prod and Preview:

Doesn't the env var need to be set on the Strapi server, not on Vercel?

dhmacs:
@danielbeardsley @masonmcelvain To clarify, they need to be set on the Strapi server, not on Vercel (frontend doesn't use these variables, just Strapi)

danielbeardsley:
> To clarify, they need to be set on the Strapi server,

Derp!   Corrected, I set the token var to a random string and the adminurl var to blank. Current deployed strapi now has these variables.

dhmacs:
@danielbeardsley @masonmcelvain I'll let you merge so you can monitor that the redeployment goes smoothly 👍

PR #1618, opened 2023-04-26
Footer cleanup: Add props to change list test font size

aburke07:
We want to be able to pass in a fontsize prop in the navigation column in iFixit/ifixit that can change the fontsize of the list items. This adds a props that we pass through to the list items.

This should have been included in #1610 but I overlooked these changes.

## QA 
The footer should look exactly the same.


aburke07:
QA :alien: Footer looks the same still :+1:

ianrohde (approved):
CR ⚪

Issue #1583, opened 2023-04-14
Only show Add to cart bar when the user is below the "real" Add to cart button 


erinemay:
### Problem

Currently, the Add to cart bar shows up whenever the "real" Add to cart button is not in view.

The index the bar-button scrolls to does not have the add to cart button in the viewport, and the user is stuck clicking a button that says "Add to cart" that doesn't add the item to their cart. 

Users can repeatedly click "Add to cart" and never be able to add it to their cart. They have to scroll to hid the bar and see the real button.
RBT https://www.ifixit.com/products/repair-business-toolkit

(Updated video)

https://user-images.githubusercontent.com/72166715/234125504-ed948962-f090-4eac-95ec-a7047517a146.mp4

Browserstack recordings below from the iPhone 13, but I confirmed it's happening on the 14 Pro, 13 Pro. Some iPhone models like the iPhone Pro Max have just enough room that the "real" add to cart button is in view once scrolled to. 

<details><summary>Some additional UX recordings</summary>

- It appears to happen on products where there is at least one dropdown option. 
- When the browser's url bar is hidden, it's not a problem. The user has to do some interacting to get it to disappear. 

Fix the world T-shirt https://www.ifixit.com/products/fix-the-world-t-shirt

https://user-images.githubusercontent.com/72166715/231890558-82599572-0269-499d-a686-e33554f40681.mp4

Slightly different experience on my Pixel where clicking on the button shows just enough of the real Add to cart button that the bar isn't shown, but it's still a little odd. This is the Fix the World shirt with two dropdown options. 

https://user-images.githubusercontent.com/72166715/231893935-16419545-6ad4-4290-9ad5-ad8cad127e8c.mp4

</details>

### Do this
When a user is above the add to cart button, don't show the Add to cart bar. Once the user has scrolled past it and it is out of view, show the bar. 

### Additional Information
Pulled from https://github.com/iFixit/react-commerce/issues/1484

cc: @sterlinghirsh

PR #1622, opened 2023-04-27
Update add to cart bar button behaviour

### LINKED ISSUE #1484: Add to cart bar: Add item to cart on click
Currently, clicking on Add to Cart from the bar scrolls back up to the Add to Cart button on the page.
Single variant page: https://www.ifixit.com/products/magnetic-project-mat
Multi variant page: https://www.ifixit.com/products/google-pixel-7-screen-genuine

>I agree it should just add if there's only a single variant. Even with multiple, since they already have one selected, I think it could be fine to just add that, as long as we indicate what the selection is like "iPhone 7 Battery / Fix Kit" or something. I think we have some canonical formatting of the variant name in a metafield or something, but even if we don't, it should be doable.

https://ifixit.slack.com/archives/C01K32B4FPX/p1679076491807469?thread_ts=1679074834.298269&cid=C01K32B4FPX

cc: @sterlinghirsh @dhmacs @mmarcon91

dhmacs:
closes #1484 

## QA

1. Open [product page preview](https://react-commerce-git-update-add-to-cart-bar-behavior-ifixit.vercel.app/products/repair-business-toolkit)
2. Scroll to until you see the add to cart bar
3. Click the add to cart button on the bar
4. Verify the the selected product variant is added to the cart

### DIFF
--- a/frontend/templates/product/sections/ProductOverviewSection/AddToCart/AddToCartBar.tsx
+++ b/frontend/templates/product/sections/ProductOverviewSection/AddToCart/AddToCartBar.tsx
@@ -1,20 +1,18 @@
-import { Box, Button, Flex, HStack, Link, Text } from '@chakra-ui/react';
-import { Rating } from '@components/ui';
-import { shouldShowProductRating } from '@ifixit/helpers';
-import { Wrapper } from '@ifixit/ui';
-import { PRODUCT_OVERVIEW_SECTION_ID } from '../constants';
+import { Box, Button, Flex, HStack, Text } from '@chakra-ui/react';
+import { ProductVariantPrice, Wrapper } from '@ifixit/ui';
+import type { ProductVariant } from '@models/product';
 
 interface AddToCartBarProps {
    title: string;
-   rating?: number | null;
-   reviewsCount?: number | null;
+   variant: ProductVariant;
    active?: boolean;
+   onClickAddToCart: () => void;
 }
 
 export function AddToCartBar({
    title,
-   rating,
-   reviewsCount,
+   variant,
+   onClickAddToCart,
    active = false,
 }: AddToCartBarProps) {
    return (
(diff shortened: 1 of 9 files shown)


deltuh-vee:
dev_block ✊ 
I can't seem to add products to the cart and this error is showing up.
image

dhmacs:
> dev_block ✊ I can't seem to add products to the cart and this error is showing up. image

@deltuh-vee I'm not able to reproduce this error. Can you provide me the exact steps, or maybe a screen recording?

> I don't see anything in the discussion about removing the product rating in the bar, though so long as we're happy with the look, the code seems correct.

@danielbeardsley I've followed the updated [mockup](https://www.figma.com/file/BolmFORO6ZfHXePOSyEGSS/iFixit---Page-sections?node-id=91-9490), the most recent versions are at the bottom. I think the reason @mmarcon91 removed it is because there's not enough space to show the product variant details, the price and the rating

deltuh-vee:
@dhmacs 

https://user-images.githubusercontent.com/5289715/234950126-f4f990a0-f3de-4808-af59-fabfeedea4f6.mp4

dhmacs:
@deltuh-vee can you check if there's a failed network request when you click add to cart? Does it fail also with the in-page add to cart button?

erinemay:
I think there's something wrong with that product. It's happening on the main branch and cominor as well. Other products get added ok. un_dev_block 🧷

erinemay:
It functions as spec'd. QA 🦖 
- Pro and sale discounts are shown and easily fit on the second line of the bar
- The add to cart button adds the currently selected item to the cart
- The details in the bar correspond to the currently selected variant

deploy_block 💭 on some thoughts about the mobile experience. What do y'all think ⬇️?

I noticed some oddness on mobile mostly from the fact that we see the bar above the add to cart button ([issue already written](https://github.com/iFixit/react-commerce/issues/1583)) and the width on mobile is so small. 

For a lot of products, because the product title is so long, no variant details are visible. The only thing that changes on mobile is the addition of the price. Is there another way to make it clear what item is being added to cart?

Example: https://react-commerce-git-update-add-to-cart-bar-behavior-ifixit.vercel.app/products/iphone-12-pro-12-pro-max-dual-sim-card-tray

main:
image
this branch:
image

It's particularly weird on pages whose options are obscured by the bar. Before this, clicking add to cart took the user to the options. Now it adds the default to cart. 

For something like the [sim card tray noted above](https://react-commerce-git-update-add-to-cart-bar-behavior-ifixit.vercel.app/products/iphone-12-pro-12-pro-max-dual-sim-card-tray), it's easy to assume there's only one option available. 

<details><summary>Screenshot</summary>

image

</details>

Should customers know to scroll down to see potential options? I think this problem will be solved by fixing https://github.com/iFixit/react-commerce/issues/1583 anyways, but still worth noting.

The t-shirt is another one. At least part of the variant options make it onto the bar so the user has an idea that one is already selected.

<details><summary>Screenshot</summary>

image

</details>




All my screenshots are from the iPhone 13, but the behavior persists across different mobile sizes.

mmarcon91:
We currently have these 3 proposals, each with pros and cons
Screenshot 2023-05-02 at 11 05 32

### 1 
Line 1: Title; 
Line 2: Review; 
Action Text: Add to cart;
Action: Scroll to buy box

👍 Pro:
Clear selection with all the info
👎 Cons:
Can't add product quickly from the bar

### 2
Line 1: Title; 
Line 2: Variant; 
Action Text: Add to cart + {Price};
Action: Add to cart variant

👍 Pro:
Add to cart from the bar
👎 Cons:
You can't have the original price crossed out if there is a discount

### 3
Line 1: Title + Variant; 
Line 2: Orginal price + Price; 
Action Text: Add to cart;
Action: Add to cart variant

👍 Pro:
Add to cart from bar
👎 Cons:
Can't read the variant on small screens


<hr>

1. More adaptable solution
2. Variant selection clear but could probably convert less because you can't see if there is a discount
3. Might convert better but might cause problems for the user because they might add the wrong variant

We have to handle it this way, another alternative could be that the add to cart button opens a modal for variant selection, however I see it a bit forced.

dhmacs:
cc @erinemay

erinemay:
Yeah, I think this is probably the best of our options. Having the discount present is important. I think most of the issues here will be alleviated by https://github.com/iFixit/react-commerce/issues/1583. If a customer is forced to scroll past the options, then they're aware that certain options are selected.

sterlinghirsh:
We could let the bar get taller when the product title is very long. It's not ideal, but I don't think it looks that bad. (I took off the overflow and line-clamp to make this.)

image

There's also the idea of hiding the product title and just showing the variant title when there's not enough room:

image

This might be a little weird if it just says "New" but maybe it's not so bad. I don't think it will be too confusing to have the add to cart bar without showing the product title if it doesn't fit, especially if it has the variant. Not sure if it would look terrible, but we might be able to put the price on the same line as the variant, though that makes the line even longer and more likely to break.

I think no matter what, we may end up with a title or variant string that's too long, at which point we either have to truncate or break the line. I think truncating is more jarring than breaking the line, so it might just make sense to let the design accommodate more lines, though I'm willing to be overruled if everyone else hates it.

mmarcon91:
@sterlinghirsh  I am afraid that having multiple lines reduces the portion of the screen a bit too much and makes the user experience worse.
Putting only the the text of the variant would work, but we should make conditions in case there is only one variant available and show the title.

In my opinion in this specific case there is not much we could do, we could test all versions at the same time and choose the one that converts better. It's about showing or hiding text fields. 

Because we think one is better than the other but maybe it converts less. Can we start an A/B test?

sterlinghirsh:
I'd like to A/B test it, but we still don't have a good setup for that yet. I'm fine with picking one option and going with it for now, and then we can A/B test it once that's set up.

erinemay:
I think the one we have now is ok.

(I, too, would like to be able to a/b test things like this)

dhmacs:
I'll merge this version then un_deploy_block 👍

PR #1624, opened 2023-04-27
Rename excludeFromHierarchyDisplay to hideFromParent 

federicobadini:
qa_req 0

connects #1406

❗❗❗This contains Strapi changes ❗❗❗

This PR is renaming `excludeFromHierarchyDisplay` to `hideFromParent` to implement the logic described in #1406

PR #1625, opened 2023-04-27
Allow hiding children from product lists parents

### LINKED ISSUE #1406: Add flag to hide product list from its parent children
In #1364 we discussed about Strapi product list field `excludeFromHierarchyDisplay` not being used and @ifixitmaxb found a use case for that. Since the name might be missleading, let's rename it to `hideFromParent` and implement the filter on the frontend.

federicobadini:
🚫 deploy_blocked waiting for #1624

closes #1406 

This PR hide product list children from their parent collection according to their `hideFromParent` flag.

### QA

1. Visit [Vercel preview](https://react-commerce-git-allow-hiding-children-from-pro-4cf252-ifixit.vercel.app/Tools)
2. Try to change the `hideFromParent` flag value
3. Verify that children are correctly hidden from the product list children section when their flag is set to true
4. Verify that the breadcrumbs in the child collection are still correctly referencing the parent collection


masonmcelvain (approved):
CR ✌🏻

deltuh-vee:
QA 🎬 
* Children are correctly hidden from the product list children section when `hideFromParent`
 is set to true.
* Breadcrumbs in a hidden child collection still correctly reference the parent collection.

PR #1626, opened 2023-04-27
ESLint: Turn on `no-unused-vars`

masonmcelvain:
Turns on the [`no-unused-vars`](https://eslint.org/docs/latest/rules/no-unused-vars) rule, and offers prefixing with an underscore as an escape hatch.

qa_req 0


batbattur (approved):
CR 🚛

PR #1629, opened 2023-04-28
Improve Docker compose file to persist media files across re-deployments

### LINKED ISSUE #1628: Persist media files in Govinor Strapi instances
The current setup doesn't keep the uploaded files after the Strapi instance is redeployed.

dhmacs:
closes #1628 
qa_req 0

deploy_block 🛑 on confirming that changing the docker compose file doesn't affect the Strapi production instance (i.e. we're not using the Docker compose file to deploy production, correct?)

cc @masonmcelvain @danielbeardsley @sterlinghirsh

### DIFF
--- a/backend/docker-compose.yaml
+++ b/backend/docker-compose.yaml
@@ -18,12 +18,7 @@ services:
       ports:
          - '${HOST_PORT:-1337}:1337'
       volumes:
-         - uploads:/opt/app/public/uploads
-         - ./config:/opt/app/config
-         - ./src:/opt/app/src
-         - ./package.json:/opt/package.json
-         - ./yarn.lock:/opt/yarn.lock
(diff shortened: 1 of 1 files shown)


masonmcelvain (approved):
CR ✌🏻 AFAIK we only use [`build-docker-and-push`](https://github.com/iFixit/react-commerce/blob/main/backend/build-docker-and-push) to deploy Strapi in production, which doesn't use `docker compose` at all, so this shouldn't affect production Strapi I think.

deploy_block ✌🏻 I'd like @danielbeardsley to confirm though, since he recently implemented the auto-deploy process for Strapi.

djmetzle:
Confirmed, no docker-compose for the production `strapi` deploy.

Issue #589, opened 2022-08-05
Collections: have title, tagline and description copy on Item Type collections


ifixitmaxb:
## Problem
We'd like to be able to edit Title, Tagline, Description, MetaDescription, and MetaTitle fields on Item Type collections.

## Do This
Try adding a repeatable section to the ProductList type that lets you specify Item Type, Title, Tagline, Description, MetaDescription, and MetaTitle. e.g. on iPhone, we could create a repeatable section "Item Type Override" like:
```
Item Type: Screens
Title: Best iPhone 7 Replacement Screens
Tagline: They're the best!
Description:
MetaDescription:
MetaTitle:
```
This would make it so when a user navigates to the /Parts/iPhone_7/Screens they will see "Best iPhone 7 Replacement Screens"and "They're the best!" instead of the default title and tagline. Same for the other fields specified. 

Template values may be specified by using `[DEVICE]` and `[ITEM]`, which will be replaced with the currently selected device and item type. e.g.
```
Item Type: 
Title: Best [DEVICE] [ITEM]
Tagline: They're the best [ITEM] for your [DEVICE]!
Description:
MetaDescription:
MetaTitle:
```
This would produce a title like "Best iPhone 7 Batteries" and tagline "They're the best Batteries for your iPhone 7".

When Item Type is blank, the values will be used as fallbacks for when the requested item type matches no other section. When another field is empty, it should fall back in this order: Matched Item Type, Blank Item Type, Page Default

This issue has been merged with #590

ifixitmaxb:
Related: https://github.com/iFixit/react-commerce/issues/590

ifixitmaxb:
"This new Strapi field is rad thanks Daniel and the rest of the team!" -Evan Gattuso, 5/16/23

PR #1630, opened 2023-04-28
Implement Item Type Overrides

danielcliu-ifixit:
## Overview

The issue is probably better to read for full context, but this pull is implementing the code to allow item type product list pages to have custom title, metatitle, description, metadescription, and tagline attributes. It allows people to set custom item type data in strapi which then gets rendered on the page. 

A couple of notes: 
1.) the Item Types override for the facet item tag they match. So the "Screens" will use override data for the itemTypeOverride with itemType = "Screens". If there are multiple itemTypeOverrides with the itemType="Screens" it will use the last one. If there is no itemType provided for the itemTypeOverride it will be a "fallback" override, which will be used on any item type product list page that doesn't have a specfic override for it.
2.) item overrides will allow for the dynamic replacement of devices and item names on the client side, but only if the string of the attribute uses `[DEVICE]` or `[ITEM]`

## CR

There is a test added but it is being skipped due to issues with our playwright mocking. It will be unskipped when that is fixed.

## QA

You'll want to have a product list page, modify the akeneo data to have itemTypeOverride(s), making sure that specific and fallback options all work and are rendered on the screen correctly. 

Closes: https://github.com/iFixit/react-commerce/issues/589


deltuh-vee:
QA 🎬
* Itemtypes without overrides use a blank or default override
* Itemtypes with multiple overrides will use the last one
* Itemtypes can have override values set

❓ Is the metaTitle supposed to be somewhere on the page?

danielcliu-ifixit:
@deltuh-vee good catch!

deltuh-vee:
QA 🎬 
Metatitle is now present on the page

sterlinghirsh:
dev_block :zap: I think we still need to split this into backend/frontend pulls.

frontend/models/product-list/server.ts
```diff
@@ -428,6 +431,48 @@ function createProductListSection(
...
+      (itemOverride): ProductListItemTypeOverride | null => {
+         if (
+            itemOverride == null ||
+            itemOverride.__typename !== 'ComponentProductListItemTypeOverride'
+         ) {
+            return null;
+         } else {
+            return {
+               id: itemOverride.id,
+               itemType: itemOverride.itemType ?? null,
```
sterlinghirsh, line 463: Why are these all `?? null`? I would have expected these fields are always present if blank strings.

frontend/models/product-list/server.ts
```diff
@@ -428,6 +431,48 @@ function createProductListSection(
...
+   const formatedOverrides = itemOverrides.map(
+      (itemOverride): ProductListItemTypeOverride | null => {
+         if (
+            itemOverride == null ||
+            itemOverride.__typename !== 'ComponentProductListItemTypeOverride'
+         ) {
+            return null;
+         } else {
+            return {
+               id: itemOverride.id,
```
sterlinghirsh, line 462: Do we need this id?

frontend/templates/product-list/MetaTags.tsx
```diff
@@ -88,17 +77,15 @@ export function MetaTags({ productList }: MetaTagsProps) {
...
             </>
          )}
-         <title>{title}</title>
-         <meta property="og:title" content={title} />
+         <title>{metaTitle}</title>
+         <meta property="og:title" content={metaTitle} />
          <meta
             name="og:description"
-            content={productList.metaDescription || productList.description}
+            content={productList.metaDescription ?? undefined}
```
sterlinghirsh, line 88: This seems odd. Is this because we're explicitly handling the metaDescription -> description fallback elsewhere? In that case, it seems like it'd make more sense to conditionalize the meta tag (and the one on line 80) entirely rather than include the tag with an empty description.

frontend/templates/product-list/ProductListView.tsx
```diff
@@ -25,17 +30,28 @@ export function ProductListView({ productList }: ProductListViewProps) {
...
+   const page = pagination.currentRefinement + 1;
+   const productListModified = updateProductListWithItemTypeOverrides(
+      productList,
+      page,
+      itemType
+   );
+
    return (
       <>
          <SecondaryNavigation productList={productList} />
```
sterlinghirsh, line 44: Seems like it would make sense to use the same productListModified everywhere. It's hard to tell from here where it matters or not, but it seems like a good approach in general to be consistent.

sterlinghirsh: One idea would be to have ProductListView be a thin wrapper that does the item type overrides and then have the bulk of this be in another component that only has a single productList in scope.

sterlinghirsh:
CR :zap: dev_block :zap: still stands. I left a few comments, most of which are pretty minor. Aside from splitting this into frontend/backend pulls, I'm most concerned with the one where we use productList some places and productListModified elsewhere. Seems like that will lead to some kind of weird hard-to-find bug down the line.

frontend/templates/product-list/MetaTags.tsx
danielcliu-ifixit: its because we convert everything into strings or null and content only takes strings or undefined, not null.

frontend/models/product-list/server.ts
danielcliu-ifixit: was trying to keep it consistent with the other types in `types.ts` but we can make them undefinable too.

frontend/templates/product-list/ProductListView.tsx
danielcliu-ifixit: ok so because we use productList.title in the secondaryNavigation im not sure we should be passing it to that, it seems to make breadcrumbs wrong.

danielcliu-ifixit:
Gonna split the pull into backend and frontend ones

PR #1631, opened 2023-05-01
Timing: Log time spent in synchronous component code

danielbeardsley:
As part of next.js process, they evaluate our entire component tree from the page on down. JSX compiles into synchrnous functions that perform logic, then return something like:

    React.createElement('div', attrs, children)

This process effectively involves a single run-through of our entire component tree. React then takes this resultant structure and translates it directly into html. We have a lot of control over the component code and shape. We don't have much control over (nor chance to optimize) the translation to html.

So, let's log the time take for our part on every page.

### DIFF
--- a/frontend/layouts/default/index.tsx
+++ b/frontend/layouts/default/index.tsx
@@ -17,6 +17,7 @@ import {
    faMagnifyingGlass,
 } from '@fortawesome/pro-solid-svg-icons';
 import { useAppContext } from '@ifixit/app';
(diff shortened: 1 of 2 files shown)


andyg0808 (approved):
CR :+1:

danielbeardsley:
QA :+1: 

Notably, this cost is *not* paid when hitting next's data api (navigation via links using `next/link` component) which just returns json props and not html.

image

Issue #1611, opened 2023-04-24
Product List Hero is fuzzy


sterlinghirsh:
On the page https://www.ifixit.com/Parts/Nokia_G22 ([strapi](https://strapi.ifixit.com/admin/content-manager/collectionType/api::product-list.product-list/2947)), it looks like the page is using this image https://ifixit-strapi-uploads.s3.amazonaws.com/large_nokia_topdown_header_v2_crop2_dc24e1ab4d.jpg but strapi says we're using this one https://ifixit-strapi-uploads.s3.amazonaws.com/nokia_topdown_header_v2_crop2_dc24e1ab4d.jpg which is much bigger. Let's figure out why it's starting with a smaller version of the image and see if we can fix it.

federicobadini:
The process through which Strapi handles uploads is the following:

### Ingestion 

Strapi is currently set with Responsive friendly upload

<img width="1723" alt="Screenshot 2023-04-27 at 10 39 08" src="https://user-images.githubusercontent.com/7805759/234820266-124db4aa-3f04-4214-bfb2-1a5f3cbeeedf.png">

So whenever an image is uploaded it is scaled down to the following sizes: 

Name | Largest Dimension
-- | --
large | 1000px
medium | 750px
small | 500px

I think there is also a `thumbnail` size even though it is not showing up in the documentation.

### Retrieval

Whenever an image is queried from Strapi we retrieve a structure like this:

```
{
  "alternativeText": null,
  "url": "/uploads/lifetime_guarantee_background_9c0e9b9baf.webp",
  "formats": {
    "thumbnail": {
      "name": "thumbnail_lifetime-guarantee-background.webp",
      "url": "/uploads/thumbnail_lifetime_guarantee_background_9c0e9b9baf.webp"
      ...
    }
  }
}
```

`url` points to the original image - `formats` points to the scaled image set

### Processing

[This util](https://github.com/iFixit/react-commerce/blob/7ebe0d21bd38f6aa42148b27432b8837208995ee/frontend/helpers/strapi-helpers.ts#L11) is used to parse the image data described above. Images processed via that util are then capped below 1000px.

### Wrap-up
All this flow was useful given that we were using `unoptimized` Next images.
Given that now we wanna try exploiting the next optimization, we could remove that util (and also remove the strapi optimizer and the formats part in the query) and just rely on Next.

federicobadini:
no performance degradation has been recorded since the PR was merged. Closing

PR #1632, opened 2023-05-01
Rely on Next image instead of Strapi formats for some Strapi provided images

federicobadini:
connects #1611

This PR removes the capping of hero images due to Strapi formats.
We remove the capping only on this section to verify that NextJS images scaling performs well and do not produce delays when downscaling large images.
If this proves effective for the hero we can think to rely more on Next image, removing the Strapi formats capping also on other Strapi provided images.

### QA

1. Visit Vercel preview
2. Verify that hero images are not capped anymore to 1000px width
3. Verify that hero images are scaled down by Next image to an acceptable size even when the originals are quite large

### DIFF
--- a/frontend/helpers/strapi-helpers.ts
+++ b/frontend/helpers/strapi-helpers.ts
@@ -10,13 +10,15 @@ interface Image {
 
 export function getImageFromStrapiImage(
    image: Pick<UploadFile, 'formats' | 'alternativeText' | 'url'>,
-   format: StrapiImageFormat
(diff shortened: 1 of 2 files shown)


sterlinghirsh (approved):
CR :zap:

deltuh-vee:
QA 🎬 
* hero images are not capped anymore to 1000px width
* hero images are scaled down by Next image to an acceptable size even when the originals are quite large

PR #1642, opened 2023-05-04
Add search to compatibility section

### LINKED ISSUE #1437: Product Page Compatibility: Add simple model search
## Problem
Along the lines of #1436, the list of model numbers on highly compatible products like https://www.ifixit.com/products/frigidaire-241798231-ice-maker-assembly#compatibility becomes ridiculously long (thousands of lines). We still want to have all these numbers visible to Google, but it's a bad UX to have such a long list. #1436 hides most of the model numbers, but it's still not great to make the user dig through thousands model numbers or rely on their browser search.

## Do This
Add a search bar at the top of the compatibility section that filters the displayed devices and model numbers. The search should update as the user types. It should rely on the data from the page without having to use Algolia or any external service.

Example:
Suppose we show a part compatible with 2 devices, each with 3 variants:
- Frigidaire Refrigerator
  - ABC123
  - ABC124
  - ABC245
- Electrolux Refrigerator
  - EFG345
  - EFG346
  - EABC123

Search terms should be matched in both the device and model. Searching "Frigidaire" should return the Frigidaire Refrigerator with all its variants, hiding the Electrolux Refrigerator. Searching "ABC" returns the Frigidaire with all its variants and the Electrolux with variant EABC123. Any hidden variants (hidden by the changes in #1436) should be shown, but we'll still want to limit to the first 10 matching variants per device unless the user decides to select the full list.

federicobadini:
closes #1437 

This PR also fixes some minor details on the functionality to show hidden variants.
The disclose button has been moved to the bottom of the list when the hidden variants are shown.

### QA

1. Visit Vercel preview
2. Verify that the implementation is coherent with the [mockup](https://www.figma.com/file/BolmFORO6ZfHXePOSyEGSS/iFixit---Page-sections?node-id=159-5560&t=3SorVnQp6rEmhV8U-0)

### DIFF
--- a/frontend/components/common/CompatibleDevice.tsx
+++ b/frontend/components/common/CompatibleDevice.tsx
@@ -1,6 +1,16 @@
-import { chakra, Flex, FlexProps, Img, Link, Text } from '@chakra-ui/react';
+import {
+   Button,
+   chakra,
+   Flex,
+   FlexProps,
+   Img,
+   Link,
+   Text,
+} from '@chakra-ui/react';
 import { ConditionalWrapper } from '@ifixit/ui/misc';
 import type { Product } from '@pages/api/nextjs/cache/product';
+import React from 'react';
+import TextWithHighlight from './TextWithHighlight';
 
 export interface CompatibleDeviceProps extends FlexProps {
    device: Omit<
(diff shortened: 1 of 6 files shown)


masonmcelvain (approved):
CR 🙌🏻  The implementation looks great, nice job!

This may be outside the scope of this PR, but now that we are no longer using a `<details>` element we can get rid of some of the event handlers in `CompatibleDevice.tsx`.

For example, the top level `Flex` has an `onClick` handler that was intended to toggle the details element when the flex box was clicked. Within the component, both links and the two flex boxes containing the variants have `onClick` handlers to stop propagation to the details element. These could all be deleted now.

@sterlinghirsh ccing you to confirm that no longer using a details element is OK now that we have our own search box.

sterlinghirsh:
I think it should still use the details element. We discussed it this morning and it's annoying that long lists will cause the "hide" button to go all the way to the bottom of the list.

federicobadini:
> There is a white background in this preview, and a background that matches the rest of the webpage in the mock-up

There is an additional panel that shows the usage when the background is not white. See here:
https://www.figma.com/file/BolmFORO6ZfHXePOSyEGSS/iFixit---Page-sections?type=design&node-id=1363-16640&t=0d3vYg5gGrb7AuMe-0

un_deploy_block

PR #1645, opened 2023-05-04
Use new Strapi import/export CLI

### LINKED ISSUE #878: Enable seed plugin to import deeper structures for nested repeatable components
Categories in store home page browse section are not imported correctly by the seed process due to the algorithm to going deeper in parsing the imported json. Let's improve this so that we don't have to manually fill those values for every new branch

dhmacs:
closes #878 

## QA

1. Open the Govinor branch instance
... WIP


deltuh-vee:
QA 🎬
Data can now be imported from another Strapi instance.

backend/src/plugins/addons/server/services/seed/backup/download.ts
```diff
@@ -0,0 +1,72 @@
...
+         .on('error', (error) => {
+            fs.unlink(dest, (err) => {
+               console.log("Couldn't delete file: ", err);
+            });
+            reject(error);
+         });
+   });
+}
+
+function ensureDirectoryExistence(filePath: string) {
```
masonmcelvain, line 66: This seems to be a duplicate of `ensureBackupFolderExists` in `services/seed/backup/export.ts`, we could extract a helper to `utils` but definitely not necessary.

backend/src/plugins/addons/server/services/seed/backup/download.ts
```diff
@@ -0,0 +1,72 @@
...
+      console.error(err);
+      throw new Error(`Failed to download backup from: ${exportURL.href}`);
+   }
+}
+
+interface GetImportPathInput {
+   isEncrypted: boolean;
+}
+
+export function getBackupFilePath({ isEncrypted }: GetImportPathInput) {
```
masonmcelvain, line 36: This seems like a duplicate function from `import.ts`. Since `seed.ts` looks like it expects the downloaded backup and import path to be the same, perhaps `downloadBackup()` could return a file path, and `importBackup()` could accept a file path as a param/option.

backend/src/plugins/addons/admin/src/pages/HomePage/index.tsx
```diff
@@ -39,80 +101,62 @@ function HomePage() {
...
+            </Box>
+         )}
+      </Flex>
    );
 }
 
-function requireShopifyDomain(value: FormDataEntryValue | null) {
+const URL_REGEX = /^https?:\/\//i;
+
+function requireStrapiDomain(value: FormDataEntryValue | null) {
```
masonmcelvain, line 154: Another sort-of-duplicate with `seed.ts`

masonmcelvain (approved):
CR ✌🏻 Thanks for making this new Strapi feature more accessible!

I made a few non-blocking observations. Based on reading the code, it seems to me that in order to use this feature one must first create a backup on the source Strapi instance, then import that backup on a target Strapi instance.

We should consider adding a reminder to the UI or the README to first create a backup on the source instance before importing.

Alternatively, we could automatically export a backup on the source instance at import-time. This might look like an API request from the target instance to the source instance requesting a backup to be exported, immediately followed by an import of that backup. That way, users wouldn't have to remember to create a backup on the source instance.

deploy_block ✌🏻 on considering adding a reminder to the UI, or making the backups automatic.

backend/src/plugins/addons/server/services/seed/backup/download.ts
dhmacs: Great suggestion! 👏

dhmacs:
@masonmcelvain I've applied the suggested refactoring.

> We should consider adding a reminder to the UI or the README to first create a backup on the source instance before importing.
> Alternatively, we could automatically export a backup on the source instance at import-time. This might look like an API request from the target instance to the source instance requesting a backup to be exported, immediately followed by an import of that backup. That way, users wouldn't have to remember to create a backup on the source instance.

This is all valid feedback. I'd like to make the process seamless as you suggest, but I'm planning to do that in another PR. This is just PR is just part of the work. The next batch of work is to make this automatic when the Strapi server boots, just like the current seeding process is working 👍 

un_deploy_block 🍏

masonmcelvain (approved):
CR ✌🏻 Thanks for the changes! And thank you for clarifying that this is part of a series of PR's, that makes sense to me 😄

deltuh-vee:
dev_block ✊ 
I tried importing from the backup and I think it broke the Govinor instance. I'm seeing a blank screen when reloading the page.
image

dhmacs:
un_dev_block 👍 The refactor introduced a small bug, I've fixed it.

@ardelato playwright tests are now failing due to being unable to fetch data from cominor. It seems like cominor is down maybe?

ardelato:
@dhmacs we should be good now.

I had to restart the `dev-server` for `www.cominor.com`

masonmcelvain (approved):
CR ✌🏻

k0rvusk0r4x:
I can't seem to log into strapi. The credentials on overlord keep being rejected.

deltuh-vee:
Carry Over QA 🎬

PR #1648, opened 2023-05-04
Use fontawesome env var in npmrc

masonmcelvain:
Using process.env in  #1636 seems to have broken preview deployments, see https://github.com/iFixit/react-commerce/pull/1647 for an example.

qa_req 0 If it passes CI

CC @ianrohde @danielbeardsley


danielbeardsley (approved):
CR :+1:

PR #1640, opened 2023-05-03
Add itemTypeOverrides to backend schema.

danielcliu-ifixit:
## Overview
The issue is probably better to read for full context, but this pull is implementing the code to allow item type product list pages to have custom title, metatitle, description, metadescription, and tagline attributes. It allows people to set custom item type data in strapi which then gets rendered on the page. This only implements the backend changes needed.

## QA

Do product list pages still work?

Connects: https://github.com/iFixit/react-commerce/issues/589

Just the backend code from https://github.com/iFixit/react-commerce/pull/1630

PR #1649, opened 2023-05-04
Item type overrides  frontend

danielcliu-ifixit:
## Overview

Follow up to https://github.com/iFixit/react-commerce/pull/1640

The issue is probably better to read for full context, but this pull is implementing the code to allow item type product list pages to have custom title, metatitle, description, metadescription, and tagline attributes. It allows people to set custom item type data in strapi which then gets rendered on the page. 

A couple of notes: 
1.) the Item Types override for the facet item tag they match. So the "Screens" will use override data for the itemTypeOverride with itemType = "Screens". If there are multiple itemTypeOverrides with the itemType="Screens" it will use the last one. If there is no itemType provided for the itemTypeOverride it will be a "fallback" override, which will be used on any item type product list page that doesn't have a specfic override for it.
2.) item overrides will allow for the dynamic replacement of devices and item names on the client side, but only if the string of the attribute uses `[DEVICE]` or `[ITEM]`

## CR

There is a test added but it is being skipped due to issues with our playwright mocking. It will be unskipped when that is fixed.

## QA

You'll want to have a product list page, modify the akeneo data to have itemTypeOverride(s), making sure that specific and fallback options all work and are rendered on the screen correctly. 

Closes: https://github.com/iFixit/react-commerce/issues/589


deltuh-vee:
dev_block ✊ 
The itemtype overrrides aren't showing up for me. Here's the Device I was using:
https://item-type-overrides--frontend.govinor.com/admin/content-manager/collectionType/api::product-list.product-list/2010?plugins[i18n][locale]=en
https://react-commerce-git-item-type-overrides-frontend-ifixit.vercel.app/Parts/iPhone_13/Adhesives?disableCacheGets

danielcliu-ifixit:
un_dev_block :v:

deltuh-vee:
QA 🎬
- Itemtype overrides can show up on the page
- Itemtypes without overrides use a blank or default override
- Itemtypes with multiple overrides will use the last one
- Itemtypes can have override values set

danielcliu-ifixit:
qa :v:  carryover just a merge of `main`

masonmcelvain (commented):
I didn't have time to finish my CR, but here are my initial thoughts. Nice work!

frontend/models/product-list/server.ts
```diff
@@ -425,6 +428,47 @@ function createProductListSection(
...
+>[0];
+
+function formatItemTypeOverrides(
+   itemOverrides: ApiProductListItemOverrides[] | undefined
+): ProductListItemTypeOverrideIndexed {
+   if (!itemOverrides) return {};
+   const convertedOverrides =
+      convertToProductListItemTypeOverrides(itemOverrides);
+   return convertedOverrides.reduce((result, item) => {
+      result[item?.itemType || '*'] = item;
```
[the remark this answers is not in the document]
danielcliu-ifixit: yes '*' means fallback

frontend/helpers/product-list-helpers.ts
```diff
@@ -109,3 +117,165 @@ export function getAdminLinks({
...
+   itemType?: string
+): ProductList {
+   const productListCopy = { ...productList };
+   productListCopy.title = getTitle(productList, itemType);
+   productListCopy.metaTitle = getMetaTitle(productList, itemType);
+   productListCopy.description = getDescription(productList, page, itemType);
+   productListCopy.metaDescription = getMetaDescription(productList, itemType);
+   productListCopy.tagline = getTagline(productList, page, itemType);
+   return productListCopy;
+}
```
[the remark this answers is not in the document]
danielcliu-ifixit: >The copy gets references to array/object fields of the original due to the shallow destructure. If the copy's referential fields get modified, so do the original's. This can be unexpected and hard to debug.

Good point, although would you be fine with this solution if we deep copied instead? Because i don't think thats too hard

>This function encourages the caller to create a variable like productListModified. It's easy to accidentally use productList where you intend to use productListModified, and vice versa, since they are the same type.

Ok i think the intent was to move away from modifing data conditionally in various components and have all the necessary data in the productlist. @sterlinghirsh do you have a particular response to which way you want the data to be?

Personally i feel like this is segmented enough from the productList being used, but im open to others thoughts.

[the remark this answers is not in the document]
sterlinghirsh: @danielcliu-ifixit and I went back and forth on this design for a while. I think there are tradeoffs either way. First of all, I don't think it's necessary to use a deep copy here since we shouldn't be modifying these values anyway. Using a deep copy on the off chance that someone starts modifying values in the future seems wasteful. I am more sympathetic to the problem of having productList vs productListModified. One option is making the default value be "productListOriginal" or making the productListModified var in an outer component that passes it in as productList to the inner component to keep the naming consistent.

I'm not sure I understand the comment about defining the fallback logic near the callers, but I'm not attached to doing it in the product-list-helpers file if it's only used in one place. I would rather not put the fallback + template logic in each place the title, desc, etc are referenced. Centralizing that logic at a high level and simplifying the inner components makes more sense to me.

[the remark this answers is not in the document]
masonmcelvain: You both make good points and changed my mind about certain things. I agree that a deep copy is overkill.

> I'm not sure I understand the comment about defining the fallback logic near the callers

Some of this data is only ever used in one place, e.g. the `metaTitle` is only used in `<MetaTitle>`. So I was suggesting that we could move the fallback logic from here to MetaTitle.

However, I agree that repeating fallback logic for fields used in multiple places isn't a good idea either, e.g. the `description` is used in `<HeroDescription>` and `<ProductListCard>`.

Another idea:
We could add item type specific data to the product model after we fetch it from the cache, since the model is stored in an item type agnostic way. That way, we could have access to the original data and the item type specific data in our UI. (e.g. we use the original product list `title` for related posts, but the item type `title` in the `<HeroSection>`). All the UI has to do is pick which one it's interested in (perhaps `productList.itemTypeOverrides.metaTitle ?? productList.metaTitle`)

```ts
type ProductListWithItemType = {
   itemTypeOverrides: {
      title: string,
      metaTitle: string,
      ...
   },
   ... // the rest of the base productlist fields
};
```

Or another idea:
We store item type override data fore every item type on the product list model (can be cached too). That way we don't need a new `ProductList` type either.

```ts
type ProductList = {
   itemTypeOverrides?: {
      Batteries: {
         title: string,
         metaTitle: string,
         ...
      },
      Screens: { ... },
      ...
   },
   ... // the rest of the base productlist fields
};
```

Then the UI can choose fallbacks like:
```ts
const itemType = useDevicePartsItemType();
productList.itemTypeOverrides?.[itemType]?.metaTitle ?? productList.metaTitle
```

frontend/tests/playwright/product/product-list.spec.ts
```diff
@@ -0,0 +1,207 @@
+import { test, expect } from '../test-fixtures';
+import { createGraphQLHandler } from '../msw/request-handler';
+
+test.describe('Product List Data', () => {
+   test.skip(true, 'Only run once mocks to getProductList are fixed.');
```
masonmcelvain, line 5: Is this test intentionally skipped?

masonmcelvain (changes requested):
CR ✌🏻 Nice job, thanks for adding tests too!

Blocking on unskipping the test. I would like to come up with a way to add these overrides to the product list model, but since this seems to work I won't block on it.

frontend/tests/playwright/product/product-list.spec.ts
danielcliu-ifixit: yes. there is an issue where mocks with strapi are completely busted. waiting for that to be fixed, but ive checked without the mocking to make sure the tests will pass.

frontend/helpers/product-list-helpers.ts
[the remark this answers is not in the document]
sterlinghirsh: Storing it for every item type requires knowing every possible item type, which will be based on the Algolia result. Might be possible, but I don't really see the advantage of precalculating these values for the dozen item types the user isn't viewing.

frontend/helpers/product-list-helpers.ts
```diff
@@ -118,18 +118,19 @@ export function getAdminLinks({
...
    itemType?: string
 ): ProductList {
-   const productListCopy = { ...productList };
-   productListCopy.title = getTitle(productList, itemType);
-   productListCopy.metaTitle = getMetaTitle(productList, itemType);
-   productListCopy.description = getDescription(productList, page, itemType);
-   productListCopy.metaDescription = getMetaDescription(productList, itemType);
-   productListCopy.tagline = getTagline(productList, page, itemType);
-   return productListCopy;
+   productList.overrides = {
```
sterlinghirsh, line 126: I'm not sure I understand why we're back to modifying the original. I think this will introduce rendering problems by modifying the props. If we are intentionally modifying it, this should be a void function, but I think this is asking for trouble.

frontend/helpers/product-list-helpers.ts
```diff
@@ -179,16 +180,16 @@ function getTagline( // taglines don't exist if page > 1
...
-      !!productList.tagline && page === 1 ? productList.tagline : null;
+      !!productList.tagline && page === 1 ? productList.tagline : undefined;
    return itemType && productList.deviceTitle
       ? getItemOverrideAttribute(
            productList.itemOverrides,
            itemType,
            'tagline',
            productList.deviceTitle
-        )
+        ) ?? undefined
```
sterlinghirsh: I don't understand this change.

frontend/templates/product-list/sections/HeroSection.tsx
```diff
@@ -27,16 +27,18 @@ export function HeroSection({ productList }: HeroSectionProps) {
    return (
       <Flex direction="column">
          <HeroTitle>
-            {productList.title}
+            {productList?.overrides?.title}
             {page > 1 ? ` - Page ${page}` : ''}
          </HeroTitle>
-         {productList.tagline && (
+         {productList?.overrides?.tagline && (
```
sterlinghirsh, line 33: Do we ever use `productList.tagline` now? Is it only used in the fallback logic?

frontend/templates/product-list/sections/HeroWithBackgroundSection.tsx
```diff
@@ -63,17 +63,17 @@ export function HeroWithBackgroundSection({ productList }: HeroSectionProps) {
...
                <Text as="h2" fontWeight="medium">
-                  {productList.tagline}
+                  {productList?.overrides?.tagline}
                </Text>
             )}
-            {productList.description && (
+            {productList?.overrides?.description && (
                <HeroDescription mt="4">
-                  {productList.description}
+                  {productList?.overrides?.description}
```
sterlinghirsh: I think this was cleaner before.

sterlinghirsh:
CR :zap: dev_block :zap: I think the `.overrides` change is a valid approach, but I think in this case it was cleaner before [c731718](https://github.com/iFixit/react-commerce/pull/1649/commits/c7317180c9278f2a54c6bb1e06c3bd34145b7478). Seems like we only need access to the original title and maybe meta description.

frontend/templates/product-list/sections/HeroSection.tsx
```diff
@@ -25,33 +23,22 @@ export interface HeroSectionProps {
...
-      page === 1 &&
-      !itemType;
-
-   const title = getProductListTitle(productList, itemType);
 
    return (
       <Flex direction="column">
          <HeroTitle>
-            {title}
+            {productList?.overrides?.title}
```
masonmcelvain, line 30: I'm not sure why we went from `productList.` to `productList?.` here and in several other files. It seems like `productList` will always be defined in these contexts.

masonmcelvain (commented):
I know we've gone back and forth and I apologize if my suggestions complicated things, but I would like to note that I don't feel too strongly about `.overrides`. It's how I would have implemented this, but there are tradeoffs both ways and I think the initial approach was functional and valid.

frontend/templates/product-list/MetaTags.tsx
```diff
@@ -77,15 +77,15 @@ export function MetaTags({ productList }: MetaTagsProps) {
...
+                  content={productList?.overrides?.metaDescription}
                />
             </>
          )}
          <title>{metaTitle}</title>
          <meta property="og:title" content={metaTitle} />
          <meta
             name="og:description"
-            content={productList.metaDescription ?? undefined}
+            content={productList?.overrides?.metaDescription}
```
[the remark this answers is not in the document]
danielcliu-ifixit: metaDescription, but metaDescription falls  back to description so youre guaranteed to have one because description isn't nullable on the productList type

frontend/templates/product-list/sections/HeroSection.tsx
danielcliu-ifixit: yes

danielcliu-ifixit:
main merge QA :v: carryover

masonmcelvain (approved):
CR ✌🏻

PR #1651, opened 2023-05-04
Header 20th logo

aburke07:
:tada: Happy 🥳 20th 🎂 iFixit! 

## QA
Please check that the header now uses the appropriately sized logo from [here](https://github.com/iFixit/ifixit/issues/47658#issue-1692997866). The logo is a tad larger than the current one, but that is to account for the extra things in the design. The actual iFixit circle logo and 'IFIXIT' parts should still be the same size.

Please also check the hamburger side menu on small screens because we show the logo here.

Closes iFixit/ifixit#47658


ianrohde (dismissed):
CR ⚪ 

Out of curiosity, is there any particular reason we aren't using an `Image` component to load the SVG?

This is totally optional, but there's a right padding applied to the parent anchor that makes the focus state look odd when you click the header:

<details>
  <summary>Video</summary>

https://user-images.githubusercontent.com/1634505/236351959-939f5dc2-79c9-419b-a5a8-683fbaf0b253.mov

</details>

aburke07:
> Out of curiosity, is there any particular reason we aren't using an Image component to load the SVG?

We use an icon for the old header, so I don't see a huge reason to change it. It also seems like the preferred way to use [svgs in chakra ui, as it renders as an svg element](https://chakra-ui.com/docs/components/icon#using-the-icon-component).

> This is totally optional, but there's a right padding applied to the parent anchor that makes the focus state look odd when you click the header

Good catch! Fixed in [13d9c9c](https://github.com/iFixit/react-commerce/pull/1651/commits/13d9c9c70a342b6a1247942f221649d89093bcb1)

ianrohde (approved):
CR ⚪

deltuh-vee:
QA 🎬 
* header now uses the appropriately sized anniversary logo
* Hamburger side menu logo still looks good

aburke07:
Deploy_block :alien: On making sure we are ready to deploy the ifixit side change too.

ianrohde:
CR ⚪

PR #1336, opened 2023-02-06
Compute Strapi govinor handle to support longer branch names

dhmacs:
closes #1330 
closes #1270 

This branch is longer than 63 characters to test the url abbreviation works correctly

Changes on the Govinor side have already been deployed.

## QA

1. Open [Vercel preview](https://react-commerce-git-generate-govinor-url-handles-f-be9310-ifixit.vercel.app/Parts)
2. Verify that `/Parts` product list page renders with data from the [govinor Strapi instance](https://generate-govinor-url-handles-for-git-branches-with-max-50db9bd7.govinor.com/admin/auth/login)

PR #1657, opened 2023-05-05
Sentry: Send DSN to the client

### LINKED ISSUE #1654: Ensure client side errors in components show up in sentry
From Post-Mortem: #1653

A client side exception was causing blank pages for some users and yet the errors did not show up in sentry. We didn't find out about it until members of our dev team saw it and let us know.

Specfically, `user.data.teams` was undefined and thus calling `.length` threw an exception:
https://github.com/iFixit/react-commerce/blob/11d654df11d30da2f537704e1e322c3bc0e7217c/frontend/layouts/default/index.tsx#L405-L409

masonmcelvain:
This fixes a bug where no client side errors were being reported to Sentry because the DSN wasn't available to the browser as an [environment variable](https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#exposing-environment-variables-to-the-browser).

Closes #1654

### DIFF
--- a/frontend/.env.development
+++ b/frontend/.env.development
@@ -4,7 +4,7 @@ NEXT_PUBLIC_APP_ORIGIN=http://localhost:3000
 NEXT_PUBLIC_STRAPI_ORIGIN=http://localhost:1337
 NEXT_PUBLIC_SHOPIFY_STOREFRONT_VERSION=2022-10
 STRAPI_IMAGE_DOMAIN=ifixit-dev-strapi-uploads.s3.us-west-1.amazonaws.com
-SENTRY_DSN=https://95ab6917c0234f80b99bb0be8e720355@o186239.ingest.sentry.io/6475315
+NEXT_PUBLIC_SENTRY_DSN=https://95ab6917c0234f80b99bb0be8e720355@o186239.ingest.sentry.io/6475315
 NEXT_PUBLIC_MATOMO_URL=https://matomo.ubreakit.com
 NEXT_PUBLIC_GA_KEY=UA-30506-1
 # If we want to test Google Analytics in dev DEBUG should be true and URL should be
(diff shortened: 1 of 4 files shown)


jarstelfox (approved):
CR 🌵 https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#exposing-environment-variables-to-the-browser

jarstelfox:
FYI It's unclear to me if this ever worked. [An old blame](https://github.com/iFixit/react-commerce/commit/d2fd741bcc3b5eaecb1bd4d4ca7f74ee5bc9a691) seems to indicate it was never exposed to the front end

masonmcelvain:
> FYI It's unclear to me if this ever worked. [An old blame](https://github.com/iFixit/react-commerce/commit/d2fd741bcc3b5eaecb1bd4d4ca7f74ee5bc9a691) seems to indicate it was never exposed to the front end

Yea I don't think it ever did 😞 

Surprisingly, we have both `SENTRY_DSN` and `NEXT_PUBLIC_SENTRY_DSN` in the react-commerce-prod environment variables, but we weren't using it ever.

danielbeardsley:
Looks like we've only ever had 1 error get reported (though across several users... I wonder how they got the sentry DSN set in their env?)

https://ifixit.sentry.io/issues/?project=6469069&query=+next.runtime%3Aclient&referrer=issue-list&statsPeriod=14d

danielbeardsley:
That can't be true... remember when we had a bunch of problems with react 18 complaining about the hydration step resulting in a different schema? It was around new-years. This definitely used to work.

danielbeardsley:
Found it I think. This pull removed the explicit env var setting the next config:

https://github.com/iFixit/react-commerce/pull/1336/files#diff-e064de65d4610383e665a32e7e32ee1f520c466a6a63ecea8925ff8337f08215L57

PR #1661, opened 2023-05-05
Use PSK for dev server access

andyg0808:
This allows code-under-test to access dev server resources which are ordinarily behind a list of permitted IPs.

qa_req 0
This shouldn't have any effect on the system when there's no corresponding server-side code deployed; I expect CI should catch any problems.


packages/ifixit-api-client/index.tsx
```diff
@@ -60,13 +60,21 @@ export class IFixitAPIClient {
...
       const url = `${this.origin}/api/${this.version}/${endpoint}`;
       const response = await timeAsync(
          `ifixit-api.${init?.method?.toLowerCase() || 'get'}.${statName}`,
          () =>
             fetch(url, {
                credentials: 'include',
                ...init,
+               headers: {
+                  ...authHeader,
+                  ...init?.headers,
```
danielbeardsley, line 76: I don't think this is safe with a value that may be null or undefined:

image

Huh... but the previous `...init` should have been unsafe too... why wouldn't typescript complain then?

danielbeardsley: TIL It's safe with splatting into objects:
image

packages/ifixit-api-client/index.tsx
```diff
@@ -60,13 +60,21 @@ export class IFixitAPIClient {
    }
 
    async fetch(endpoint: string, statName: string, init?: RequestInit) {
+      const pskToken = process.env.DEV_API_AUTH_TOKEN;
+      const authHeader: { Authorization?: string } = pskToken
```
danielbeardsley, line 64: In this case, isn't this type implicit in the expression?

andyg0808: Here's the TypeScript error if that type's not included:
```
node_modules/@ifixit/analytics/node_modules/@ifixit/cart-sdk/node_modules/@ifixit/ifixit-api-client/index.tsx:72:16 - error TS2322: Type '{ Authorization: string; } | { length: number; toString(): string; toLocaleString(): string; pop(): [string, string] | undefined; push(...items: [string, string][]): number; concat(...items: ConcatArray<...>[]): [...][]; concat(...items: ([...] | ConcatArray<...>)[]): [...][]; ... 29 more ...; Authorization: string;...' is not assignable to type 'HeadersInit | undefined'.
  Type '{ Authorization?: undefined; }' is not assignable to type 'HeadersInit | undefined'.
    Type '{ Authorization?: undefined; }' is not assignable to type 'undefined'.

72                headers: {
                  ~~~~~~~
```

I'm guessing the problem is that the empty object doesn't get inferred to be `{Authorization?: string}` with `Authorization` unset. Instead, it gets picked up as `{}`, and then we access the unknown property `Authorization` on it, getting `undefined`. That's my best guess; it seems like something better inference could eliminate, but perhaps TypeScript's not able to handle that yet.

danielbeardsley (approved):
CR :+1:

PR #1663, opened 2023-05-08
Rename Upsell to Cross Sell

danielcliu-ifixit:
## Overview

We aren't upselling, technically. We are cross selling. So lets have the names be accurate.

The remaining references to "upsells" are because we don't want the code to break as we change the ifixit api response to be "crossSellProducts". Basically just so we can smoothly swap things out. 

## QA 

Does the upsell section in the cart still work fine? 

Connects https://github.com/iFixit/ifixit/issues/46262


deltuh-vee:
QA 🎬 
Cross sell section is still working

Issue #1627, opened 2023-04-27
Tests: Fix graphql handler for strapi/local queries 


batbattur:
## Summary 
Background https://ifixit.slack.com/archives/C01DBJK0H7U/p1682617412312469.

When we add a graphql handler using `createGraphQLHandler()` for a strapi endpoint, it cancels all other strapi graphql queries at `http://localhost:1337/graphql`.

For example, adding the following handler for `getProductList` graphql query, causes other strapi graphql queries to fail with the following error: `FetchError: request to http://localhost:1337/graphql failed, reason: socket hang up`.

https://github.com/iFixit/react-commerce/blob/f75297458f15757e84eb3b8e0e24b68d344ac28e/frontend/tests/playwright/product/product-list.spec.ts#L27-L38

<details> <summary> example log </summary> 

`pnpm playwright:run --project="Desktop Chrome" --headed product-list.spec.ts`

```
...
[MSW] Warning: captured a request without a matching request handler:

  • query getGlobalSettings (POST http://localhost:1337/graphql)

If you still wish to intercept this unhandled request, please create a request handler for it.
Read more: https://mswjs.io/docs/getting-started/mocks
[MSW] Warning: captured a request without a matching request handler:

  • query getStoreList (POST http://localhost:1337/graphql)

If you still wish to intercept this unhandled request, please create a request handler for it.
Read more: https://mswjs.io/docs/getting-started/mocks
[MSW] Warning: captured a request without a matching request handler:

  • query findStore (POST http://localhost:1337/graphql)

If you still wish to intercept this unhandled request, please create a request handler for it.
Read more: https://mswjs.io/docs/getting-started/mocks
[MSW] Warning: captured a request without a matching request handler:

  • GET https://www.cominor.com/api/2.0/cart/part_collections/devices/Microsoft_Surface_Laptop_2

If you still wish to intercept this unhandled request, please create a request handler for it.
Read more: https://mswjs.io/docs/getting-started/mocks
strapi.getProductList => 28.71ms
FetchError: request to http://localhost:1337/graphql failed, reason: socket hang up
    at NodeClientRequest.<anonymous> (/Users/batwork/react-commerce/node_modules/.pnpm/next@12.2.3_t7ss3ubh4wigfvyfclbvqff3gm/node_modules/next/dist/compiled/node-fetch/index.js:1:65763)
    at NodeClientRequest.emit (node:events:538:35)
    at NodeClientRequest.emit (node:domain:475:12)
[...]

PR #1666, opened 2023-05-09
Fix MSW Handler Race Condition

ardelato:
## Description
There was an issue with the msw request handlers for `strapi` requests. https://github.com/iFixit/react-commerce/issues/1627

After digging into the issue, it seems to be an issue with `node-fetch` or `strapi` itself. The trigger for it was 3 requests being made directly to `strapi`: `strapi.getGlobalSettings`, `strapi.getStoreList`, and `strapi.findStore`. These requests were not matched by any of the msw request handlers, so they were passed through to the real `strapi` server.

Then a `strapi.getProductList` request was made, which was matched by the msw request handler, so it was intercepted and mocked. However, this request was immediately returned while the previous 3 requests were still pending. This caused the `FetchError: socket hang up` error to be thrown by `node-fetch` as the queue of pending requests was closed since the lastest request was already returned. In order to fix this, we need to delay the mocked responses to ensure that the previous requests have been returned before the next request is mocked. This also mimics the behavior of real requests, which are not returned immediately.

### CR

Worked with @batbattur on this.

In order to figure out the issue behind #1627, we hooked onto the `life-cycle` events from the MSW server[^1] and modified the source code, within the `node_modules` directory, of the MSW `SetupServerAPI` class to log out steps.

From there, we noticed that the `interceptor.on('response'...)` event [^2] was called when there were no handlers defined for `strapi`, but once the `getProductList` handler was defined, the other `strapi` requests errored and never triggered this event.

### QA:

Unskip the tests in https://github.com/iFixit/react-commerce/blob/item-type-overrides--frontend/frontend/tests/playwright/product/product-list.spec.ts#L5.

- [ ] Confirm a `socket hangup` error is no longer being thrown like in the original branch.
	> ⚠️ The tests will still fail as the tests themselves need to be fixed.

**cc:** @danielcliu-ifixit 

Closes: #1627 

[^1]: https://mswjs.io/docs/extensions/life-cycle-events
[^2]: https://github.com/mswjs/msw/blob/c32241627da09af8b34c4068cc17dd1941ff3b0d/src/node/SetupServerApi.ts#L105

### DIFF
--- a/frontend/tests/playwright/msw/request-handler.ts
+++ b/frontend/tests/playwright/msw/request-handler.ts
@@ -167,10 +167,20 @@ export const createGraphQLHandler = ({
       '*',
       customResolver ??
          ((req, res, ctx) => {
+            const delayAmount = 2000; // 2 seconds
             if (passthrough) return req.passthrough();
-            if (once) return res.once(ctx.status(status), ctx.data(body));
+            if (once)
+               return res.once(
+                  ctx.delay(delayAmount),
+                  ctx.status(status),
+                  ctx.data(body)
+               );
 
-            return res(ctx.status(status), ctx.data(body));
+            return res(
+               ctx.delay(delayAmount),
+               ctx.status(status),
+               ctx.data(body)
+            );
          })
    );
 };
@@ -202,7 +212,8 @@ export const createRestHandler = ({
          ((req, res, ctx) => {
             if (passthrough) return req.passthrough();
 
-            const transformers = [ctx.status(status)];
+            const delayAmount = 2000; // 2 seconds
+            const transformers = [ctx.delay(delayAmount), ctx.status(status)];
 
             if (headers) {
                transformers.push(ctx.set(headers));


deltuh-vee:
Was the [product-list.spec.ts](https://github.com/iFixit/react-commerce/blob/item-type-overrides--frontend/frontend/tests/playwright/product/product-list.spec.ts#L5) broken up?

ardelato:
> Was the [product-list.spec.ts](https://github.com/iFixit/react-commerce/blob/item-type-overrides--frontend/frontend/tests/playwright/product/product-list.spec.ts?rgh-link-date=2023-05-09T23%3A58%3A00Z#L5) broken up? I don't see it anymore.

I see it in the same location as in the base branch?
https://github.com/iFixit/react-commerce/blob/fix-msw-handler-race-condition/frontend/tests/playwright/product/product-list.spec.ts#L5

danielbeardsley:
This makes me sad. I'm glad you dove into it, but I'm sad that our current way of dealing with it is "add an arbitrary 2 second delay". I don't really understand why the timing of requests and responses will result in something breaking, this sounds like a bug in msw perhaps?

I'm also not convinced that we should add this delay on all tests runs when it seems this issue only affects iterating on these tests in development (where the next.js server reloads). Do I have that right?

mlahargou:
I agree with @danielbeardsley that adding a 2 second delay to all these tests is sad. Is there any way we can somehow just block until we have 0 in-flight requests before resolving the mocked request?

ardelato:
dev_block ☁️ I'm going to dig a bit deeper to see if we can find another solution besides the arbitrary delay

ardelato:
> I'm also not convinced that we should add this delay on all tests runs when it seems this issue only affects iterating on these tests in development (where the next.js server reloads). Do I have that right?

I should note that this issue also affects CI, we just aren't seeing it because we are currently skipping the test that exposed the issue: https://github.com/iFixit/react-commerce/blob/item-type-overrides--frontend/frontend/tests/playwright/product/product-list.spec.ts#L5

In addition, this delay will only affect tests that use an MSW handler, therefore there won't be a significant increase in the runtime of the tests.

----

un_dev_block ⚡ **TL;DR** This is an issue with the underlying https://github.com/mswjs/interceptors package.

`mswjs/msw` uses `mswjs/interceptors` to intercept and handle requests. This package has a class that extends a Node `ClientRequest` in which it will call `agent.destroy` when a mocked response finishes.

https://github.com/mswjs/interceptors/blob/aff787ee28e58737662b31b93e2b7f24faad7beb/src/interceptors/ClientRequest/NodeClientRequest.ts#L353-L362

However, this agent also has references to the sockets from the non-matched requests to `localhost:3000/graphql`. As a result, those sockets are forcibly closed before they have finished and cause the `socket` hangup.

image


image

I confirmed that by not running `agent.destroy()` the sockets would not hang and the page would render successfully.

That said, the only way to resolve this issue on our side is to delay the mocked response or mock every `strapi` request. The former is easier to implement while the latter would require appropriately snapshotting every `strapi` request to mock accordingly for each page.

mlahargou (approved):
CR 📱

danielbeardsley:
Good job diving into this!

>  the only way to resolve this issue ...

I think it's appropriate to open an issue on their repo. Seems like a bug that mocking *some* requests can sometimes cause unexpected failures.

deltuh-vee:
QA 🎬 
No longer getting `socket hangup` errors.

danielcliu-ifixit:
Thanks for diving into this @ardelato @batbattur

PR #1670, opened 2023-05-10
Bump @core-ds/primitives to `2.5.2`

ianrohde:
This version removes `Liberation Mono` from the mono font stacks: https://github.com/iFixit/core-primitives/pull/40

TL;DR it improves the vertical alignment of mono fonts on Linux.

qa_req 0 - tested in https://github.com/iFixit/core-primitives/pull/40 and represents less than 1% of users.


davidrans (approved):
CR :coffee:

Issue #1643, opened 2023-05-04
Investigate raise in CLS


federicobadini:
A real macro reason for this raise in CLS has not emerged yet.
The fact that this happens only on Algolia related pages may suggest the hypothesis that this CLS originate from delays between page interaction and dom updates.

If this proves to be the final answer, one of the mitigation strategy we have is reducing all sorts of delays introduced between the user interaction and the execution of an action.
For example, the search field in product collections introduces a 300ms debounce time, and that could be problematic.

federicobadini:
CLS is falling fast after #1770 
Here are the reported pagespeed insights values for mobile across the last 10 days:
0.71 —> 0.65 —> 0.61 —> 0.59 —> 0.56 —> 0.52 —> 0.48 —> 0.43 —> 0.37 —> 0.34

Desktop is improving too.

Search console already turned green for all /Parts and /Tools pages:

<img width="1709" alt="Screenshot 2023-07-04 at 10 22 52" src="https://github.com/iFixit/react-commerce/assets/7805759/4e1cb4fe-3883-42a1-b2bc-8eea47c6af04">

We can close this since it seems this is going to be solved soon.

PR #1674, opened 2023-05-11
Mitigate cls on page

federicobadini:
connects #1643 

First attempt to mitigate CLS deriving from on page user interactions.

We suspect that the root cause of this might be some slow network requests.
Google leaves only 500ms from a user interaction and the connected layout shifts. 
So if a response from the API comes back slow, any shift deriving from that response would count towards CLS.

To mitigate the problem we are removing induced delays like the debounceTime of the search field on product list pages.

Moreover, we noticed that `productListChildren` were wrongly removed if a specific search query did not match any result on itemType collections, we fixed that condition.

### QA

1. Visit Vercel Preview
2. The product list page ui should behave like before with a reduced debounce time for the search input
3. Product list children should not disappear anymore even if a particular search query produces no results on an itemType collection


masonmcelvain (approved):
CR ✌🏻

PR #1681, opened 2023-05-15
Lint: fix outstanding linting issues

danielbeardsley:
The linter is commenting about these on every pull. Let's fix them so it's quiet.

Also, prevent inclusion of more undefined vars by making them errors.

qa_req 0


ianrohde (approved):
CR ⚪

PR #1684, opened 2023-05-15
Vulcan: Fix styling on bulleted lists

ianrohde:
## Summary
Clean up the styling on bulleted lists. Match Figma design indentation and get order lists following same indentation.

> The bulleted lists on the NextJS page aren't currently indented as they are on our wiki pages. Compare against https://www.ifixit.com/Wiki/Dryer_Not_Spinning and https://www.ifixit.com/Wiki/Electric_Dryer_Not_Heating (notice the bullets floating to the right of the image at the bottom of the page).

## Modifications
1. [sync ol and ul margins to align and match design](https://github.com/iFixit/react-commerce/commit/d5c1d031d5b2b50325aad5e2247d854ba730d642)
   - ol, moved styling up to parent to also apply to pseudo element numbers
   - conditionally load `intro.heading` only if it exists

## CR/QA Notes
http://localhost:3000/Vulcan/Dryer_Not_Spinning

<details>
  <summary>Alignment Screenshot</summary>

Screenshot 2023-05-15 at 4 37 08 PM

</details>

Closes: https://github.com/iFixit/ifixit/issues/47910

### DIFF
--- a/frontend/templates/troubleshooting/index.tsx
+++ b/frontend/templates/troubleshooting/index.tsx
@@ -114,16 +114,18 @@ const Wiki: NextPageWithLayout<{
 
 function TableOfContents({ solutions }: { solutions: Section[] }) {
    return (
-      <OrderedList marginBottom="24px" spacing="8px">
+      <OrderedList
+         marginBottom="24px"
+         spacing="8px"
+         sx={{ marginInlineStart: '2em' }}
+      >
          {solutions.map((solution, index) => (
-            <ListItem key={solution.heading}>
-               <Link
-                  href={`#solution-${index + 1}`}
-                  color="brand.500"
-                  fontWeight="bold"
-               >
-                  {solution.heading}
-               </Link>
+            <ListItem
+               color="brand.500"
+               fontWeight="bold"
+               key={solution.heading}
+            >
+               <Link href={`#solution-${index + 1}`}>{solution.heading}</Link>
             </ListItem>
          ))}
       </OrderedList>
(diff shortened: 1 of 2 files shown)


aburke07 (approved):
CR :alien:

ianrohde:
@deltuh-vee the Heading alignment is correct and has not been changed. I indented the list on this pull to account for the list items that are double-digits. From https://github.com/iFixit/ifixit/issues/47910:

> List of problem jump link numbers extend to the LEFT of the cards when they are greater than one digit:
> image

Looking into option for getting the ordered list numbers linked.

ianrohde:
> Should the numbers in a numbered list also be part of the clickable link? If they're not black, I would expect them to be clickable.

@mmarcon91 what do you think about this situation? We're at a cross-roads with HTML semantics and design. An anchor should not be a direct child of a list, but the Bat is correct that if we make the list item numbers the same color, it will be assumed they are links.

<details>
  <summary>Screenshot</summary>

Screenshot 2023-05-16 at 11 58 52 AM

</details>

mmarcon91:
> @mmarcon91 what do you think about this situation? We're at a cross-roads with HTML semantics and design. An anchor should not be a direct child of a list, but Bat is correct that if we make the list item numbers the same color, it will be assumed they are links.

Yes, the numbers should be part of the link. We could use css counter variables to solve.

ianrohde:
In https://github.com/iFixit/react-commerce/pull/1684/commits/4839fa0c811240d34828bb93256b5610de215819 I used the index we already used to write out the number, but had to break from the default left-alignment due to the single digit values. I set a `minWidth="3ch"` on the count container to account for double-digits and get them inline with the other unordered list on the page above.

ianrohde:
un_dev_block 🟢 
un_deploy_block 🟢

aburke07 (approved):
CR :alien:

deltuh-vee:
QA 🎬 
Numbers are now part of the clickable link.

PR #1689, opened 2023-05-17
Algolia: count facet results after doing distinct

danielbeardsley:
We'll now be using the "attributeForDistinct" feature. This means that for some products there will be more than one algolia document that represents it. We'll be turning on this attributeForDistinct feature which will ensure we only get one result per product in the result set. However, the result counts in the facets will show the number of documents instead of the number of products until we set this option.

See: https://www.algolia.com/doc/api-reference/api-parameters/facetingAfterDistinct/

Connect https://github.com/iFixit/ifixit/issues/47734

### DIFF
--- a/frontend/models/product-list/server.ts
+++ b/frontend/models/product-list/server.ts
@@ -495,6 +495,7 @@ async function findDevicesWithProducts(devices: string[]) {
          facets: ['device'],
          filters: `public = 1${deviceFilterSuffix}`,
          maxValuesPerFacet: 1000,
+         facetingAfterDistinct: true,
          hitsPerPage: 0,
       });
       return facets?.device ? facets?.device : {};
--- a/frontend/templates/product-list/ProductListView.tsx
+++ b/frontend/templates/product-list/ProductListView.tsx
@@ -44,7 +44,11 @@ export function ProductListView({ productList }: ProductListViewProps) {
          <SecondaryNavigation productList={productListWithOverrides} />
          <Wrapper py={{ base: 4, md: 6 }}>
             <VStack align="stretch" spacing={{ base: 4, md: 6 }}>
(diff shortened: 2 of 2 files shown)


danielbeardsley:
### QA
Ensure this url on both the dev and prod preview deployments shows the correct facet result counts: `/Parts/Appliance/Ice_Makers`

erinemay:
QA 👍 The dev preview matches the prod preview in that counts in the filter facet options match the actual available filter results. 
Mixing filters, like `/Parts/iPhone/Cables?filter%5Bfacet_tags.Device%20Type%5D=iPad` - all have appropriate counts.

`/Parts/Appliance/Ice_Makers` doesn't include duplicates like on the main preview. Other similar instances of duplicates showing incorrectly in the facet counts aren't happening on this branch.

Non-duplicate facets remain the same across main, dev preview, and prod preview - facet counts match the actual available filter results.

davidrans (approved):
CR :coffee:

frontend/models/product-list/server.ts
```diff
@@ -495,6 +495,7 @@ async function findDevicesWithProducts(devices: string[]) {
          facets: ['device'],
          filters: `public = 1${deviceFilterSuffix}`,
          maxValuesPerFacet: 1000,
+         facetingAfterDistinct: true,
```
masonmcelvain, line 498: Hm I wonder if we actually need this here, since this query only collects unique facets and not hits. I doubt it will make a difference functionally, though I'm curious if we'll see any performance change.

danielbeardsley: > though I'm curious if we'll see any performance change.

True. I don't know. But I can't say with any certainty which way would be faster.

PR #1692, opened 2023-05-18
Use DeviceTitle As Device Replacement in Item Type Overrides 

danielcliu-ifixit:
## Overview

Lets use the deviceTitle to replace `[DEVICE]` in all cases. If deviceTitle doesn't exist for that productlist, we will no longer skip the override we will instead replace `[DEVICE]` with `title` instead.

## QA
Pretty much just check that it does what the overview says it will.

Connects: https://github.com/iFixit/react-commerce/issues/589


masonmcelvain (approved):
CR ✌🏻

PR #1694, opened 2023-05-19
Add meta viewport tag to default layout

ianrohde:
## Summary
We're missing this meta tag on commerce. Let's add it so mobile pages behave consistently.

https://developer.mozilla.org/en-US/docs/Web/HTML/Viewport_meta_tag

## Modifications
1. [add meta viewport tag to default layout](https://github.com/iFixit/react-commerce/commit/f221666f718af905d5368aaeb1d479b4b847ad0d)

## CR/QA Notes
This is intentionally added at the default layout level. We want it applied on all pages.

There should be no visual difference.

### DIFF
--- a/frontend/layouts/default/index.tsx
+++ b/frontend/layouts/default/index.tsx
@@ -165,6 +165,11 @@ const DefaultLayoutComponent = function ({
                   href="https://assets.cdn.ifixit.com/static/icons/ifixit/safari-pinned-tab.svg"
                   color="#5bbad5"
                />
+
+               <meta
+                  name="viewport"
+                  content="width=device-width, initial-scale=1"
+               />
             </Head>
             <Flex direction="column" minH="100vh">
                <Header>


jordycosta:
QA 🟢 

The meta `viewport` tag is present on all pages _(Store, Parts, Tools, Products)_ 👍 

firefox_PW0YDQEJhN

masonmcelvain:
This seems like a good thing to add, though I notice that it's already present on our Next.js pages. I wonder what's adding it, if we're not already doing it in our code 🤔 

image
_https://www.ifixit.com/Parts_
It looks like the default for initial-scale is 1, based on the MDN docs linked in the PR comment.

ianrohde:
I should have checked the generated code, not the src! https://nextjs.org/docs/app/api-reference/functions/generate-metadata#viewport

masonmcelvain:
Ah ok. Thanks for finding the link!

PR #1705, opened 2023-05-22
Troubleshooting: 404 on missing wiki

jyee27:
404 instead of crashing the page when visiting a troubleshooting wiki that doesn't exist, and handle bad product/guide links by not making a card for them.

Example 404 wiki: `/Vulcan/asdfasdfasdf`
Example wiki with bad product/guide links: `/Vulcan/troubleshooting_products_test`

Closes: https://github.com/iFixit/ifixit/issues/48034


danielbeardsley:
This is the first time I've seen this code. Can you provide some example urls that should succeed and some that should 404?

frontend/templates/troubleshooting/server.tsx
```diff
@@ -53,25 +54,33 @@ export const getServerSideProps: GetServerSideProps<
...
+   let troubleshootingData: TroubleshootingApiData;
+   try {
+      troubleshootingData = await client.get<TroubleshootingApiData>(
+         `Troubleshooting/${wikiname}`,
+         'troubleshooting',
+         {
+            credentials: 'include',
+         }
+      );
+   } catch (e) {
```
danielbeardsley, line 79: I think it's a bad idea here to swallow all exceptions. Do 404s throw exceptions? Can we narrow this handling to *just* deal with 404s?

frontend/templates/troubleshooting/server.tsx
```diff
@@ -53,25 +54,33 @@ export const getServerSideProps: GetServerSideProps<
...
-         credentials: 'include',
-      }
-   );
+   let troubleshootingData: TroubleshootingApiData;
+   try {
+      troubleshootingData = await client.get<TroubleshootingApiData>(
+         `Troubleshooting/${wikiname}`,
+         'troubleshooting',
+         {
+            credentials: 'include',
```
danielbeardsley, line 76: This... shouldn't matter I don't think. This is server-side code, there's no cookies to send. I know your pull didn't add this, but I hadn't seen this code before. I'll likely open another issue to handle this.

danielbeardsley:
dev_block :+1: 

I'm glad you're doing this! We shouldn't be 500ing on bad user input, so this is a good direction. That said... I don't think we should be swallowing all the errors for all the API requests made here.

frontend/templates/troubleshooting/server.tsx
jyee27: The exception comes from here: https://github.com/iFixit/react-commerce/blob/0ff5ad629f7070ce81339a87e920a597d578c77d/packages/ifixit-api-client/index.tsx#L80-L82
We could check against the response's status text to serve a different error page. Maybe just re-throw the exception which should get us the error page: https://github.com/iFixit/react-commerce/blob/c55d58bf0b0f150baff1486b85738a773c6cc6c0/frontend/pages/_error.tsx

jyee27: Re-throwing the exception should also show the error page for the exceptions that `fetch()` can throw. https://developer.mozilla.org/en-US/docs/Web/API/fetch#exceptions

jyee27: [`f6e86bc` (#1705)](https://github.com/iFixit/react-commerce/pull/1705/commits/f6e86bc1b690baa9c39121f38c6cbf57ad030429)

frontend/templates/troubleshooting/server.tsx
```diff
@@ -53,25 +54,37 @@ export const getServerSideProps: GetServerSideProps<
...
+         }
+      );
+   } catch (e) {
+      // If fetch() doesn't error, e.message is the response's statusText
+      if (typeof e === 'object' && (e as Error)?.message === 'Not Found') {
```
danielbeardsley, line 81: Let's turn this into a function like:

`function rethrowUnless404(e) { if (404) { return } else { throw e} }`

Then we can do `rethrowUnless404(e)` and use that in the several places here where we are using catch-alls.

danielbeardsley:
Pattern seems reasonable, let's do the same for the other catch-alls.

CR :+1:

frontend/templates/troubleshooting/server.tsx
jyee27: [44045fc](https://github.com/iFixit/react-commerce/pull/1705/commits/44045fc915e90cd2a1a70e0a05e089b5df396e44)

jyee27:
un_dev_block :coffee:

k0rvusk0r4x:
- We are now getting a proper 404 instead of a crash.
- Vulcan pages are working.

QA 😺

frontend/templates/troubleshooting/server.tsx
andyg0808: Issue created: https://github.com/iFixit/ifixit/issues/48080

andyg0808 (commented):
CR :+1: 
I really dislike having to use `let` as seen on line 85, but I'm not seeing a better way to implement this.

andyg0808 (approved):
CR :+1:

andyg0808 (approved):
CR :+1:

k0rvusk0r4x:
Still working as [here](https://github.com/iFixit/react-commerce/pull/1705#issuecomment-1560072876).

QA 😺

Issue #1532, opened 2023-04-04
Add banners sections to product page


dhmacs:
Implement banners section, which can display one or multiple banners

### Single banner

<img width="901" alt="Screenshot 2023-04-04 at 22 09 44" src="https://user-images.githubusercontent.com/4640135/229908874-dddced31-d434-4cf9-b423-e50464a34045.png">

### Multiple banners

<img width="905" alt="Screenshot 2023-04-04 at 22 10 28" src="https://user-images.githubusercontent.com/4640135/229909020-cbdc5301-21d1-415c-98c9-d3192b4f3b8f.png">

PR #1706, opened 2023-05-22
Page: Add banner link fallback spacing

### LINKED ISSUE #1703: Vertically center page header content
## Problem

When the page header button is omitted, the rest of the text is not vertically centered.

image

image

[Slack thread](https://ifixit.slack.com/archives/C02KMFEHX62/p1684794713835469)

masonmcelvain:
Keeps the banner text vertically aligned if the link is missing.

Closes #1703

## QA

- Visit `/products/repair-business-toolkit`
- On https://page-header-alignment.govinor.com/admin, get rid of the banner call to action for the repair business toolkit page.
- Verify that the height of the image remains unchanged, and that the text is vertically aligned.
- With the call to action, the UI should match `main`.

### DIFF
--- a/frontend/components/sections/BannersSection/SingleBanner.tsx
+++ b/frontend/components/sections/BannersSection/SingleBanner.tsx
@@ -30,38 +30,42 @@ export function SingleBanner({ id, banner }: SingleBannerProps) {
                />
             </Box>
          )}
-         <Wrapper>
-            <Box textAlign="center" position="relative">
-               {banner.label && (
-                  <Text color="white" mb="3" fontSize="sm">
-                     {banner.label}
-                  </Text>
-               )}
-               {banner.title && (
-                  <SectionHeading color="white" mb="2.5">
-                     {banner.title}
-                  </SectionHeading>
-               )}
-               {banner.description && (
-                  <SectionDescription
-                     richText={banner.description}
-                     color="white"
-                     maxW="750px"
-                     mx="auto"
-                  />
-               )}
-               {banner.callToAction && (
-                  <Box position="absolute" w="full" bottom="-20">
-                     <SmartLink
-                        as={LinkButton}
-                        href={banner.callToAction.url}
-                        colorScheme="brand"
-                     >
-                        {banner.callToAction.title}
-                     </SmartLink>
-                  </Box>
-               )}
-            </Box>
+         <Wrapper position="relative" textAlign="center">
+            {banner.label && (
+               <Text color="white" mb="3" fontSize="sm">
+                  {banner.label}
+               </Text>
+            )}
+            {banner.title && (
+               <SectionHeading color="white" mb="2.5">
+                  {banner.title}
+               </SectionHeading>
+            )}
+            {banner.description && (
+               <SectionDescription
+                  richText={banner.description}
+                  color="white"
+                  maxW="750px"
+                  mx="auto"
+               />
+            )}
+            {banner.callToAction && (
+               <Box
+                  position="absolute"
+                  w="full"
+                  bottom="-20"
+                  left="50%"
+                  transform="translateX(-50%)"
+               >
+                  <SmartLink
+                     as={LinkButton}
+                     href={banner.callToAction.url}
+                     colorScheme="brand"
+                  >
+                     {banner.callToAction.title}
+                  </SmartLink>
+               </Box>
+            )}
          </Wrapper>
       </Center>
    );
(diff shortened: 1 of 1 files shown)


frontend/components/sections/BannersSection/SingleBanner.tsx
```diff
@@ -69,6 +69,8 @@ export function SingleBanner({ id, banner }: SingleBannerProps) {
                   >
                      {banner.callToAction.title}
                   </SmartLink>
+               ) : (
+                  <Box h="10" mt="10" />
```
dhmacs, line 73: This looks a bit a hacky solution... I think it should be possible to convert the container to flex and vertically center that way? 🤔

masonmcelvain: I agree it's a bit hacky/unmaintainable. It's tricky due to the use of padding on the outermost container to determine the height of the image.

I did consider making the container a flex like you suggested. However, when the CTA is omitted the height of the image is reduced. We could conditionalize the padding on the presence of the CTA, but that is just as unmaintainable as this Box solution.

I will spend a little more time with this today to see if there is a cleaner solution. dev_block ✌🏻

masonmcelvain:
dev_block ✌🏻

masonmcelvain:
un_dev_block ✌🏻 It's not a flexbox, but this solution is more maintainable than https://github.com/iFixit/react-commerce/commit/c9c6c0bf0419ee8c95b04b81fe9f198aa2e96fbe

frontend/components/sections/BannersSection/SingleBanner.tsx
```diff
@@ -13,7 +13,7 @@ interface SingleBannerProps {
 
 export function SingleBanner({ id, banner }: SingleBannerProps) {
    return (
-      <Box as="section" id={id} position="relative" w="full" pt="36" pb="16">
+      <Box as="section" id={id} position="relative" w="full" py="144px">
```
ianrohde, line 16:
```suggestion
      <Box as="section" id={id} position="relative" w="full" minHeight="310px" display="flex" placeItems="center">
```
We can do this to make it dynamic. The minHeight is 1px less than the old padding height total (design OCD doesn't allow single-digit values 😅), and `display="flex" placeItems="center"` handle the centering.

We also have an opportunity to clean the code up elsewhere by replacing:
```
top="0"
left="0"
bottom="0"
right="0"
```
with the awesome new shorthand `inset`:
```
inset="0"
```

masonmcelvain:
dev_block ✌🏻 on @ianrohde's suggestion. Thanks!

masonmcelvain:
un_dev_block ✌🏻 Thanks @ianrohde!

ianrohde (approved):
CR 🤘

jordycosta:
QA 🟢 

The header text is vertically aligned, and no longer shifts when omitting the call to action link 👍 

[Strapi Test Page](https://page-header-alignment.govinor.com/admin/content-manager/collectionType/api::banner.banner/5?plugins[i18n][locale]=en)

firefox_NGbHTg8eW9

ianrohde:
dev_block 🔴 on making sure this is what we want. Should the entire content me centered when the button is present? It doesn't look like it is.

frontend/components/sections/BannersSection/SingleBanner.tsx
```diff
@@ -40,38 +30,43 @@ export function SingleBanner({ id, banner }: SingleBannerProps) {
...
+               />
+            )}
+            {banner.callToAction && (
+               <Box
+                  position="absolute"
+                  w="full"
+                  bottom="-20"
+                  left="50%"
+                  transform="translateX(-50%)"
+               >
```
ianrohde, line 59: If we want all content including the button to vertically center, we should remove this Box component. It has `position: absolute` applied, so it's removed from the page stacking order.

masonmcelvain:
I couldn't find a figma mockup for this feature, but this is the issue about first implementing the banner section: https://github.com/iFixit/react-commerce/issues/1532

We might ask @mmarcon91 to be sure.

mmarcon91:
I had put a comment in this [slack](https://inkofpixel.slack.com/archives/C02KMFEHX62/p1684836147764109?thread_ts=1684794713.835469&cid=C02KMFEHX62) discussion.

> I think the best solution is to give the option in strapi to choose the content alignment (top/middle/bottom).
It works when there is a button because the purpose is to get the user to scroll. Without CTA it looks better centered

frontend/components/sections/BannersSection/SingleBanner.tsx
masonmcelvain: The call-to-action button was excluded from the "centered" content before, and it continues to be excluded in this pull. I did my best to keep the spacing the same.

### Main



### Feature

masonmcelvain:
un_dev_block ✌🏻 

I am going to merge as-is to meet the [initial request](https://ifixit.slack.com/archives/C02KMFEHX62/p1684794713835469) in a timely manner.

I opened an [issue](https://github.com/iFixit/react-commerce/issues/1725) to add an option to Strapi to choose the alignment.

PR #1709, opened 2023-05-24
Strapi SDK: Manual input validation

### LINKED ISSUE #1708: Typescript-validation-schema error in generating types
From the main branch, by just running the project, I see that all the references to non scalar types are removed from the validation.ts file. @dhmacs experienced the same.

Screenshot 2023-05-23 at 16 53 27

masonmcelvain:
Closes #1708

Stops auto-generating Strapi input validation schemas in favor of maintaining them by hand.

### Pros
- Freedom to be more or less restrictive in our validation logic
- Custom error messages

### Cons
- Potentially more work keeping these schemas aligned with the Strapi API if it ever changes (though maintaining a schema generation system is also work)
  - Changes to the Strapi API are likely rare, and if anything they will add more operators, since changing/removing operators would be a breaking API change.

qa_req 0

CC @federicobadini


danielbeardsley:
Does this mean there will be no build-time failures if we we typo some name in the schema (i.e. a query name or an operator name)?

frontend/lib/strapi-sdk/zod.ts
```diff
@@ -0,0 +1,190 @@
...
+      lte: z.boolean().nullish(),
+      ne: z.boolean().nullish(),
+      not: z.lazy(() => BooleanFilterInputSchema().nullish()),
+      notContains: z.boolean().nullish(),
+      notContainsi: z.boolean().nullish(),
+      notIn: z.array(z.boolean().nullable()).nullish(),
+      notNull: z.boolean().nullish(),
+      null: z.boolean().nullish(),
+      or: z.array(z.boolean().nullable()).nullish(),
+      startsWith: z.boolean().nullish(),
```
danielbeardsley, line 39: Seems like this section is repeated 3 or 4 times and could easily be extracted to a variable / constant.

masonmcelvain:
> Does this mean there will be no build-time failures if we we typo some name in the schema (i.e. a query name or an operator name)?

Yes, no build time failures. For the current use case, it would mean that product list cache gets for product lists with offending `filters` would always be misses.

danielbeardsley:
>  product lists with offending filters would always be misses.

If the passed in types don't match the zod schema then typescript would throw a fit I think because Zod produces valid TS types that we'd use for the input arguments. I'm talking about if the zod schema for the inputs doesn't match the actual inputs accepted by GraphQL. There's nothing to keep them in sync other than testing and CR, right?

masonmcelvain:
> product lists with offending filters would always be misses.

Derp, this isn't right. I was thinking of the values schema, but the `ProductListFiltersInputSchema` is part of the variables schema.

> There's nothing to keep them in sync other than testing and CR, right?

masonmcelvain:
@danielbeardsley see an alternative solution here: https://github.com/iFixit/react-commerce/pull/1716

federicobadini:
This is still being discussed or it was superseded by #1716 ?

masonmcelvain:
@danielbeardsley and I opted for the solution in #1716, since we perceived this solution to be error-prone and we didn't see a redeeming level of value for schema and error message customization in our current use case.

However, it's definitely not closed to discussion and change, so if you have feedback we would love to hear it.

PR #1711, opened 2023-05-24
Troubleshooting: Drop unneeded `credentials: include`

andyg0808:
Does what it says on the tin.

Closes https://github.com/iFixit/ifixit/issues/48080


andyg0808 (commented):
QA :+1:
The page loads. Since the affected `fetch()` is the primary one for the entire troubleshooting page, it appears to still be working.

CI will also validate this.

aburke07 (approved):
CR :alien:

PR #1712, opened 2023-05-24
Troubleshooting: Update test to require Vulcan canonical link

andyg0808:
This ensures the canonical link points at the new page and not at the original WIKI the data is sourced from. This test already passes; I just wanted to add something to confirm that it works as-expected.

qa_req 0
This only changes a test.

Closes https://github.com/iFixit/ifixit/issues/48082


aburke07 (approved):
CR 🧆

PR #1716, opened 2023-05-25
Strapi SDK: Generate graphql schema

### LINKED ISSUE #1708: Typescript-validation-schema error in generating types

masonmcelvain:
Implements a workaround for the zod input schema becoming `definedNonNullAnySchema` due to the graphql endpoint being `http`, not `https`. Described in more detail here: https://github.com/Code-Hex/graphql-codegen-typescript-validation-schema/issues/131#issuecomment-1344439278

This is an alternative solution to https://github.com/iFixit/react-commerce/pull/1709

Closes #1708 

## QA

Run `pnpm dev`. Make sure that no new diff is generated by codegen, and that all the codegen logs indicate that it's successfully running.

CC @federicobadini @danielbeardsley


danielbeardsley:
What would you think about *not* committing the 2,000 line file and letting it get downloaded each time code gen runs?

masonmcelvain:
Seems sane. We could `.gitignore` it too.

masonmcelvain:
Oof, I just noticed that the `schema.graphql` is not generated when running `pnpm dev`. I think it fails because we've now introduced a dependency chain- the validation file depends on the schema file being generated first. It works if we comment out the validation file config, and then uncomment it.

I'm not sure codegen supports this with the yaml file, but they do have a [JS api](https://the-guild.dev/graphql/codegen/docs/advanced/programmatic-usage) we could use to sequence our generated files.

masonmcelvain:
dev_block ✌🏻 on that ^

masonmcelvain:
un_dev_block ✌🏻

deltuh-vee:
QA 🎬 
`pnpm dev` is no longer making edits to validation.ts

frontend/codegen.ts
```diff
@@ -0,0 +1,67 @@
...
+               HTML: 'string',
+               URL: 'string',
+               Decimal: 'string',
+            },
+         },
+      },
+   },
+};
+
+export default generate(schemaConfig, true).then(() => config);
```
danielbeardsley, line 67: Hmm... are there things that codegen does besides generate? i.e. now that the config file *always* generates, are there some cli commands (like `--help`) that used to just load the config and now they will end up running it (and generating files).

Perhaps this shouldn't be a config file, but instead a task: `generate(schemaConfig).then(() => generate(config))`

masonmcelvain: I think codegen only evaluates `codegen.ts` when we pass the config file to the command as `--config` option. Like:

https://github.com/iFixit/react-commerce/blob/1c27e290a225e0980152f07d4a812858f1029848/frontend/package.json#L13

So we could run `pnpm graphql-codegen --help`, and the config file doesn't get evaluated.

That being said, this implementation _is_ a bit hacky. We're exporting a promise, though it seems to work despite that.

danielbeardsley: My point was that if all we need is this file to be executed, we can have it call `generate()` twice and then run it directly (however that's done with `.ts` files) instead of having it get inadvertently "run" as part of being loaded as a config file.

Not mandatory, but it's rather weird to have a config file that ends up executing the program it's trying to configure as part of configuration.

danielbeardsley: > Another option would be defining the two configurations separately, and running

I'd choose this option. I like that it's straight forward and the config doesn't execute the program it's intending to config.

masonmcelvain:
un_dev_block ✌🏻 Thanks for the suggestion!

danielbeardsley (approved):
CR :+1: 

Nice!

danielbeardsley:
> No loader specified for type ".ts"

image

dev_block :+1: 

I'll see if I can fix it.

danielbeardsley:
un_dev_block :+1: 

A fresh `pnpm install:all` cured that and it now works.

QA :+1:

PR #1718, opened 2023-05-26
Algolia getServerState(): optimize to only what's needed

### LINKED ISSUE #1641: Possible Performance Improvement: reduce getServerState() work
We call algolia's `getServerState()` which effectively evaluates all your react components and then extract's some state (effectively refinements and results) from the algolia components.

This takes about 40% of our total response time in next.js (for pages that use algolia).

Average of a whole day's worth of data:
image

### Do This
Evaluate this hypothesis: Only components that actually *make* the aloglia requests or set refinements need to be present or rendered in that pass. Having some components not rendered may significantly reduce the time that `getServerState()` takes.

Consider and experiment with passing a "algoliaMinimalMode" prop to the `AppContext` or similar flag to let parts of the page know they can render in a minimal mode and see how it affects timing.

danielbeardsley:
Algolias `getServerState()` does a pre-render phase and then extracts the result
set and facets from the internal state so they can be included /
captured in the props.

Because of this, we don't need any of the rest of the page, just the
minimal components needed to fire off the correct query. Here we add a
"minimalMode" that tries to use the same hooks and filtering, but renders
as few UI components as possible.

In testing locally, this dropped about 100ms of component time from *every*
request while providing the same results.

Closes #1641

### DIFF
--- a/frontend/components/common/AppProviders.tsx
+++ b/frontend/components/common/AppProviders.tsx
@@ -52,7 +52,6 @@ export type AppProvidersProps = {
    algolia?: AlgoliaProps;
    ifixitOrigin?: string;
    adminMessage?: string;
-   minimalMode?: boolean;
 };
 
 export function AppProviders({
@@ -60,7 +59,6 @@ export function AppProviders({
    algolia,
    ifixitOrigin,
    adminMessage,
-   minimalMode,
 }: React.PropsWithChildren<AppProvidersProps>) {
    const markup = (
       <ChakraProvider theme={customTheme}>{children}</ChakraProvider>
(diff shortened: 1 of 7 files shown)


sterlinghirsh (dismissed):
CR :zap:

I'm curious to get this out so we can see if it does actually work. In a future revision, I might want to change the name from "minimalMode" to something that references Algolia (algoliaComponentsOnly or something?), since it's not very clear why minimalMode would be used and there's no comment describing it.

This is definitely getting into hacksville, but 70ms is worth it. We may be able to drop this in the future with some other optimizations we've discussed, but this diff is minimal (get it?) enough that it doesn't really bother me.

danielbeardsley:
QA :+1:

In prod the gains look to be about... 65ms (though it's higher if we drop anomalies). I loaded the same page 10 times (bumping a query param each time to ensure we didn't get a CDN cache hit).

image

danielbeardsley:
> . In a future revision, I might want to change the name from "minimalMode" to something that references Algolia 

I went ahead and did this here... and dropped it from the context as well.

masonmcelvain (approved):
CR ✌🏻

danielbeardsley:
Success!

image

danielbeardsley:
We saved about ~80ms

image

PR #1720, opened 2023-05-26
Link to legacy featured product pages from cross sell section

dhmacs:
closes #1719 

This fixes #1719 as well as a couple of minor React warnings. I've also done some refactoring on the path helpers

## QA

1. Open [this product page preview](https://react-commerce-git-fix-product-links-ifixit.vercel.app/products/tesa-61395-tape)
2. Scroll to the Frequently Bought Together section and click on the pro tech toolkit card
3. Verified that you are redirected to the legacy php product page

PR #1721, opened 2023-05-26
Use smart link for cross sell

masonmcelvain:
Fixes a bug where cross sell links to the pro tech toolkit and manta driver kit would 404 when the app was served on www.ifixit.com.

Related to #1720

It seems like when we use `https://www.ifixit.com/Store/Tools/Pro-Tech-Toolkit/IF145-307` with `<NextLink>`, Next.js tries to serve the page (when on www.ifixit.com) and 404s. A smart link will use a plain link for for urls that start with `https`.

CC @deltuh-vee


davidrans (approved):
CR :coffee:

masonmcelvain:
Looks like the links no longer 404 live 👍🏻

PR #1722, opened 2023-05-26
Add react-i18next package and test it out!

aburke07:
## Overview
This is an example use of react-i18next to try and get it to translate some strings at render time.

This also has some examples of different ways to use interpolation, plurals, and components with translations.

This is just a reference pull for future work, so it should not be merged.

Also, _please excuse my french_ 🤠 

Connects to #https://github.com/iFixit/ifixit/issues/48088


packages/translations/lang/fr.json
```diff
@@ -0,0 +1,12 @@
...
+   "Hello!": "Bonjour!",
+   "welcome {{ userName }}": "Bonjour {{ userName }}",
+   "There are {{ count }} items_one": "There is {{ count }} item",
+   "There are {{ count }} items_other": "There are {{ count }} items",
+   "super <strong>strong</strong>": "Bonjour <strong>strong</strong>",
+   "linkTestTag": "<1>Link's</1> Awakening",
+   "linkTest": "<5>Link's</5> Awakening",
+   "nameTitle": "This is your name",
+   "messagesCount_one": "Bonjour <1>{{userName}}</1>, you have {{count}} unread message. <5>Go to message</5>.",
+   "messagesCount_other": "Bonjour <1>{{userName}}</1>, you have {{count}} unread messages.  <5>Go to messages</5>."
```
sctice-ifixit, line 11: I'm concerned that this isn't really the way that our translation system has worked until now. We always use the English translation as the translation key. I don't see an example of interpolating a React component into a translation where the translation key is the English translation (rather than a separate bespoke key). Is that something we can do?

More to the point, we don't want translators to have to figure out `<1>…</1>`, `{{userName}}`, `{{count}}`, and `<5>Go to messages</5>` based on a translation key like `messagesCount_other`. Up to this point, we've never created English translations because the translation key _is_ the English translation. I wouldn't want to have to start having developers create English translations in order to establish the appropriate variables/placeholders.

aburke07: > Is that something we can do?

I tested this out multiple ways but wasn't able to get react components to translate using the English string as the key. I will look into it more and see if it is possible.

aburke07: I am still getting a re-hydration error at render time though, so I still need to look into that.

PR #1723, opened 2023-05-26
Algolia: optimize by sharing results of `useHits()`

danielbeardsley:
useHits() is expensive (> 1ms per call) and we called it in quite a few
places. Calling it once instead saves ~10ms or so.

Note this savings happens twice on the server (in getServerState()) and
in the next.js render-to-string phase, then it happens client side after
every change in the search query.

Connect #1641


dhmacs:
First of all, thanks for the work you're doing in identifying potential performance optimizations.
I have a few question about these changes.

1. Could you please elaborate a bit on the method you used to measure the time taken by useHits()?
2. Can we dive a bit deeper into why useHits() is expensive? Is it due to the complexity of computations?
3. Adding a React context can sometimes cause the app to render more components than necessary. While your solution does seem to be a logical way to prevent multiple calls to useHits(), it might introduce other costs that we are not considering. Can we use the same method you used for profiling useHits() to verify that the proposed solution is indeed more efficient, considering all possible impacts?

I'm asking these questions because I'm a bit surprised that adding a context saves time in this scenario, as generally adding contexts introduces more computation

danielbeardsley:
> Could you please elaborate a bit on the method you used to measure the time taken by useHits()?

It's a synchronous call, so I just used `time()`: https://github.com/iFixit/react-commerce/blob/main/packages/helpers/generic-helpers.ts#L89

```
const t = time("useHits()");
const { hits } = useHits();
t();
```

> Can we dive a bit deeper into why useHits() is expensive? Is it due to the complexity of computations?

Yes. It apparently "escapes" the html on all the results. This is partly why it's slow.... oh hmmm.. I thought I linked to the issue on aloglia, guess I forgot: https://github.com/algolia/instantsearch/issues/5237

> Can we use the same method you used for profiling useHits() to verify that the proposed solution is indeed more efficient, considering all possible impacts?

We can use the same mechanism because `useHits()` is synchronous while rendering everything is async. But timing it is indeed important. I did lots of tests like this with 10 samples and I'll do it again with the changes in context and report back.
image


> Adding a React context can sometimes cause the app to render more components than necessary.

Yeah, I considered this. We didn't add a context, but we added new values to the context and moved it up one level. A change in context results in every component under the Provider re-rendering. We already expect every component in the `<InstantSearchPRovider>` to be re-rendered when the algolia query or filters change, so, on the surface, I don't expect this to result in more renders of any components. You're right though, it's certainly worth measuring.

danielbeardsley:
Here's the results of this branch when looking at just one page 10 times: `Parts/Android_Tablet` on vercel's infrastructure (using the react-commerce-prod deployment). I added a query param and changed the value of it so each time was a CDN cache miss.

image


If we drop the worst datapoint from both data sets, the difference is 45ms savings. If we drop the two worst datapoints the savings is still 34ms.

I'll now do some experimentation around quantity of renders.

danielbeardsley:
Closing in favor of #1729

PR #1729, opened 2023-05-30
Algolia: prefer usePagination()

danielbeardsley:
useHits() is an expensive call on the server (> 1ms per call) and we
called it in quite a few places (some are in loops). Calling
usePagination() is < 0.1ms and provides us the thing we need most often
(hit count). This saves us ~10ms or so.

I've noticed that useHits() is not slow on the client strangely enough.

Note this savings happens twice on the server (in getServerState() and
in the next.js render-to-string phase.

This is a minimal version of pull #1723

### DIFF
--- a/frontend/components/common/InstantSearchProvider.tsx
+++ b/frontend/components/common/InstantSearchProvider.tsx
@@ -62,7 +62,7 @@ export function InstantSearchProvider({
          algoliaApiKey,
          getClientOptions(logContextName)
       );
-   }, [algoliaApiKey]);
+   }, [algoliaApiKey, logContextName]);
 
    const facets = useFacets();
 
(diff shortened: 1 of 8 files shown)


danielbeardsley:
@mmarcon91 and @dhmacs -- This is a simpler way to get the performance gains from #1723 -- Thoughts?

The gains aren't quite as big, but it's a bit hard to tell given the volatility:
image

danielbeardsley:
Turns out... I think all algolia hooks are just achingly slow on the server side. Calling any of the `use...()` hooks is just slow (about 1ms a piece). I've been playing with node debugger and performance profiler.

dev_block :+1: on a bit more experimentation. I think the previous context approach might have been better if we can work out how to reduce the number of renders client-side.

sterlinghirsh:
I think this is out of date enough that I'm going to close it.

PR #1731, opened 2023-06-01
localStorage: require a schema validator

### LINKED ISSUE #1656: localStorage: Can we do zod schema validation on value when they come out of cache?
From Post-Mortem: #1653

### Problem
We cached user data in local storage. The schema changed, the cached values didn't match the new schema, but we don't do validation, so the result was an exception when our components used the data. The exception caused a blank page until you refreshed

### Do This
It's a good idea to make a similar schema-validating cache layer for localStorage so that changing the schema will invalidate the cache and we won't have to remember to bump a cache key.

Definitely add validation. Maybe implement a custom solution instead of zod if the bundle size increase is significant.

danielbeardsley:
Now corrupt values in localStorage won't be used if they don't pass the validator (which is required to return the same type as the data or null).

Closes #1656

### DIFF
--- a/packages/ui/hooks/index.ts
+++ b/packages/ui/hooks/index.ts
@@ -196,7 +196,7 @@ export function useLocalPreference<Data = any>(
          if (serializedData != null) {
             const data = validator(JSON.parse(serializedData));
             if (data !== null) {
-               setData(data as Data);
+               setData(data);
             }
          }
       } catch (error) {}
--- a/packages/ui/hooks/index.ts
+++ b/packages/ui/hooks/index.ts
@@ -136,7 +136,8 @@ export function useSSRBreakpointValue<Value>(
 export function useExpiringLocalPreference<Data = any>(
    key: string,
    defaultData: Data,
-   expireInDays: number
+   expireInDays: number,
+   validator: (data: any) => Data | null
 ): [Data, (data: Data) => void] {
    type ExpiringData = {
       value: Data;
@@ -153,8 +154,9 @@ export function useExpiringLocalPreference<Data = any>(
             const expiresAt = Number.isInteger(data?.expires)
                ? data.expires
                : 0;
-            if (expiresAt && Date.now() < expiresAt) {
-               setData(data?.value);
+            const validData = validator(data?.value);
+            if (validData !== null && expiresAt && Date.now() < expiresAt) {
+               setData(validData);
             } else {
                localStorage.deleteIem(key);
             }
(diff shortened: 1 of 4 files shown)


packages/ui/hooks/index.ts
```diff
@@ -153,8 +154,9 @@ export function useExpiringLocalPreference<Data = any>(
             const expiresAt = Number.isInteger(data?.expires)
                ? data.expires
                : 0;
-            if (expiresAt && Date.now() < expiresAt) {
-               setData(data?.value);
+            const validData = validator(data?.value);
+            if (validData !== null && expiresAt && Date.now() < expiresAt) {
+               setData(validData);
```
danielcliu-ifixit, line 159: we could treat the validator returning anything besides null as valid and then `setData(data?.value)` so we wouldn't need the ternary operations in the validators anymore

packages/ui/hooks/index.ts
```diff
@@ -183,16 +185,19 @@ export function useExpiringLocalPreference<Data = any>(
...
 
    React.useEffect(() => {
       try {
          const serializedData = localStorage.getItem(key);
          if (serializedData != null) {
-            const data = JSON.parse(serializedData);
-            setData(data as Data);
+            const data = validator(JSON.parse(serializedData));
+            if (data !== null) {
+               setData(data as Data);
```
danielcliu-ifixit, line 199: do we need this type validation still if we have a validator?

danielcliu-ifixit (dismissed):
CR :v: but left some small suggestions

packages/ui/hooks/index.ts
danielbeardsley: You're right, we don't!

packages/ui/hooks/index.ts
danielbeardsley: I don't really understand, it seems like what we're doing here: 

```
if (validData !== null) { setData(data) }
```

danielcliu-ifixit: sorry what i meant to propose is having the validators return a boolean value so like instead of 
```php
data === ProductViewType.List || data === ProductViewType.Grid
            ? data
            : null
```

we could just have 
```php
data === ProductViewType.List || data === ProductViewType.Grid
```

danielcliu-ifixit (approved):
CR :v:

packages/ui/hooks/index.ts
danielbeardsley: Oh, I forgot that typescript can do assertion functions...,..

Update: I tried but I couldn't get typescript to enforce that the validator function was checking the type. It's easier to enforce that it returns the right type.


For instance, using a validator type of:
`validator: (data: any) => data is Data`
and a validation function of:
`(data: any): data is ProductViewType => data`

Results in typescript not complaining. It apparently doesn't validate that your assertion function actually asserts anything.

deltuh-vee:
QA 🎬 
Invalid data in Local Storage no longer crashes the page.

PR #1739, opened 2023-06-07
framer-motion: upgrade to v10

davidrans:
This gets us on the same version as iFixit/ifixit.

QA
---
This is mostly used via Chakra, but we do have a couple explicit usages in the Cart Drawer code. Let's double-check the cart drawer animations and make sure nothing's busted.

### DIFF
--- a/packages/footer/package.json
+++ b/packages/footer/package.json
@@ -19,7 +19,7 @@
       "@chakra-ui/react": ">=2",
       "@core-ds/primitives": ">2.4",
       "@emotion/react": ">=11.0.0 <12.0.0",
-      "framer-motion": ">=3.0.0 <4.0.0 || >=4.0.0 <5.0.0 || >=5.0.0 <6.0.0 || >=6.0.0 <7.0.0",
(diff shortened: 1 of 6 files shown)


packages/footer/package.json
```diff
@@ -19,7 +19,7 @@
       "@chakra-ui/react": ">=2",
       "@core-ds/primitives": ">2.4",
       "@emotion/react": ">=11.0.0 <12.0.0",
-      "framer-motion": ">=3.0.0 <4.0.0 || >=4.0.0 <5.0.0 || >=5.0.0 <6.0.0 || >=6.0.0 <7.0.0",
+      "framer-motion": ">=4.0.0 <11.0.0",
```
davidrans, line 22: I changed the minimum version to 4 cause I think that's the minimum version supported by Chakra

masonmcelvain (approved):
CR ✌🏻

deltuh-vee:
QA 🎬 
Cart drawer animations still look good.

PR #1759, opened 2023-06-14
GA4: Install gtag tracking code

jarstelfox:
Our old GA account is sunsetting on June 30th. This old account is GA's Universal Analytics (UA). The new version is GA4.

Google seems to have made the transition easy enough:

You can run both the UA and GA4 code side by side
The existing events from the UA version are sent automagically

This will enable GA analytics (old and new)

QA
---
* Ensure you receive console logs for all the usual GA events
	* This tests the old UA is still working
* Ensure [the new debugviewer](https://analytics.google.com/analytics/web/?authuser=1#/a30506p383872824/admin/debugview/overview) shows the events sent to the GA4 instance
	* You may need to ping me for access to view this URL ^
* Also, ensure Matomo is receiving events as well

Connects: [#48129](https://github.com/iFixit/ifixit/issues/48129)


ianrohde (dismissed):
CR ⚪

deltuh-vee:
dev_block ✊ 
I'm not seeing events from react-commerce in the debug [viewer](https://analytics.google.com/analytics/web/?authuser=1#/a30506p383872824/admin/debugview/overview).

frontend/.env.development
```diff
@@ -9,8 +9,9 @@ NEXT_PUBLIC_MATOMO_URL=https://matomo.ubreakit.com
 NEXT_PUBLIC_GA_KEY=UA-30506-1
 # If we want to test Google Analytics in dev DEBUG should be true and URL should be
 # https://www.google-analytics.com/analytics_debug.js
-NEXT_PUBLIC_GA_DEBUG=false
+NEXT_PUBLIC_GA_DEBUG=true
```
masonmcelvain, line 12: You need to set this in `.env.production` too so that the previews don't dump data into product GA. Then in the one production deployment, we will override the env var

jarstelfox: >then in the one production deployment, we will override the env var

Could you expound a bit? Is there a third place for env vars? Is there a UI in vercel?

frontend/.env.development
```diff
@@ -9,8 +9,9 @@ NEXT_PUBLIC_MATOMO_URL=https://matomo.ubreakit.com
...
 NEXT_PUBLIC_GA_URL=
+NEXT_PUBLIC_GTAG_ID=G-5ZXNWJ73GK
```
masonmcelvain, line 14: If this will be the same everywhere, we can put in the `.env` file (we can create it if it doesn't exist yet)

https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#default-environment-variables

frontend/.env.development
masonmcelvain: Yea, sorry for my brevity. There is a UI on vercel here to override env vars: https://vercel.com/ifixit/react-commerce-prod/settings/environment-variables

image

Vercel uses `.env.production` for the previews and the production deployment. If we need more granular control than that, we've been putting the preview version of the env var in `.env.production`, and overriding the production value in Vercel's UI.

For example, we only want the production deployment to report to production Sentry. So we put the dev sentry url in `.env.production`, and override it in vercel's ui for the production deployment.

image

masonmcelvain: We also use the vercel ui for env vars we'd like to keep private and not commit to a public repo

jarstelfox:
un_dev_block 🌵 

deploy_block 🌵 @masonmcelvain I assume we will want to set the env var in Vercel's UI before deploy?

masonmcelvain (approved):
CR ✌🏻

masonmcelvain:
>  I assume we will want to set the env var in Vercel's UI before deploy?

Yep! Thanks for the reminder. I added the env var override to react-commerce-prod:

image

un_deploy_block ✌🏻

frontend/components/analytics/GoogleAnalytics.tsx
```diff
@@ -22,8 +22,22 @@ export function GoogleAnalytics() {
       };
    }, [router?.events]);
 
-   return GA_URL ? (
+   return GA_URL || GTAG_ID ? (
       <>
+         <Script
+            id="gtag-ga4"
+            strategy="afterInteractive"
+            src={`https://www.googletagmanager.com/gtag/js?id=${GTAG_ID}`}
```
danielbeardsley, line 30: If `GTAG_ID` is falsy, then don't we want to *not* include this script tag?

As written, this will always include the tag if either the `GA_URL` OR `GTAG_ID` is set.

danielbeardsley (changes requested):
dev_block :+1:  on a bug.

frontend/components/analytics/GoogleAnalytics.tsx
jarstelfox: Fair point. Do we want to treat them as independent or both required?

masonmcelvain (approved):
CR ✌🏻

deltuh-vee:
QA 🎬
- The new debugviewer shows the events sent to the GA4 instance
- Matomo is still receiving events

❓ Were there any console logs for GA in react-commerce?

jarstelfox:
>Were there any console logs for GA in react-commerce?

No, it seems is was already using the new API (GA4), but with the old account (GA3) I am slightly unsure how to test it.

I was wondering if ship it and revert if required was enough

PR #1761, opened 2023-06-16
Cross Sell: Display Multiple Items

danielcliu-ifixit:
## Overview

We want to have mulitple x-sell items show up in minicart. This makes it so that happens. 

It also does away with some unnecessary `upsell` diction that i forgot to clean up previously.

## QA

For now, just make sure the cross sell works the same as before

Connects: https://github.com/iFixit/ifixit/issues/48134


deltuh-vee:
QA 🎬 
Cross Sells is still working.

sterlinghirsh (approved):
CR :zap:

PR #1768, opened 2023-06-21
Id is required for inline script

federicobadini:
Hotfix for a small issue on main causing build failures.

Due to a typo the id was placed on the `Script` tag with src instead on the one with inline content.
According to Nextjs documentation it is needed only for inline scripts.

See https://nextjs.org/docs/messages/inline-script-id

qa_req 0


danielbeardsley (approved):
CR :+1:

PR #1770, opened 2023-06-22
Correct algolia integration url handling

### LINKED ISSUE #1764: Filter facets: On first visit, item type facets represented in url with param
## Description
When I first select an item type filter, it's applied to the URL as a search param. Upon reload, or if I re-do the same steps, I get the item type added to the path consistently. 

https://github.com/iFixit/react-commerce/assets/72166715/b113df1d-5ffa-410e-ba53-0fe727d45512

## Additional details
Found while testing https://github.com/iFixit/react-commerce/pull/1763#issuecomment-1599349142

### LINKED ISSUE #1765: Product list pages: Intermediate states shown to user on page load
## Details
There are a lot of intermediate states shown to the user when moving from a parent to a child collection. 
 
All the screen-recordings and shots here are when moving from a page with an item type selected to a child collection. Similar behavior is present when moving from a parent to a child without any item-type filters present, the item type changes just aren't present. 

<details><summary>Screen-recording</summary>

https://github.com/iFixit/react-commerce/assets/72166715/61d0bb8e-80ba-462b-99f9-25d5b6f6e306

</details>

Screenshots showing all the different states changed on load:
Starting at: https://www.ifixit.com/Parts/Kindle_Tablet/Buttons
image
image
image
image
Ending at: https://www.ifixit.com/Parts/Kindle_Paperwhite_4

## Additional details
Found while testing https://github.com/iFixit/react-commerce/pull/1763#issuecomment-1599349142

### LINKED ISSUE #1766: Product list: Filter facets not cleared out when moving to child collection
## Details
I noticed that this isn't happening consistently after the first try. Maybe caching issue?
Steps:
1. Go to a collection page
2. Add an item type facet
3. Click on a child page
4. See the item type is still applied to the page

https://github.com/iFixit/react-commerce/assets/72166715/a204ceeb-75c1-4e44-b45a-4657560c7fd9

## Additional details
Found while testing https://github.com/iFixit/react-commerce/pull/1763#issuecomment-1599349142

federicobadini:
closes #1764

RouterOptions' createURL and parseURL functions were not pure anymore due to the references to `productListType`. InstantSearch seems to live across page render and to preserve the first read `productListType` value even after navigation.
It was reverted to the old logic.

closes #1765 

Intermediate states are due to the fact that Algolia share the provider across renders and if we move via Next links Algolia reacts to the url changes only after the effective page change. 
These shifts might also have an impact on CLS since page content is moved around with a not so small delay from user interaction (connects #1643)

A possible solution to this was to force regenerating the Algolia provider on every url change (as suggested [here](https://github.com/algolia/instantsearch/issues/5241#issuecomment-1438318404)).
This proved to be a valid solution for intermediate states but still showed issues in correctly dealing with browser back button (see [here](https://github.com/algolia/instantsearch/issues/5622) and the video below)

https://github.com/iFixit/react-commerce/assets/7805759/984bcdfe-34aa-43c3-b98a-e5e5f1eda0e7

To solve all this issue `productListChildren` were moved from Next links to basic anchor tags.

closes #1766

### DIFF
--- a/frontend/components/common/InstantSearchProvider.tsx
+++ b/frontend/components/common/InstantSearchProvider.tsx
@@ -10,7 +10,7 @@ import {
 import { useAuthenticatedUser } from '@ifixit/auth-sdk';
 import { assertNever } from '@ifixit/helpers';
 import { usePrevious } from '@ifixit/ui';
-import { FacetWidgetType, ProductListType } from '@models/product-list';
+import { FacetWidgetType } from '@models/product-list';
 import { useFacets } from '@templates/product-list/sections/FilterableProductsSection/facets/useFacets';
 import algoliasearch from 'algoliasearch/lite';
 import singletonRouter from 'next/router';
@@ -29,7 +29,6 @@ type InstantSearchProviderProps = React.PropsWithChildren<AlgoliaProps>;
 export type AlgoliaProps = {
    url: string;
    indexName: string;
-   productListType: ProductListType;
    serverState?: InstantSearchServerState;
    apiKey: string;
    logContextName: string;
@@ -47,7 +46,6 @@ export function InstantSearchProvider({
    children,
    url,
    indexName,
-   productListType,
    serverState,
    apiKey,
    logContextName,
@@ -89,8 +87,7 @@ export function InstantSearchProvider({
             .filter((part) => part !== '');
          const firstPathSegment = pathParts.length >= 1 ? pathParts[0] : '';
          const deviceHandle = pathParts.length >= 2 ? pathParts[1] : '';
-         const isDevicePartsPage =
-            productListType === ProductListType.DeviceParts;
+         const isDevicePartsPage = firstPathSegment === 'Parts' && deviceHandle;
 
          let path = `/${firstPathSegment}`;
          if (deviceHandle) {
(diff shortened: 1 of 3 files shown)


masonmcelvain (approved):
CR ✌🏻 Thanks @erinemay for catching these bugs and @federicobadini for fixing them! It's unfortunate that we have to get rid of the NextLink on product list children, but it seems to be the easiest working solution based on what is described.

Perhaps we can find another workaround eventually that still allows us to use NextLink. Since the live production UX is poor though, we should deploy this fix if it works.

deltuh-vee:
dev_block ✊ 
For https://github.com/iFixit/react-commerce/issues/1766, @sterlinghirsh says we want to retain filters if possible. 

https://github.com/iFixit/react-commerce/assets/5289715/0985f84b-5421-4e0d-b05b-e08099b430f7

federicobadini:
If @sterlinghirsh agrees, I would suggest to open another issue for [this](https://github.com/iFixit/react-commerce/issues/1766#issuecomment-1601371946).

Preserving filters ONLY if results are present seems not that easy and will require some additional investigation.

Moreover, filters preservation is currently just an unexpected side effect that happens if the user follows some precise browsing flows. 
For instance if you follow this flow - `All parts --> Mac Parts --> apply filter --> Mac Desktop Parts` - the filter is preserved.
But if you follow this flow - `Mac Parts --> apply filter --> Mac Desktop Parts` - the filter is not preserved.

This PR primarily aims at fixing the current situation restoring a consistent UX under all browsing flows.

sterlinghirsh:
CR :zap: un_dev_block :zap: I think this change is an improvement. I still would like to go back and try to get child device filtering working since I was pretty sure it worked before but I agree it's a different issue and this fixes the bug.

deltuh-vee:
QA 🎬 
* The Itemtype filters in the url are now more consistent. 
* Fewer intermediate states are shown upon refresh

PR #1778, opened 2023-06-27
Troubleshooting: Use `id` values from server

andyg0808:
This both ensures that jumplinks point at the right place and avoids the need to define a client-side function to generate ID values.

## QA Notes
Please confirm that jumplinks point at the right places on the page.
Here's an example page which links to another solution on the same page in the text of the first solution: https://react-commerce-git-troubleshooting-use-id-values-daf02c-ifixit.vercel.app/Vulcan/Dryer_Making_Loud_Noise

Closes https://github.com/iFixit/ifixit/issues/48525


djmetzle (approved):
Seems reasonable. CR :ticket:

deltuh-vee:
QA 🎬
Jumpllinks point to the right places on the page.

hcutrone (approved):
CR :burrito:

PR #1779, opened 2023-06-28
Upgrade to node 18

### LINKED ISSUE #1756: Upgrade to node `18.x`
image

federicobadini:
closes #1756

This PR addresses and closes issue #1756 by standardizing Node.JS to v18 for all environments (development, preview, and production). The upgrade extends to both the frontend and Strapi.

#### Frontend
We compensated for a change brought in by Node.js 17 that deprioritized ipv4 addresses over ipv6 by introducing the env variable --dns-result-order ipv4first across various commands; see [here](https://nodejs.org/api/dns.html#dnssetdefaultresultorderorder). Node.JS 17 made this [change](https://github.com/nodejs/node/issues/40702#issuecomment-958157082) without introducing a happy-eyeballs implementation, which led to addresses often resolving to the ipv6 loopback address ::1 instead of 127.0.0.1, thus breaking builds. Notably, Node 20 included a [happy-eyeballs implementation](https://github.com/nodejs/node/pull/44731), and we anticipate being able to remove these env variables in the future.

#### Strapi
Our current version of Strapi advises the use of Node 18 as per their documentation [here](https://docs.strapi.io/dev-docs/installation/cli). To reflect this, we have updated the backend's package.json to only permit the use of the advised Node.JS versions ("node": ">=14.x.x <=18.x.x") and we updated the Strapi docker base image.

### QA

1. Visit Vercel preview.
2. Validate that no discernible changes in backend, frontend, or test functionalities are present.

Node.JS 18 is already set as the build image for the react-commerce Vercel project. Please note that react-commerce-prod will also be updated just before this PR is deployed.
So deploy_block 🛑 for the above reason.


k0rvusk0r4x:
All looks well.

QA 😺

backend/Dockerfile
```diff
@@ -1,7 +1,5 @@
-FROM node:16-alpine3.14
-# Installing libvips for sharp compatibility
-RUN apk --update-cache add vips-dev \
-  && rm -rf /var/cache/apk/*
+FROM node:18-alpine3.18
+RUN apk update && apk add python3 build-base
```
masonmcelvain, line 2: I don't entirely understand why we removed 

```
# Installing libvips for sharp compatibility
RUN apk --update-cache add vips-dev \
  && rm -rf /var/cache/apk/*
```

and added

```
RUN apk update && apk add python3 build-base
```

I think we use docker to build Strapi, so I don't think these changes should have any runtime implications as long as the builds are still successful.

masonmcelvain (approved):
CR ✌🏻 With a comment about something I don't understand. Thanks for doing this!

k0rvusk0r4x:
This continue to look good.

QA 😺

erinemay:
deploy_block 🛑 I think deploy blocks in the pull description don't register.

backend/Dockerfile
federicobadini: Vips-dev was originally included because otherwise the yarn install command errored out (sharp dependency). 

Upon upgrading to Node 18, my intent was to declutter the install process by eliminating those additional packages, keeping the install process as barebone as possible.

However, running the build without any additional package produced errors since both Python and make commands were missing. That's why python and build-base were reintroduced.

danielbeardsley:
un_deploy_block :+1: Looks like react-commerce-prod is already at node 18.

image

Also, the strapi docker image is built and deployed with every relevant merge, so no manual work is needed here.

PR #1783, opened 2023-06-28
Prerendered: Support SVG icons from FontAwesome

andyg0808:
We're considering adding inline SVG icons to wikitext under some conditions; this adds the necessary CSS to support them.

## QA Notes
This is tricky to test, because we need to render some SVGs in order to see if the CSS is working. There's a parallel PR which can be checked out to start generating these SVG icons: https://github.com/iFixit/ifixit/pull/48569
With that checked out (and possibly with some caches busted), loading a problem page with links which open in a new tab should show nice icons to the right of each link.

qa_req 0 https://github.com/iFixit/react-commerce/pull/1783#issuecomment-1622529267

### DIFF
--- a/frontend/templates/troubleshooting/prerendered.tsx
+++ b/frontend/templates/troubleshooting/prerendered.tsx
@@ -213,7 +213,7 @@ const renderStyles: SystemStyleObject = {
       },
    },
 
-   '.svg-icon': {
+   '.fa-svg-icon': {
       svg: {
          width: '1em',
          height: '1em',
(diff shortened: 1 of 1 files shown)


danielbeardsley (approved):
CR :+1: 

Looks sensible!

erinemay:
I'm happy to qa_req 0 this. It's not breaking the page, so further testing or development isn't hindered by merging. We can confirm it works on Vulcan pages once https://github.com/iFixit/ifixit/pull/48569 is merged.

PR #1793, opened 2023-06-30
Vulcan: fix missing linkcard images

aburke07:
## Overview
Some link cards were missing images because the product variant didn't have an image set. The variant image is not guaranteed to exist, and we have always fell back to the first product image. We didn't do that here so we were missing images.

This also removes duplicate link cards from the same solution.

## QA
1. View a Vulcan wiki with a product link card that is missing an image.
Ex: http://localhost:3000/Vulcan/Nintendo%20Switch%20Blue%20Screen%20of%20Death#display-damage (the [`Nintendo Switch Screen OLED Assembly`](http://localhost:3000/products/nintendo-switch-oled-screen-assembly) product card)
2. The link card should now have an image.
3. Viewing a solution with duplicate links should no longer show duplicate link cards
Ex: http://localhost:3000/Vulcan/Nintendo%20Switch%20Blue%20Screen%20of%20Death#Section_CPU_Disconnect

Closes https://github.com/iFixit/ifixit/issues/48545

### DIFF
--- a/frontend/templates/troubleshooting/Resource.tsx
+++ b/frontend/templates/troubleshooting/Resource.tsx
@@ -60,7 +60,11 @@ export function ProductResource({ product }: { product: Product }) {
       <Resource
          href={productUrl}
          title={product.title}
-         imageUrl={selectedVariant.image?.url}
+         imageUrl={
+            selectedVariant.image?.url ||
+            product.images[0]?.thumbnailUrl ||
+            product.images[0]?.url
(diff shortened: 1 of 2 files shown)


hcutrone (approved):
CR :burrito:

deltuh-vee:
QA 🎬 
* Product links with missing images now have images
* Duplicate links in a solution no longer show multiple link cards

PR #1799, opened 2023-07-03
Troubleshooting: Redirect to relative URL

andyg0808:
The `Location` header supports relative URLs; let's use those to allow the browser to handle the protocol/host matching for the redirect.

This code seems to have been causing weird redirect problems in prod; I'm not sure what's going wrong, but simplifying it seems to be a good first thing to try.

Connects https://github.com/iFixit/ifixit/issues/48621

### DIFF
--- a/frontend/templates/troubleshooting/server.tsx
+++ b/frontend/templates/troubleshooting/server.tsx
@@ -107,15 +107,11 @@ export const getServerSideProps: GetServerSideProps<TroubleshootingProps> =
       }
 
       const canonicalUrl = new URL(troubleshootingData.canonicalUrl);
-      canonicalUrl.protocol = /localhost/.test(context.req.headers.host || '')
-         ? 'http'
-         : 'https';
-      canonicalUrl.host = context.req.headers.host || canonicalUrl.host;
 
       if (context.resolvedUrl !== canonicalUrl.pathname.toString()) {
          return {
             redirect: {
-               destination: canonicalUrl.toString(),
+               destination: canonicalUrl.pathname.toString(),
                permanent: true,
             },
          };
--- a/frontend/templates/troubleshooting/server.tsx
+++ b/frontend/templates/troubleshooting/server.tsx
@@ -107,11 +107,22 @@ export const getServerSideProps: GetServerSideProps<TroubleshootingProps> =
       }
 
       const canonicalUrl = new URL(troubleshootingData.canonicalUrl);
+      /*
+       * Since `resolvedUrl` doesn't include a hostname or protocol,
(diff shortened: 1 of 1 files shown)


jarstelfox (approved):
CR 🌵

frontend/templates/troubleshooting/server.tsx
```diff
@@ -107,15 +107,11 @@ export const getServerSideProps: GetServerSideProps<TroubleshootingProps> =
...
       const canonicalUrl = new URL(troubleshootingData.canonicalUrl);
-      canonicalUrl.protocol = /localhost/.test(context.req.headers.host || '')
-         ? 'http'
-         : 'https';
-      canonicalUrl.host = context.req.headers.host || canonicalUrl.host;
 
       if (context.resolvedUrl !== canonicalUrl.pathname.toString()) {
          return {
             redirect: {
-               destination: canonicalUrl.toString(),
```
danielbeardsley, line 118: Before we perma-redirect, I think we need to compare the *pathname* to the *pathname*. It appears here that we're comparing the *url* (i.e. including query string) to the *pathname*. If we know that `context.resolvedUrl` *doesn't* include the query string, then I'd suggest putting in a local `const currentPathname = context.resolvedUrl`.

danielbeardsley (changes requested):
CR :+1: 

While this change seems reasonable, I may see a problem in url redirection, so dev_block :+1: on looking into that.

deltuh-vee:
QA 🎬 
Redirects are still working in react-commerce. I think we'll have to check this once it's merged to see if the redirects are working again in cominor.

frontend/templates/troubleshooting/server.tsx
jarstelfox: You're right. I'd suggest turning both in to URL objects and comparing `url.pathname` only (or explicitly checking the params as well)

andyg0808: Oh, nice point. I'll switch that.

andyg0808:
Sorry for force-push; it was only to amend the last commit to include the correct query string.

andyg0808:
un_dev_block :+1: 
@danielbeardsley, thanks for the catch! I think that's fixed now.

jarstelfox (approved):
CR 🌵 this sounds correct to me

erinemay:
QA 🦖 
On this branch, `/Vulcan/Dryer_Not_Spinning` is redirected to `/Vulcan/Dryer%20Not%20Spinning` in production and dev links, but it is on main as well. Will need to check on cominor after merge.

PR #1801, opened 2023-07-05
Add bulk operations to addons plugin

### LINKED ISSUE #266: Enable bulk update of product list attributes
The marketing team needs to bulk update product list fields.

An example of that would we replacing "parts" with "replacement parts" in every product list metadescription
Strapi currently doesn't have a native interface for bulk updates.

## To do

Research a solution to enable bulk updates. Probably this would have to be done via a custom plugin

cc @jeffsnyder

dhmacs:
closes #266 

This PR adds bulk operations to the Strapi addons plugin. As part of this the addons plugin UI has been updated with a navigation sidebar.
This PR introduces also some changes to Strapi Typescript config that are necessary for zod to work correctly and to improve types overall.

## QA

1. Open the [Strapi preview instance](https://add-bulk-operations.govinor.com/admin/plugins/addons/bulk-operations) for this branch
2. Navigate to the addon plugin and click "Bulk operations" on the secondary nav bar
3. Try to export a collection type (e.g. "product list"): this should give you a link to download the CSV
    Screenshot 2023-07-05 at 22 12 21

4. Open the downloaded CSV with Google Spreadsheet and update the entries value. Add also new entries (new entries are rows with the id column left blank) and you can delete the rows that you don't need to update.
5. Export the updated file as a CSV
6. In the addons plugin "Bulk Operations" page, click on the "import" tab and follow the instructions to upload the file and start the import
    Screenshot 2023-07-05 at 22 13 17

7. Once the import finishes, navigate to the Strapi Content Manager and verify that the changes you've made to the CSV have been applied to the collections


deltuh-vee:
Could we exclude LegacyPageID and LegacyDescription?

deltuh-vee:
dev_block ✊ 
I wasn't able to create a new productList. All I have to do is leave `id` empty, right?
image
Here's the CSV I was trying to use:
[new_productlist.csv](https://github.com/iFixit/react-commerce/files/12005645/new_productlist.csv)

dhmacs:
@deltuh-vee It took me a while to get this, but apparently the product list db table sequence got out of sync. I've fixed the issue by running:

```sql
SELECT setval('product_lists_id_seq', (SELECT MAX(id) FROM product_lists));
```

I'm not quite sure how that happened exactly, it might be due to some Strapi bug. 

Can you retry the QA steps that you did and see if it works now?

un_dev_block 🟢

backend/src/plugins/addons/server/services/bulk-operations/import-csv.ts
```diff
@@ -0,0 +1,271 @@
...
+      for (const name of requiredComponentAttributesNames) {
+         input[name] = {};
+      }
+      for (const name of requiredDynamicZoneAttributesNames) {
+         input[name] = [];
+      }
+   }
+   if (typeof record.published === 'string') {
+      if (isTrue(record.published)) {
+         input.publishedAt = new Date().toISOString();
```
masonmcelvain, line 182: Should we set `updatedAt` here too, so that it matches with `publishedAt`? Probably doesn't matter that much, but the consistency might be nice for end users.

masonmcelvain (approved):
CR ✌🏻 Nice work this looks good! None of my comments are blocking.

backend/src/plugins/addons/admin/src/components/bulk-operations/ImportSection.tsx
```diff
@@ -0,0 +1,352 @@
...
+   return stats;
+}
+
+function parseCSVLines(csvString: string) {
+   const lines: string[] = [];
+
+   let inQuote = false;
+   let line = '';
+
+   for (var i = 0; i < csvString.length; i++) {
```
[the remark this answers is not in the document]
dhmacs: Because `fast-csv` is based on Node.js APIs so it requires a node environment

backend/src/plugins/addons/server/services/bulk-operations/import-csv.ts
dhmacs: I think `updatedAt` autoupdates whenever the record is updated 🤔

masonmcelvain (dismissed):
CR ✌🏻 Thanks for the changes!

erinemay:
QA 🦖 the product lists I added are added in strapi admin. 


deploy_block 🧷  on some questions and weirdness around non-productlist uploads. If we don't have use cases for these right now, they're probably good as follow ups.

1. I'm curious if this feedback is correct. 
I made a change to 1 item and added 2 new items. Is this feedback just because it updates the whole productlist set based on the new CSV?
image

2. It's setting my newly added productlists `defaultShowAllChildrenOnLgSizes`, `forceNoindex`, and `forceNoindex` to false automatically. Here's the exported CSV after I made the change showing the values as FALSE after I uploaded mine with those fields blank. Just want to make sure this is expected.
```csv
id,type,handle,title,h1,description,brandLogoWidth,filters,tagline,metaDescription,metaTitle,deviceTitle,forceNoindex,hideFromParent,legacyPageId,sortPriority,legacyDescription,defaultShowAllChildrenOnLgSizes,locale,published
2381,parts,test-new-collection-test,New Collection Test,,test description goes here,,some tagline,TestMETA,device title here,FALSE,FALSE,,0,,FALSE,en,TRUE
2382,parts,test-non-published,Unpublished Collection,,test unpublished description here,,another tagline,UNpub meta,device tiele here,FALSE,FALSE,,0,,FALSE,en,FALSE

```

3. A [productlist](https://add-bulk-operations.govinor.com/admin/content-manager/collectionType/api::product-list.product-list/2381?plugins[i18n][locale]=en) that I added and set up to be published 404s https://react-commerce-git-add-bulk-operations-ifixit.vercel.app/Parts/test-new-collection-test?disableCacheGets=1. Should an added productlist entry be pubished? Is there something wrong with the data I added? Here're the rows I added. 
```csv
id,type,handle,title,h1,description,brandLogoWidth,filters,tagline,metaDescription,metaTitle,deviceTitle,forceNoindex,hideFromParent,legacyPageId,sortPriority,legacyDescription,defaultShowAllChildrenOnLgSizes,locale,published
,parts,test-new-collection-test,New Collection Test,,test description goes here,,some tagline,TestMETA,device title here,,,,,0,,FALSE,en,TRUE
,parts,test-non-published,Unpublished Collection,,test unpublished description here,,another tagline,UNpub meta,device tiele here,,,,,0,,FALSE,en,FALSE
```

4. I don't know if we have use cases for other collection types, but I had some difficulty adding a new social post. 
The downloaded CSV had empty titles and uploading with empty titles triggered a note that the title field had to be unique. Makes sense, so I added unique titles. When I upload this CSV (below), I get the following error. 
(I get the same thing when I do the same with FAQs)
```
id,title,author,url,published
1,safasd,@iFixit,,TRUE
2,dgdfg,@iFixit,,TRUE
3,;dkdlf,@iFixit,,TRUE
4,kjkdg,@iFixit,,TRUE
5,jjgjg,@iFixit,,TRUE
6,dslkf,@iFixit,,TRUE
7,lksdl,@iFixit,,TRUE
,some title,test author,,FALSE
```
image

5. When I take the FAQs exported CSV, and try to upload it as a banner upload, it appears to validate it and do the upload. I would expect it to say something like the headers aren't correct. It doesn't appear to actually update the banners (they remain unchanged), but it appears successful from the upload view.

dhmacs:
Hey @erinemay, thank you for going through all the edge cases with your review 🙌 Here's the answers to the feedback:

> 1. I'm curious if this feedback is correct.
>    I made a change to 1 item and added 2 new items. Is this feedback just because it updates the whole productlist set based on the new CSV?

Ideally the import CSV should only contain the items that you have edited or added. So whatever there is in the CSV will be considered an update (unless the id is blank, in that case is considered a new item). I'm not doing any strict comparison with what we have in the db to check what's been edited and what not, the plugin just updates based on the data of the CSV.
That said, there shouldn't be any issues uploading the entire thing. If the data is the same, than virtually nothing changes, even though the plugin goes trough each item in the CSV an tries to update it.

> 2. It's setting my newly added productlists `defaultShowAllChildrenOnLgSizes`, `forceNoindex`, and `forceNoindex` to false automatically. Here's the exported CSV after I made the change showing the values as FALSE after I uploaded mine with those fields blank. Just want to make sure this is expected.

If boolean fields are left blank, they will take whatever default value they have in Strapi (usually false)

> 3. A [productlist](https://add-bulk-operations.govinor.com/admin/content-manager/collectionType/api::product-list.product-list/2381?plugins%5Bi18n%5D%5Blocale%5D=en) that I added and set up to be published 404s https://react-commerce-git-add-bulk-operations-ifixit.vercel.app/Parts/test-new-collection-test?disableCacheGets=1. Should an added productlist entry be pubished? Is there something wrong with the data I added? Here're the rows I added.

The reason for the 404 is that the product list is set to be a "parts" product list. Parts product lists require a device title (there's a field for that in Strapi), and that title should basically be the stylized version of the `handle`. Check out this [product list](https://add-bulk-operations.govinor.com/admin/content-manager/collectionType/api::product-list.product-list/2374?plugins[i18n][locale]=en) for reference. Ideally there should also be a corresponding device wiki.
The easiest way to test a product list is to create a "marketing" one, because it doesn't have these requirements. I've just switched yours to marketing, and you can check it here: https://react-commerce-git-add-bulk-operations-ifixit.vercel.app/Shop/test-new-collection-test?disableCacheGets=1 (they're under the `/Shop` prefix)

> 4. I don't know if we have use cases for other collection types, but I had some difficulty adding a new social post.
>    The downloaded CSV had empty titles and uploading with empty titles triggered a note that the title field had to be unique. Makes sense, so I added unique titles. When I upload this CSV (below), I get the following error.
>    (I get the same thing when I do the same with FAQs)

This was affected by the same issue mentioned here https://github.com/iFixit/react-commerce/pull/1801#issuecomment-1633961307. Now I've fixed all collections sequences, so they should all work if you want to try again.

> 5. When I take the FAQs exported CSV, and try to upload it as a banner upload, it appears to validate it and do the upload. I would expect it to say something like the headers aren't correct. It doesn't appear to actually update the banners (they remain unchanged), but it appears successful from the upload view.

We talked about this with @sterlinghirsh during our last call. We could get a little bit more fancy with validation, trying to match the CSV headers against the collection schema, but we can't guarantee anything, because if two collections have fields with the same names, there won't be a way to tell them apart. Keep in mind that the plugin works even if you don't specify all the fields in the schema (the plugin updates only the attributes it finds in the CSV, the others will be left unchanged).

deltuh-vee:
Carry Over QA 🎬

PR #1809, opened 2023-07-06
Vulcan Guides: Limit type to required props

jarstelfox:
We will change the API response so it is much smaller and only contains the required props

qa_req 0

Connects: https://github.com/iFixit/ifixit/issues/48100


ianrohde (approved):
CR ⚪

PR #1812, opened 2023-07-07
Strapi: Fix deploy process (docker build was broken)

danielbeardsley:
We upgraded to node 18 in #1779. This actually broke our strapi build process. I'm not quite sure how that pull passed all the tests if strapi wasn't able to build.

This line was in the failed deploy log:

> prebuild-install warn install No prebuilt binaries found (target=18.16.1 runtime=node arch=x64 libc=musl platform=linux)

It then precedes to try to build `better-sqlite3` and fails. @masonmcelvain discovered that:

> Looks like our version of better-sqlite3 (we're on 7.4.6) doesn't support node 18 prebuilds.
> That's introduced in 7.5.2
> https://github.com/WiseLibs/better-sqlite3/releases/tag/v7.5.2

So, we update that package to 7.5.2. In the process I fixed the nvmrc version format.


jarstelfox (approved):
CR 🌵

danielbeardsley:
After some investigation...

An earlier deployment for the node-18 branch [did fail](https://govinor.com/deployments/4696). I think it left the docker image (can it do that?) in a state that the next build didn't re-attempt that part of the install process (for some yet-unknown reason) and [succeeded](https://govinor.com/deployments/4768).

masonmcelvain (approved):
CR 🙌🏻 Thanks!

danielbeardsley:
QA :+1: Made it through this part of the process:
image

federicobadini:
I've not completely understood the discussion here:

The warning message you referred to simply indicates that no prebuilt library was located:
```
prebuild-install warn install No prebuilt binaries found (target=18.16.1 runtime=node arch=x64 libc=musl platform=linux)
```

Subsequently, the process continued trying to build the package, but it all ended in failure due to non-availability of Python:

```
gyp ERR! stack Error: Could not find any Python installation to use
```
This was not the last commit on the `upgrade-to-node-18` branch though.
A [commit](https://github.com/iFixit/react-commerce/pull/1779/commits/a0725340ab2c38d4f1d6bf00aa8cb69816e7c028) which included the necessary libraries to build the package was issued after this build. Consequently, the following builds were able to pass the installation step. Once all the build step issues were addressed too, a series of successful builds was achieved. The last of them is [this](https://govinor.com/deployments/4768).

The post-merging [issue](https://govinor.com/deployments/4776) seems not related to code but to the inability to locate an intermediate image
```
No such image: sha256:b62dd75cb6d52e61e71e3c74c44b62357aed3843293b508c8b4a4ac8ad313aa1, Code: 1
```
It happens just after 9s which is much less than needed to reach the install step in the docker build process. 

Also notice that: 
- building the last commit of the upgrade-to-node-18 branch with --no-cache on localhost succeeds. 
- building the `test-sqlite-version` branch that revert this PR on Govinor [succeeds](https://govinor.com/deployments/4795). 

I've no visibility on the production logs, but if they are not showing anything different from what is visible in Govinor this seems much more a docker/cache related issue rather than a code issue.

masonmcelvain:
Hey @federicobadini, thanks for looking into this.

Here is the bulk of the Slack conversation, for context: https://ifixit.slack.com/archives/C04TFU3QFD5/p1688764351743119

Here is the [production log file](https://cimpler-logs.s3-us-west-1.amazonaws.com/strapi-deploy/main--f9cc08ede15a4925ad320c69ed1e760aca94fea8--1688664842024.log) we were referencing. This output is produced when [building and pushing](https://github.com/iFixit/react-commerce/blob/main/backend/build-docker-and-push) the docker image.

The error we saw in that log file during the install step was

```
10:35:12 make: Entering directory '/opt/node_modules/better-sqlite3/build'
10:35:12 make: printf: Operation not permitted
10:35:12 make: *** [deps/locate_sqlite3.target.mk:16: b857c92884e9598d609f6be182a2595df7a8e00f.intermediate] Error 127
```

which was a bit obtuse to us. However, we convinced ourselves that we could avoid the sqlite build by bumping the version to make a prebuild available.

I'm not sure what to make of the deploy you linked that couldn't locate an intermediate image. I'm also not super familiar with docker and caching, so I'll defer to @danielbeardsley (though he's on vacation at the moment).

danielbeardsley:
Here was the failing strapi deploy log:
https://cimpler-logs.s3-us-west-1.amazonaws.com/strapi-deploy/main--157016cd185e672200c30de5213a30f2aecd69ac--1688765548703.log

Python is found but the build fails later.
```
14:33:42 gyp info using node@18.16.1 | linux | x64
14:33:42 gyp info find Python using Python version 3.11.4 found at "/usr/bin/python3"
....
14:33:42 make: Entering directory '/opt/node_modules/better-sqlite3/build'
14:33:42 make: printf: Operation not permitted
14:33:42 make: *** [deps/locate_sqlite3.target.mk:16: b857c92884e9598d609f6be182a2595df7a8e00f.intermediate] Error 127
14:33:42 make: Leaving directory '/opt/node_modules/better-sqlite3/build'
```

Bumping this dependency resulting in using the prebuilt binary and thus it [deployed successfully](https://cimpler-logs.s3-us-west-1.amazonaws.com/strapi-deploy/main--547fccd5bdbb097c579b08b0ba1ce5067186291e--1688767165134.log).

federicobadini:
This seems a low level kind of error which might be due to the fact that there are discrepancies between the production instance and the containers that run the tests (or the dev machines).

I notice that the Github actions are always running on ubuntu-latest.
The versions of the pre-installed software, including Docker, are set and maintained by GitHub, and I don't think we can directly control them.
On the contrary the production machine is reasonably pinned to specific versions of docker/OS.

Such differences in Docker versions or the Operating System may introduce inconsistencies in the build process across different instances.

This release note dating back to alpine 3.14 seems worth reading to be sure we are not falling into this
https://wiki.alpinelinux.org/wiki/Release_Notes_for_Alpine_3.14.0#faccessat2

federicobadini:
@danielbeardsley @masonmcelvain what are your thoughts on this?
It seemed promising to me because we were experiencing a printf permission error which seems strange.
Reading that newly introduced syscall could lead to permission errors if the environment is not up to date made me think that this might be the case. 
Even though before we were running on 3.16 (and this issue seems to date back to 3.14) the availability of a pre-built image for 3.16 might have hidden this issue.

danielbeardsley:
I appreciate your investigation into our environment, but I'm not sure what you're suggesting or why.

If there's an additional problem, I think a new issue that describes it is appropriate. As for this problem (missing pre-built binary for a development dependency causes a compilation during install that fails), we've solved it by updating the dependency so a pre-built binary is available.

I don't want to pour more effort into trying to get a build process for a dev-only dependency to work across many environments unless it's hindering progress in some way.

federicobadini:
I'm suggesting to check the reason why this was building correctly on every environment (local dev and CI) but not the production machine. 

Having found a pre-build binary is a sort of escape hatch but the ideal scenario is having consistent build in CI to avoid that a build works in CI and not in production. 
The release notes I posted suggest that an environment not up to date could bring to permission errors like the ones we were seeing. I proposed to check the production machine to verify if our environment is aligned with the requirements of the release notes. 

If not that might explain why we were seeing the error only in production.

danielbeardsley:
>  but not on the production machine.

I'm confused. Strapi failing to build in production is precisely the issue here. The deploy to strapi was failing which is why we made the issue and then made this pull.

In the pull description: "This actually broke our strapi build process.... This line was in the failed deploy log:"

federicobadini:
Perfect, if you can copy paste here the before and after result of `docker version` I think it will suffice to determine if we were falling into that pitfall or not.

Here were the requirements and I think most of these parameters are outputted by the docker version command (just one of these needed to be satisfied to comply)

```
1. runc v1.0.0-rc93
if using Docker's Debian repositories, this is part of containerd.io 1.4.3-2
if using Docker Desktop for Windows or Mac, this is part of Docker Desktop 3.3.0

2. Docker 20.10.0 (which contains [moby commit a181391]
(https://github.com/moby/moby/commit/a18139111d8a203bd211b0861c281ebe77daccd9)) or greater, AND libseccomp 2.4.4 (which contains backported [libseccomp commit 5696c89](https://github.com/seccomp/libseccomp/commit/5696c896409c1feb37eb502df33cf36efb2e8e01)) or greater. In this case, to check if your host libseccomp is faccessat2-compatible, invoke scmp_sys_resolver faccessat2. If 439 is returned, faccessat2 is supported. If -1 is returned, faccessat2 is not supported. Note that if runc is older than v1.0.0-rc93, Docker must still be at least version 20.10.0, regardless of the result of this command.

3. As a workaround, in order to run under old Docker or libseccomp versions, [the moby default seccomp profile](https://github.com/moby/moby/blob/master/profiles/seccomp/default.json) should be downloaded and on line 2, defaultAction changed to SCMP_ACT_TRACE, then --seccomp-profile=default.json can be passed to dockerd, or --security-opt=seccomp=default.json passed to docker create or docker run. This will cause the system calls to return ENOSYS instead of EPERM, allowing the container to fall back to faccessat.
```

federicobadini:
@danielbeardsley just a reminder for tracking down the before and after `docker version`

sterlinghirsh:
Just ran `docker version` on a Strapi machine that was booted 2023-05-11:

```
 docker version
Client:
 Version:           19.03.11
 API version:       1.40
 Go version:        go1.14.3
 Git commit:        42e35e6
 Built:             Sun Jun  7 21:16:58 2020
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          19.03.11
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.14.3
  Git commit:       42e35e6
  Built:            Sun Jun  7 00:00:00 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.3.3
  GitCommit:
 runc:
  Version:          spec: 1.0.2-dev
  GitCommit:
 docker-init:
  Version:          0.18.0
  GitCommit:
```

I'm not 100% sure what you meant by before and after. If you mean after the machine is rebuilt, then I guess we'll have to wait. But it looks like we're good on runc (1.0.2-dev > 1.0.0-rc93) though maybe not containderd (1.3.3 < 1.4.3-2). However, sure enough, our docker version looks like it might be the culprit (19.03.11 < 20.10.0).

federicobadini:
> I'm not 100% sure what you meant by before and after

Just that Danny informed me that they were working on upgrading the OS in production, so I was curious to know if docker or other elements in the system were being impacted too. 

I'm surprised that we are on runc 1.0.2 since the release notes of containerd say that runc 1.0.0rc93 is part of containerd.io 1.4.3-2. Maybe runc was updated separately!
Anyway, considering that even a single docker image layer built in a non compliant env could cause issue to Alpine, I'm pretty positive that we found the root origin of the problem.

danielbeardsley:
Here's what it likely will be after the OS upgrade:

```
▶  docker version                                
Client:                                          
 Version:           20.10.14                     
 API version:       1.41                         
 Go version:        go1.16.15                    
 Git commit:        485636f                      
 Built:             Mon Apr 18 14:09:17 2022     
 OS/Arch:           linux/amd64                  
 Context:           default                      
 Experimental:      true                         
                                                 
Server:                                          
 Engine:                                         
  Version:          20.10.14                     
  API version:      1.41 (minimum version 1.12)  
  Go version:       go1.16.15                    
  Git commit:       e2f740d                      
  Built:            Mon Apr 18 14:09:17 2022     
  OS/Arch:          linux/amd64                  
  Experimental:     false                        
 containerd:                                     
  Version:          1.6.2                        
  GitCommit:                                     
 runc:                                           
  Version:          1.1.0                        
  GitCommit:        eae618f-dirty                
 docker-init:                                    
  Version:          0.19.0                       
  GitCommit:                                     
```

PR #1813, opened 2023-07-10
Add top level error boundary

### LINKED ISSUE #1655: Make an error in a component not result in a blank page
From Post-Mortem #1653

### Problem
A js error client-side in a user-menu component (#1654) caused the entire page to go blank.

### Do This
Start with one react error boundary at the top

masonmcelvain:
Closes #1655

Adds a catch-all error boundary at the layout level that renders a minimal fallback UI. Client errors will still be reported to Sentry.

## Bundle Size

This will increase the bundle size on all pages due to the installation of `react-error-boundary`. Bundlephobia [approximates](https://bundlephobia.com/package/react-error-boundary@4.0.10) it is 1.1 kb gzipped, which lines up with the ~1kb difference we see in the bundle stats.

## CR

Hiding whitespace will make the diff easier to view. The fallback UI is inspired by the [cart drawer fallback ui](https://github.com/iFixit/react-commerce/blob/547fccd5bdbb097c579b08b0ba1ce5067186291e/packages/ui/cart/drawer/CartDrawer.tsx#L120-L143).

## QA

Make sure unhandled client side errors render a page that looks like the following. Make sure the errors show up in the console and are reported to Sentry.

I tested this locally by throwing an error in a random component.

```ts
export function FeaturedProductsSection({...}: FeaturedProductsSectionProps) {
   throw new Error('uh oh');
```

CC @danielbeardsley @dhmacs

### DIFF
--- a/frontend/layouts/default/LayoutErrorBoundary.tsx
+++ b/frontend/layouts/default/LayoutErrorBoundary.tsx
@@ -1,0 +1,50 @@
+import { Alert, AlertDescription, AlertTitle, Center } from '@chakra-ui/react';
+import { FaIcon } from '@ifixit/icons';
+import { faCircleExclamation } from '@fortawesome/pro-solid-svg-icons';
+import * as Sentry from '@sentry/nextjs';
+import { ErrorBoundary } from 'react-error-boundary';
+
+type LayoutErrorBoundaryProps = React.PropsWithChildren;
+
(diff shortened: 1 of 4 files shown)


danielbeardsley (approved):
CR :+1:

deltuh-vee:
QA 🎬 
* Unhandled client side errors render a generic error page
* Errors are showing up in the console and in Sentry

PR #1815, opened 2023-07-11
Migrate to using Verdaccio

ardelato:
## Description
Similar to iFixit/ifixit#48389, we want react-commerce to point to our Verdaccio proxy registry.

### CR Notes:

I added the token to GitHub actions as well as Vercel.

### QA Notes
You can use the token we use in `iFixit/ifixit`.

- [ ] Confirm packages can still be successfully installed
- [ ] Smoke test that local dev still works
- [ ] CI should pass

Closes: https://github.com/iFixit/ifixit/issues/48354


.github/workflows/nextjs_bundle_analysis.yml
```diff
@@ -54,7 +55,7 @@ jobs:
          # Here's the first place where next-bundle-analysis' own script is used
          # This step pulls the raw bundle stats for the current bundle
          - name: Analyze bundle
-           run: npx -p nextjs-bundle-analysis@0.5.0 report
+           run: pnpm --package=nextjs-bundle-analysis@0.5.0 dlx report
```
mlahargou, line 58: What's the reasoning for this change?

mlahargou (approved):
CR 📱 Nice!

.github/workflows/nextjs_bundle_analysis.yml
ardelato: Mainly for consistency. We are running `pnpm install` in an earlier step but we are then using `npm exec` here.

deltuh-vee:
QA 🎬 
* Packages can still be successfully installed
* Local dev still works
* CI is still passing

PR #1817, opened 2023-07-13
Catch localstorage Access Errors

danielcliu-ifixit:
## Overview
This pull does 2 things.
1.) Move localstorage calls into try block.
2.) Replace "deleteIem" with removeItem()

## QA

Not quite how to qa this. I was able to make sure the removeItem() works now on chrome by just loading a product page while logged in. I think the errors in question have have to do with third party cookie blocking, which makes testing it on the test sites hard.

Closes: #1798
qa_req 0

### DIFF
--- a/packages/cart-sdk/package.json
+++ b/packages/cart-sdk/package.json
@@ -7,9 +7,10 @@
    "dependencies": {
       "@ifixit/analytics": "workspace:*",
       "@ifixit/app": "workspace:*",
+      "@ifixit/auth-sdk": "workspace:*",
       "@ifixit/helpers": "workspace:*",
       "@ifixit/ifixit-api-client": "workspace:*",
-      "@ifixit/auth-sdk": "workspace:*",
+      "@ifixit/local-storage": "workspace:*",
       "@ifixit/shopify-storefront-client": "workspace:*"
    },
    "peerDependencies": {
--- a/packages/local-storage/index.tsx
+++ b/packages/local-storage/index.tsx
@@ -1,0 +1,50 @@
+interface SafeStorage {
+   getItem(key: string): string | null;
+   setItem(key: string, value: string): void;
+   removeItem(key: string): void;
+   key(index: number): string | null;
+   clear(): void;
+}
+
+export function useSafeLocalStorage(): SafeStorage {
+   const safeLocalStorage = {
+      setItem: function (key: string, value: string): void {
+         try {
+            localStorage.setItem(key, value);
+         } catch (error) {}
+      },
+
+      getItem: function (key: string): string | null {
+         try {
+            const item = localStorage.getItem(key);
+            if (item === null || item === 'null') {
+               return null;
+            }
+            return item;
+         } catch (error) {
+            return null;
+         }
+      },
+
+      removeItem: function (key: string): void {
+         try {
+            localStorage.removeItem(key);
+         } catch (error) {}
+      },
+
+      key: function (index: number): string | null {
+         try {
+            return localStorage.key(index);
+         } catch (error) {
+            return null;
+         }
+      },
+
+      clear: function (): void {
+         try {
+            localStorage.clear();
+         } catch (error) {}
+      },
+   };
+   return safeLocalStorage;
+}
--- a/packages/ui/hooks/index.ts
+++ b/packages/ui/hooks/index.ts
@@ -1,4 +1,5 @@
 import { useBreakpointValue } from '@chakra-ui/react';
+import { useSafeLocalStorage } from '@ifixit/local-storage';
 import * as React from 'react';
 
 export function useDebounce<Value = any>(value: Value, delay: number): Value {
@@ -147,21 +148,18 @@ export function useExpiringLocalPreference<Data = any>(
    const [data, setData] = React.useState(defaultData);
 
    React.useEffect(() => {
-      try {
-         const serializedData = localStorage.getItem(key);
-         if (serializedData != null) {
-            const data = JSON.parse(serializedData) as ExpiringData;
-            const expiresAt = Number.isInteger(data?.expires)
-               ? data.expires
-               : 0;
-            const validData = validator(data?.value);
-            if (validData !== null && expiresAt && Date.now() < expiresAt) {
-               setData(validData);
-            } else {
-               localStorage.removeItem(key);
-            }
+      const safeLocalStorage = useSafeLocalStorage();
+      const serializedData = safeLocalStorage.getItem(key);
+      if (serializedData != null) {
+         const data = JSON.parse(serializedData) as ExpiringData;
+         const expiresAt = Number.isInteger(data?.expires) ? data.expires : 0;
+         const validData = validator(data?.value);
+         if (validData !== null && expiresAt && Date.now() < expiresAt) {
+            setData(validData);
+         } else {
+            safeLocalStorage.removeItem(key);
          }
-      } catch (error) {}
+      }
    }, []);
 
    const setAndSave = (data: Data) => {
@@ -171,7 +169,8 @@ export function useExpiringLocalPreference<Data = any>(
          expires: Date.now() + expireInDays * 1000 * 86400,
       } as ExpiringData;
       const serializedData = JSON.stringify(expiringData);
-      localStorage.setItem(key, serializedData);
+      const safeLocalStorage = useSafeLocalStorage();
+      safeLocalStorage.setItem(key, serializedData);
    };
 
    return [data, setAndSave];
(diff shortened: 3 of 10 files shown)


packages/ui/hooks/index.ts
```diff
@@ -158,10 +158,10 @@ export function useExpiringLocalPreference<Data = any>(
             if (validData !== null && expiresAt && Date.now() < expiresAt) {
                setData(validData);
             } else {
-               localStorage.deleteIem(key);
```
masonmcelvain, line 161: Dang I can't believe we were using a nonexistent `deleteIem` method and silencing the error.

masonmcelvain (commented):
We could merge this pull and carry on our happy way.

However, we've been aware that the `catch (error) {}` approach could silence real bugs, and this pull turned up one of them.

Assuming we are only worried about browsers that prevent us from using `localStorage`, another solution to this problem might be wrapping `localStorage` with a `useSafeLocalStorage` hook or something, and telling eslint to complain when we use `localStorage` directly.

This would discourage developers from wrapping chunks of code in `try { ... } catch (error) {}`, when we really only need to wrap the localStorage methods.

Something like

```ts
// safeLocalStorage.ts
const safeLocalStorage = localStorage;

safeLocalStorage.getItem = (key: string): string | null => {
   try {
      return localStorage.getItem(key);
   } catch (error) {}
};

export safeLocalStorage;
```

masonmcelvain:
However, that would be more work than addressing this single callsite.

At the very least, I'd like to change this code to only wrap the `getItem` and `removeItem` calls with a `try { ... } catch (error) {}`, to minimize the code we are silencing. It's a little more verbose, but inspires a lot more confidence that our code is doing what we intend it to do.

masonmcelvain:
I also see we have an unwrapped `setItem` in this file. Maybe we should wrap that too?

https://github.com/iFixit/react-commerce/blob/106d971e1a2ae0494c14afe251a3ee7fbef294ba/packages/ui/hooks/index.ts#L174

packages/localStorage/package.json
```diff
@@ -0,0 +1,21 @@
+{
+   "name": "@ifixit/localStorage",
+   "version": "0.0.0",
+   "main": "./index.tsx",
+   "types": "./index.tsx",
+   "license": "MIT",
+   "exports": {
+      ".": "./index.tsx",
+      "./logger": "./logger.ts",
+      "./nextjs": "./nextjs.tsx"
```
masonmcelvain, line 10:
```suggestion
```

packages/localStorage/package.json
masonmcelvain, line 1: I think this `packages/localStorage` directory also needs a tsconfig.json

masonmcelvain: Also, super minor, but if we want to follow the kebab-case naming convention for these packages, we could name the directory `local-storage` or `localstorage` instead of `localStorage`

packages/localStorage/index.tsx
```diff
@@ -0,0 +1,45 @@
+export function useSafeLocalStorage(): Storage {
+   const safeLocalStorage = {} as Storage;
+   safeLocalStorage.setItem = (key: string, value: string): string | null => {
```
masonmcelvain, line 3: I'm pretty sure `setItem` has a `void` return type

packages/localStorage/index.tsx
```diff
@@ -0,0 +1,45 @@
+export function useSafeLocalStorage(): Storage {
+   const safeLocalStorage = {} as Storage;
```
masonmcelvain, line 2: I think the type coercion could mask potential type errors.

```suggestion
   const safeLocalStorage = {};
```

masonmcelvain (changes requested):
CR ✌🏻 Thanks for implementing my suggestion! This is close, I think as long as we get `tsc` to pass we should be fine.

packages/localStorage/package.json
danielcliu-ifixit: i did consider the naming thing originally but thought it would be weird to hyphenate a name for a distinct thing. But maybe youre right it makes more sense to keep the style

danielcliu-ifixit:
Sorry due to the naming change i rebased/squashed the changes, sorry if it is confusing/inconvenient for your review i just thought this would be cleaner

masonmcelvain (approved):
CR ✌🏻 This looks great thanks @danielcliu-ifixit

packages/local-storage/index.tsx
```diff
@@ -0,0 +1,50 @@
+interface SafeStorage {
+   getItem(key: string): string | null;
+   setItem(key: string, value: string): void;
+   removeItem(key: string): void;
+   key(index: number): string | null;
+   clear(): void;
+}
+
+export function useSafeLocalStorage(): SafeStorage {
+   const safeLocalStorage = {
```
danielbeardsley, line 10: I don't see any reason why we need this const to be created in here, it doesn't close over variables, it doesn't use any state. I'd put it at the top level and then just export the object.

masonmcelvain: I suggested this pattern because I thought of JS objects as mutable, making it safer to create a new instance for each caller in case one caller mutates the object (very unlikely, but possible). Thinking about it again though, I'm not sure how es6 modules affect mutability...

TIL that default exports are "copies" and named exports are references (I think?)
https://stackoverflow.com/questions/58437520/mutable-difference-between-es6-named-and-default-exports

packages/cart-sdk/package.json
```diff
@@ -7,9 +7,10 @@
    "dependencies": {
       "@ifixit/analytics": "workspace:*",
       "@ifixit/app": "workspace:*",
+      "@ifixit/auth-sdk": "workspace:*",
       "@ifixit/helpers": "workspace:*",
       "@ifixit/ifixit-api-client": "workspace:*",
-      "@ifixit/auth-sdk": "workspace:*",
+      "@ifixit/local-storage": "workspace:*",
```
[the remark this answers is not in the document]
masonmcelvain: We wanted to use this code in `packages/` and `frontend/`. This could have been added to an existing package (like `helpers` maybe), but I'm not aware of any disadvantages to this approach so I thought this was fine.

packages/local-storage/index.tsx
danielbeardsley: If we're worried about importers messing up the things they import, doesn't that same concern apply to *every* import in the whole ecosystem?

packages/cart-sdk/package.json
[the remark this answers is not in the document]
danielbeardsley: > but I'm not aware of any disadvantages to this approach so I thought this was fine

To evaluate the cost of things like this, I think about expanding a pattern. Would I want every single file to be turned into a package that each has it's own package.json? In my opinion the answer there is no because it's more boilerplate, more futzing with extra files when adding an `import ...` statement (must declare dependencies in two places), and likely slower installs.

I think there is utility in some amount of separation of responsibilities, but in cases like this (a thin safety wrapper around a built-in) I don't think there's any gains in separating it into a new package.

packages/local-storage/index.tsx
danielcliu-ifixit: i guess i could have just returned without creating a variable. Do you want another pull to address this though?

masonmcelvain: Yea I guess it does. Your suggestion seems like a nice simplification

packages/cart-sdk/package.json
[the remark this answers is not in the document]
masonmcelvain: I agree with you, and I don't think I would have made this a new package if I had implemented it. At the same time, there is a cost to another round of CR/QA and I chose to accept this working solution at the time.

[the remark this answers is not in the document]
danielbeardsley: > there is a cost to another round of CR/QA and I chose to accept this working solution at the time.

I agree, let's not open a new pull.

packages/local-storage/index.tsx
danielbeardsley: I think for things like this (effectively a library with no state, no hooks, no capturing, no config) a plain export makes the most sense as it's the most predictable and friendly way of exposing a library and it let's the caller know that there's no complexity under the hood.

That said, I don't think it's worth it to make a new pull to change this.

packages/cart-sdk/package.json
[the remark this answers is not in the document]
danielbeardsley: > is there a disadvantage to adding a new package?

Not very much for a single new package, but I enumerated the reasoning in my [previous comment](https://github.com/iFixit/react-commerce/pull/1817#discussion_r1265722416)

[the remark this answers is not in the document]
danielcliu-ifixit: is there a place where in an existing package that you think a localstorage wrapper logically belongs in? because to me i think having a sensible/readable file hierarchy is valuable too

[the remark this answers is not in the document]
danielbeardsley: I'd probably vote for the `helpers` package, but it doesn't matter that much.

PR #1825, opened 2023-07-17
Troubleshooting: Update tests for new canonical URLs

andyg0808:
This updates an existing test to ensure that it doesn't break when we change what the canonical URLs look like. It also adds a new test to ensure that the Troubleshooting pages redirect to the canonical URL when it doesn't match the page URL. This is important (mildly) because it's the mechanism by which our existing (non-public) Vulcan URLs will continue to end up at the right URLs after we start using the final URLs as the canonical URLs.


jordycosta:
QA 🟢 

The updated and new test are passing during CI 👍 

[See workflow](https://github.com/iFixit/react-commerce/actions/runs/5581601515/jobs/10199924505?pr=1825)

JWg9Shtp6o

PR #1826, opened 2023-07-18
React-commerce: Add startup script

aburke07:
## Overview
The react commerce repo setup is starting to get long. Lets make a script that can handle all of it for a more streamlined dev experience.

## Usage
1. Add your algolia api key to the `algoliaApiKey` variable
2. Run `source startReactCommerce.sh`

## QA
1. Make sure the script successfully runs on both a newly cloned repo and an existing repo.
2. Try out making a new Verdaccio account and logging into an existing account.

### DIFF
--- a/startReactCommerce.sh
+++ b/startReactCommerce.sh
@@ -1,0 +1,80 @@
+#!/bin/bash
+
+git checkout main
+git pull
+git checkout -
+rm -rf node_modules
+nvm use
+npm install -g pnpm@8
+npm install -g yarn
+cp .env.local.example .env.local
+
+## Verdaccio setup
+read -p "Do you have a verdaccio account? (y/n)? [y]: " userInput
+
+# Remove leading/trailing whitespace and convert to lowercase
+userInput=$(echo "${userInput}" | tr '[:upper:]' '[:lower:]' | xargs)
+
+# Check the user's input
+while [[ ! "${userInput}" =~ ^(y|n)$ ]]; do
+    if [[ -z "${userInput}" ]]; then
+        userInput="y"  # Set default value to "yes" if no input is provided
+        break
+    fi
+    echo "Invalid input. Please enter 'y' or 'n'."
+    read -p "Do you have a verdaccio account? (y/n)? [y]: " userInput
+    userInput=$(echo "${userInput}" | tr '[:upper:]' '[:lower:]' | xargs)
+done
+
+# Continue based on the user's input
+if [[ "${userInput}" == "y" ]]; then
+    # Login to exisiting user:
+    pnpm login --registry=https://verdaccio.ubreakit.com    
+else
+    # Add new verdaccio user:
+    pnpm adduser --registry=https://verdaccio.ubreakit.com
+fi
+
+npmrcPath=~/.npmrc
+envFilePath=.env.local
+
+# Extract token from ~/.npmrc
+token=$(cat ~/.npmrc |  grep -o '\/\/verdaccio\.ubreakit\.com\/:_authToken="[^"]*"' | sed 's/\/\/verdaccio\.ubreakit\.com\/:_authToken="//' | sed 's/"$//')
+
+if [ -z "$token" ]; then
+    echo "Token not found in ~/.npmrc"
+    return 1
+fi
+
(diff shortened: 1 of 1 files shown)


ianrohde (approved):
CR ⚪

deltuh-vee:
QA 🎬 
* Script successfully runs on both existing and newly cloned repos
* Verdaccio accounts can be created and logged into

startReactCommerce.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
+
+git checkout main
```
sterlinghirsh, line 3: Seems odd to always force the user to the main branch.

startReactCommerce.sh
```diff
@@ -0,0 +1,80 @@
...
+# Update .env.local with VERDACCIO_AUTH_TOKEN
+awk -v token="\"$token\"" '/^VERDACCIO_AUTH_TOKEN=/{$0="VERDACCIO_AUTH_TOKEN=" token}1' .env.local > .env.local.tmp && mv .env.local.tmp .env.local
+
+echo -e "Token successfully copied to .env.local\n\n"
+
+cp backend/.env.example backend/.env
+
+envFilePath="frontend/.env.local"
+exampleEnvFilePath="frontend/.env.local.example"
+algoliaApiKey="\"12345-your-key-here\""
```
sterlinghirsh, line 58: In general, if you have a field you're supposed to edit in a shell script, I'd recommend putting it at the top since finding a random variable to change in the middle of the file is unintuitive. But that said, since we have interactive questions regarding verdaccio anyway, we might just want to ask interactively (though that could get logged) or use an environment variable. Otherwise this sets up a risk of committing with the change. It might also make switching branches weird (like we do at the top of the file) since we will have a changed file.

startReactCommerce.sh
```diff
@@ -0,0 +1,80 @@
...
+# Check if ALGOLIA_API_KEY value is empty
+if [ -z "$apiKey" ]; then
+    echo "ALGOLIA_API_KEY is empty. Filling in the value..."
+    awk -v placeholder="$algoliaApiKey" '/^ALGOLIA_API_KEY=/{gsub(/=.*/, "=" placeholder)} 1' "$envFilePath" > "$envFilePath.tmp" && mv "$envFilePath.tmp" "$envFilePath"
+fi
+
+pnpm install:all
+
+## We can remove the verdaccio token after exporting it to .env.local
+pnpm logout --registry=https://verdaccio.ubreakit.com
```
sterlinghirsh, line 78: Is there a reason to remove the token? Won't this cause problems for updating later?

startReactCommerce.sh
aburke07: Pull open to fix this: https://github.com/iFixit/react-commerce/pull/1831

startReactCommerce.sh
aburke07: We `git checkout -` to return the user to their old branch afterwards. The point is to make sure their main branch is up to date in case they merge or want to checkout a non-local branch.

startReactCommerce.sh
aburke07: According to the readme:

⚠️ **_If you don't want to use the Verdaccio registry outside of this repo, you can run the following:_**

-  ```
   pnpm logout --registry=https://verdaccio.ubreakit.com
   ```
   -  This will remove the Verdaccio auth token from your `~/.npmrc`, and so long as you've exported the token to `.env.local`, you will still be able to install packages from Verdaccio for this repo.
   
   

We probably don't want to force the verdaccio config on everyone's computer so I think it's better to remove it. We will always export the verdaccio token on startup with this script so I don't think there's a reason so save it.

startReactCommerce.sh
sterlinghirsh: TIL about `git checkout -`.

startReactCommerce.sh
sterlinghirsh: Makes sense though I doubt people will use this script every time they want to run an update. Seems like it would take a very long time to delete node_modules and reinstall everything from scratch. But if it will work for a normal incremental install then it's probably fine.

PR #1827, opened 2023-07-18
Vulcan: Fix data table layouts

ianrohde:
## Issue
Previous Vulcan table layout was geared towards image tables and used flex. We need to handle data tables, and flex is a no-no for this.

## CR/QA
This is a temp change to get the data tables layout under control so we can asses the situation and plan how to deal with displaying both table types well.

Linked to https://github.com/iFixit/ifixit/pull/48817


ianrohde:
dev_block 🛑 I want to spend a little more time to see what else I can do in the interim.

ianrohde:
un_dev_block 🟢

Austin and I discussed tables and the current state of Vulcan v1.0. It looks like the handling of images with captions are the primary concern, with the data tables being much lower on the list. We're good to proceed ([slack thread](https://ifixit.slack.com/archives/CDFNZ4N4A/p1689796070638169))

k0rvusk0r4x:
- Captions on images now look like the "After" on the issue.
- Text outside the caption is no longer lining up with it.
- Images without captions continue to look good.

QA 😺

aburke07 (approved):
CR :alien:

ianrohde:
QA 🟢 carryover, https://github.com/iFixit/react-commerce/pull/1827/commits/3fb0783bcd675620f1ab6c9b6ac0f071b3b41433 was added cleanly

PR #1830, opened 2023-07-19
Fix conditional use of hooks

### LINKED ISSUE #1829: Avoid conditional use of hooks
See https://github.com/iFixit/react-commerce/blob/547fccd5bdbb097c579b08b0ba1ce5067186291e/packages/footer/components/Newsletter.tsx#L87-L102

federicobadini:
closes #1829 

qa_req 0

### DIFF
--- a/packages/footer/components/Newsletter.tsx
+++ b/packages/footer/components/Newsletter.tsx
@@ -84,10 +84,6 @@ const NewsletterForm = ({
    subscription: Subscription;
    subscribe: (email: string) => void;
 }) => {
-   if (isSubscribed) {
-      return null;
-   }
-
    const inputRef = React.useRef<HTMLInputElement>(null);
 
    const onSubscribe = React.useCallback(
@@ -100,6 +96,10 @@ const NewsletterForm = ({
       },
       [subscribe]
    );
+
+   if (isSubscribed) {
+      return null;
+   }
 
    return (
       <Stack


danielbeardsley (approved):
CR :+1: 

Thanks for catching this.

Seems like this could be found by static analysis. When this is executed and the component is rendered with a different number of hooks from the last time it was rendered, what happens? Does React catch this and yell at us in the console? Does React throw an exception and stop rendering entirely?

I'm trying to gauge how important it is to catch this before deploy.

packages/footer/components/Newsletter.tsx
```diff
@@ -101,6 +97,10 @@ const NewsletterForm = ({
       [subscribe]
    );
 
+   if (isSubscribed) {
```
danielbeardsley, line 100: Why bother even passing in this prop though? Is there a reason we don't have the caller do `{!isSubscribed && <Newsletter>}` instead of returning null here?

federicobadini: It can be also done that way.
However there was at least another component that had the same `return null` approach below in the same file (`NewsletterSubscribed`). 
Since this was just a hotfix I didn't changed that logic. Maybe it was done on purpose to keep the NewsletterComponent easier to read but I agree with you that we can also avoid rendering the component at all.

federicobadini:
> Seems like this could be found by static analysis. When this is executed and the component is rendered with a different number of hooks from the last time it was rendered, what happens? Does React catch this and yell at us in the console? Does React throw an exception and stop rendering entirely?

In general React prescribes that hooks are not used conditionally because it relies on their order inside a component to correctly reconcile state. Here is an example of what could otherwise happen: https://legacy.reactjs.org/docs/hooks-rules.html#explanation.

By the way there should be a linter rule to prevent this. 
In the frontend folder the rule is correctly enforced but packages folder file seems not to be checked against this.

PR #1838, opened 2023-07-20
Refactor product list ancestry code

### LINKED ISSUE #1837: Refactor product list ancestors to exclude store home page
The ancestors array in our product list model currently includes the store home page, making it act more like breadcrumbs than true ancestry. This misrepresentation complicates tasks that only require product list ancestors, such as querying for reusable sections (a PR I'm working on now).

I propose to refactor the product list ancestors code to only include relevant product lists, simplifying related coding tasks. The breadcrumbs will be derived from the ancestry in the component (including the store home breadcrumb)

dhmacs:
closes #1837 

## QA

1. Open a [product list preview](https://react-commerce-git-refactor-product-list-ancestors-ifixit.vercel.app/Parts/Android_Phone)
2. Test that product list breadcrumbs work and verify that they should include the store home page breadcrumb


frontend/models/product-list/component/product-list-ancestor.ts
```diff
@@ -0,0 +1,85 @@
...
+
+   const type = productListTypeFromStrapi(attributes.type);
+
+   return ancestors.concat({
+      deviceTitle: attributes.deviceTitle ?? null,
+      title: getProductListTitle({
+         title: attributes.title,
+         type,
+      }),
+      type: productListTypeFromStrapi(attributes.type),
```
masonmcelvain, line 47: It seems like we already compute this above on line 39. I see that it was this way in the previous implementation, but I don't see harm in reusing the variable.

```suggestion
      type,
```

frontend/models/product-list/component/product-list-ancestor.ts
```diff
@@ -0,0 +1,85 @@
...
+      title = 'All';
+      handle = 'Parts';
+   }
+
+   return {
+      deviceTitle: title,
+      title: `${title} Parts`,
+      type,
+      handle,
+   };
```
masonmcelvain, line 84: This may be personal preference, but I find this immutable implementation easier to read.

```suggestion
   if (title === 'Root') {
      return {
         deviceTitle: title,
         title: 'All',
         type: ProductListType.AllParts,
         handle: 'Parts',
      };
   }
   return {
      deviceTitle: title,
      title: `${title} Parts`,
      type: ProductListType.DeviceParts,
      handle: '',
   };
```

masonmcelvain (dismissed):
CR ✌🏻 I think this is right, thank you for simplifying this!

frontend/models/product-list/component/product-list-ancestor.ts
dhmacs: I agree 👍  `title` should be `All Parts` though in this case to match the behaviour

masonmcelvain: Ah good catch thank you

masonmcelvain (approved):
CR ✌🏻

deltuh-vee:
QA 🎬 
Product List breadcrumbs still work and include the store home page breadcrumb.

Issue #1840, opened 2023-07-20
Lighthouse: Render Box if no `href` prop exists


ianrohde:
## Issue
This was flagged on a Lighthouse run for HTML semantics. We are rendering NavTabs without href attributes.

Let's render a Box component when the href URL doesn't exist.

PR #1841, opened 2023-07-20
Lighthouse: Render a Box instead of a Link when no `href` prop exists

ianrohde:
## Issue
This was flagged on a Lighthouse run for HTML semantics. We are rendering NavTabs without href attributes.

## CR/QA
This change renders a `Box` component instead of a `Link` when the href URL doesn't exist.

<details>
<summary>After Screenshot</summary>

Screenshot 2023-07-20 at 2 03 02 PM

</details>

Closes https://github.com/iFixit/react-commerce/issues/1840

### DIFF
--- a/frontend/templates/troubleshooting/index.tsx
+++ b/frontend/templates/troubleshooting/index.tsx
@@ -516,7 +516,7 @@ function NavTabs({
 
          {deviceGuideUrl ? (
             <Link {...notSelectedStyleProps} href={deviceGuideUrl}>
-               Parts
+               Guides
             </Link>
          ) : (
             <Box className="isDisabled" {...notSelectedStyleProps}>
--- a/frontend/templates/troubleshooting/index.tsx
+++ b/frontend/templates/troubleshooting/index.tsx
@@ -504,20 +504,26 @@ function NavTabs({
 
    return (
       <Flex {...props} gap={1.5} height="100%">
-         <Link
-            className={devicePartsUrl ? '' : 'isDisabled'}
-            {...notSelectedStyleProps}
-            href={devicePartsUrl}
-         >
-            Parts
-         </Link>
-         <Link
-            className={deviceGuideUrl ? '' : 'isDisabled'}
-            {...notSelectedStyleProps}
-            href={deviceGuideUrl}
-         >
-            Guides
-         </Link>
+         {devicePartsUrl ? (
+            <Link {...notSelectedStyleProps} href={devicePartsUrl}>
+               Parts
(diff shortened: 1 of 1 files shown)


jordycosta:
Since this removes the `Link` component if no href exists, I think we could drop the _isDisabled_ prop

ianrohde:
isDisabled provides specific styling. There was a conditional that rendered it on the Link, so with that changed, we need to apply it on the Box.

deltuh-vee:
QA 🎬 
When Parts is disabled, we're no longer getting penalized on Lighthouse runs for having an uncrawlable link.

deltuh-vee:
dev_block ✊ 
Parts is appearing twice.
image

frontend/templates/troubleshooting/index.tsx
```diff
@@ -504,20 +504,26 @@ function NavTabs({
...
+            </Link>
+         ) : (
+            <Box className="isDisabled" {...notSelectedStyleProps}>
+               Parts
+            </Box>
+         )}
+
+         {deviceGuideUrl ? (
+            <Link {...notSelectedStyleProps} href={deviceGuideUrl}>
+               Parts
```
jordycosta, line 519:
```suggestion
               Guides
```

ianrohde:
Thanks for catching that!

frontend/templates/troubleshooting/index.tsx
```diff
@@ -504,20 +504,26 @@ function NavTabs({
...
+         {deviceGuideUrl ? (
+            <Link {...notSelectedStyleProps} href={deviceGuideUrl}>
+               Guides
+            </Link>
+         ) : (
+            <Box className="isDisabled" {...notSelectedStyleProps}>
+               Guides
+            </Box>
+         )}
+
```
aburke07, line 526: Completely optional, but we can get rid of the duplicate ternary if we wrap the two links in a react fragment and the two boxes in a react fragment.

aburke07 (approved):
CR :alien: with an optional comment

frontend/templates/troubleshooting/index.tsx
ianrohde: The conditionals target `devicePartsUrl` and `deviceGuideUrl`, so unfortunately it's not that simple. I was going to get fancy and break this out into a function, but in this instance, I just didn't think it was worth it.

deltuh-vee:
QA 🎬 
A Box component is rendered instead of a Link when the href URL doesn't exist for Parts.

PR #1846, opened 2023-07-24
Set swcMinify To False

danielcliu-ifixit:
## Overview

We've had some errors show up on sentry like [this](https://github.com/iFixit/react-commerce/issues/1808) since we updated `next.js`. Based on a github issue comment on the next js repo, turning `swcMinify` to `false` may stop these errors. You can read more about `swc` [here](https://swc.rs/) and [here](https://nextjs.org/docs/architecture/nextjs-compiler) but basically next js started using it to compile and it could be causing this error. Because `next.js` just changed the default from `false` to `true` and now this is changing it back, im not too worried about potential problems with this change, outside of slower build times and the like. Nevertheless, we should keep an eye on sentry when this deploys. 

## QA

Make sure the vercel preview seems fine. 

Connects: https://github.com/iFixit/react-commerce/issues/1808

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -160,6 +160,7 @@ const moduleExports = {
          disableClientWebpackPlugin: true,
       }),
    },
+   swcMinify: false,
 };
 
 // Make sure adding Sentry options is the last code to run before exporting, to


danielbeardsley:
Does this mean it will use whatever minifier it *used to* before using swc by default?

The bundle size is only *slightly* bigger, so this seems fine.

danielbeardsley (approved):
CR :+1:

danielcliu-ifixit:
> Does this mean it will use whatever minifier it _used to_ before using swc by default?

based on [this](https://github.com/iFixit/react-commerce/pull/1846#issuecomment-1648746069) it seems like it falls back to terser, which im assuming it was using before next 13 by default.

federicobadini:
Before merging this I think we should consider upgrading Next.js to the latest version.
As far as it can be seen in the changelog, Swc is being updated continously.
 
I'd just try to bump Next to the latest version before switching off the new default.

jordycosta:
QA 🟢 

The [Vercel preview](https://react-commerce-prod-git-turn-swcminify-off-ifixit.vercel.app/Parts/Android_Phone/Cameras) appears fine 👍

danielcliu-ifixit:
un_deploy_block :v: we tried the upgrade and the error didn't go away

Issue #265, opened 2022-04-21
Enable bulk assignment of banners to product list pages


dhmacs:
During promotions the marketing team needs to add banners to several product lists. 

Examples:
- Add a banner to every product list
- Add a banner to "iphone" product lists

Currently Strapi does not have a native interface for bulk updates.

## To do

Research what's the best option we have here. Probably we need to build a custom plugin that enables that.

cc @jeffsnyder

jeffsnyder:
I'd like to use the same assignment method for banners as we build for FAQs. We will encounter this same category of problem with many other content types. I'd like to develop a single method applicable to multiple content/page types. There is a brief mention of the plan Sterling and I came up with in the FAQ Epic: https://github.com/iFixit/react-commerce/issues/772

PR #1848, opened 2023-07-26
Refactor product list item type overrides

dhmacs:
Another piece of the work being done in #265 

## QA

To QA this let's do a smoke test on product lists with particular focus on how the content of the hero and meta tags is affected when an item type is selected. Try to add an item type override to a product list in Strapi e see how it behaves. The behaviour should be the same as on `main` (you can use https://react-commerce.vercel.app/ and https://main.govinor.com/admin to do the comparison)


frontend/templates/product-list/MetaTags.tsx
```diff
@@ -11,92 +11,141 @@ import {
...
 
 export function MetaTags({ productList }: MetaTagsProps) {
-   const appContext = useAppContext();
-   const currentRefinements = useCurrentRefinements();
+   const itemTypeOverrides = useItemTypeOverrides(productList);
+   const metaTitle = useMetaTitle(productList);
+   const metaDescription =
+      itemTypeOverrides?.metaDescription ??
+      productList.metaDescription ??
+      itemTypeOverrides?.metaDescription ??
```
sterlinghirsh, line 27: I think you were going for this:
```suggestion
      itemTypeOverrides?.description ??
```

dhmacs:
> Before, we were able to do all the item type override replacement in one place, then pass the modified values where they were needed. Now, it looks like we call useItemTypeOverrides in quite a few places, which will run through the override and replacement logic every time if I'm understanding correctly. For example, MetaTags calls useItemTypeOverrides, then immediately calls useMetaTitle which calls useItemTypeOverrides again. This seems kind of wasteful. We could pass the item type overrides into useMetaTitle to save on a call to that, but then why not pass the overrides to MetaTags? Seems like passing a prop in one place is simpler (and more performant?) than adding a hook that gets rerun multiple times. Maybe it's the case that these hooks are very fast, so we don't care so much, but I figured I'd call it out just to check.

I agree with you, lifting up state is a best practice in situations like this, and while refactoring this I though of just doing it that way. The pattern that I ended up using is meant to keep props structure simpler. Whenever I have some data that can be derived, I just redo the computation instead of passing to the components both the source data and the derived data (this is an approach that I learned from Redux https://redux.js.org/style-guide/#keep-state-minimal-and-derive-additional-values). In this approach, when performance is an issue, you can use memoization.
With that said, I don't feel strongly about this, so if you prefer I can use props instead 👍  

> I also liked the idea of keeping all the override logic in one place, so that deeper components didn't have to worry so much about which title, tagline, etc. to use. @danielcliu-ifixit and I came to this design (in which calculateProductListOverrides produces a modified productList with an overrides property) after some iteration, but I actually favored an even more centralized approach. In my approach, we'd produce a modified productList object where title, tagline, etc. would be the modified values. In the cases where we'd want to access the original title, we would still be able to add a field like title_orig or something. I'm not sure which approach is the best balance of DX and performance, but my gut tells me that the override and fallback logic for these fields should be centralized and deduplicated to simplify the deeper components.

I like your approach, it seemed logical to me as well. The only reason this code came to my attention is that I was trying to understand which Strapi fields were being used, and it was not straightforward to get that information. Reading through the `calculateProductListOverrides` function and subroutines, I think the reason for this complexity is that we're trying to do too many things there. Symptoms of this is code like:

```ts
   const { h1, title, ...productListMinusH1 } = productList;
   return (
      metaTitle ||
      getTitle(
         {
            h1: null,
            title: realTitle,
            ...productListMinusH1,
         },
         itemType
      ) + ' | iFixit'
   );
```  

Here we're removing h1 from the product list being passed to `getTitle` because `h1` is supposed to be used only by the page h1, not by the meta title. 

The idea behind the refactor is that I wanted the component to look like this:

```tsx
 <HeroSection
   title={
      itemTypeOverrides?.title ??
      productList.h1 ??
      productList.title
   }
   tagline={
      itemTypeOverrides
         ? itemTypeOverrides.tagline
         : productList.tagline
   }
   description={
      itemTypeOverrides
         ? itemTypeOverrides.description
         : productList.description
   }
/>
```

I think pushing the customizations as closely as possible to the components that are using them, makes it easier to see what's being affected. Following this approach, I've also moved other aspects closer to where they were being used, such as the description being shown only if it's page 1. 

> I don't see the connection between bulk modifications and using hooks.

I'm refactoring the product list model to have a section array that includes all the sections (including the static one such as the hero section). This is to able to sort sections according to Strapi content.
While doing this refactor I've encountered this code about item type overrides on the hero section, so I was trying to understand how to model the hero section. Admittedly I went a little off on a tangent here, and I was a bit torn when deciding whether to take a chance at refactoring this or not, because this refactor is not strictly required. Ultimately I thought it was worth it, but If you prefer the previous version, we can dismiss this PR 👍 

un_dev_block 🏎️ 
deploy_block 🛑  on discussing this change

danielcliu-ifixit:
> Reading through the calculateProductListOverrides function and subroutines, I think the reason for this complexity is that we're trying to do too many things there. Symptoms of this is code like:

while i agree that code may be a bit dense if we aren't using another metatitle elsewhere im not sure if i agree with surfacing the construction. That goes for most of the other values too

I think the question we need to answer is are item type overrides pagewide or are they context specific? Because if we only want the title override in one spot then surfacing the construction, simplifying the override code, and even decentralizing the override creation all make sense. However, if the metatitle is always going to found be checking for the metatitle override then the default metatitle then the title override then the default title, I think we should just do that calculation once. Im not sure why we would want to repeat code for that functionality in multiple places, opening up maintainability overhead for essentially the exact same functionality.

dhmacs:
> I think the question we need to answer is are item type overrides supposed to have precedent pagewide or are they context specific? Because if we only want the title override in one spot then surfacing the construction, simplifying the override code, and even decentralizing the override creation all make sense. However, if the metatitle is always going to found be checking for the metatitle override then the default metatitle then the title override then the default title, I think we should just do that calculation once. Im not sure why we would want to repeat code for that functionality in multiple places, opening up maintainability overhead for essentially the exact same functionality.

I agree that some of these things would rightly be page wide concepts, and ideally would be computed in the server model if Algolia routing wasn't flaky. After all, item type product list are in essence like a separate product list page. 
Our current approach is to send a product list object that basically contains information about both the current product list page and all its item type overrides. Many of these fields, like `h1`, are context specific (only be applies to the `h1` tag). The meta title is only being used in the meta tags component. The description is only being used in the hero. So probably not worth an abstraction.
If these fields were to be used in multiple places, then I'd create small helpers, or hooks (e.g. `useProductListTitle(productList)`, `useProductListMetaTitle(productList)`, etc).

I think there's also a good approach that resembles more `calculateProductListOverrides` (and perhaps is what @sterlinghirsh was thinking) that is to build a hook like this:

```ts
export function useProductListWithItemTypeOverrides(productList: ProductList) {
   const selectedItemType = useDevicePartsItemType(productList);

   if (selectedItemType == null) return productList;

   const replacements = {
      DEVICE: productList.deviceTitle ?? productList.title,
      ITEM: selectedItemType,
   };

   const overrides = productList.itemOverrides[selectedItemType];
   const allItemTypesOverrides = productList.itemOverrides['*'];

   const overrideTitle = replacePlaceholders(
      overrides?.title ?? allItemTypesOverrides?.title,
      replacements
   );

   const overrideMetaTitle = replacePlaceholders(
      overrides?.metaTitle ?? allItemTypesOverrides?.metaTitle,
      replacements
   );

   const overrideDescription = replacePlaceholders(
      overrides?.description ?? allItemTypesOverrides?.description,
      replacements
   );

   const overrideMetaDescription = replacePlaceholders(
      overrides?.metaDescription ?? allItemTypesOverrides?.metaDescription,
      replacements
   );

   const overrideTagline = replacePlaceholders(
      overrides?.tagline ?? allItemTypesOverrides?.tagline,
      replacements
   );

   return {
      ...productList,
      title:
         overrideTitle ??
         getDefaultTitleOverride(productList, selectedItemType),
      metaTitle: overrideMetaTitle ?? productList.metaTitle,
      h1: overrideTitle ?? productList.h1,
      description: overrideDescription,
      metaDescription: overrideMetaDescription,
      tagline: overrideTagline ?? productList.tagline,
   };
}
```

This way the code would still be fairly quick to parse.

danielbeardsley:
> This way the code would still be fairly quick to parse.

I think having different data (post overrides) stored in the same shape (depending on which hook you call) is more confusing when reading the code than having the overrides already applied and an explicit `original_title` attribute.

dhmacs:
> I think having different data (post overrides) stored in the same shape (depending on which hook you call) is more confusing when reading the code than having the overrides already applied and an explicit `original_title` attribute.

@danielbeardsley I'm not sure I understand, can you elaborate on the approach you're proposing? Are you suggesting to directly mutate the `productList` prop?

Just to be clear, the intended use of the hook above is like this:

```tsx
export function ProductListView({
   productList,
   algoliaSSR,
}: ProductListViewProps) {
   const productListWithOverrides = useProductListWithItemTypeOverrides(productList);

   // Want to access override title?
   productListWithOverrides.title

   // Want to access "original" title?
   productList.title
}
```

And the rationale is to treat an item type product list just like any other product list. In this scenario there's no need to add `original` attributes, because you already have the original product list at hand (and likely we won't need those attributes anyway, since they effectively would be data of a different product list).

sterlinghirsh:
After looking it over again, I think I'm fine with the change here. If we work with it and decide we hate it or run into a performance issue, then we can always change it back, but I think it'll be fine.

sterlinghirsh (approved):
CR :zap:

danielcliu-ifixit:
I personally think that the newest iteration 

```php
export function useProductListWithItemTypeOverrides(productList: ProductList) { 
   const selectedItemType = useDevicePartsItemType(productList);

   if (selectedItemType == null) return productList;

   const replacements = {
      DEVICE: productList.deviceTitle ?? productList.title,
      ITEM: selectedItemType,
   };

   const overrides = productList.itemOverrides[selectedItemType];
   const allItemTypesOverrides = productList.itemOverrides['*'];

   const overrideTitle = replacePlaceholders(
      overrides?.title ?? allItemTypesOverrides?.title,
      replacements
   );
   ...
   ```
   
   is the ideal solution

**Comment by erinemay** [comment]
Selecting an item type:
- Hero title and meta title are both updated to show `<device> <item type>` (example /Parts/Mac/Batteries was Mac Batteries or Mac Batteries | iFixit). This is the same as on main.
- Changing between different item type selections and using the back button to remove them or move between works and updates the page title and hero title accordingly
- Mixing item type and other facets still works.

Item type overrides
- The following are all displayed on the page as expected from the override
   - meta title
   - title
   - tagline
   - description

dev_block 🧷 
The following is as-is. No item type override present. 
On this branch, when you add an item type, we lose the tagline.
https://react-commerce-git-refactor-product-list-item-typ-e4b1ab-ifixit.vercel.app/Parts/MacBook_Air/Accessories
image
image

https://react-commerce.vercel.app/Parts/MacBook_Air/Accessories
image
image

/Tools is missing the title 

image
https://react-commerce.vercel.app/Tools
image

**Comment by dhmacs** [comment]
@erinemay 

> On this branch, when you add an item type, we lose the tagline

While fixing this, I checked the spec here: https://github.com/iFixit/react-commerce/issues/589

it says:

```
When another field is empty, it should fall back in this order: Matched Item Type, Blank Item Type, Page Default
```

In this case an item override is not present, which means that the tagline should be present, but the description should be present as well. Right?

cc @ifixitmaxb 

---

> /Tools is missing the title

I cannot reproduce this. If I visit the preview I see the title 🤔 Can you try with `disableCacheGets` query params to see if it's a caching issue?

Screenshot 2023-08-01 at 21 10 14

erinemay:
> I cannot reproduce this. If I visit the preview I see the title 🤔 Can you try with disableCacheGets query params to see if it's a caching issue?

Hmm ok I also can't produce it now. Must be caching. 

So just waiting to see if both the description AND the tagline should be present?

sterlinghirsh:
If there's no item type override specified in Strapi, then an item type page should get the default description and tagline.

ifixitmaxb:
Per Massimo tagging me here, I'm pretty sure I agree with him and Sterling: "the description AND the tagline should be present"

dhmacs:
Ok, implemented according to spec 👍 

I took the chance to also refactor the solution into the item type product list hook as discussed above, so the new solution isn't wasting any additional computation. (cc @danielcliu-ifixit) If we'd like to keep it as before then I can just remove the last commit.

un_dev_block 🏎️ 
un_deploy_block 🟢

erinemay:
dev_block ❌ 
Sorry @dhmacs, I missed this in my last pass:
Breadcrumbs on this branch have the item type attached to the parent-no-item-type breadcrumb. Breadcrumbs look ok on this branch without an item type selected.
https://react-commerce-git-refactor-product-list-item-typ-e4b1ab-ifixit.vercel.app/Parts/MacBook_Air/Accessories
image
https://react-commerce.vercel.app/Parts/MacBook_Air/Accessories
image

The override title is sitting in that spot too:
https://react-commerce-git-refactor-product-list-item-typ-e4b1ab-ifixit.vercel.app/Parts/MacBook_Air/Cables?disableCacheGets=1
image

<details><summary>Override details</summary>

https://refactor-product-list-item-type-overrides.govinor.com/admin/content-manager/collectionType/api::product-list.product-list/2162?plugins[i18n][locale]=en

image

</details>

Everything else looks and works well ✔️ 
Selecting an item type:
- Hero title and meta title are both updated to show `<device> <item type>` (example /Parts/Mac/Batteries was Mac Batteries or Mac Batteries | iFixit). This is the same as on main.
- Changing between different item type selections and using the back button to remove them or move between works and updates the page title and hero title accordingly
- Mixing item type and other facets still works.
- /Tools remains unaffected.
- Tagline and description are both now present when there is an item type selection.

Item type overrides - if there is an item type override, the fields in the override in govinor are displayed on the page for that item type.
- The following are all displayed on the page as expected from the override
   - meta title
   - title
   - tagline
   - description

dhmacs:
My bad @erinemay, I think it's a regression I've introduced in my last commit. We should be good now 

un_dev_block 👍

erinemay:
QA 👍 That fixed it. The breadcrumbs issue https://github.com/iFixit/react-commerce/pull/1848#issuecomment-1664658246 is fixed and everything that was working there is still working now. /Parts and /Shop pages remain unaffected.

sterlinghirsh (approved):
CR :zap:

PR #1860, opened 2023-07-27
Vulcan: Fix device image flash on mobile

jyee27:
Fix the device image showing on first render on mobile on Troubleshooting pages.
Any server-side breakpoints can be incorrect on the first render, causing the image to flash in/out.
Directly adding the media breakpoint with sx fixes this.

Closes: https://github.com/iFixit/ifixit/issues/48975

### DIFF
--- a/frontend/templates/troubleshooting/index.tsx
+++ b/frontend/templates/troubleshooting/index.tsx
@@ -114,21 +114,26 @@ const Wiki: NextPageWithLayout<{
                   borderBottom="1px"
                   borderColor="gray.300"
                >
-                  <Hide below="sm">
-                     <Image
-                        src={mainImageUrl}
-                        onClick={onOpen}
-                        cursor="pointer"
-                        alt={title}
-                        htmlWidth={120}
-                        htmlHeight={90}
-                        objectFit="contain"
-                        borderRadius="md"
-                        outline="1px solid"
-                        outlineColor="gray.300"
-                        marginRight={3}
-                     />
-                  </Hide>
+                  <Image
+                     sx={{
+                        display: 'none',
+                        // sm breakpoint
+                        '@media (min-width: 576px)': {
+                           display: 'block',
+                        },
+                     }}
+                     src={mainImageUrl}
+                     onClick={onOpen}
+                     cursor="pointer"
+                     alt={title}
+                     htmlWidth={120}
+                     htmlHeight={90}
+                     objectFit="contain"
+                     borderRadius="md"
+                     outline="1px solid"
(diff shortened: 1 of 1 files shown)


frontend/templates/troubleshooting/index.tsx
```diff
@@ -114,21 +114,26 @@ const Wiki: NextPageWithLayout<{
...
-                        marginRight={3}
-                     />
-                  </Hide>
+                  <Image
+                     sx={{
+                        display: 'none',
+                        // sm breakpoint
+                        '@media (min-width: 576px)': {
+                           display: 'block',
+                        },
```
jarstelfox, line 123: Could we do:

```
const smBreakpoint = useToken('breakpoints', 'sm');

...

sx={{
                        display: 'none',
                        // sm breakpoint
                        `@media (min-width: smBreakpoint)`: {
                           display: 'block',
                        },
```

This would reduce the use of the token being a one-off magic number

jyee27: We could do this:
```
   const foo: any = {
      display: 'none',
   };
   foo[`@media (min-width: ${smBreakpoint})`] = {
      display: 'block',
   };
```

jarstelfox: I think I lean that direction. It's less readable, which stinks, but it is slightly cleaner code IMO.

jyee27: [a81cd4f](https://github.com/iFixit/react-commerce/pull/1860/commits/a81cd4f359ae3588913099e44189694d04abfe6d)

frontend/templates/troubleshooting/index.tsx
```diff
@@ -84,6 +85,13 @@ const Wiki: NextPageWithLayout<{
       </>
    );
 
+   const imageSx: any = {
```
jarstelfox, line 88: I think you can import `SystemProps` from chakra and that will add the type

jarstelfox (approved):
CR 🌵

frontend/templates/troubleshooting/index.tsx
jyee27: I tried both `SystemProps` and `SystemStyleProps` and got this error:
```
Element implicitly has an 'any' type because expression of type '"@media (min-width: sm)"' can't be used to index type 'SystemProps'.
  Property '@media (min-width: sm)' does not exist on type 'SystemProps'.ts(7053)
```

deltuh-vee:
QA 🎬 
Device image is no longer show flashing on mobile.

PR #1863, opened 2023-07-28
Vulcan Hreflangs: Get rendering

jarstelfox:
The hreflangs component was working as expected. The correct links were rendered,
but when viewing the page no links were actually rendered. It ends up we
are running into an edge case in next/head: vercel/next.js#8384

This commit adds a workaround for this issue which was suggested from the
issue mentioned above.

The fix is to invert the usage such that <Head> is used in your component.
Multiple uses of <Head> will be merged together.

Closes: https://github.com/iFixit/ifixit/issues/48856

### DIFF
--- a/frontend/templates/troubleshooting/index.tsx
+++ b/frontend/templates/troubleshooting/index.tsx
@@ -110,10 +110,8 @@ const Wiki: NextPageWithLayout<{
                flexShrink="1"
                id="main"
             >
-               <Head>
-                  {metadata}
-                  <HreflangUrls urls={wikiData.hreflangUrls} />
-               </Head>
+               <Head>{metadata}</Head>
+               <HreflangUrls urls={wikiData.hreflangUrls} />
                <HStack
                   spacing={0}
                   mt={{ base: 3, sm: 8 }}
@@ -539,11 +537,11 @@ function NavTabs({
 function HreflangUrls({ urls }: { urls: Record<string, string> }) {
    const hreflangs = Object.entries(urls);
    return (
-      <>
+      <Head>
          {hreflangs.map(([lang, url]) => (
             <link rel="alternate" key={lang} hrefLang={lang} href={url} />
          ))}
-      </>
+      </Head>
    );
 }
 
--- a/frontend/templates/troubleshooting/index.tsx
+++ b/frontend/templates/troubleshooting/index.tsx
@@ -100,6 +100,7 @@ const Wiki: NextPageWithLayout<{
                flexShrink="1"
                id="main"
             >
+               <TagManager />
                <Metadata
                   metaDescription={metaDescription}
                   metaKeywords={metaKeywords}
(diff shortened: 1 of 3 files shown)


frontend/templates/troubleshooting/index.tsx
```diff
@@ -539,11 +532,34 @@ function NavTabs({
...
+   canonicalUrl: string;
+}) {
+   return (
+      <Head>
+         <meta name="description" content={metaDescription} />
+         <meta name="title" content={title} />
+         <meta name="keywords" content={metaKeywords} />
+         <meta name="robots" content="index, follow" />,
+         <link rel="canonical" href={canonicalUrl} />
+         <TagManager />
```
dhmacs, line 561: Isn't the same issue happening for this component as well?

frontend/templates/troubleshooting/index.tsx
```diff
@@ -110,10 +100,13 @@ const Wiki: NextPageWithLayout<{
...
-                  {metadata}
-                  <HreflangUrls urls={wikiData.hreflangUrls} />
-               </Head>
+               <Metadata
+                  metaDescription={metaDescription}
+                  metaKeywords={metaKeywords}
+                  canonicalUrl={canonicalUrl}
+                  title={title}
+               />
+               <HreflangUrls urls={wikiData.hreflangUrls} />
```
dhmacs, line 109: What makes hreflang urls not be part of metadata?

jarstelfox: Our intent was to have many small components with a single responsibility. The metadata was originally on `<meta>` tags. I personally am fine keeping these three separate. (TagManager, Metadata, HrefLangs), but if you have opinions please voice them.

frontend/templates/troubleshooting/index.tsx
jarstelfox: Oh shoot, I didn't notice that.

Yes, you are correct. I will fix it shortly.

erinemay:
QA 🦕 
When I got to a [troubleshooting page](https://react-commerce-git-vulcan-hreflangs-fix-ifixit.vercel.app/Troubleshooting/Google_Phone/Google+Pixel+Overheating/479437) that has other language translations of the wiki, I see correct hreflang tags. EN has the troubleshooting URL and the other languages have the appropriate wiki url. The languages present represent all the languages that the wiki has.
image

jyee27 (approved):
CR ☕

frontend/templates/troubleshooting/index.tsx
dhmacs: Not a big deal, primarily I was just curious why they were designed this way. I just thought these components are so simple that they could all fit into a single metadata component. 
I'd be interesting to see if there's enough logic that it'd be useful to share them between templates 🤔

PR #1866, opened 2023-08-01
Replace custom seeding with Strapi native seeding

### LINKED ISSUE #1644: Use Strapi CLI import to seed when server starts
This completes the work started by #878

In #878 we built a new interface to transfer data between Strapi instances using the new Strapi import/export commands. This works manually, and now we need to make it automatic during server bootstrap phase, replacing the older implementation

dhmacs:
closes #1644 

This PR switches the custom seed algorithm with the builtin Strapi import/export feature. 
I've found out that the existing seed algorithm has a bug that breaks the database table id sequence generators, so this PR will fix that as well. 

### Note for the reviewer

The core change is in this file: `backend/src/plugins/addons/server/bootstrap.ts`, everything else pretty much follows from it. I've also made a couple of minor changes to the plugin UI.

qa_req 0


dhmacs:
ℹ️ Status update: there seems to be a Strapi bug that prevents images from being imported, and https://github.com/strapi/strapi/issues/17557 might be related. I'm keeping an eye on the issue and evaluating my next steps to move this forward

dhmacs:
This is no longer need since https://github.com/iFixit/react-commerce/pull/2142 has been merged

PR #1868, opened 2023-08-02
Bump E2E workflow timeout from 15 to 20 minutes

ardelato:
## Description

It looks like we've been barely skirting the 15-minute timeout for a while now, where in some cases the timeout is hit. Let's increase the timeout to 20 minutes to give us a bit more breathing room. https://github.com/iFixit/react-commerce/actions/runs/5741240564/job/15561767476

https://app.datadoghq.com/dashboard/3xd-yax-e2h/github-actions?from_ts=1690993556686&to_ts=1690997156686&live=true
image

qa_req 0

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,7 +5,7 @@ env:
    npm_config_userconfig: './.npmrc'
 jobs:
    playwright-run:
-      timeout-minutes: 15
+      timeout-minutes: 20
       strategy:
          fail-fast: false
          matrix:


mlahargou:
Is there any logical way to split up these tests into different jobs? 15 minutes is getting up there and it may be nicer to have a few different jobs that each take ~5 minutes.

CR 📱

mlahargou (approved):

ardelato:
> Is there any logical way to split up these tests into different jobs? 15 minutes is getting up there and it may be nicer to have a few different jobs that each take ~5 minutes.

Possibly? We could also just drop the `Firefox` run as that seems to be the main offender. I'm creating a follow-up issue for this though.

Issue #1867, opened 2023-08-02
Upgrade Strapi to v4.12.1


masonmcelvain:
Security patch

https://github.com/strapi/strapi/releases/tag/v4.12.1

PR #1869, opened 2023-08-02
chore(deps): bump strapi to `v4.12.1`

masonmcelvain:
Closes https://github.com/iFixit/react-commerce/issues/1867

## QA

Make sure strapi still works and is on the correct version, and smoke test product pages and product list pages.

Try running `pnpm install:all` locally and confirm you don't get a diff.

### DIFF
--- a/backend/docker-compose.yaml
+++ b/backend/docker-compose.yaml
@@ -4,7 +4,7 @@ services:
       build: .
       environment:
          NODE_ENV: production
-         STRAPI_VERSION: 4.11.5
+         STRAPI_VERSION: 4.12.1
          SEED_DB: 'true'
          STRAPI_ADMIN_ENABLE_ADDONS_DANGEROUS_ACTIONS: '${STRAPI_ADMIN_ENABLE_ADDONS_DANGEROUS_ACTIONS}'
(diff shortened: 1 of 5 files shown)


dhmacs (dismissed):
@masonmcelvain can you update also the packages in the addons plugin, as well as the version references in the docker compose? I suggest to perform a project search for the version we're using, to make sure we're replacing it everywhere

dev_block 🔢

masonmcelvain:
Good catch! Will do

masonmcelvain:
un_dev_block ✌🏻 There are now no results for `git grep "4\.11\.5"`

dhmacs (approved):
Thanks for the PR!

CR 👍

deltuh-vee:
QA 🎬 
* Strapi works and is on the correct version
* Product Pages and Product List Pages still work
* Running`pnpm install:all` doesn't produce a diff

I think there's something broken with the Strapi deploy cause I see this when visiting /Parts from the preview link.
image

masonmcelvain:
CR ✌🏻 carry forward. I'll carry forward QA assuming the strapi deployment succeeds

PR #1879, opened 2023-08-03
Do not bundle algolia if not needed

federicobadini:
Only bundle Algolia if the template requires it.
(Saves ~30kb gzipped in non algolia templates)

### QA

1. Visit Vercel preview
2. Verify that Algolia templates are still working correctly (navigation, browser back button etc)


sterlinghirsh (approved):
CR :zap:

deltuh-vee:
Are those page size increases anything to worry about?

danielbeardsley:
> Are those page size increases anything to worry about, even though the global bundle size decreased?

dev_block :+1: 

Yes. I think "bundle size" is a mostly useless metric as it means effectively the total of all the javascript used across all pages. The individual pages are more interesting and the numbers here seem to go against the point of the pull so I think it's worth investigating.

frontend/components/common/AppProviders.tsx
```diff
@@ -9,10 +9,17 @@ import {
...
 import * as React from 'react';
-import { AlgoliaProps, InstantSearchProvider } from './InstantSearchProvider';
+import type {
+   AlgoliaProps,
+   InstantSearchProviderProps,
+} from './InstantSearchProvider';
 
 const customTheme = extendTheme(theme);
+const InstantSearchProvider = dynamic<InstantSearchProviderProps>(() =>
+   import('./InstantSearchProvider').then((mod) => mod.InstantSearchProvider)
```
danielbeardsley, line 21: It seems like an import at run time will result in a longer time before the asset beings downloadin. *This file* has to be downloaded and ran before `./InstantSearchProvider` is even requested.

I'd be more inclined to have an additional component (an optional wrapper) here (outside `AppProviders` or inside it) that only includes `InstantSearchProvider`.

Actually.. since `<InstantSearchProvider>` is really simple, let's not put it in a new component, but merely use it directly on the pages that need it. Thoughts?

federicobadini:
I think the next bundle analysis report above was a little bit misinterpreted.
What is indicated as global bundle size is the javascript bundle that loads alongside every page. 
So it is the most important metric because it impacts all pages.

I usually base my analysis on the Nextjs build output since, being measured by the framework itself, should be the most precise metric. Their terminology is much clearer: `First Load JS shared by all`.

That being said here I've compared the main branch with the current branch stats (taken from vercel build).
I even expected zero gain for pages that included Algolia - there is no reason to suspect a ~14/16kB gain on those.
The real gains are on pages that did not needed Algolia to work (~30 kB).

Route (pages) | First Load JS kB (main) | First Load JS kB (branch) | Savings
-- | -- | -- | --
┌   /_app | 370 | 338 | 32
├ λ /[...slug] | 397 | 365 | 32
├ ○ /404 | 388 | 355 | 33
├ λ /api/nextjs | 370 | 338 | 32
├ λ /api/nextjs/cache/product | 370 | 338 | 32
├ λ /api/nextjs/cache/product-list | 370 | 338 | 32
├ ● /CommunityNext | 376 | 343 | 33
├ λ /Parts | 430 | 414 | 16
├ λ /Parts/[...deviceHandleItemType] | 430 | 414 | 16
├ λ /products/[handle] | 430 | 399 | 31
├ λ /Shop/[handle] | 430 | 414 | 16
├ λ /Tools | 430 | 414 | 16
├ λ /Tools/[handle] | 430 | 414 | 16
├ λ /Troubleshooting/[device]/[problem]/[wikiid] | 403 | 371 | 32
└ λ /Vulcan/[wikiname] | 403 | 371 | 32
-- | -- | -- | --
First Load JS shared by all |   |   |  
├ chunks/framework-2bc61a9deeafa95a.js |   |   |  
├ chunks/main-a476c930690fa3dd.js |   |   |  
├ chunks/pages/_app-5119a48b3bc8f6d2.js |   |   |  
├ chunks/webpack-b004282169e5ec8b.js |   |   |  
└ css/011d829062655983.css |   |   |

frontend/components/common/AppProviders.tsx
federicobadini: Actually bundles (also dynamically linked ones) are linked via script tag in the page html at SSR time. 
So the client immediately sees the entire list of chunks required by a specific page.
There shouldn't be the need to parse the code and then decide if an additional js chunk needs to be downloaded or not

danielbeardsley: > There shouldn't be the need to parse the code and then decide if an additional js chunk needs to be downloaded or not

Maybe I'm missing something, but if a chunk isn't conditionally downloaded, then what makes the import "dynamic"? And how would making it dynamic and conditionally including (as you've done here) reduce the size of the downloaded JS?

danielbeardsley:
> What is indicated as global bundle size is the javascript bundle that loads alongside every page. So it is the most important metric because it impacts all pages.

I see. Seems like it'd be more useful to show the true delta on each page instead of requiring readers to add up two deltas.

> That being said here I've compared the main branch with the current branch stats (taken from vercel build).

I'm glad you did that. Hmm.. Two things make me question my assumptions / reduce my trust in this:

1. There are reported "first load" sizes for pages that have no front-end component or bundle (`api/*`)
2. Bundle size decreased for pages I wouldn't have expected it to
   - Perhaps this is correct in that some JS is deferred, but then it means it will be loaded *later* and thus incur additional time before the second requests are made?
   
That said, here's a waterfall of all JS requests on this branch. Note they all start at the same instant except for the one analytics.js, so in practice I don't see any dependency chains:
image

frontend/components/common/AppProviders.tsx
federicobadini: Next.js dynamic import are some sort of code splitting hints.
At compile time dynamically imported dependencies are bundled as separate chunks.
At runtime, the imported module is evaluated on the server side. 
If the dynamic component is conditionally loaded based on a falsey condition, it won't be included in the initial bundle.
Otherwise if the condition in truthy, it is included along with the initial server-side rendered HTML.

federicobadini:
> There are reported "first load" sizes for pages that have no front-end component or bundle (`api/*`)

I think the default analyser is treating every single file under pages as a proper "page" in terms of the internal build process, and thus may include it in that analysis, even when these server-side files indeed don't contribute to the client-side bundle.

> Bundle size decreased for pages I wouldn't have expected it to

Dynamic imports in Next.js might depend on variables that may only be known at runtime. As such, it isn't possible to definitively predict whether some dependencies will be included in the bundle before analyzing the specific request.
Regarding /Tools and /Parts pages, you'll notice an increase in size smaller than the decrease observed in _app. This can be explained upon inspecting their respective bundles. Both pages statically import some code related to Algolia, which inevitably increases their bundle size. However, a portion of Algolia's dependencies is imported dynamically, and can't be incorporated by default in these stats.

When you load these pages in the browser, you might even observe a slight increase in the transferred JavaScript due to the additional code splitting. But this minor increase is a small trade-off considering the overall kilobytes saved on other pages that do not use Algolia.

danielbeardsley:
@federicobadini -- Thank you for your thoughtful and detailed responses!

> When you load these pages in the browser, you might even observe a slight increase in the transferred JavaScript due to the additional code splitting. But this minor increase is a small trade-off considering the overall kilobytes saved on other pages that do not use Algolia.

I agree.

> As such, it isn't possible to definitively predict whether some dependencies will be included in the bundle before analyzing the specific request.

This makes sense. Does this now mean that there is some javascript (about 16kb worth) that is no longer represented in the bundle size analysis (because it's conditionally included)? Thus future changes to the conditionally included chunks will now be invisible to bundle analysis entirely?

danielbeardsley:
I certainly like this direction (don't bundle algolia stuff if it isn't needed on the page), but I'm wondering if it'd be better to be more explicit about it instead of using a conditional import:

> I'd be more inclined to have an additional component (an optional wrapper) here (outside AppProviders or inside it) that only includes InstantSearchProvider.

Thoughts?

danielbeardsley (approved):
CR :+1: 

Nice!

federicobadini:
un_dev_block ✌🏻

erinemay:
QA 🦖 
Looks good:
- Applying filters and item types
- Stepping forward and backwards via forward/back buttons, collection buttons, and breadcrumbs with and without filters and item types
- Clearing filters

I basically did as many combinations as I could think of using the filters/facets and item types navigating up and down the tree. I looked at the URL structure, titles on the page, and filters in the filter bar. Each action was the same as on master.

PR #1882, opened 2023-08-03
Refactor: DisableCacheGets move to cache control middleware

jarstelfox:
I will want to expand on when the cache is disabled to aid with https://github.com/iFixit/ifixit/issues/48965. However, I noticed the caching behavior was controlled in multiple places. This pull moves the caching to a single middleware.

QA
---

See: https://github.com/iFixit/react-commerce/pull/1882#issuecomment-1664680136


jarstelfox:
QA 🌵 

I checked a url without DSC has the correct cache control:
image

The same URL with DSC has cache-control 0:

image

ianrohde (approved):
CR ⚪

PR #1883, opened 2023-08-03
Revisionid param: Set to no cache if value is HEAD

jarstelfox:
This related back to https://github.com/iFixit/ifixit/issues/48965#issue-1823307688

We will want to send editors to this url to bypass vercel's caching
and view the content with the edits immediately.

QA
---

Same as https://github.com/iFixit/react-commerce/pull/1882#issuecomment-1664680136 But ensure `?revisionid=HEAD` also bypasses the cache only on troubleshooting pages

Connects: https://github.com/iFixit/ifixit/issues/48965

### DIFF
--- a/frontend/helpers/cache-control-helpers.ts
+++ b/frontend/helpers/cache-control-helpers.ts
@@ -30,7 +30,7 @@ export function hasDisableCacheGets(context: GetServerSidePropsContext) {
 }
 
 function getCacheString(options: CacheControlOptions) {
-   if ('disabled' in options && options.disabled) {
+   if ('disabled' in options) {
       return CACHE_CONTROL_DISABLED;
    }
 
--- a/frontend/helpers/cache-control-helpers.ts
+++ b/frontend/helpers/cache-control-helpers.ts
@@ -30,7 +30,7 @@ export function hasDisableCacheGets(context: GetServerSidePropsContext) {
 }
 
 function getCacheString(options: CacheControlOptions) {
-   if ('forceDisabled' in options && options.forceDisabled) {
+   if ('disabled' in options && options.disabled) {
       return CACHE_CONTROL_DISABLED;
    }
 
--- a/frontend/helpers/cache-control-helpers.ts
+++ b/frontend/helpers/cache-control-helpers.ts
@@ -16,8 +16,14 @@ export function hasDisableCacheGets(context: GetServerSidePropsContext) {
       return true;
    }
 
-   const disableCacheGets = context.query.disableCacheGets !== undefined;
-   return disableCacheGets;
+   const { disableCacheGets, revisionid } = context.query;
+
+   if (disableCacheGets !== undefined) {
+      return true;
+   }
+
+   const wantsHeadRevision = revisionid?.toString().toUpperCase() === 'HEAD';
+   return wantsHeadRevision;
 }
 
 function getCacheString(options: CacheControlOptions) {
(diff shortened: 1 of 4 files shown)


ianrohde (dismissed):
CR ⚪

frontend/helpers/cache-control-helpers.ts
```diff
@@ -16,8 +16,14 @@ export function hasDisableCacheGets(context: GetServerSidePropsContext) {
...
 
-   const disableCacheGets = context.query.disableCacheGets !== undefined;
-   return disableCacheGets;
+   const { disableCacheGets, revisionid } = context.query;
+
+   if (disableCacheGets !== undefined) {
+      return true;
+   }
+
+   const wantsHeadRevision = revisionid?.toString().toUpperCase() === 'HEAD';
```
sctice-ifixit, line 25: If we can do it, I think it would be better to _not_ centralize this logic and instead add a "hook" when setting up caching in the controller. We don't want to add this behavior to every route that sets up caching.

jarstelfox: I wasn't sure if opt-in or always made the most sense. Easy enough to change.

jarstelfox:
dev_block 🌵 on https://github.com/iFixit/react-commerce/pull/1883#discussion_r1283783982

sctice-ifixit:
dev_block :hand: on considering making the cache behavior more configurable in the "controller" so that we can localize the logic to where it's relevant

jarstelfox:
un_dev_block 🌵 I like the pattern I set out here to allow more configuration of the caching

frontend/helpers/cache-control-helpers.ts
```diff
@@ -3,11 +3,21 @@ import { GetServerSidePropsMiddleware } from '@lib/next-middleware';
...
 
+interface DisabledCacheControlOptions {
+   disabled: true;
+}
+
+type CacheControlOptions =
+   | EnabledCacheControlOptions
+   | DisabledCacheControlOptions;
+
+type withCacheProps = GetCacheControlOptions | CacheControlOptions;
```
danielbeardsley, line 19: I know it's not mandated, but I think we generally `TitleCase` all types.

frontend/helpers/cache-control-helpers.ts
```diff
@@ -3,11 +3,21 @@ import { GetServerSidePropsMiddleware } from '@lib/next-middleware';
 import { Duration } from '../lib/duration';
 import { GetServerSidePropsContext } from 'next';
 
-interface CacheControlOptions {
+interface EnabledCacheControlOptions {
    sMaxAge: number;
    staleWhileRevalidate: number;
 }
 
+interface DisabledCacheControlOptions {
```
danielbeardsley, line 11: Maybe this should be a const instead of a type? Not entirely sure that'd work out, but since it's prescriptive in its key and value, seems like a const.

danielbeardsley (approved):
CR :+1: 

Nice work @jarstelfox !

frontend/helpers/cache-control-helpers.ts
jarstelfox: Yea, I took a look at using `false` or an object like `const obj = {disabled: true}` and it felt worse to use with the type system then this odd type. I am inclined to leave it and we can take a stab at changing it later.

deltuh-vee:
dev_block ✊ 
If `?revisionid=HEAD` isn't set, should the cache be bypassed?
image

jarstelfox:
@deltuh-vee ate you testing with the prod version? The dev link will always be cache disabled

deltuh-vee:
QA 🎬 
* `?revisionid=HEAD` bypasses the cache on just troubleshooting pages
* `?disableCacheGets` still works

PR #1888, opened 2023-08-07
Fix Govinor clean up and re-deployment by running Strapi as non-root

### LINKED ISSUE #1887: Fix Strapi docker permissions
Currently, Strapi docker container is configured to run as root. This leads to a conflict when a Strapi instance performs any file-system change (e.g., bulk exporting files, uploading an image, etc). The conflict arises because the Govinor platform does not run as root, which consequently restricts its ability to operate on the deployment directory whenever such changes are made by Strapi.

## Do this

Run Strapi as a non-root user

dhmacs:
closes #1887 

qa_req 0


djmetzle:
I took a look at the [launch-script](https://github.com/iFixit/server-templates/blob/master/specific/strapi/startup-scripts/launch-strapi) and don't see anything likely to break. I think we can update, and if anything breaks on machine launch, we can fix it forward.

dhmacs:
@danielbeardsley feel free to merge it if you have time to keep an eye that production deploys smoothly 👍

danielbeardsley (approved):
CR :+1:

PR #1891, opened 2023-08-07
Prevent errors from being cached

masonmcelvain:
Closes https://github.com/iFixit/ifixit/issues/49029

This instructs browsers to not cache 404s and 500s. I think it will also tell cloudfront not to cache 404s for static assets. SSR pages shouldn't be cached by cloudfront anymore https://github.com/iFixit/ifixit/pull/49130).

## QA

Make sure errors still get reported to Sentry. You can test this by running the project locally and throwing an error somewhere in the application code.

We could also check any product, product list, or troubleshooting route that 404s on the previews and make sure the correct `Cache-Control` header is set on the response. (I think this will work based on the [docs](https://nextjs.org/docs/pages/building-your-application/routing/custom-error#more-advanced-error-page-customizing), but perhaps we will have to run the project locally in production mode for this to work correctly).

> pages/_error.js is only used in production. In development you’ll get an error with the call stack to know where the error originated from.

### DIFF
--- a/frontend/helpers/cache-control-helpers.ts
+++ b/frontend/helpers/cache-control-helpers.ts
@@ -1,7 +1,7 @@
 import { CACHE_DISABLED } from '@config/env';
 import { GetServerSidePropsMiddleware } from '@lib/next-middleware';
 import { Duration } from '../lib/duration';
-import { GetServerSidePropsContext } from 'next';
+import { GetServerSidePropsContext, NextPageContext } from 'next';
 
 interface EnabledCacheControlOptions {
    sMaxAge: number;
@@ -21,7 +21,9 @@ type WithCacheProps = GetCacheControlOptions | CacheControlOptions;
 export const CACHE_CONTROL_DISABLED =
    'no-store, no-cache, must-revalidate, stale-if-error=0';
 
-export function hasDisableCacheGets(context: GetServerSidePropsContext) {
+export function hasDisableCacheGets(
+   context: GetServerSidePropsContext | NextPageContext
+) {
    if (CACHE_DISABLED) {
       return true;
    }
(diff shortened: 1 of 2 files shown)


djmetzle (dismissed):
Very nice. CR :no_entry:

frontend/pages/_error.tsx
```diff
@@ -27,6 +28,8 @@ const CustomErrorComponent: NextPage<ErrorProps> = (props) => {
 };
 
 CustomErrorComponent.getInitialProps = async (contextData) => {
+   contextData.res?.setHeader('Cache-Control', CACHE_CONTROL_DISABLED);
```
jarstelfox, line 31: It might be nice to updated `withCache` to be able to handle this case and just wrap this function.

jarstelfox (dismissed):
CR 🌵

frontend/pages/_error.tsx
masonmcelvain: It would be nice to keep all the cache control logic in one place.

It might be tricky to reuse `withCache` here because it is designed to work with `getServerSideProps`, not `getInitialProps`. I didn't want to go down the rabbit hole of abstracting the `GetServerSidePropsMiddleware` types. But if you have any suggestions I'm all ears

jarstelfox: I was thinking it might be easy to make two separate functions but yea I took a very quick stab at it and agree this is easy for now.

djmetzle: Hmm. If it's possible, it might be worth trying our best to consolidate the caching logic _as much as possible_. I suspect that keeping the consts for `no-cache, no-store`, et alia, near the caching middleware is a good breadcrumb. That'd be nice to let that module evolve, even if we can't use the same exact HoC.

masonmcelvain:
Hm I just noticed that https://www.ifixit.com/products/asdf already has `cache-control: private, no-cache, no-store, max-age=0, must-revalidate`. Were we ever able to reproduce a cache-control header on an error response that had caching enabled? I'm reading through the slack thread and I can't find anything definitive. cc @danielbeardsley

danielbeardsley:
No, but I assumed that 500s had different behavior. Though we should probably experiment.

jarstelfox:
@masonmcelvain I pushed a WIP commit that should get us mostly the way there to shared code. Feel free to use it or force-push it off.

masonmcelvain:
I ran `main` in production mode locally and caused a 500 by throwing an error in `getServerSideProps` for the product template. In this scenario, caching _is_ enabled by the Cache-Control header. I'm not sure if this experiment is sufficient, but we can test this feature against it to see if it makes a difference.

image

frontend/helpers/cache-control-helpers.ts
```diff
@@ -18,10 +18,12 @@ type CacheControlOptions =
 
 type WithCacheProps = GetCacheControlOptions | CacheControlOptions;
 
-const CACHE_CONTROL_DISABLED =
+export const CACHE_CONTROL_DISABLED =
```
masonmcelvain, line 21:
```suggestion
const CACHE_CONTROL_DISABLED =
```

masonmcelvain:
Thanks @jarstelfox those changes look great!

frontend/pages/_error.tsx
```diff
@@ -26,13 +26,16 @@ const CustomErrorComponent: NextPage<ErrorProps> = (props) => {
...
+      await Sentry.captureUnderscoreErrorException(contextData);
 
-   // This will contain the status code of the response
-   return NextErrorComponent.getInitialProps(contextData);
-};
+      // This will contain the status code of the response
+      return NextErrorComponent.getInitialProps(contextData);
+   },
+   { disabled: true }
+);
```
jarstelfox, line 39: Super minor: Thoughts on swapping the params for readability?

```suggestion
CustomErrorComponent.getInitialProps = withInitialCacheValue<ErrorProps>(
   { disabled: true },
   async (contextData) => {
      // In case this is running in a serverless function, await this in order to give Sentry
      // time to send the error before the lambda exits
      await Sentry.captureUnderscoreErrorException(contextData);

      // This will contain the status code of the response
      return NextErrorComponent.getInitialProps(contextData);
   }
);
```

jarstelfox: > Sure! I might rename `func` to be more descriptive too if you don't mind.

Totally go for it!

PS I my actual hope here was to do:

```
export default withInitialCacheValue<ErrorProps>(MyCacheValue, CustomErrorComponent)
```

But I didn't get that far last night.

masonmcelvain:
I'm pretty happy as-is, but if you want to refactor this more @jarstelfox go ahead! (In response to your suggest above https://github.com/iFixit/react-commerce/pull/1891#discussion_r1287447106 and on slack)

> Also feel free to add a named export like:
> `export CacheDisabledOption = { disabled: true };`

jarstelfox (approved):
CR 🌵

PR #1900, opened 2023-08-09
iFixit API: Force no caching

jarstelfox:
OMG Next overrides fetch and defaults to caching:
https://nextjs.org/docs/app/api-reference/functions/fetch

Worst of all, they drop many of the configuration values.

As far as Danny and I could tell we would want the mdn value of "default" but Next does not support this value.

We basically get "on or off" and "For how long?" as a global cache even between deployments on vercel.

Lastly we may need to discuss overriding fetch globally in next to avoid this problem in other places.

CC @danielbeardsley

### DIFF
--- a/packages/ifixit-api-client/index.tsx
+++ b/packages/ifixit-api-client/index.tsx
@@ -85,7 +85,7 @@ export class IFixitAPIClient {
                credentials: 'include',
                ...init,
                headers: headers,
-               cache: 'no-cache',
+               cache: 'no-store',
             })
       );
       if (!response.ok) {
--- a/packages/ifixit-api-client/index.tsx
+++ b/packages/ifixit-api-client/index.tsx
@@ -85,6 +85,7 @@ export class IFixitAPIClient {
                credentials: 'include',
                ...init,
                headers: headers,
+               cache: 'no-cache',
             })
       );
       if (!response.ok) {


dhmacs:
Hey @jarstelfox, I'm searching for confirmation in the docs, but I don't think this affects us currently. The fetch cache should only apply to React Server Components (so when using the app router). We're still using pages router, so it should not be a concern for us.

Screenshot 2023-08-09 at 09 04 11

dhmacs:
@jarstelfox Here's another piece in the doc where they mention under which conditions caching is done:

Screenshot 2023-08-09 at 12 33 14

Source: https://nextjs.org/docs/app/building-your-application/caching#data-cache

I've also put together a demo to test this out, check it here: https://next-i18n-demo-gamma.vercel.app/pages-fetch

In the "pages" folder implementation the fetch call is never cached (the time updates on every page refresh).
In the app router the behaviour is different, as the first fetch request is cached (as you can see by the fact that time 1 is always the same). Interestingly if a fetch request is made after the first one, it's not cached.

From my understanding, the `fetch` override happens only within a React tree and is done by React itself, hence why it affects only React Server components and not the legacy stuff.

jarstelfox:
@dhmacs / @danielbeardsley I wonder if we are using https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#fetchcache which is the same basic behavior, but for the legacy "page routing" as far as I understand it.

danielbeardsley:
image

I think it's crazy to make this soo complicated. Especially since they are opting for a non-standard ignore-the-cache-control-header operation. The conditions under which this behavior is the new default seems to be very convoluted. I'd much rather opt in to *not replacing the default behavior* everywhere if possible.

dhmacs:
> @dhmacs / @danielbeardsley I wonder if we are using https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#fetchcache which is the same basic behavior, but for the legacy "page routing" as far as I understand it.

@jarstelfox Where does it say that this is a pages router option?

packages/ifixit-api-client/index.tsx
```diff
@@ -85,6 +85,7 @@ export class IFixitAPIClient {
                credentials: 'include',
                ...init,
                headers: headers,
+               cache: 'no-cache',
```
sterlinghirsh, line 88: Is this supposed to be no-store?
```suggestion
               cache: 'no-store',
```

jarstelfox:
Maybe this isn't the fix:


Visited https://react-commerce-prod-git-fetch-bust-caching-ifixit.vercel.app/Troubleshooting/Acer_Chromebook/Not+Charging/514224?revisionid=HEAD

Then https://www.ifixit.com/Wiki/Edit/Acer_Chromebook_Not_Charging and [made an edit](https://www.ifixit.com/Wiki/History/Acer_Chromebook_Not_Charging?revidA=1068505&revidB=1069994)

Then refreshed https://react-commerce-prod-git-fetch-bust-caching-ifixit.vercel.app/Troubleshooting/Acer_Chromebook/Not+Charging/514224?revisionid=HEAD

I am still seeing the outdated revision.

Response Headers:
```
Age:
0
Cache-Control:
no-store, no-cache, must-revalidate, stale-if-error=0
Content-Encoding:
br
Content-Type:
text/html; charset=utf-8
Date:
Wed, 09 Aug 2023 22:50:12 GMT
Etag:
W/"hh0ebxz2zt59md"
Server:
Vercel
Strict-Transport-Security:
max-age=63072000; includeSubDomains; preload
X-Matched-Path:
/en-US/Troubleshooting/[device]/[problem]/[wikiid]
X-Powered-By:
Next.js
X-Robots-Tag:
noindex, nofollow
X-Vercel-Cache:
MISS
X-Vercel-Id:
sfo1::iad1::bblzv-1691621412018-c3e696e4ea39
```

The ifixit API is updated: ` curl https://www.ifixit.com/api/2.0/Troubleshooting/wikiid/514224 | grep TEST` does find the revision

danielbeardsley:
> I am still seeing the outdated revision.

Was that before or after the pushed commit was deployed? It looks like it was *after* but it's a bit hard to tell in retrospect.

I'd vote for adding extensive logging (body and headers) to these requests (on this branch) and trying again.

jarstelfox:
> Was that before or after the pushed commit was deployed? It looks like it was _after_ but it's a bit hard to tell in retrospect.

It was after 👍🏻

jarstelfox:
dev_block 🌵

jarstelfox:
1) @dhmacs you are correct and this setting is not the bug. We can close this issue.

2) The actual bug was found via

`ROLES=lb cap prod shell` then `varnishlog -q "ReqURL ~ '/api/2.0/Troubleshooting/wikiid/514224'"`

3) The bug: This branch does not have https://github.com/iFixit/react-commerce/pull/1899 which correctly bypasses varnish.

We found that was the issue: We were not bypassing varnish.

My testing was with a logged in user which automatically bypasses varnish even with some testing from Thunder Client in VSCode.

I am going to close this as well as claim https://github.com/iFixit/react-commerce/pull/1899 fixes the edit bugs we have seen

PR #1901, opened 2023-08-09
Robots Header: Stop no-indexing in prod

jarstelfox:
We think the recent proxying change that sends *all* client headers through to the origin may have resulted in the user-agent no longer being `Amazon Cloudfront`. Thus we are now no-indexing all of prod :-(

### DIFF
--- a/frontend/helpers/next-helpers.ts
+++ b/frontend/helpers/next-helpers.ts
@@ -3,7 +3,7 @@ import { setSentryPageContext } from '@ifixit/sentry';
 import { withTiming } from '@ifixit/helpers';
 import type { GetServerSidePropsMiddleware } from '@lib/next-middleware';
 import * as Sentry from '@sentry/nextjs';
-import { GetServerSidePropsContext } from 'next';
+import { GetServerSidePropsContext, NextPageContext } from 'next';
 
 export const withLogging: GetServerSidePropsMiddleware = (next) => {
    return withTiming(`server_side_props`, async (context) => {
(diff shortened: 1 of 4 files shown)


danielbeardsley (dismissed):
CR :+1: 

Let's follow this up with a fix. I'll work with @jarstelfox on this.

danielbeardsley (dismissed):
CR :100:

danielbeardsley:
Confirmed, prod currently is 100% no-indexed. So, this will reverse that. It will also index domains we *don't* want to index, but we'll fix that shortly.

frontend/helpers/next-helpers.ts
```diff
@@ -24,16 +24,42 @@ export const withLogging: GetServerSidePropsMiddleware = (next) => {
...
+      context.res?.setHeader('X-Robots-Tag', restrictRobots);
+   }
+}
+
+export function noindexDevDomains(context: ContextType) {
+   if (context.req.headers['user-agent'] !== PROD_USER_AGENT) {
+      // return RestrictRobots.RESTRICT_ALL;
+   }
+
+   return undefined;
```
danielbeardsley, line 64:
```suggestion
   return RestricRobot.ALLOW_ALL;
```

jarstelfox:
CC @danny https://vercel.com/guides/are-vercel-preview-deployment-indexed-by-search-engines

It looks like that no index is vercel handling previews themselves

danielbeardsley:
un_dev_block :+1:

frontend/helpers/next-helpers.ts
danielbeardsley: Doesn't that apply to this change regardless of which way this goes (i.e. it needs to line up?)

jarstelfox: Fair point. That's why I started refactoring it. The hope was we could have a more general way of getting the indexing string out of a shared function so both the headers and meta tags can reused the code.

danielbeardsley:
> do you have thoughts on QA?

I think curling localhost will be an effective QA (doesn't have vercel preview headers).

jarstelfox:
```
(base)  JARREDS-XPS-13-9310  developer 10:39:39 (no-index-fix) ~/Documents/react-commerce 
▶   curl -i -s http://localhost:3000/products/spudger | grep -i x-robots
(base)  JARREDS-XPS-13-9310  developer 10:39:57 (no-index-fix) ~/Documents/react-commerce E:1
▶  
```

QA 🌵 No x-robots header is sent in local dev

danielbeardsley (approved):
CR :+1:

davidrans:
This deployment error seems relevant, though I don't see that error in the diff:

https://vercel.com/ifixit/react-commerce-prod/82zQjjP5QV2XzjV3AawPDUVs1y1Z
> Type error: Cannot find name 'RestricRobot'. Did you mean 'RestrictRobots'?

jarstelfox:
@davidrans It was relevant before my force push to walk back that commit: https://github.com/iFixit/react-commerce/compare/38fa30621927ee8891205a3150a6c03cfe33f937..ececf31978e8219f52da30608bb23e336278a1ac

PR #1903, opened 2023-08-09
No Index: re-add for all non-cloudfront requests

danielbeardsley:
Recently we foudn that we were no-indexing *everything* from next.js accidentally. We quickly removed that entirely:
(https://github.com/iFixit/react-commerce/pull/1901)

Now we re-add the fixed functionality.

Specifically, we want to no-index all responses that come from requests that *are not* being proxied through cloudfront. In other words, we only want www.ifixit.com/* responses to be indexed.

In the case of Troubleshooting pages, this directive conflicts with the html metatag but we've found evidence that google will listen tot he more restricive directive (noindex in this case).


jarstelfox (approved):
CR 🌵

danielbeardsley:
QA :+1: 

image

PR #1905, opened 2023-08-10
noindex all responses

danielbeardsley:
@jarstelfox and I discovered this while testing something else and came up with this solution.

The short of it is, since vercel caches responses based on url only (seemingly), we can't have it changing the response based on a header. Currently, we have code here to add a "X-Robots-Tag: noindex" header if there are Cloudfront headers on a request. This works some of the time but we end up caching (and serving) responses generated for the same url, but different sets of headers.

Effectively, the requests from CF and the the requests directly to https://react-commerce-prod.vercel.app share an http cache in vercels CDN. Because of this, we need to not change our vercel responses based on headers as they aren't part of the cache key.

The most straightforward way to disallow indexing on all requests except for those coming through CF is to always add the no-index directive and expicitly have CF remove it from all responses it deals with.

See https://github.com/iFixit/ifixit/pull/49259

### DIFF
--- a/frontend/helpers/next-helpers.ts
+++ b/frontend/helpers/next-helpers.ts
@@ -2,7 +2,6 @@ import { setSentryPageContext } from '@ifixit/sentry';
 import { withTiming } from '@ifixit/helpers';
 import type { GetServerSidePropsMiddleware } from '@lib/next-middleware';
 import * as Sentry from '@sentry/nextjs';
-import { GetServerSidePropsContext } from 'next';
 
 export const withLogging: GetServerSidePropsMiddleware = (next) => {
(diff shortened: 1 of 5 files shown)


jarstelfox (approved):
CR 🌵

erinemay:
How can we test that this is working?

danielbeardsley:
Curling the preview url seems to give me the correct header on all requests:

image

QA :+1:

danielbeardsley:
un_deploy_block :+1: 
Other pull is merged and applied.

danielbeardsley:
I deployed this and @masonmcelvain and I validated that it's currently working as expected with caching / cloudfront / preview urls.

PR #1906, opened 2023-08-10
Add IntlDate component

### LINKED ISSUE #1904: Fix troubleshooting template date hydration error
When you visit a troubleshooting page (e.g. `/Troubleshooting/Acer_Chromebook/Not+Charging/514224?revisionid=HEAD`) during development, you see the following error:

Screenshot 2023-08-10 at 09 36 25

This happens because there's an hydration mismatch between the server date locale format, and the browser date locale format.

dhmacs:
closes #1904 

## QA

1. Open a [troubleshooting preview page](https://react-commerce-git-fix-troubleshooting-date-ssr-bug-ifixit.vercel.app/Troubleshooting/Acer_Chromebook/Not+Charging/514224?revisionid=HEAD)
2. Verify that the last updated at date still renders

### DIFF
--- a/frontend/components/ui/IntlDate.tsx
+++ b/frontend/components/ui/IntlDate.tsx
@@ -1,7 +1,7 @@
 /**
  * This component renders a localized date in a way that is compatible with React hydration.
- * The first render will use a non localized format which is the same on the server and client.
- * The second render will use the localized format.
+ * The first render will not render the date to avoid FOUC.
+ * The second render (which happens in the client) will use the localized format.
  *
  * There are other ways to do this, for example by taking advantage of the Accept-Language header
  * to know the user's locale in the server. However this approach is the one that is more cache friendly.
@@ -21,9 +21,9 @@ export function IntlDate({
    options,
    className,
 }: IntlDateProps) {
-   const [stableFormattedDate, setStableFormattedDate] = useState(
-      value.toDateString()
-   );
+   const [stableFormattedDate, setStableFormattedDate] = useState<
+      string | null
+   >(null);
    const isoString = value.toISOString();
    const formattedDate = value.toLocaleDateString(locales, options);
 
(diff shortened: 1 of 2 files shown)


dhmacs:
> There is a brief flash of the day of the week being shown and then going away. There also seems to be a date difference

@jordycosta Unfortunately this is expected. If this bothers us we could just hide the date until the client is hydrated. Would you prefer that?

erinemay:
> > There is a brief flash of the day of the week being shown and then going away. There also seems to be a date difference
> 
> @jordycosta Unfortunately this is expected. If this bothers us we could just hide the date until the client is hydrated. Would you prefer that?

Would that appear like the recording, but not show the initial date flash? And show the final rendered date around the same time it updates in that recording?

dhmacs:
@erinemay 
> Would that appear like the recording, but not show the initial date flash? And show the final rendered date around the same time it updates in that recording?

Yes, I can do it and if you don't like it we revert back to this if you want

sterlinghirsh (dismissed):
CR :zap: I was curious about trying an approach like this. Not a huge fan of the visual change on page load but it's a tradeoff. We could probably get away with just using the US date format or matching the date format to the view language. But for now I'm down to try it and if we decide the flash is too annoying we can revert it. There's a similar concern for localizing currency display.

sterlinghirsh:
I want to make sure the date difference is expected before we unblock this.

dhmacs:
> I want to make sure the date difference is expected before we unblock this.

Yes, it's due to timezone differences. I've implemented a version that avoids FOUC. Let me know if you prefer this

un_ deploy_block 🏎️ 

cc @sterlinghirsh @erinemay @jordycosta

jordycosta:
QA 🟢 

I think this is much better 👍 

firefox_VhSRtHxjnD

sterlinghirsh (approved):
CR :zap:

PR #1907, opened 2023-08-10
Add generated types

dhmacs:
In #1869 we upgraded Strapi and apparently some slight api changes have been introduced that trigger the sdk code generation. 

This quick PR is to push these changes into the repo. I've also taken the chance to declare zod schema for unhandled scalars so the next dev cli will not keep showing warnings about unhandled scalars. I'm using `z.unknown()` to preserve the current behaviour.

qa_req 0


masonmcelvain (approved):
CR ✌🏻 Thanks!

PR #1908, opened 2023-08-10
Add reusable sections

dhmacs:
closes #265 
closes #1746 

## Problem

To summarize the tasks the lead to this PR: The marketing team needs to be able to apply the same unit of content (i.e. page section) across different pages. For example marketing team wants to create the testimonials section once, and apply it in several pages. To do that currently they would have to insert the same data over and over again, for every page.

## Solution

This PR introduces reusable sections, that are sections that can be reused and applied across multiple pages. In this first version reusable sections can only be applied to product list template pages. They can be placed at the top, after the products section, or at the bottom. The also have an associated "priority" field that can be used to show a reusable section before the others. The higher the priority, the higher the section will show up within its group (top, after products, bottom).

The first 4 supported reusable sections are:

- Banners
Screenshot 2023-08-17 at 15 55 09
- Split with image
Screenshot 2023-08-17 at 16 01 17
- Quote gallery
Screenshot 2023-08-17 at 16 02 52
- Press quotes
Screenshot 2023-08-17 at 16 03 19

## QA

1. Open the [strapi preview](https://add-reusable-sections.govinor.com/admin/content-manager/collectionType/api::reusable-section.reusable-section?page=1&pageSize=10&sort=title:ASC) instance for this branch
2. Open also the [next.js preview](https://react-commerce-git-add-reusable-sections-ifixit.vercel.app/Parts) for this branch
3. Check how the reusable sections are configured, and verify that the next.js preview displays them according to the Strapi data

⚠️ Bundle size changes are mainly due to quote gallery and press quotes that use the `Slider` component

cc @jeffsnyder @sterlinghirsh

PR #1912, opened 2023-08-14
Add reusable sections to Strapi

dhmacs:
This PR ships the Strapi backend changes of #1908 so when we ship the frontend the Strapi API will be already available.

qa_req 0


sterlinghirsh (approved):
CR :zap:

PR #1916, opened 2023-08-14
HeroSection: Show tagline and description on page 2+ with image

rickisXP:
Closes https://github.com/iFixit/ifixit/issues/48955

This change includes:
- Shows the Hero tagline and description on pages 2+ when a backgroundImage is also shown
- If there isn't a backgroundImage, Hero tagline and description won't be shown on pages 2+
- Using non-breaking spaces between the ` - Page <num>`

### DIFF
--- a/frontend/templates/product-list/sections/HeroSection.tsx
+++ b/frontend/templates/product-list/sections/HeroSection.tsx
@@ -127,7 +127,7 @@ function HeroTitle({
    children,
    page,
 }: React.PropsWithChildren<{ page: number }>) {
-   // Non-breaking spaces added for all spaces in ` - Page <num>`
+   // Place non-breaking space between 'Page' and page number
    return (
       <Heading
          as="h1"
@@ -137,7 +137,15 @@ function HeroTitle({
          data-testid="hero-title"
       >
          {children}
-         {page > 1 ? ` - Page ${page}` : ''}
+         {page > 1 ? (
+            <>
+               {' - Page'}
+               <span dangerouslySetInnerHTML={{ __html: '&nbsp;' }} />
+               {page}
+            </>
+         ) : (
+            ''
+         )}
       </Heading>
    );
 }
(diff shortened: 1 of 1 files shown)


jordycosta:
QA 🟢 

The tagline/description of the hero section is present past page 1 👍 

### Screen Capture:

firefox_zw5VZ6Imaa

---

deploy_block 🟨 on a question

Should the Page # be on the next line? I noticed in [the issue](https://github.com/iFixit/ifixit/issues/48955) it was breaking line in between the Page and #, but now it is breaking line after the -

I think that this still looks good, but wanted to confirm that this was expected

rickisXP:
> Should the Page # be on the next line? I noticed in [the issue](https://github.com/iFixit/ifixit/issues/48955) it was breaking line in between the Page and #, but now it is breaking line after the -

This is fine. We wanted to add a non-breaking space between the page and number to prevent cases where the `Page <#>` could break in the middle, causing the 'Page' to be on one line and the '<#>' to be on the next.

un_deploy_block :star:

frontend/templates/product-list/sections/HeroSection.tsx
```diff
@@ -140,7 +137,15 @@ function HeroTitle({
          data-testid="hero-title"
       >
          {children}
-         {page > 1 ? ` - Page ${page}` : ''}
+         {page > 1 ? (
+            <>
+               {' - Page'}
+               <span dangerouslySetInnerHTML={{ __html: '&nbsp;' }} />
```
aburke07, line 143: If we just don't want this to wrap at small sizes, can we just increase the width of the parent element?

aburke07 (approved):
CR :alien: with a comment

aburke07:
Dev_block :alien: on above

frontend/templates/product-list/sections/HeroSection.tsx
rickisXP: I'm sure we could, but I think the motivation to use a non-breaking space here is so that the 'Page' and page number are not split off into separate lines.

PR #1919, opened 2023-08-15
Product page: Compatibility notes api update

masonmcelvain:
Connects https://github.com/iFixit/ifixit/issues/49081

Now that the iFixit product api only returns an array of compatibility note device strings, we can narrow our types.

## QA

Make sure the compatibility notes section works the same as on `main`.

Here is a product that has compatibility notes: https://www.cominor.com/products/ice-bucket-assembly-da97-14504c

There are many products without them

### DIFF
--- a/frontend/lib/ifixit-api/productData.ts
+++ b/frontend/lib/ifixit-api/productData.ts
@@ -5,7 +5,7 @@ export type ProductDataApiResponse = {
    variantOptions: {
       [o_code: string]: string[];
    };
-   compatibilityNotes: string | string[];
+   compatibilityNotes: string[];
 };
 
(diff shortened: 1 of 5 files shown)


jordycosta:
QA 🟢 

The compatibility notes looks the same compared to [this branch](https://github.com/iFixit/react-commerce/tree/product-page-compatibility-api-update) and [`main`](https://github.com/iFixit/react-commerce/tree/main) 👍 

firefox_8yantyrMu2

---

### Additional Notes:

> I went through and confirmed the other 297 compatible devices are the same

hcutrone (approved):
CR :burrito:

PR #1920, opened 2023-08-16
FAQ fix: Render markdown

masonmcelvain:
Slack conversation: https://ifixit.slack.com/archives/C01GMQXJC95/p1692141525169339

This change correctly renders markdown from Strapi FAQ answers on featured product pages.

## Bundle size diff

snarkdown is about 1kb and we import it on the product page in this change, so the +1kb bundle size diff on product pages makes sense.

## QA

Add some markdown syntax to a strapi faq on https://product-page-faq-markdown.govinor.com/admin/content-manager/collectionType/api::faq.faq

Link the faq to a featured product on https://product-page-faq-markdown.govinor.com/admin/content-manager/collectionType/api::product.product

When you visit the featured product page on this preview, you should notice that the markdown text is rendered correctly when the appropriate faq is expanded.

### DIFF
--- a/frontend/components/sections/FAQsSection.tsx
+++ b/frontend/components/sections/FAQsSection.tsx
@@ -2,16 +2,16 @@ import {
    Accordion,
    AccordionButton,
    AccordionItem,
-   AccordionItemProps,
    AccordionPanel,
    Box,
-   forwardRef,
 } from '@chakra-ui/react';
 import { faMinusCircle, faPlusCircle } from '@fortawesome/pro-solid-svg-icons';
+import { markdownToHTML } from '@helpers/ui-helpers';
 import { isPresent } from '@ifixit/helpers';
 import { FaIcon } from '@ifixit/icons';
 import { Wrapper } from '@ifixit/ui';
 import type { FAQ } from '@models/components/faq';
+import { useMemo } from 'react';
 import { SectionDescription } from './SectionDescription';
 import { SectionHeading } from './SectionHeading';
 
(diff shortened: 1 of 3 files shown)


sterlinghirsh (approved):
CR :zap: QA :zap: Thanks for the quick fix.

frontend/components/sections/FAQsSection.tsx
```diff
@@ -53,67 +53,52 @@ export function FAQsSection({
...
-                                    >
-                                       {faq.answer}
-                                    </AccordionPanel>
-                                 </>
-                              );
-                           }}
-                        </FAQAccordionItem>
-                     );
-                  })}
+                  {faqs.map((faq, index) => FAQAccordionItem({ faq, index }))}
```
danielbeardsley, line 56: I don't remember if this syntax is better than using jsx? I think `<FAQAccordionItem key={index} .../>` is a bit better cause it puts the "key" prop right here in the loop, rather than having to pass it into the function.

dhmacs: I think @danielbeardsley is right, using components as functions is definitely not a conventional approach in React

danielbeardsley: Technically, this is no different than a typical map function (which is just a function that returns a component). The main difference is that the function body is outside of the scope and thus can't access any other local vars.

masonmcelvain: Thanks for pointing this out, we rushed this one out the door. I'll open a fix PR 👍🏻

masonmcelvain: https://github.com/iFixit/react-commerce/pull/1922

PR #1923, opened 2023-08-17
Vulcan: Add TOC (Desktop/ Mobile) and Scroll Indicator

jarstelfox:
Some choices made:

* TOC items are stored via their name
	* This means no duplicate names are currently allowed, the last one wins
	* We may want to take a second pass and force them to require ids and for them to be unique, but we found the API may not return ids for all headings right now

CR
---

There was a lot of back and forth after about the TOC logic after 38c2ab6b7540f92bd87d67358d5357d18ce1a6f7.

I would CR commit by commit up to there then start looking at the new files TOC / TOCContext as greenfield and look at the overall diff.

 
QA
---
* Ensure TOC is **off** on the `react-commerce` deployment
* Ensure TOC is **on** on the `react-commerce-prod` deployment
* Ensure TOC behaves well on all screen sizes
* Ensure TOC matches [figma](https://www.figma.com/file/jRG4EyQXRDMGfyLqut08KP/iFixit---Working-files?node-id=13998%3A321530&mode=dev) and has behavior like the [example app](https://rwddn4.csb.app/)
* Here is a [good example of small many toc items](https://react-commerce-prod-git-scroll-add-ui-ifixit.vercel.app/Troubleshooting/Chainsaw/Overheating/484453)
* Here is [a more normal example](https://react-commerce-prod-git-scroll-add-ui-ifixit.vercel.app/Troubleshooting/Television/TV+Has+Sound+But+No+Picture/493422)
* Ensure the Desktop TOC is rendered server-side
	* No flash in
* Ensure the Mobile TOC will pop in once scrolled to
	* This will pop in on refresh
	* Ensure no page shift / CLS occurs
* Ensure Desktop TOC activates the correct item on page load
	* Ensure no URL hash, scroll down and refresh the page so you load at the same scroll position
* Ensure Scroll indicator /TOC are removed if we entirely scroll past content (Mobile mostly, but code doesn't care)
* Ensure selecting an item on the TOC scrolls you to it and that item is selected
	* The code cares about scroll positioning only to highlight so this is checking the math to where to scroll is correct
* Ensure selecting an item is the TOC highlights it briefly
	* Not part of the spec, but I liked it
* Ensure the TOC never selects and item off screen
* Ensure the TOC is scrollable if required
	* Ensure a scroll fade occurs on desktop / mobile TOCs when the TOC is scrollable
	* Ensure no fade if not scrollable
* Ensure the TOC selects the active item as you scroll down the page
* Ensure every possible TOC item is selected when scrolling up / down a page
* Ensure clicking a TOC item updates the URL hash (and browser history) to the new item
	* Test history by pressing the back button

Connects: https://github.com/iFixit/ifixit/issues/49284


djmetzle (approved):
Looks great @jarstelfox, awesome work! CR :scroll:

kwiens:
Awesome, very exciting!

@mmarcon91 What do you think of the double TOC when the list is long?
image

deltuh-vee:
dev_block ✊ 
I found it very difficult to scroll through the Mobile TOC. I can only go a couple of items at a time. This is on a physical version of the iPhone SE 2022.

https://github.com/iFixit/react-commerce/assets/5289715/6558d01d-25c2-420a-916f-c09ba54c590a

jarstelfox:
@deltuh-vee I had seen this myself too. It ends up it's a chakra bug: https://github.com/chakra-ui/chakra-ui/issues/7325

I tried to upgrade chakra, but it seemed to break some other components. I will try again tomorrow

mmarcon91:
> Awesome, very exciting!
> 
> @mmarcon91 What do you think of the double TOC on desktop when the list is long? image

In this case it looks like duplication, on desktop I would probaibly remove it

https://www.figma.com/file/jRG4EyQXRDMGfyLqut08KP/iFixit---Working-files?type=design&node-id=13996-318155&mode=design&t=Jg970l1E8rwIHZjT-4

ianrohde:
> I found it very difficult to scroll through the Mobile TOC. I can only go a couple of items at a time. This is on a physical version of the iPhone SE 2022.

Since this is a Chakra issue, can we increase the height of the dropdown as a stop-gap, so the scrolling requirement is reduced?

jarstelfox:
Okay, I have not made any changes to fix the UI, but I spent much time today standing up an entire feature flag system that behaves well on both the client and server.

Here's how it works:

As a dev you [append this enum](https://github.com/iFixit/react-commerce/blob/9a8439fe5e93cd016e32326486557ea8fa5dc7f9/frontend/components/common/Feature.tsx#L12-L14).

This will auto-magically create a new toggle here: http://localhost:3000/dev/features

This toggle takes effect immediately and some SSR pages may complain until you refresh.

The feature state is set in a cookie with the same name.

We either have the cookie (and it's set to true) or we remove it (being nice on cleanup).

We also automatically listen to `?<valueOfEnum>`.

EX: `table-of-contents` is the enum key, `?table-of-contents` does turn it on.

We also allow `?table-of-contents={true|false|0|1}` to turn it off.

Both client and server log the state of features/changes.

danielbeardsley:
> We either have the cookie (and it's set to true) or we remove it (being nice on cleanup).
> Both client and server log the state of features/changes.

Will it be a problem that Cookies don't make it through to the backend? Or.. it's possible they may now if we're sending all the headers, but since the cookie is not part of the cache key in vercel's CDN cache, there may be trouble.

Now that I think about it, I think we started sending request headers through to vercel. Changing our responses based on headers (including Cookie) will cause trouble if we're also caching those pages. I don't think we can realistically force ourselves to only look at headers for urls that we know won't be cached.

I'm excited about a feature flag system though!

jarstelfox:
> Now that I think about it, I think we started sending request headers through to vercel. Changing our responses based on headers (including Cookie) will cause trouble if we're also caching those pages. I don't think we can realistically force ourselves to only look at headers for urls that we know won't be cached.

Good point on the caching front. I am not sure how to best handle that.

Right now, my main concern was for deving / QA passes until a feature is ready.

Currently, the code effectively defaults all features to off if the cookie is missing.

I think the concern would be someone hitting dev page in prod and then setting the cookie/response to a bad version.

djmetzle:
We're not currently forwarding Cookies to the Vercel origin:
https://github.com/iFixit/ifixit/blob/e5d37e303d8487ecdead353f35f2f1ff45c6d33e/terraform/cloudfront/prod/nextjs-dynamic-policies.tf#L8-L10

jarstelfox:
Oh idea💡!

we already have to inspect the ssr context why don’t we just always no-store if any feature cookie is set?

jarstelfox:
dev_block 🌵

frontend/components/common/FlexScrollGradient.tsx
```diff
@@ -0,0 +1,243 @@
...
+function getScrollDirection(el: HTMLElement): FlexScrollDirection {
+   const { flexDirection } = window.getComputedStyle(el);
+   if (flexDirection === 'column') {
+      return FlexScrollDirection.COLUMN;
+   } else {
+      return FlexScrollDirection.ROW;
+   }
+}
+
+function getGradientsBaseStyle(
```
danielbeardsley, line 45: It's hard to tell exactly what this component does...  but it feels very complex for what seemingly amounts to making a gradient slightly bigger or smaller depending on the browser width. I don't doubt that it looks cool but it strikes me as a lot of complexity (and thus the maintenance) for what it buys us.

frontend/templates/troubleshooting/index.tsx
```diff
@@ -95,125 +124,153 @@ const Wiki: NextPageWithLayout<{
...
+                        <ModalContent
+                           width="auto"
+                           maxWidth="calc(100% - 64px)"
+                           background="none"
+                        >
+                           <VisuallyHidden>
+                              <ModalHeader>{title}</ModalHeader>
+                           </VisuallyHidden>
+                           <ModalCloseButton />
+                           <VisuallyHidden></VisuallyHidden>
```
danielbeardsley, line 203: I assume this empty "hidden" component was unintentional?

jarstelfox: @andyg0808?

I did not add this. The diff only shows this because I added a container and the code shifted inward with whitespace.

frontend/components/common/FlexScrollGradient.tsx
danielbeardsley: > This is copied from code in the monorepo.

Ok. Though that doesn't refute the point I was trying to make. Potentially makes it worse as we have at least two copies of this complexity.

> Although, I hear you, we have this implemented in a few different ways in the mono repo and all versions behave differently

[:trollface:](https://xkcd.com/927/)

> I can port the full test suite for this component across as well. If that helps?

I'd rather not. It doesn't help my concerns of complexity.

jarstelfox:
un_dev_block 🌵

frontend/templates/troubleshooting/index.tsx
```diff
@@ -57,9 +57,23 @@ import { HeadingSelfLink } from './components/HeadingSelfLink';
...
+} from './toc';
 import { ViewStats } from '@components/common/ViewStats';
 import { IntlDate } from '@components/ui/IntlDate';
 
+const RelatedProblemsTitle = 'Related Problems';
+
+const FlaggedTOC = onlyShowIfTOCFlagEnabled(TOC);
+const FlaggedScrollPercent = onlyShowIfTOCFlagEnabled(ScrollPercent);
+const FlaggedTOCContextProvider =
+   onlyShowIfTOCFlagEnabledProvider(TOCContextProvider);
```
djmetzle, line 75: This is the beating heart of the stubbing right here? Disabled in production.

jarstelfox: Yep!

frontend/templates/troubleshooting/index.tsx
```diff
@@ -95,125 +124,153 @@ const Wiki: NextPageWithLayout<{
...
+                              <SectionCard
+                                 key={solution.heading}
+                                 index={index + 1}
+                                 solution={solution}
+                              />
+                           ))}
+                        </Stack>
+                     </Box>
+                  )}
+                  <Conclusion conclusion={filteredConclusions} />
```
djmetzle, line 265: Will this change drop the `Related Pages` from the live version? Is that okay?

frontend/templates/troubleshooting/index.tsx
```diff
@@ -703,24 +762,26 @@ function IntroductionSection({ intro }: { intro: Section }) {
...
          <Prerendered html={conclusion.body} />
       </>
    );
-}
+};
 
 function Conclusion({ conclusion: conclusions }: { conclusion: Section[] }) {
-   const filteredConclusions = conclusions.filter(
-      (conclusion) => conclusion.heading !== 'Related Pages'
-   );
```
djmetzle, line 720: Disregard! Just moved.

frontend/templates/troubleshooting/scrollPercent.tsx
```diff
@@ -0,0 +1,115 @@
...
+            (window.scrollY - container.offsetTop) /
+               (container.offsetHeight - window.innerHeight)
+         );
+         setScrollPercent(scrollPercent);
+         setScrolledPast(
+            scrollPercent === 1 &&
+               window.scrollY > container.offsetTop + container.offsetHeight
+         );
+      };
+   };
```
djmetzle, line 37: Neat.

frontend/templates/troubleshooting/index.tsx
jarstelfox: Yes, this is already a Business Rule, I just moved it up the stack one component.

djmetzle (approved):
I tried to check through the force-pushes, and most seem reasonable. Looks you pushed away the separate `/dev/feature` route.

Otherwise these changes seem mostly the same to version we pair-reviewed.

If this is disabled in production, lets get this merged so that we can continue to iterate against master, without such overwhelming diffs.

CR :gear:

frontend/templates/troubleshooting/index.tsx
jarstelfox: It looks like this was added here: https://github.com/iFixit/react-commerce/commit/00f6c9f2468d5f22f13df2bff55b449cb2e054fe

I think we can clean it up, but I'd love to handle that as a follow up.

deltuh-vee:
QA 🎬 
* TOC is off for the react-commerce deployment and on for the react-commerce-prod deployment.
* TOC behaves well on all screen sizes
* TOC matches [figma](https://www.figma.com/file/jRG4EyQXRDMGfyLqut08KP/iFixit---Working-files?node-id=13998%3A321530&mode=dev) and has behavior like the [example app](https://rwddn4.csb.app/)
* The Desktop TOC is rendered server-side and doesn't flash in
* The Mobile TOC pops in once scrolled to
* No page shift / CLS occurs
* The Desktop TOC activates the correct item on page load(at least on the more normal example)
* Without a URL hash, refreshing loads you back at the same scroll position
* Scroll indicator and TOC are removed if we entirely scroll past content (Mobile mostly, but code doesn't care)
* Selecting an item on the TOC scrolls you to it and that item is selected
* Selecting an item in the TOC highlights it briefly
* The TOC never selects an item off screen
* The TOC is scrollable if required
    * A scroll fade occurs on desktop / mobile TOCs when the TOC is scrollable
    * No fade appears if not scrollable
* the TOC selects the active item as you scroll down the page
* Every possible TOC item is selected when scrolling up / down a page
* Clicking a TOC item updates the URL hash (and browser history) to the new item

☝️ Refreshing the page causes the progress bar to disappear
CleanShot 2023-08-22 at 13 51 35@2x
☝️ When scrolling past content, the mobile TOC disappears before the progress bar. Could we keep them a little more synced towards the end?
CleanShot 2023-08-22 at 14 17 20@2x

danielbeardsley:
When viewing the preview, the TOC list for mobile jumps around a ton while scrolling within it.

dev_block :+1: 

This is especially noticeable on the ("Many small items" example](https://react-commerce-prod-git-scroll-add-ui-ifixit.vercel.app/Troubleshooting/Chainsaw/Overheating/484453)

jarstelfox:
@danielbeardsley any reason why you dev blocked when the goal was to iterate this in a follow up: https://github.com/iFixit/react-commerce/pull/1923#issuecomment-1688679330?

sctice-ifixit:
un_deploy_block :flags: If we're confident this is behind a feature flag that won't be relevant in prod, let's continue and just make sure that we note all of the current issues for follow-up work.

danielbeardsley:
> any reason why you dev blocked when the goal was to iterate this in a follow up

Nope! Sorry I didn't see that comment!

frontend/components/common/FlexScrollGradient.tsx
danielbeardsley: When examining how this looks, I think I prefer the look of a fixed-height gradient (or none at all) over one that has a dynamic height.

danielbeardsley:
> let's continue and just make sure that we note all of the current issues for follow-up work.

I understand that long pulls can get unruly and comment threads can be harder to follow, but chopping them off by merging and wanting discussion to continue someplace other than the diff of the code has its own disadvantages. Threads either get lost, or must continue in a context separate from the changes they are discussing.

That said, I'm fine with merging this!

frontend/components/common/FlexScrollGradient.tsx
jarstelfox: I think some more general discussion about the desired fade behavior might be in order.

The other feature taken into account here is that the fade slowly shifts its opacity as you scroll until you pass its end.

Here is what I am talking about:

[Screencast from 2023-08-22 16-59-54.webm](https://github.com/iFixit/react-commerce/assets/857362/836c87d4-6ab0-4cae-89b0-452731f0d71e)

Note: I made the fade red to help visibility.

Here's what it looks like normally

[Screencast from 2023-08-22 17-01-35.webm](https://github.com/iFixit/react-commerce/assets/857362/1a97e5f2-34ca-46b7-9a87-0a97b27cb2b3)

frontend/templates/troubleshooting/index.tsx
jarstelfox: @ianrohde Showed me this approach which does work, but only in chrome and it puts the box shadow behind text, not in front of it: https://daverupert.com/2023/08/animation-timeline-scroll-shadows/

frontend/components/common/FlexScrollGradient.tsx
jarstelfox: @ianrohde Showed me this approach which does work, but only in chrome and it puts the box shadow behind text, not in front of it: https://daverupert.com/2023/08/animation-timeline-scroll-shadows/

Issue #1871, opened 2023-08-02
GitHub Actions: Investigate Increase in "Build and Cache" step for Playwright tests


ardelato:
## Description

It seems the `Build and Cache` step (https://github.com/iFixit/react-commerce/blob/main/.github/workflows/main.yml#L44) has been gradually taking longer which has caused us to be hitting the 15-minute timeout. As such, we've bumped up the timeout a bit to give us more breathing room (#1868). Regardless we should investigate why it has been increasing over the past few months as this trend shows it will continue to take longer.

![image](https://github.com/iFixit/react-commerce/assets/22064420/60516742-b24f-423a-9ee6-2a476e94976a)

### To Do

- [ ] Investigate why the step is taking longer and longer
- [ ] Decide whether we should drop the `Firefox` test as it is one of the main culprits behind the timeouts.

**References:** 
- https://github.com/iFixit/react-commerce/pull/1868#issue-1833622098
- https://app.datadoghq.com/notebook/5656966/react-commerce-times?range=172800000&view=view-mode&start=1690825657531&live=true

batbattur:
I couldn't find anything suspicious that increased the build time. The caching seems to be working fine for the `build and cache frontend for test` step. Although, noticed that it fluctuates anywhere from `~2m` to `~4m` randomly. 

Now, with `Firefox` removed, we are getting the following times:
- <img width="253" alt="Screen Shot 2023-08-18 at 4 39 35 PM" src="https://github.com/iFixit/react-commerce/assets/58952979/079928b3-ae90-46b2-a754-f75312e89293">

Going to close this issue.

PR #1924, opened 2023-08-18
Tests: Drop Firefox from playwright projects

batbattur:
## Summary
Drops `Firefox` from the playwright projects list.

Playwright tests running on `Firefox` take the longest time and usually cause
timeout which leads to failing CI. `Firefox` has been historically slower than `Chrome`
and only `~3.4%` of our users use Firefox, see https://github.com/iFixit/react-commerce/issues/1871#issuecomment-1684389877. 

## QA
- `Desktop Firefox` should no longer be part of `E2E` CI checks. 
- Passing is enough. 

connects #1871

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,13 +9,7 @@ jobs:
       strategy:
          fail-fast: false
          matrix:
-            project:
-               [
-                  'Desktop Chrome',
-                  'Desktop Firefox',
-                  'Mobile Chrome',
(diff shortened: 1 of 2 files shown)


mlahargou (approved):
CR 📱

jordycosta:
QA 🟢 

**Desktop Firefox** is no longer apart of E2E CI checks ✔️ 

firefox_8teDD4gjDy

Issue #1928, opened 2023-08-22
Malformed data from `store/product/${productHandle}` in dev and preview


federicobadini:
The iFixit product-data endpoint located at `store/product/${productHandle}` has been returning improper data, which subsequently leads to the condition label being incorrectly displayed in both the development and preview environments.

The above service produces a list of variant options that are used to filter the Shopify product variants (see [here](https://github.com/iFixit/react-commerce/blob/40f042f94b594d032081082a5b0691109a633307/frontend/models/product/components/product-option.ts#L19-L21)). However, the aforementioned endpoint is currently returning the value `New-TEsstss`, instead of the expected value `New`. This discrepancy results in condition being shown as a select with no option instead of a label.

<img width="585" alt="Screenshot 2023-08-22 at 09 38 28" src="https://github.com/iFixit/react-commerce/assets/7805759/6855c87f-446b-452b-ac0c-3b1c8bb578f6">

masonmcelvain:
I changed the text back to `New` in [Akeneo](https://akeneo.cominor.com/#/configuration/attribute/condition_code/edit). I checked a few pages on the preview and the problem seems resolved.

PR #1932, opened 2023-08-24
Migrate product pages to app router

federicobadini:
This PR migrates product pages from `pages` to `app` folder, marking a substantial shift in our structure as it introduces React Server Components (RSC). This is more than just a file relocation—it opens up a different caching strategy, a distinct routing approach, a possible large reduction in client bundle size and several other core changes.

In order to ensure a seamless transition, the changes in this PR were kept lean and minimalistic.
Future optimizations are in sight, and they involve transforming a significant segment of the UI into RSCs to prevent unnecessary code from permeating into the client bundle. This however, lies beyond the scope of this specific PR and will be addressed subsequently.

Most of the changes you will see in this PR are linked to these 4 requirements

- Data Retrieval: We had to transition from leveraging the getServerSideProps function to using a plain JavaScript function invoked within a React Server Component (RSC).
- SEO and Metadata: Since the `<Head/>` component is no longer available, metadata is now populated using the [metadata API](https://nextjs.org/docs/app/building-your-application/optimizing/metadata).
- Routing Changes: The new router has a different API. For example the request query and headers are not returned by the `useRouter` hook by default. Those data are now readable only via specific functions (`headers()` and `useSearchParams()`). 
- Caching Headers: given that headers retrieved via `headers()` function are now read-only, we are precluded from setting caching headers as we previously did with getServerSideProps.

A couple of things to monitor:

- Shallow Navigation: As highlighted in this [github discussion](https://github.com/vercel/next.js/discussions/48110#discussioncomment-6481618), shallow navigation is not yet available. Thus, we experience a slight UI delay due to re-rendering whenever URL parameters get altered via the new router.
- Data Cache: As mentioned in the documentation, a fetch call that follows a headers call will disable data caching. Therefore, should we decide to use data caching in the future, we must consider this aspect.

### QA

1. Visit Vercel preview
2. Check the new product page to highlight potential differences with the previous one.
3. If you try to make a performance comparison, consider that preview branches might include additional js due to the Vercel UI for commenting previews.


federicobadini:
deploy_block 🔴

frontend/models/product/components/product-option.ts
```diff
@@ -15,7 +15,13 @@ export function productOptionFromShopify(
    variants: ProductVariant[],
    iFixitOptions?: string[]
 ): ProductOption {
-   const values = iFixitOptions?.length ? iFixitOptions : option.values;
+   // TODO: cleanup
+   const values = iFixitOptions?.length
+      ? iFixitOptions.map((option) =>
+           option === 'New-TEsstss' ? 'New' : option
+        )
+      : option.values;
```
jarstelfox, line 23: huh?

danielbeardsley:
@jarstelfox and I are trying to read the docs and understand the reality of the app router. At the moment, this change removes the cache-control header we used to send back with the response.

Reading the docs and many issues, it appears we can *no longer control* the cache header that's sent back and it seems like it probably be "no-store" for every response.

Regarding `fetch()`, the behavior described seems crazy complex and flips HTTP caching semantics on it's head (The requestor decides how long to cache something) and `cache-control` headers in the response are ignored. I think this behavior is somewhat present in the pages router too, but it's not entirely clear. Whether or not `fetch()` has caching forced on it is dependent on code "discovery order" though it's not clear what that means.

There's a lot of talk about "static" routes and "dynamic" routes. I think we can safely ignore "static" and assume every route is "dynamic" because apparently this means "uses information about the request to issue the response". Because we take in params in the url, this implies that our routes are dynamic.

There's a lot here (in the app router docs) and it's going to take me a long time to absorb it all.

frontend/models/product/components/product-option.ts
federicobadini: This hack was necessary because of this: https://github.com/iFixit/react-commerce/issues/1928
Since some time, the Shopify option values are also filtered by iFixit options provided by Akeneo. However Akeneo was returning some malformed data that was causing a faulty display of the variant selector.
I left the `Todo` and deploy_blocked this PR to remember to clean this up.

Since it is now resolved I'm going to remove it immediately

jarstelfox:
> Reading the docs and many issues, it appears we can _no longer control_ the cache header that's sent back and it seems like it probably be "no-store" for every response.

I mostly agree with this, but I found these docs: https://nextjs.org/docs/app/building-your-application/caching#duration-3

>The cache is stored in the **browser's temporary memory**. Two factors determine how long the router cache lasts:
>
> * Session: The cache persists across navigation. However, it's cleared on page refresh.
>
> * Automatic Invalidation Period: The cache of an individual segment is automatically invalidated after a specific time. The duration depends on whether the route is [statically](https://nextjs.org/docs/app/building-your-application/rendering/server-components#static-rendering-default) or [dynamically](https://nextjs.org/docs/app/building-your-application/rendering/server-components#dynamic-rendering) rendered:
**Dynamically Rendered: 30 seconds**
Statically Rendered: 5 minutes

danielbeardsley:
> Two factors determine how long the router cache lasts:

FYI, that's about the "router" cache, an in-memory cache in JS that seems to cache responses the requests for "segments" (not sure what those are). That part of the docs isn't about HTTP caching.

erinemay:
It looks like there's still some ongoing discussion. I'm gonna drop the `external_block` label on here to sort this to the bottom of the QA list so we don't test it 'til everything's decided. Feel free to remove it if you want this tested as is.

federicobadini:
> FYI, that's about the "router" cache, an in-memory cache in JS that seems to cache responses the requests for "segments" (not sure what those are). That part of the docs isn't about HTTP caching.

I can confirm that is a browser level JS cache that prevents request from hitting the server.
For more context, here's the route segment definition: https://nextjs.org/docs/app/building-your-application/routing#route-segments

jarstelfox:
@federicobadini We are trying to determine what the HTTP cache control header will look like for this page now. Do you have any ideas?

federicobadini:
The mechanism according to which the app router is working is kinda different wrt the one we were used to before. 
With pages router setting caching headers was left to the programmer.
With app router, the `Cache-Control` header is defined automatically according to the data requirements of the page.

I've performed a few tests to verify how HTTP cache can be controlled.
We have 3 main headers to control caching:
[Vercel-CDN-Cache-Control](https://vercel.com/docs/edge-network/headers#vercel-cdn-cache-control)
[CDN-Cache-Control](https://vercel.com/docs/edge-network/headers#cdn-cache-control)
[Cache-Control](https://vercel.com/docs/edge-network/headers#cache-control)

The latter as said is overwritten for pages by Next.js, but both `Vercel-CDN-Cache-Control` and `CDN-Cache-Control` should be good to control how content is delivered to the user

The following two urls display the exact same page:
- https://next-app-router-tests.vercel.app/headers-read
- https://next-app-router-tests.vercel.app/headers-read-no-cache

Given that both pages are reading headers via `headers()` function, Next.js by default consider those pages as dynamic server rendered page. So `Cache-control` should be forced to `no-cache` or some equivalent value.
However the former has some additional headers added via `next.config.js`.
Here are the headers that are added to the `headers-read` page:

```
{
  source: "/headers-read",
  headers: [
    {
      key: "Cache-Control",
      value: "max-age=60",
    },
    {
      key: "CDN-Cache-Control",
      value: "max-age=60",
    },
    {
      key: "Vercel-CDN-Cache-Control",
      value: "max-age=60",
    },
  ],
}
```

And here are the headers that reach the browser for the same page:
```
cache-control: private, no-cache, no-store, max-age=0, must-revalidate
cdn-cache-control: max-age=60
x-vercel-cache: HIT
```
As expected `Vercel-CDN-Cache-Control` was used and stripped from the CDN, `CDN-Cache-Control` is still present while `Cache-Control` was overwritten.

`headers-read-no-cache` as expected is always a MISS.

frontend/app/(defaultLayout)/helpers.tsx
jarstelfox, line 1: Why not use the function here: https://github.com/iFixit/react-commerce/blob/6f47191f1a63fdc1860988e04f5046d77e4e27e1/frontend/helpers/path-helpers.ts#L131

this removes the need for this entire file

frontend/app/(defaultLayout)/helpers.tsx
```diff
@@ -0,0 +1,24 @@
...
+   let item = iterator.next();
+   while (!item.done) {
+      const [key, value] = item.value;
+      result[key] = value;
+      item = iterator.next();
+   }
+   return result;
+}
+
+export function ifixitOriginFromHostWrapper(): string {
```
danielbeardsley, line 18: I'm also a little confused why we have this. The iFixit origin that should be used is in an ENV variable I thought.

frontend/app/(defaultLayout)/components/GoogleAnalytics.tsx
jarstelfox, line 1: Why are you re adding this component when it already exists on main?

frontend/components/common/AppProviders.tsx
```diff
@@ -1,3 +1,5 @@
+'use client';
```
jarstelfox, line 1: Why are you re adding this when it already exists on main?

frontend/layouts/default/index.tsx
jarstelfox, line 1: I think you need to revert all changes to this file as we have forked the layout for App / Pages routers and the App router version already handles the changes required

frontend/app/(defaultLayout)/components/Matomo.tsx
jarstelfox, line 1: Why are you re adding this component when it already exists on main?

frontend/app/(defaultLayout)/layout.tsx
jarstelfox, line 1: Why are you re adding this component when it already exists on main?

jarstelfox (commented):
dev_block :cactus: I think you intended to merge in origin/main, but have not. I’ve CRd just in case the current code was intended, but it seems like there is a lot of repeated work

federicobadini:
The alignment with the `main` branch was yet to be completed since I paused the reconciliation due to the fact that I was iterating a bit more on the `hello-world-app-router` branch. 
That branch was a good starting point also for an activity that @dhmacs was going to start, so we decided to add a couple more commits there.
I will open a PR for that and then I will continue the reconciliation here.

I'll remove the block as soon as this is ready for rewiew/discussion

frontend/app/(defaultLayout)/layout.tsx
```diff
@@ -9,14 +11,16 @@ export default async function DefaultLayout({
 }: {
    children: ReactNode;
 }) {
+   const readonlyHeaders = headers();
```
danielbeardsley, line 14: I'm not entirely sure if this is the only place, but the only reason we're using `headers()` here is to know if we're being proxied in development so we can use relative API urls (i.e. to `{user}.cominor.com`)

This functionality was added in #1025. It is only needed in development and we could definitely turn this (extracting ifixit-origin from host header) on and off via an env variable if that would help.

frontend/app/(defaultLayout)/products/[handle]/helpers.ts
```diff
@@ -0,0 +1,156 @@
...
+
+   const { handle } = params;
+   const { images, metaTitle, shortDescription } = product;
+
+   const defaultVariantId = getDefaultVariantId(product);
+   const searchParamVariantId = getSearchParamVariantId(searchParams.variant);
+   let variant = product.variants.find((v) => v.id === searchParamVariantId);
+   if (variant == null) {
+      variant = product.variants.find((v) => v.id === defaultVariantId);
+   }
```
Comment (danielbeardsley) [inline frontend/app/(defaultLayout)/products/[handle]/helpers.ts:67]: This bit is identical to the part in `createLdJsonSupport()` below. How about extracting this to a function `getVariant(product, searchParams)`?

frontend/app/(defaultLayout)/layout.tsx
sterlinghirsh: I think that's a fair point, though we'll want to be careful about having different behavior in dev/prod.

PR #1933, opened 2023-08-24
Fix cart drawer animations

### LINKED ISSUE #1930: Checkout section is missing sometimes
### Preview:

https://github.com/iFixit/react-commerce/assets/95656772/5a449350-7488-4ce0-9d48-dd98d02cd504

dhmacs:
closes #1930 

I took this bug as an opportunity to fix other cart drawer animations bugs as well. 

Given that the problem seems to be a bug in framer motions components I replaced the animation with CSS animations. This change down the road will help us get rid of framer motion ([~40kb min/gzip](https://bundlephobia.com/package/framer-motion@10.16.2)).

## QA

Open [Vercel preview](https://react-commerce-git-fix-crat-drawer-animation-bug-ifixit.vercel.app/products/pro-tech-toolkit) and verify that bug described by #1930 is fixed


jordycosta:
QA 🟢 

The checkout section is now consistently rendering 👍 

---

### Testing Procedure:

1. Visit [a product page](https://react-commerce-git-fix-crat-drawer-animation-bug-ifixit.vercel.app/products/google-pixel-6a-screen-genuine)
2. Add the product to your cart and remove it
3. Repeat step 2.)
4. Ensure that the checkout section is always being rendered

dhmacs:
@jordycosta I think in the process of verify that the bug is solved you should have seen that adding to cart and removing to cart still works. I don't have anything else in mind.

packages/ui/animations/Collapse.tsx
```diff
@@ -0,0 +1,61 @@
...
+               ? height == null
+                  ? undefined
+                  : transition.state
+               : transition.state
+         }
+         hidden={transition.state === 'hide'}
+         sx={{
+            ...sx,
+            '--slide-height': `${height}px`,
+            '&.hide': {
```
masonmcelvain, line 35: Is `props.sx` intended to be an override, or a mechanism to add additional styling? Destructuring it first prevents the caller from overriding the default styling, but perhaps that is intentional.

(Relevant to `Fade.tsx` too)

packages/ui/animations/useAnimatedList.tsx
```diff
@@ -0,0 +1,100 @@
...
+}
+
+export function useAnimatedList<T>({ items, getId }: UseAnimatedListProps<T>) {
+   const animatedItems = useAnimatedItems({ items, getId });
+   const transitionItems = useTransitionItems(animatedItems);
+
+   const allItems = useMemo(() => {
+      const all = items.slice();
+      for (const indexedItem of transitionItems.exiting) {
+         all.splice(indexedItem.index, 0, indexedItem.item);
```
masonmcelvain, line 16: Does this duplicate exiting items? I don't understand why this is needed

packages/ui/animations/AnimatedList.tsx
```diff
@@ -0,0 +1,104 @@
...
+import React, { PropsWithChildren } from 'react';
+import { slideIn, slideOut } from './keyframes';
+import { useAnimatedList } from './useAnimatedList';
+import { useSize } from './useSize';
+
+interface AnimatedListProps<T = any> {
+   items: T[];
+   renderItem: (item: T) => React.ReactNode;
+   getItemId?: (item: T) => string;
+   debug?: string;
```
masonmcelvain, line 11: Is this a debugging artifact? I don't see it used

masonmcelvain (approved):
CR ✌🏻 Thanks for going above and beyond on this!

packages/ui/animations/useAnimatedList.tsx
dhmacs: `all` is a copy of the current `items` while `transition.exiting` contains items that are no longer present in `items`.
The purpose of this is to keep exiting items around for the time needed to the exit animations to complete (otherwise they'll be unmounted as soon as they're dropped from the array).

List animations in React are tricky because you have to keep exiting items around to be able to animate them, and this hook is accomplishing that in a transparent way.

packages/ui/animations/AnimatedList.tsx
dhmacs: Yes, I was using that while developing the animations, I thought I dropped it all, I'm quite surprise that this one managed to survived because I've reviewed the code multiple times before submitting the review 😅

dhmacs:
@masonmcelvain removed the debug line and fixed conflicts 👍 

cc @jordycosta

packages/ui/animations/Collapse.tsx
dhmacs: Yes intentional, albeit not even necessary for now. I'd say let's not worry about this now and come back to this if we ever need to

masonmcelvain (approved):
CR 👏🏻  Thanks for your responses!

jordycosta:
QA 🟢 See https://github.com/iFixit/react-commerce/pull/1933#issuecomment-1701405144

PR #1935, opened 2023-08-24
Minicart: Use retail discount from api

masonmcelvain:
Closes https://github.com/iFixit/ifixit/issues/45708

The retail discount is a new field that factors in the retail (aka compare at) price when it is set, otherwise it is the same as the `discount` field.

## QA

Test all combinations of discounts and make sure that the Next.js product page, the Next.js cart drawer, and checkout all align with each other (except for employee discounts, which only appear in checkout).

You can edit a product in akeneo.cominor.com to change it's price and retail (aka "compare at") price, as well as price tier overrides. Default pro percent off discounts are defined in `ifixit_cart.customer_discounts`, and apply to products when "Reseller Discount" is enabled in Akeneo.

### DIFF
--- a/packages/cart-sdk/hooks/use-cart.ts
+++ b/packages/cart-sdk/hooks/use-cart.ts
@@ -37,8 +37,8 @@ function isValidCartPayload(data: any): data is CartAPIResponse {
 function createCart(input: APICart): Cart {
    const lineItems = input.products.map<CartLineItem>((product) => {
       const priceAmount = parseFloat(product.subPrice);
-      const singleItemDiscount = product.discount
-         ? parseFloat(product.discount)
+      const singleItemDiscount = product.retailDiscount
+         ? parseFloat(product.retailDiscount)
          : 0;
       const price: Money = {
          amount: priceAmount,
--- a/packages/cart-sdk/types.ts
+++ b/packages/cart-sdk/types.ts
@@ -92,6 +92,7 @@ export interface APICartProduct {
    maxToAdd: number;
    name: string;
    quantity: number;
+   retailDiscount: string;
    subPrice: string;
    subPriceStr: string;
    subTotal: string;


packages/cart-sdk/hooks/use-cart.ts
```diff
@@ -37,8 +37,8 @@ function isValidCartPayload(data: any): data is CartAPIResponse {
 function createCart(input: APICart): Cart {
    const lineItems = input.products.map<CartLineItem>((product) => {
       const priceAmount = parseFloat(product.subPrice);
-      const singleItemDiscount = product.discount
-         ? parseFloat(product.discount)
+      const singleItemDiscount = product.retailDiscount
+         ? parseFloat(product.retailDiscount)
```
danielbeardsley, line 41: This already falls back to the user discount, correct? So we don't have to do any fallbacks here?

danielbeardsley (approved):
CR :+1:

packages/cart-sdk/hooks/use-cart.ts
masonmcelvain: Correct, see

https://github.com/iFixit/ifixit/blob/bd505ceda2824dd50f5c6b5b0de0d329b0918a05/Libs/Cart/BasePricingLib.php#L66-L68

danielbeardsley: I have confirmed that this is the case.

erinemay:
QA 👍 
To non-pro, non-employee users, discounts are identical across:
- [x] Product and collection pages
- [x] Cart drawer
- [x] Checkout

I wanted to see what discounts look like as a [pro user](https://overlord.ifixit.com/Wiki/Software_QA_iFixit.com_test_users#Section_Pro_Users), and I'm a little surprised that when logged in as a pro user, I see the discounted price and not my pro price. Not a blocker because it's present on the main branch. I think there's an issue with our test pro users. I'm getting a 401 from https://www.cominor.com/api/2.0/user.
https://react-commerce-git-cart-pro-discount-fix-ifixit.vercel.app/products/fix-the-world-sticker
https://akeneo.cominor.com/#/enrich/product/8af3a3ab-f4f5-4264-b74d-607ea1bd15e1
image
image


deploy_block 🌵 
@masonmcelvain, should this have any affect on how pros view prices? I can't really test that right now.

masonmcelvain:
Thanks @erinemay!

> I'm getting a 401

For the iFixit user api request to succeed, we need to preview on `cominor.com`. So try visiting on https://react-commerce-git-cart-pro-discount-fix-ifixit.cominor.com/products/fix-the-world-sticker

Something to do with CORS or the Origin header, but I don't remember the specifics

erinemay:
un_deploy_block 🌵 
That's working. 
image
image
image

pro prices are correct on page and in cart

PR #1939, opened 2023-08-25
Hide cross sell when there are no available items

### LINKED ISSUE #1938: Frequently Bought Together: Hide section when empty
When empty of suggestions, Frequently Bought Together should be hidden. The logic should be the same as previous. When a criss-sell is out-of-stock, do not list the item in the Frequently Bought Together widget. And when there are no recommendations in the Frequently Bought Together widget, hide the section. 

Example product: https://www.ifixit.com/products/nt140whm-n41-lcd-panel

image

CC @dhmacs @masonmcelvain

dhmacs:
close #1938 

qa_req 0


masonmcelvain (approved):
CR ✌🏻

Issue #1884, opened 2023-08-04
Proxying: Test out using vercel's infra to proxy to our php app


danielbeardsley:
### Motivation
We've been using Cloudfront as means to service both next.js and our php app from the same domain. Requests flow like this:

```mermaid
flowchart TD
A[Internet] --> B[Cloudfront]
B --> C[Vercel CDN]
B --> D[PHP App]
C --> E[Vercel next.js]
```
This was familiar to set up, but has some Cons:
* CF and Vercel CDN are on different platforms (Vercel CDN uses cloudflare)
   * Thus requests may not take an optimal route between these two systems
* Having two complex caching proxies between our servers and users has caused surprising problems
* Dealing with session cookies through Cloudfront has required a lot of thinking and we still haven't done it
* 

### Proposal: Vercel as a Proxy
Have Vercel CDN do the front-end proxying of our `www.ifixit.com` domain and send php-app requests directly to us-east-1.

There's several options here:
#### Option 1 (Edge middleware with proxylbs)
Keep our proxylbs, let vercel's CDN proxy to `proxylb.ifixit.com`.

```mermaid
flowchart TD
A[Internet] --> B[Vercel CDN: global]
B --> C[PHP App: proxylbs]
C --> E[PHP App: us-east-1]
B --> D[Vercel next.js: us-east-1]
```

#### Option 2 (Edge middleware with one LB)
Have vercel's CDN send requests directly to one (or two) that are in us-east `lb.ifixit.com`.

```mermaid
flowchart TD
A[Internet] --> B[Vercel CDN: global]
B --> C[PHP App: us-east-1]
B --> D[Vercel next.js: us-east-1]
```

#### Option 2 (Next proxy)
Have vercel CDN send all requests to next.js and have the proxying to our PHP App happen within the next.js process. both next and our app are in us-east so the network time between them is less than 4ms apart.
```mermaid
flowchart TD
A[Internet] --> B[Vercel CDN global]
D --> |proxy in next| C[PHP App: us-east-1]
B --> D[Vercel next.js: us-east-1]
```

More to come ...

danielbeardsley:
I was going to say I'm sure Vercel's SLA is good... but ooof. They aim for 99.99% uptime (probably realistic) but that doesn't include this :point_down: 

> Scheduled Downtime will not exceed eight (8) hours per month and will be scheduled in advance during off-peak hours (based on PT). We will notify you via email of any Scheduled Downtime that will exceed two (2) hours.

[...]

PR #1943, opened 2023-08-28
/Store pages: Drop catch-all route

danielbeardsley:
For potential vercel proxying, we won't be able to use a catch-all route. Since we've only ever used `/Store`, let's limit the catch-all route to `/Store/*` (and `/Store`).

Note: This means that "pages" in strapi will only be matched if they start with `/Store`. No changes are needed for the existing /Store page.

This way, we can test a fallback that will proxy to ifixit.com on unmatched pages.

Connect: #1884


frontend/templates/page/server.tsx
```diff
@@ -16,8 +16,9 @@ export const getServerSideProps: GetServerSideProps<PageTemplateProps> = async (
       };
    }
 
-   const slug = context.params?.slug;
+   const slug = context.params?.slug || [];
    invariant(Array.isArray(slug), 'page slug param is missing');
```
sterlinghirsh, line 20: This is kind of pointless now.

danielbeardsley: I think it was kind of pointless before :-) if this wasn't here the next line would throw an exception.

danielbeardsley: That said, it helps typescript to know that this is a `string[]` instead of `string`.

danielbeardsley:
`/Store` is accessible on the preview and looks the same as that page on the ifixit.com domain.

frontend/pages/Store/index.tsx
```diff
@@ -0,0 +1,2 @@
+export { default } from '@templates/page';
+export { getServerSideProps } from '@templates/page/server';
```
dhmacs, line 2: This file is not necessary.. We can make the catch all segment [optional](https://nextjs.org/docs/pages/building-your-application/routing/dynamic-routes#optional-catch-all-segments) by renaming `frontend/pages/Store/[...slug].tsx` to `frontend/pages/Store/[[...slug]].tsx`

danielbeardsley: TIL! But I'd still rather have this two line file than have someone need to know about the particulars of what double-brackets means in a file name. Actually, for that reason, I think I'll rename this to just `pages/Store.tsx` so it's even more obvious.

danielbeardsley:
QA :+1: 

`/Store` still works

frontend/pages/Store/index.tsx
sterlinghirsh: I forgot about it too but we shouldn't get in the habit of creating extra files just in case someone doesn't know a feature of the framework we're using. I'm OK with putting it at just /Store for now since we're not using other pages, but we'll probably want to revisit this eventually when we do want to put up more pages.

sterlinghirsh (approved):
CR :zap: I'm going to let this merge to move it along, but it feels a little wrong since the server.tsx file doesn't really make sense anymore. We should decide what we want to do with these arbitrary pages and clean this up or use the double brackets or something. I think keeping a /Store prefix is a good compromise to let us make new ad-hoc pages without a global catchall.

danielbeardsley:
> but it feels a little wrong since the server.tsx file doesn't really make sense anymore.

Not following. What part isn't making sense?

> We should decide what we want to do with these arbitrary pages and clean this up or use the double brackets or something

Not sure what "clean this up" means except maybe "remove the feature" ?

> I think keeping a /Store prefix is a good compromise to let us make new ad-hoc pages without a global catchall.

Agreed.

sterlinghirsh:
> What part isn't making sense?

It has code to accept arbitrary slugs from path variables but now we force inject Store and don't accept any path variables.

> Not sure what "clean this up" means except maybe "remove the feature" ?

Yeah either remove the feature or make it work with path variables again.

danielbeardsley:
> and don't accept any path variables.

? Maybe I missed something? I certainly didn't intended to break that.

Looking at the diff again, it looks like I intended it to look, The catch-all route is limited to `/Store/*` and `/Store` still works and uses the same code.

Maybe there was a mis-communication?

image

sterlinghirsh:
Yeah you're right I think I misread it. I'd still be down to use the double brackets but since we still have /Store and /Store I'm less concerned.

PR #1944, opened 2023-08-28
Troubleshooting TOC: Polish

jarstelfox:
This pull addresses many minor issues noticed when working on #1923

Some notes for CR:

@danielbeardsley brought up [a convo here](https://github.com/iFixit/react-commerce/pull/1923#discussion_r1301954469), which is still worth discussing.

CR with ignoring whitespace for an easier time.

QA
---

* Ensure TOC looks like the figma
* [Ensure TOC behaves](https://github.com/iFixit/react-commerce/pull/1923#issuecomment-1688960687)
* [Ensure mobile TOC scrolling on iOS works nicely](https://github.com/iFixit/react-commerce/pull/1923#issuecomment-1683192508)
* Point out anything of concern

Note: The causes list will be hidden on desktop due to work from @andyg0808 

Closes: https://github.com/iFixit/ifixit/issues/49284


frontend/components/common/FlexScrollGradient.tsx
```diff
@@ -76,10 +76,7 @@ function getGradientsBaseStyle(
    };
 }
 
-function getGradientStyleProps(
-   el: HTMLElement,
-   gradientSizePX: number
-): InnerFlexStyling {
+function getGradientStyleProps(el: HTMLElement): InnerFlexStyling {
```
jarstelfox, line 79: @danielbeardsley I'd love to continue our convo from https://github.com/iFixit/react-commerce/pull/1923#discussion_r1301954469

Basically, your concern was that this component was complex. I agree.
The component has a few jobs and we might be able to reduce the work it is doing.

Before getting into that, I wanted to comment that I'd love a pure CSS solution to this, but there is not an exist CSS feature to allow CSS only when scrolled, not at either end, but [they are working on it](https://daverupert.com/2023/08/animation-timeline-scroll-shadows/)


Here's the jobs of this component:

* Add a before / after element iff scrolled
* Change the opacity of the before/after while scrolling
	* Before hidden at start, after hidden when scrolled fully
* Adjust for the size of the scrollbar
	* Say content is 100px, but then you add scroll auto, the browser adds a scrollbar (chrome's is 15px) and not 'width: 100%' will be `115px` Given the positioning of the fades, they would overlay the scrollbar, thus we must subtract the scrollbar
* Handle both row and column flex directions
	* All math is the same but inversed
* Adjust the size of the fade based on the available space of the scroll container

danielbeardsley: > Basically, your concern was that this component was complex. I agree.
> ... Before getting into that, I wanted to comment that I'd love a pure CSS solution to this

I don't mean to downplay what you done, it's pretty snazzy, but I tend to think it's not a problem that needs solving. It seems a bit of a stretch to effectively say "native scrolling has a bad UI that we need to fix". Every website deals with scrolling, what makes this particular scroll area unique such that it needs a solution?

jarstelfox: > It seems a bit of a stretch to effectively say "native scrolling has a bad UI that we need to fix".

Totally, the main reason this exists is that we keep designing fades (usually for mobile) so we indicate to a user that a section is scrollable.

Some examples:
image
image
image

This seems to be part of our new design language, but if we decide we don't need it / can simplify that would be great.

danielbeardsley: I think augmenting native scrolling with fades is fairly complex and really easy to mess up, potentially breaking the scroll behavior. I vote we should rely on the presence of scroll bars and the specifying a height that isn't an exact multiple of the items (so one is chopped off).

That said, I think region vertical scrolling on mobile devices is a frustrating experience because it breaks the "drag anywhere to scroll the screen" model that we're all so used to.

Thoughts @ianrohde and @federicobadini and @mmarcon91 ?

jarstelfox: Overall I agree with you Danny, but I want to point out one thing.

> I vote we should rely on the presence of scroll bars and then specifying a height that isn't an exact multiple of the items (so one is chopped off).

This isn't always possible. For example, the guide flag text is translatable and the adding/removing of flags is user-editable. It would be possible for some combination of strings/flags to be near pixel-perfect to the edge of the scrollable pane. Without the fade, it would not seem scrollable at a glance.

Similarly, I suspect the font stack could wrap text differently and have similar issues.

---

I just wanted to claim that your statement is totally reasonable, but in practice may not be possible due to many variables.

danielbeardsley: In specific, I think the little drop down is not that useful on mobile. I think a "Scroll to Table of Contents" link would be more friendly and much easier to implement.

ianrohde: The TOC matches the Figma comp, but Jarred went the extra mile to add the visual cue that the list is scrollable:

<details>

Screenshot 2023-08-29 at 11 55 32 AM

</details>

> That said, I think region vertical scrolling on mobile devices is a frustrating experience because it breaks the "drag anywhere to scroll the screen" model that we're all so used to.

I'm betting that Mattia considered this, but it's just difficult to convey in Figma.

jarstelfox: > I'm betting that Mattia considered this, but it's just difficult to convey in Figma.

There is a fade in the demo app he built on both desktop but not mobile: https://rwddn4.csb.app/

I think I had assumed the fade was intended and I had already invested time in this component from the other repo and new it was easy enough to drop in.

If not intended, I can easily remove it.

sterlinghirsh: I think fades are good, especially for horizontal scrolls. There are rarely any permanently visible scroll bars on mobile, so I don't think that's a good answer here. The fade on a single line of text seems a little hard to understand to me, so maybe it just needs to be a shorter gradient.

Alternatively, it seems like this dropdown could just be a bigger, maybe even full-screen on mobile since dropdowns are a little funky even in the best case.

mmarcon91: For me we can remove the shading in the mobile summary panel. They make reading complicated and reduce the visible space.

We can also keep the native system scrolling and remove the bars, in 90% of cases the point-title at the bottom of the panel will be cut off and the user should understand that there is content to scroll through.

ianrohde (commented):
CR ⚪ 

Nice work here 👍

erinemay:
@jarstelfox, I tried to differentiate between general TOC feedback (put in the [discussion](https://github.com/iFixit/ifixit/discussions/48967#discussioncomment-6870061)) and feedback based on changes here. 

This is good. Everything introduced here on mobile that I can tell is a fix. Just the one desktop regression noted below. 
- Refreshing the page on mobile doesn't make the progress bar disappear
- When scrolling past TOC content, the progress bar and TOC disappear in unison
- Scrolling on iOS is much smoother 👌 

Page I'm using: 
branch - https://react-commerce-git-troubleshooting-toc-polish-ifixit.vercel.app/Troubleshooting/Mac_Laptop/MacBook+Keyboard+Not+Working/505002

master - https://react-commerce.vercel.app/Troubleshooting/Mac_Laptop/MacBook+Keyboard+Not+Working/505002

dev_block 🌵 
<details><summary>The desktop TOC is now stuck to the left and we lose the right sided bounding of the text/steps</summary>

Windows + Chrome

| branch | master |
| - | - |
| image | image |

</details>

frontend/components/common/FlexScrollGradient.tsx
mmarcon91: In the prototype

mmarcon91: In the prototype, I've implemented the fade with Tailwind like this:

```html
<div class="h-24 w-full bottom-0 absolute left-0 bg-gradient-to-t from-gray-50 pointer-events-none"></div>
```

It should be possible to replicate this using Charka, right?

jarstelfox: @mmarcon91 im not quite understanding what you mean with you comment.

The code above is effectively replicating the fade you described. The only difference is the height varies based on the available space. This was done so that the fade component can be used anywhere on any design

jarstelfox:
> The desktop TOC is now stuck to the left and we lose the right sided bounding of the text/steps

I believe @mmarcon91 wanted it “stuck to the left” from a comment on the original pull

frontend/components/common/FlexScrollGradient.tsx
mmarcon91: Not necessarily the fade height should vary based on the container height. The fades that I design are based on the specific context, they don't follow some specific mathematic rules, so I don't think it's really possible to express this as a reusable component. 
We would also probably benefit in terms of performance from using a css only solution, what do you think?

mmarcon91:
@erinemay [Here](https://www.figma.com/file/QVLUEpDVoyiMM4MXEGq1L9/iFixit---Problems?type=design&node-id=201-65526&mode=design&t=2Jdk7kaM5hUkO0z6-4) (mockups on the right)  are the various versions based on screen resolution. Regarding the width of the text should be reduced with the introduction of the sidebar

frontend/components/common/FlexScrollGradient.tsx
jarstelfox: > We would also probably benefit in terms of performance from using a css only solution, what do you think?

Totally, but a CSS-only solution does not work how you might expect.

There is no CSS hook to determine if you have scrolled to the top or bottom. There is also no CSS to decide whether your element is scrollable at all.

You can totally have a fade that is always there, but floated over nothing like your example:
image


But if we want to only add the scroll if and only if the container is scrollable we need a use effect.

@ianrohde and I have taken some time to look at a CSS-only solution and we still lean towards the current solution for now.

erinemay:
un_dev_block 🌵

sctice-ifixit:
Regarding @erinemay's [comment](https://github.com/iFixit/react-commerce/pull/1944#issuecomment-1701315639) about the TOC positioning and @mmarcon91's [reply](https://github.com/iFixit/react-commerce/pull/1944#issuecomment-1704917710), the page is already wider than we'd like, and it sounds like this pull is going to make it wider. If it's an easy change, how about we leave the TOC behavior as it was prior to this pull until we have the sidebar implemented, at which point the main content will be less expansive?

ianrohde (approved):
CR ⚪

erinemay:
Is everything here safely behind a feature switch? If not, I'd love to verify appearance/behavior without the TOC.

(I got kicked out of Vercel as part of the great purge. I don't know if we necessarily need to pay for my account just to flip the switch on features.)

jarstelfox:
>it sounds like this pull is going to make it wider.

This pull is not widening the container, there will be no change of the container size, the TOC will just exist to the left of the same size container

erinemay:
> > it sounds like this pull is going to make it wider.
> 
> This pull is not widening the container, there will be no change of the container size, the TOC will just exist to the left of the same size container

Even if it's not the container, something is allowing the content to become very wide on this branch. This is at 100% zoom on my laptop screen. 

https://react-commerce.vercel.app/Troubleshooting/Mac_Laptop/MacBook+Keyboard+Not+Working/505002
image
https://react-commerce-git-troubleshooting-toc-polish-ifixit.vercel.app/Troubleshooting/Mac_Laptop/MacBook+Keyboard+Not+Working/505002
image

As far as QA. If this is all safely behind a feature switch and will not make any changes to the non-TOC version, I'm happy to stamp this. And y'all can fix the widening and everything in the discussion we want to do in follow ups.

ianrohde:
The main content has more space to fill because the TOC was moved out of what was a shared container.

We have been discussing how to handle this, but we'll need to discuss with Mattia to get clarity on how to handle the viewports that are not included in the design.

jarstelfox:
> As far as QA. If this is all safely behind a feature switch and will not make any changes to the non-TOC version, I'm happy to stamp this. And y'all can fix the widening and everything in the discussion we want to do in follow ups.

Let's do that.

If I open https://react-commerce-git-troubleshooting-toc-polish-ifixit.vercel.app/Troubleshooting/Mac_Laptop/MacBook+Keyboard+Not+Working/505002 I am not seeing the same behavior you are:

image
image


The container is the same size, the difference is that the TOC was moved out of the container. But this does not seem to be what you are indicating

erinemay:
> > As far as QA. If this is all safely behind a feature switch and will not make any changes to the non-TOC version, I'm happy to stamp this. And y'all can fix the widening and everything in the discussion we want to do in follow ups.
> 
> Let's do that.

QA 👍 I did a pretty thorough check on this branch and main and have everything I saw documented here or in the discussion. deploy_block 🌵 @jarstelfox un block if you wanna merge.

jarstelfox:
un_deploy_block 🌵 Let's continue chatting and iterating on the TOC, but in a new pull / on the discussion

PR #1948, opened 2023-08-31
Fix `chakra-border-color` not being set during SSR

federicobadini:
We were starting to adopt `chakra-border-color` as color value for borders because we noticed that sometimes borders were set to either `gray.200` or `gray.300`, and this create some subtle differences in the ui. 
However, that specific value seems not to be set during SSR (see [here](https://github.com/chakra-ui/chakra-ui/issues/6727)) and produces a flash of unstyled content.

In the image below you can see the problem in the frequently bought together component.
This PR inject an inline style to fix this issue.

Screenshot 2023-08-31 at 11 18 21

qa_req 0


federicobadini:
It is almost equivalent. The original problem lies in the fact that the `--chakra-colors-chakra-border-color` css variable is set conditionally, depending on the `data-theme` property. The theme setting is a client information so the data-theme prop is set as soon as the app is hydrated.

Screenshot 2023-09-01 at 15 04 40

That css variable is used in the global theme to set the `border-color` prop.
So overriding `borderColor` inside the theme as you did is another way to fix the issue. 

I'm not saying the two are 100% equivalent because if some Chakra component internally references the CSS variable, that would not work in SSR. But as far as I can see, I've not found any visible issue in switching my solution to yours.
I'll switch to yours since it is just using chakra apis and feels more natural

dev_block 🛑

federicobadini:
Given that as anticipated we noticed that sometimes borders were set to either gray.200 or gray.300, and this create some subtle differences in the ui I've also defined an additional Chakra color specific for borders for future use

un_dev_block 👍🏻

ianrohde (approved):
CR ⚪

PR #1950, opened 2023-08-31
App Router: Add a basic /Troubleshooting/Device page

jarstelfox:
This is the first use of the new app router.
It is a simple hello world page that uses the new app router to render a page with a dynamic path.

Note to @federicobadini: Sorry about redoing some of the work from your pull!

We are looking to use this for a new page, but found that the Layout/migration changes are required for *any* page to use the App Router.

The hope is that pull is the MVP for any page to use the newer router.

CR
---

Commit by Commit. I added a lot of context.

QA
---

* Smoke test all pages
* This should show a basic page: https://react-commerce-git-hello-world-app-router-ifixit.vercel.app/Troubleshooting/MyDevice
* This should be a 404: https://react-commerce-prod-git-hello-world-app-router-ifixit.vercel.app/Troubleshooting/MyDevice
	* This is due to the feature flag

Connects: https://github.com/iFixit/ifixit/issues/49588
Connects: https://github.com/iFixit/react-commerce/pull/1932


frontend/layouts/default/index.tsx
```diff
@@ -87,87 +88,6 @@ const DefaultLayoutComponent = function ({
             apiVersion="2020-01"
          >
             <Box>
-               <Head>
-                  <title>{title || 'iFixit'}</title>
-                  <link
-                     rel="apple-touch-icon"
-                     sizes="57x57"
-                     href="https://assets.cdn.ifixit.com/static/icons/ifixit/apple-touch-icon-57x57.png"
-                  />
```
masonmcelvain, line 96: Will this remove these tags from all pages that use the `pages` router?

jarstelfox: I was wondering this too, but did not check. It looks like this does remove the icons.

https://react-commerce-prod-git-hello-world-app-router-ifixit.vercel.app/products/spudger
https://react-commerce-prod.vercel.app/products/spudger

<details>
<summary>Branch</summary>

image

</details>

<details>
<summary>Production</summary>

image

</details>

@federicobadini I assume this was not intended?

Due to the time difference, I am going to push a commit to add these links back, but let me know if this is incorrect

masonmcelvain: This is tricky, because now we might be double-including these tags in `frontend/app/layout.tsx`. Though maybe the `<Head>` component just doesn't work/render anything in the app router?

federicobadini:
The `migrate-to-app-router` pull request is under `deploy_block` because before merging some files need to be duplicated to work consistently both in the app router and in the pages router.
For instance: 
- app router uses the router from next/navigation while the pages router uses next/router 
- the `Head` component is doing nothing under app router etc. 
- etc.

Given that the PR was already complicated on its own, In an attempt to keep it less cluttered and more readable, I temporarily omitted these duplications of the common logic.
Moreover I was planning - while we were testing the new product page -  to also migrate the product list pages, basically removing the need to disambiguate the common logic.

Nonetheless, since Algolia is not completely compatible with the app router at this point, migrating the product list may take a little longer. As a result, it is likely that I'll likely have to disambiguate the common files first.

Coming to this PR, the point is the same: we don't have to mix the apis.
So GoogleAnalytics, Matomo and layout at least need to be duplicated.

it is not 100% clear to me the reason why we are willing to push an hello-world to production though.
if the original request was to create a new page, I personally feel it may be better to work directly on the new content. It is not a big deal for me to reconcile the `migrate-to-app-router` PR with the work we will end up doing here. However I am somewhat skeptical about deploying a hello-world page that has no practical relevance to production. 

On the other hand, if the aim was different - for example testing the interaction between the app router and the various CDN layers we have in production, that's a different point.  

dev_block 🛑

jarstelfox:
> However I am somewhat skeptical about deploying a `hello-world` page that has no practical relevance to production.

Agreed here. The problem is the new page will be under the `/Troubleshooting/...` route. Since the route is already served by the pages router, we will need to upgrade the existing page before adding the new one.

The main goal here was to separate the concept of "getting ready for using the App route" and "upgrade the page which generates all revenue". The general concern was the `migrate-to-app-router` would have some churn in CR / QA and would block our team from further work.

The secondary goal was to help pull out some work from your pull and make it easier to CR / QA.

jarstelfox:
> for example testing the interaction between the app router and the various CDN layers we have in production - that's a different point.

This was also some of the point. It just gave us a very simple test bed in production to ensure all is working as expected.

The hope is to drop it within 24 hours of deploying it / migrating the troubleshooting page as a follow up

jarstelfox:
> The new page is in a conflicting position?

Yes, the new page will conflict. However, we can totally add it under a new path then work to upgrade the old one as well as create the new one.

I did not want this pull to focus on upgrading existing pages or adding a new fully featured page.

The main goal is to show that we can run both routers and unblock my work from your work

jarstelfox:
un_dev_block 🌵 

After chatting we decided on a few things:

1) Let's use the expected route for the new page we want to build
2) Let's duplicate any logic the App Router needs so we can easily support the new system

Some new notes:

* I noticed that the 404 page was Next's default, not ours so I duplicated the 404 page across as well
* The new page is behind a feature flag 👍🏻 
* We slightly changed the URL of the new page due to possible collisions with dynamic routes

frontend/app/(defaultLayout)/Troubleshooting/[device]/pageTemplate.tsx
Comment (federicobadini) [inline frontend/app/(defaultLayout)/Troubleshooting/[device]/pageTemplate.tsx:1]: Maybe also here we could try to organize the hierarchy with:
- page.tsx
- components folder - to colocate other client component that are not shared with other routes/pages

So we could transform this into a Heading client component.

I am proposing to organise the code like so, because we will have to push as far as possible client components to the leaf of the hierarchy. And having smaller components could simplify the process.

federicobadini (commented):
I left some suggestion regarding code organization. Let me know what you think

frontend/components/analytics/AppRouter/GoogleAnalytics.tsx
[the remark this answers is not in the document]
dhmacs: Agree 👍 Much of the components over time are probably better organized if colocated closer to where they're used since app router allows this

frontend/app/(defaultLayout)/Troubleshooting/[device]/pageTemplate.tsx
Comment (jarstelfox) [reply frontend/app/(defaultLayout)/Troubleshooting/[device]/pageTemplate.tsx:1]: This seems like the best way forward.

If we do end up creating client components which are shared without views we can create a `common` folder

One note is that `(defaultLayout)/Troubleshooting/components` would be routable if someone adds a `.../components/page.tsx` / `.../components/layout.tsx`

I wonder if we would want to make it a group: `(defaultLayout)/Troubleshooting/'(components)`

For now, I think a group may be overkill so I am inclined to leave it as `(defaultLayout)/Troubleshooting/components`. If you have an opinion let me know.

frontend/app/(defaultLayout)/iFixitHeaderFooter.tsx
```diff
@@ -0,0 +1,461 @@
...
+
+   return (
+      <LayoutErrorBoundary>
+         <ShopifyStorefrontProvider
+            shopDomain={shopifyCredentials.storefrontDomain}
+            storefrontAccessToken={shopifyCredentials.storefrontAccessToken}
+            apiVersion="2020-01"
+         >
+            <Box>
+               <Head>
```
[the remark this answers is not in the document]
jarstelfox: Thanks, I missed this myself.

frontend/app/(defaultLayout)/Troubleshooting/[device]/page.tsx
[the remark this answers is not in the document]
Comment (jarstelfox) [reply frontend/app/(defaultLayout)/Troubleshooting/[device]/page.tsx:1]: >Or is it already a schema of the page we wanna merge in production?

It is currently the latter. We intend to continue adding a page at this route location. However, the url spec for this page is still being sorted out. We believe this will be the final URL, but we may change it.

>the ifixitOrigin variable should be retrieved differently (headers/env var)

I noticed how you were grabbing this on your pull and it made sense, but was another 100-200 lines of code which I was hoping to not add to this pull.

I'll take a look if all of the code in the urils file is needed to port that over.

>alternates languages should not include those options

I'm not sure what you mean here.

For now, I will drop the alternates languages options. Originally, this was me messing with the new metadata API so I could understand hreflangs would be rendered

erinemay:
Here's some preliminary feedback. I can go deeper tomorrow, but here are the differences I can see between this branch preview and main.

dev_block 🌵 

- Missing header links
image
- This might just be because this was built off of an older branch, but we get these double lines: https://react-commerce-git-hello-world-app-router-ifixit.vercel.app/Troubleshooting/Washing_Machine/Has+No+Water/493438
image

Otherwise, product pages, adding to cart, checking out, vulcan pages, collection pages (including filtering) look ok.

jarstelfox:
> * Missing header links

I see this too. Nice catch. It's unclear why, but we are hitting a bail condition which has existed before this pull:


The menu data looks like:
```
[dev:frontend] [Next] menu, {
[dev:frontend] [Next]   title: 'US Main',
[dev:frontend] [Next]   items: [
[dev:frontend] [Next]     { type: 'submenu', name: 'Fix Your Stuff', submenu: null },
[dev:frontend] [Next]     { type: 'submenu', name: 'Join the Community', submenu: null },
[dev:frontend] [Next]     { type: 'submenu', name: 'Store', submenu: null }
[dev:frontend] [Next]   ]
[dev:frontend] [Next] }
```

We are hitting this condition:
https://github.com/iFixit/react-commerce/blob/6f47191f1a63fdc1860988e04f5046d77e4e27e1/frontend/layouts/default/index.tsx#L202-L204

The condition was added here: https://github.com/iFixit/react-commerce/commit/dac042382d424c14594abf2b38f422f9dac7b90d

We also load this data here:
https://github.com/iFixit/react-commerce/blob/6f47191f1a63fdc1860988e04f5046d77e4e27e1/frontend/layouts/default/server.ts#L15-L35

@dhmacs do you have any idea why would would have null `submenu`s here?

dhmacs:
> @dhmacs do you have any idea why would would have null `submenu`s here?

@jarstelfox it's because those are not set in Strapi [here](https://hello-world-app-router.govinor.com/admin/content-manager/collectionType/api::menu.menu/3?plugins[i18n][locale]=en):

Screenshot 2023-09-06 at 09 53 58

Unfortunately the Strapi seed has been broken by recent updates in Strapi, so some deeply nested fields are not imported while seeding.

I'm trying to fix it in #1866, but I'm stuck on another Strapi bug. Meanwhile we need to select those submenu manually. I'm going to do it for you in this PR 👍

jarstelfox:
un_dev_block 🌵

jyee27 (approved):
CR ☕ through [6f47191](https://github.com/iFixit/react-commerce/pull/1950/commits/6f47191f1a63fdc1860988e04f5046d77e4e27e1)

erinemay:
QA 👍 
The /Troubleshooting/Device page works in the dev preview and 404s in the prod preview.

Smoke test: 
I compared the following to main. The only differences I saw were those appropriate to a couple recent changes that haven't been merged in to this branch yet and the differences between strapi contents.
- Collection pages
   - item type facets, regular facets
   - children pages
   - breadcrumbs and category diving
   - Parts and Tools
   - search
- /Store 
- Product pages
   - adding to and modifying cart
   - checking out
   - viewing and interacting with the page
- Problem pages with various solution card contents (guide and product links)

>I'm trying to fix it in https://github.com/iFixit/react-commerce/pull/1866, but I'm stuck on another Strapi bug. Meanwhile we need to select those submenu manually. I'm going to do it for you in this PR 👍

Thank you @dhmacs 💙

jarstelfox:
@dhmacs and @federicobadini I am going to merge and will be out until Monday. If for any reason there is a concern about the structure of the project, let's handle those as follow-ups.

federicobadini:
Just for completeness, the error pointed out by sentry in the previous comment was already fixed in a subsequent PR

jarstelfox:
I actually believe the bugs shown came from my dev environment. More explanation here: https://github.com/getsentry/sentry/discussions/49996#discussioncomment-6807526

Either way the errors here are not concerning

PR #1953, opened 2023-09-01
Vulcan: Show `Causes` list on desktop and mobile

andyg0808:
I'd not realized that the ToC wasn't shipped yet, so we didn't want to hide the `Causes` list on desktop yet.

## QA Notes
Please confirm that the Causes list is visible on both Desktop and mobile

Closes https://github.com/iFixit/ifixit/issues/49627


jarstelfox (approved):
CR 🌵

andyg0808:
QA :+1: 
I confirmed that this makes the Causes list visible on Desktop.

PR #1954, opened 2023-09-01
Shopping Cart: Add cart-view page

danielbeardsley:
Create most of the cart view page according to the figma design.

This started as copying the CartDrawer (re-copied this morning) and modifying it as needed to match the figma. I also extracted a few bits that weren't likely to change.

I've opted in many cases to copy the functionality because keeping one component that can work in both places would require compromising on styling a fair amount, or increase complexity (adding a page and drawer mode).

### Features

* Cart contents
* Summary / totals
* Checkout button
* Cross sell (borrowed component from Drawer)
* Cart Alerts

This is missing some of the sections from the design and I imagine those will come as future pulls but I want to get this in the code base now as the `CartDrawer` continues to evolve.

Connect iFixit/ifixit#48798

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,1 +1,3 @@
-{}
+{
+   "typescript.tsdk": "node_modules/typescript/lib"
+}
--- a/packages/cart-sdk/hooks/use-cart.ts
+++ b/packages/cart-sdk/hooks/use-cart.ts
@@ -6,6 +6,7 @@ import {
 } from '@ifixit/helpers';
 import { useIFixitApiClient } from '@ifixit/ifixit-api-client';
 import { useIsMutating, useQuery } from '@tanstack/react-query';
+import { useMemo } from 'react';
 import {
    APICart,
    Cart,
@@ -33,6 +34,26 @@ export function useCart() {
       enabled: !isMutating,
    });
    return query;
+}
+
+export function extractCrossSells(cart: Cart | null | undefined) {
+   return useMemo(
+      () =>
+         cart
+            ? cart.crossSellProducts
+                 .filter((item) => {
+                    const isAlreadyInCart =
+                       item &&
+                       cart.lineItems.find(
+                          (lineItem) => lineItem.itemcode === item.itemcode
+                       );
+                    if (isAlreadyInCart) return null;
+                    return item;
+                 })
+                 .sort((a, b) => a.handle.localeCompare(b.handle))
+            : [],
+      [cart]
+   );
 }
 
 function isValidCartPayload(data: any): data is CartAPIResponse {
--- a/packages/cart-sdk/hooks/use-cart.ts
+++ b/packages/cart-sdk/hooks/use-cart.ts
@@ -36,7 +36,9 @@ export function useCart() {
    return query;
 }
 
-export function extractCrossSells(cart: Cart | null | undefined) {
+export function useCrossSells() {
+   const cartQuery = useCart();
+   const cart = cartQuery.data;
    return useMemo(
       () =>
          cart
--- a/packages/ui/cart/view/CartLineItem.tsx
+++ b/packages/ui/cart/view/CartLineItem.tsx
@@ -102,10 +102,10 @@ export function CartLineItem({ lineItem }: CartLineItemProps) {
             >
                {lineItem.name}
             </Link>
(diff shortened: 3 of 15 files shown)


packages/ui/cart/view/CartLineItem.tsx
```diff
@@ -0,0 +1,188 @@
...
+         w="full"
+         justify="space-between"
+         p="3"
+         spacing="3"
+         align="stretch"
+      >
+         <CartLineItemImage src={lineItem.imageSrc} alt={lineItem.name} />
+         <VStack align="flex-start" spacing="0" flexGrow="1">
+            <Link
+               href={`${appContext.ifixitOrigin}/Store/Product/${lineItem.itemcode}`}
```
sterlinghirsh, line 97: What is this link? Should this be /products/{handle}? Or is there a url property on the lineItem?

danielbeardsley: Good catch. I'm not sure, though it's the same as the current cart-drawer:

https://github.com/iFixit/react-commerce/blob/7726466185b364acee4b92f2e7a9124ade6e724d/packages/ui/cart/drawer/CartLineItem.tsx#L95

Looks like a redirecter: https://www.ifixit.com/Store/Product/IF145-307-4

Though... I can't yet figure out where it's from.

sterlinghirsh:
CR :zap: dev_block :zap: on verifying the product link. Also I want to make sure that our header link to the cart isn't a next link that will start trying to serve this. Otherwise this looks like a good start on the cart.

packages/ui/cart/view/CartLineItem.tsx
danielbeardsley: Definitely php code because I see varnish headers in the response.

danielbeardsley: I logged the route and got here:
image

Which ends up here for the redirect:
https://github.com/iFixit/ifixit/blob/72d86ea35f0b850df0848070df8eb5980f613d00/Cart/UI/parts_store.php#L120

Crazy.

packages/cart-sdk/hooks/use-cart.ts
```diff
@@ -35,6 +36,26 @@ export function useCart() {
    return query;
 }
 
+export function extractCrossSells(cart: Cart | null | undefined) {
+   return useMemo(
```
dhmacs, line 40: I noticed that this function contains a hook, therefore [by React convetions it should be a hook itselves](https://legacy.reactjs.org/docs/hooks-rules.html#only-call-hooks-from-react-functions).

I'd take this opportunity to suggest to refactor like this (taking advantage of context, so where we use this we don't have to use another hook to get the cart):

```suggestion
export function useCartCrossSells() {
   const cart = useCart();
   return useMemo(
```

packages/ui/cart/view/CartLineItem.tsx
dhmacs: I think this was here before we had the product pages in next.js and has never been updated. It think we can safely refactor it now 👍

dhmacs (commented):
Great job @danielbeardsley 👏 I left a couple of suggestions. 

What do you think about moving this work under the app router? Seems like the work in #1950 is going to get merged soon, and this looks like a good page to get started with the app router

.vscode/settings.json
```diff
@@ -1 +1,3 @@
-{}
+{
+   "typescript.tsdk": "node_modules/typescript/lib"
+}
```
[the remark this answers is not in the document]
danielbeardsley: This tells vscode to use the typescript (for tsserver) version from the repo instead of the one that's bundled in vscode. 

See: becd0b5e8856381703f48225e92e0992660d8040

packages/cart-sdk/hooks/use-cart.ts
danielbeardsley: Ok. I'm always a little leery about the performance effects of adding additional calls to contexts when the values we wan't already in scope. That said, I appreciate the encapsulation.

I'll go for it.

sterlinghirsh:
This has been paused because some other projects became higher priority. I agree that we should move this under the App router whenever we pick this up.

danielbeardsley:
This can be re-opened (if desired) in this code's new home

PR #1955, opened 2023-09-04
Enable Google genuine part banner logo

federicobadini:
As per title, this PR enables the Google genuine part banner logo, that was still commented out in code.

qa_req 0

### DIFF
--- a/frontend/templates/product/sections/ProductOverviewSection/GenuinePartBanner.tsx
+++ b/frontend/templates/product/sections/ProductOverviewSection/GenuinePartBanner.tsx
@@ -1,5 +1,6 @@
 import {
    Crucial,
+   Google,
    Hp,
    Lenovo,
    Logitech,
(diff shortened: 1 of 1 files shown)


frontend/templates/product/sections/ProductOverviewSection/GenuinePartBanner.tsx
```diff
@@ -25,7 +26,7 @@ export type GenuinePartBannerProps = {
 
 const partnerCodeToComponentMap: { [key: string]: React.FC } = {
    // Don't have permission yet
-   // google_pixel: Google,
+   google_pixel: Google,
```
masonmcelvain, line 29: I might be missing context here - do we have permission from Google to use this logo now? If so, we should get rid of the `Don't have permission yet` comment too.

masonmcelvain (approved):
CR ✌🏻

PR #1958, opened 2023-09-05
Sentry: Tag production deployment status

### LINKED ISSUE #1942: Error: GraphQL query failed to execute: Bad Request
Sentry Issue: [REACT-COMMERCE-3EF](https://ifixit.sentry.io/issues/4421858774/?referrer=github_integration)

```
Error: GraphQL query failed to execute: Bad Request
  File "app:///_next/server/chunks/478.js", line 3890, in requester
    throw new Error(`GraphQL query failed to execute: ${response.statusText}`);
  File "app:///_next/server/chunks/478.js", line 4028, in findProductList
    const result = await strapi_sdk/* strapi.getProductList */.s1.getProductList({
  File "app:///_next/server/pages/Tools.js", line 54, in getServerSideProps
    const [globalSettings, stores, currentStore, productList] = await Promise.all([(0,_models_global_settings__WEBPACK_IMPORTED_MODULE_4__/* . {snip}
...
(4 additional frame(s) were not displayed)
```

masonmcelvain:
Closes #1942

Adds a boolean tag (`current_production`) to all Sentry events indicating whether the deployment is the current production deployment.

## QA

Throw an error somewhere in local dev. Make sure the new tag (`current_production`) shows up in Sentry. It should be `false` for local dev.

### DIFF
--- a/frontend/sentry.client.config.ts
+++ b/frontend/sentry.client.config.ts
@@ -10,8 +10,15 @@ const SENTRY_DSN = process.env.NEXT_PUBLIC_SENTRY_DSN;
 
 Sentry.init({
    async beforeSend(event) {
-      const current_production = await isCurrentProductionDeployment();
-      event.tags = { ...event.tags, current_production };
+      try {
+         const current_production = await isCurrentProductionDeployment();
+         event.tags = { ...event.tags, current_production };
+      } catch (e) {
+         event.extra = {
+            ...event.extra,
+            'Exception Checking Production Deployment': e,
+         };
+      }
       return event;
    },
    dsn: SENTRY_DSN,
--- a/frontend/sentry.client.config.ts
+++ b/frontend/sentry.client.config.ts
@@ -2,12 +2,18 @@
 // The config you add here will be used whenever a page is visited.
 // https://docs.sentry.io/platforms/javascript/guides/nextjs/
 
+import { isCurrentProductionDeployment } from '@helpers/vercel-helpers';
 import * as Sentry from '@sentry/nextjs';
 import { BrowserTracing } from '@sentry/browser';
 
 const SENTRY_DSN = process.env.NEXT_PUBLIC_SENTRY_DSN;
 
 Sentry.init({
+   async beforeSend(event) {
+      const current_production = await isCurrentProductionDeployment();
+      event.tags = { ...event.tags, current_production };
+      return event;
+   },
    dsn: SENTRY_DSN,
    integrations: [new BrowserTracing()],
    sampleRate: 1.0,
--- a/frontend/sentry.client.config.ts
+++ b/frontend/sentry.client.config.ts
@@ -14,6 +14,7 @@ Sentry.init({
          const current_production = await isCurrentProductionDeployment();
          event.tags = { ...event.tags, current_production };
       } catch (e) {
+         event.tags = { ...event.tags, before_send_error: true };
          event.extra = {
             ...event.extra,
             'Exception Checking Production Deployment': e,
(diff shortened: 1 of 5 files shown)


jarstelfox (dismissed):
CR 🌵 but deploy_block 🌵 I think it would be worth while to add the sha as a release as we do in the PHP app: 

https://github.com/iFixit/ifixit/blob/596fe7a6879b33d41ef0c82eb609a53561064b89/iFixit/Framework/Errors/SentryReporter.php#L74
https://github.com/iFixit/ifixit/blob/596fe7a6879b33d41ef0c82eb609a53561064b89/iFixit/Framework/Errors/SentryReporter.php#L263-L273

masonmcelvain:
>  I think it would be worth while to add the sha as a release as we do in the PHP app

I omitted it here because it seems that Sentry is automatically detecting the sha and assigning it to a release tag.

That said, I'm happy to do it explicitly. dev_block ✌🏻

jarstelfox:
>it seems that Sentry is automatically detecting the sha and assigning it to a release tag.

I'd leave it and un_deploy_block 🌵.

The main reason is assigning commits to a release is an extra step or two that requires a GH action on deploy. If it just works let's keep it a non-issue 👍🏻

masonmcelvain:
Ok! un_dev_block ✌🏻 

I'd also like your opinion on whether a tag is suitable here, or if we should drop the event entirely (i.e. don't report to Sentry if it's a non-current production deployment). I thought that was a bit extreme, and the tag empowers us to filter in the dashboard at least.

masonmcelvain:
dev_block ✌🏻 @erinemay we're going to take a different approach

https://ifixit.slack.com/archives/C01GMQXJC95/p1693937453575549

danielbeardsley:
I feel pretty weird having our code curl github before reporting every error.

Then again, I'm not sure of another way to get this done. Perhaps we limit each page load to reporting a single error? Perhaps we cache the "current production" value for a few minutes?

masonmcelvain:
> I feel pretty weird having our code curl github before reporting every error.
> 
> Then again, I'm not sure of another way to get this done. Perhaps we limit each page load to reporting a single error? Perhaps we cache the "current production" value for a few minutes?

It is weird, but I can't think of an alternative solution with the [system flags](https://vercel.com/docs/projects/environment-variables/system-environment-variables) Vercel exposes to us.

I tried implementing the caching solution you suggested.

masonmcelvain:
un_dev_block ✌🏻 We ended up deciding this solution is good enough for now (though not ideal). Maybe we'll change the Sentry release to a datetime or timestamp in another PR.

frontend/sentry.client.config.ts
```diff
@@ -2,12 +2,25 @@
...
 Sentry.init({
+   async beforeSend(event) {
+      try {
+         const current_production = await isCurrentProductionDeployment();
+         event.tags = { ...event.tags, current_production };
+      } catch (e) {
+         event.extra = {
+            ...event.extra,
+            'Exception Checking Production Deployment': e,
+         };
```
jarstelfox, line 20: Minor: I would also add a new tag indicating that the `event.extra.'Exception Checking Production Deployment'` exists.

Something like: `SecondaryError: true`

Just makes it easy to search

masonmcelvain: Good idea! https://github.com/iFixit/react-commerce/pull/1958/commits/053774c826fb44d90ffbcad910581ee4c79d346a

jarstelfox (approved):
CR 🌵

erinemay:
QA 👍 I caused an error and saw the new tag present: 
image

It was false for local_dev

masonmcelvain:
We got a `before_send_error` on this [event](https://ifixit.sentry.io/share/issue/6b81069c611d47dc913c0b30b1fc3141/), but the exception we logged was an empty object. Not sure what to make of it 🤷🏻 

image

Issue #772, opened 2022-10-04
FAQ Epic


sterlinghirsh:
We want to add an FAQ section on product lists, but we don't want to have to make thousands of copies of the same questions and answers. We'd like to do this in a way that allows us to use inheritance to reuse the same FAQs down the collection tree.

Examples: https://docs.google.com/spreadsheets/d/1gJiMZnz6pmzzlItFBGM-YJSWIJRNJ7QgpI0XXvdeQFk/edit#gid=0

* Add a new collection type FAQs to Strapi
  * Field list
    * Category (e.g. Safety, Quality)
      * This is a key that is used for inheritance. We will only show one FAQ from each Category. Deeper devices override shallower devices.
      * e.g. iPhone 7's Quality FAQ will override Phone's Quality FAQ.
    * Product List (reference to Product List collection type, possibly one-to-many)
      * Connect the highest collection in the tree. Child collections will look for ancestors' FAQs too.
    * Product List Filter (string?)
      * A facet filter similar to an algolia filer that indicates whether we should display a certain FAQ.
      * e.g. item_type="Batteries" AND part_or_kit="Fix Kit"
    * Question (single-line text e.g. "How are iPhone Screens tested?")
    * Answer (rich text)
    * Sort Order (int)
* Update our Strapi collection query to also fetch associated FAQs
* Display FAQs on collection pages

jeffsnyder:
@sterlinghirsh I'd like to prioritize this. What spec(s) need to be added for @dhmacs to take over?

dhmacs:
@jeffsnyder I think what @sterlinghirsh has written is very clear and detailed, so it's only a matter of prioritizing this over other work. Also if you're expecting me or Federico to do it it should be on the iOP backlog since we first look into that column when taking new tasks. My current focus in on getting the caching stuff merged, but I'm happy to fill in voids with this. I'll let you and Sterling assess the priority, just let me know.

jeffsnyder:
@ifixitmaxb and I were chatting through the FAQ plan. Question: will we have FAQ as a manually definable section on ProductList pages? Or will we _only_ be assigning FAQs based on what we build in Assigning ProductList and ProductDetail Segments to Many Pages https://github.com/iFixit/react-commerce/issues/1746 ? 

[...]

PR #1961, opened 2023-09-06
Add placement fields to Strapi FAQ

dhmacs:
Part of #1940 
qa_req 0

This PR expands the Strapi FAQ collection type with the fields that are necessary to accomplish an MVP for #772.
Allowing complex filtering capabilities to determine whether an FAQ should be visible or not in a product list is tricky and can result in a complex interface for the editors. 
We've explored many solutions and during last week iOP meeting we agreed on getting started with a simple version that only allows to apply an item type filter (e.g. an FAQ that only applies to "Batteries"). This will help us ship an initial implementation and iterate afterward on filtering capabilities as needed.

### DIFF
--- a/backend/src/api/faq/content-types/faq/schema.json
+++ b/backend/src/api/faq/content-types/faq/schema.json
@@ -4,7 +4,8 @@
    "info": {
       "singularName": "faq",
       "pluralName": "faqs",
-      "displayName": "FAQ"
+      "displayName": "FAQ",
+      "description": ""
    },
    "options": {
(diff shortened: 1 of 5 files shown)


masonmcelvain (approved):
CR ✌🏻

masonmcelvain:
The strapi deploy failed for an unrelated reason, we are working on it
https://ifixit.slack.com/archives/C8KBFH1JT/p1694027011620599

masonmcelvain:
Strapi has been redeployed ✔️

PR #1962, opened 2023-09-06
Ignore: Testing Crowdin GH Integration

ifixit-systems:


aburke07:
Whoops. We should turn off this integration so it stops opening PRs.

PR #1965, opened 2023-09-07
Improve hello world app router

federicobadini:
This PR moves from the previous hello-world-app-router one (https://github.com/iFixit/react-commerce/pull/1950) and:

- removes the 'use client' directive inside `(defaultLayout)/layout.tsx` since that was meant to be a server component
- explicitly mark the page.tsx as an async server component, since there we will need to perform data fetching
- refactor a bit the file structure

qa_req 0

### DIFF
--- a/frontend/app/(defaultLayout)/Troubleshooting/[device]/components/ServerHeader.tsx
+++ b/frontend/app/(defaultLayout)/Troubleshooting/[device]/components/ServerHeader.tsx
@@ -1,0 +1,7 @@
+export type ServerHeaderProps = {
+   device: string;
+};
+
+export default function ServerHeader({ device }: ServerHeaderProps) {
+   return <span>Device: {device}</span>;
+}
--- a/frontend/app/(defaultLayout)/Troubleshooting/[device]/page.tsx
+++ b/frontend/app/(defaultLayout)/Troubleshooting/[device]/page.tsx
@@ -1,10 +1,11 @@
-import { Metadata } from 'next';
-import PageTemplate from './components/pageTemplate';
-import { RestrictRobots } from '@helpers/next-helpers';
-import { notFound } from 'next/navigation';
 import { flags } from '@config/flags';
+import { RestrictRobots } from '@helpers/next-helpers';
 import { getiFixitOrigin } from '@helpers/path-helpers';
+import { Metadata } from 'next';
 import { headers } from 'next/headers';
+import { notFound } from 'next/navigation';
+import Header from './components/Header';
+import ServerHeader from './components/ServerHeader';
 
 export type PageParams = {
    device: string;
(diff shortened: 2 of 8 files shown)


masonmcelvain (approved):
CR ✌🏻

frontend/app/(defaultLayout)/Troubleshooting/[device]/components/ServerHeader.tsx
Comment (jarstelfox) [inline frontend/app/(defaultLayout)/Troubleshooting/[device]/components/ServerHeader.tsx:1]: Why did you change the name here? We intend to update the naming and types on this page as we build it out, the change just seems unnecessary to me.

Also changing it back is unnecessary as well

frontend/app/(defaultLayout)/Troubleshooting/[device]/page.tsx
```diff
@@ -19,7 +20,12 @@ export default function Page({ params, searchParams }: PageProps) {
    ensureFlag();
 
    const pageProps = getPageProps({ params, searchParams });
-   return <PageTemplate {...pageProps} />;
+   return (
+      <>
+         <Header {...pageProps} />
+         <ServerHeader {...pageProps} />
```
Comment (jarstelfox) [inline frontend/app/(defaultLayout)/Troubleshooting/[device]/page.tsx:26]: Why is there a server header component vs a client component? This seems very unnecessary

Comment (jarstelfox) [reply frontend/app/(defaultLayout)/Troubleshooting/[device]/page.tsx:26]: I should clarify that the component was still rendered server side even with a use client statement

Comment (federicobadini) [reply frontend/app/(defaultLayout)/Troubleshooting/[device]/page.tsx:26]: The component does not have any additional functional purpose.
I intentionally created a duplicated component just to underline the difference between a server component and a server rendered client component. 
It might be not so obvious for anyone, since RSC are a pretty new topic, so I'll use your question to build a little bit more context for anyone that might end up reading this.

When Next.js receives a request, it performs a complete server-rendering of the component hierarchy. This process involves both server components and client components.
The real difference from RSC and RCC lies in the payload that is sent to the client: if you inspect the bundle, you would notice how the `ServerHeader` component is not even included in the bundle, while the `Header` component is.
Even though the impact of a single component might be tiny, we have to consider that also dependencies might or might not be included according to a component being a RSC or a RCC.

Therefore - whenever possible - we should aim to use client components only when needed and especially on leaf nodes of the component hierarchy.

If you consider the current ecommerce UI, several parts if not most parts could be made by RSC.
The real blocker at the moment is Chakra UI, that is not compatible with RSC, due to the fact that handles state internally. So, me and @dhmacs are probably going to iterate a bit more on this hello-world-app-router branch to test alternatives that could allow a wider adoption of RSC (Tailwind css)

Screenshot 2023-09-08 at 11 21 52
Screenshot 2023-09-08 at 11 22 00

frontend/app/(defaultLayout)/Troubleshooting/[device]/components/ServerHeader.tsx
Comment (federicobadini) [reply frontend/app/(defaultLayout)/Troubleshooting/[device]/components/ServerHeader.tsx:1]: Just for the duplication mentioned in the comment below.

PR #1976, opened 2023-09-13
TOC: Polish 2

jarstelfox:
QA
---

* Update feedback on https://github.com/iFixit/ifixit/discussions/48967#discussioncomment-6870061
* Ensure the Prod version looks around the same

>1. First steps only seems to be highlighted in the TOC when the header First Steps is visible. It makes for a brief flash of highlighting as you scroll down. It looks like the other sections are highlighted when their headers are out of view.

Mostly solved. This was due to the causes list appearing when the TOC was enabled on desktop. The causes list will now only appear on mobile

>2. Related problems goes nowhere and is never highlighted. I assume this is WIP, but I'm not excluding anything here.

Solved.

>3. Scrolling issue in comment above is present on desktop as well.

Not solved. This is due to the browser controlling scroll vs my code controlling scroll. If you click a header, the browser determines which pixel needs to be shown. If you click the TOC, I am determining which pixel to scroll into view.

>4. Mobile First steps #Section_First_Steps doesn't accommodate the height of the header, so it's obscured by the TOC

I am not sure this is something we can solve for. Namely, the math to scroll to an item does not take into account the height of the header. In fact, it does not know about a header existing.

>5. I don't love it when cut off text can be so close to other cut off text.

I added back a fade on mobile, but we can remove it if Mattia is not a fan. Personally, I like it as well.

>6. I can see a pixel or two of scrolling above the TOC.

I think this is solved now that we no longer have the scroll percentage at the top of the page

PR #1980, opened 2023-09-18
Troubleshooting Toc: Turn on

jarstelfox:
### QA

Let's do a cross-browser smoke test of the TOC on desktop and mobile.


jarstelfox:
deploy_block 🌵 on #1976

andyg0808 (approved):
CR :+1:

erinemay:
Checked the following URLs:

- https://react-commerce-prod-git-toc-turn-on-ifixit.vercel.app/Troubleshooting/Whirlpool_Refrigerator/Door+Not+Closing/509237
- https://react-commerce-prod-git-toc-turn-on-ifixit.vercel.app/Troubleshooting/Google_Phone/Google+Pixel+Overheating/479437
- https://react-commerce-prod-git-toc-turn-on-ifixit.vercel.app/Troubleshooting/Chainsaw/Overheating/484453
- https://react-commerce-prod-git-toc-turn-on-ifixit.vercel.app/Troubleshooting/Television/TV+Will+Not+Turn+On/489001

On the following browsers:
- Desktop
   - Windows + Chrome
   - Mac + Safari
- Mobile
   - Android + Chrome
   - iOS + Safari
-  Tablet
   - iOS + Safari

QA 🚀 Overall, it looks and feels great! @jarstelfox thanks for sticking with this through all the nitty feedback 🙌 
- Scrolling looks and feels good (scrolling a few causes away, from bottom to top,, and top to bottom) 
- Clicking on the cause links in the TOC takes me to a nice spot on all browsers (can see the cause title)
- Using and scrolling through the mobile TOC looks and feels good
- What's highlighted is intuitive for short and very long causes

deploy_block 🌵 on another case of the scroll getting interrupted. 

https://react-commerce-prod-git-toc-turn-on-ifixit.vercel.app/Troubleshooting/Chainsaw/Overheating/484453
This is the Google Pixel Chrome. I can reliably reproduce it by going to the introduction and then to chain too tight.
I can't reproduce it on main https://react-commerce-prod.vercel.app/Troubleshooting/Chainsaw/Overheating/484453

https://github.com/iFixit/react-commerce/assets/72166715/a64b5944-719a-4c75-8406-e15967b0caa0

Also, I'm starting to think it's worth it to override the browser's scrolling location. Every time I opened a page on mobile, my initial reaction was to click on a cause link. After reaching a cause link, I ended up using the TOC, but my first click always took me to a partially hidden cause. Not sure if it needs to be a blocker for release. 
image

jarstelfox:
>Also, I'm starting to think it's worth it to override the browser's scrolling location. Every time I opened a page on mobile, my initial reaction was to click on a cause link.

Fixed!

I spent a lot of time digging into this and found an easy way! The browser's scroll event will always fire, but if we fire another scroll event ours will take precedence. Thus, I am looking at the hash to match a record in our toc code. If we find it, then we call the nice scroll logic, otherwise the browser does it's thing

>another case of the scroll getting interrupted.

Fixed!

Whew! I think I found the real cause of the bug I have been seeing and found a real solution to the problem. I was able to remove my hacky code and all seems to be fixed for me.

frontend/templates/troubleshooting/toc.tsx
```diff
@@ -105,45 +106,36 @@ export function MobileTOC({
...
+
    return (
-      <Collapse in={Boolean(showMobileTOC && activeItem)} unmountOnExit={true}>
-         <Flex {...props}>
-            <MobileTOCMenu activeItem={activeItem} items={items} />
-         </Flex>
-      </Collapse>
+      <AnimatePresence>
+         {isVisible && (
+            <Flex
```
andyg0808, line 132: > Note: Direct children must each have a unique key prop so AnimatePresence can track their presence in the tree.

I'm guessing this doesn't matter because there's only one child, but this may be worth being aware of (the simple demos don't seem to include `key`, so I'm guessing it's fine in this case.)

https://www.framer.com/motion/animate-presence/##exit-animations

frontend/templates/troubleshooting/toc.tsx
```diff
@@ -144,6 +144,27 @@ export function MobileTOC({
...
+
+   useEffect(() => {
+      if (isMounted) {
+         return;
+      }
+
+      scrollTo && scrollTo();
+
+      setIsMounted(true);
+   }, [isMounted, scrollTo]);
```
andyg0808, line 165: Pretty sure this'll work, because `useEffect` runs on mount, and passing no dependencies means we never re-run:
```suggestion
   useEffect(() => {
      scrollTo && scrollTo();
}, []);

andyg0808 (commented):
CR :+1: 
deploy_block :+1: on a couple comments.

andyg0808 (approved):
CR :+1:

frontend/templates/troubleshooting/toc.tsx
jarstelfox: Yea, I saw that, but everything seemed fine when testing. I saw their demos do not as well.

frontend/templates/troubleshooting/toc.tsx
jarstelfox: Oh, that's a good call. I believe that is the same. Let me double check and I'll commit if so

kthaler:
I've confirmed everything mentioned in this comment https://github.com/iFixit/react-commerce/pull/1980#issuecomment-1728566523

> on another case of the scroll getting interrupted.
https://react-commerce-prod-git-toc-turn-on-ifixit.vercel.app/Troubleshooting/Chainsaw/Overheating/484453
This is the Google Pixel Chrome. I can reliably reproduce it by going to the introduction and then to chain too tight.
I can't reproduce it on main https://react-commerce-prod.vercel.app/Troubleshooting/Chainsaw/Overheating/484453

This is fixed on Pixel Chrome and I have not seen this issue on any other browser that @erinemay listed above.

QA 👍 

I can also confirm that the TOC feels really good and responsive, and the scroll is not being interrupted.

deploy_block 🌵 

I am encountering a bug on iPhone Safari where the table of contents gets caught on screen when you tap and hold something in the TOC dropdown then begin swiping in the main page. I can recreate this fairly reliably.

https://github.com/iFixit/react-commerce/assets/145375039/4b10397b-1262-4677-b0eb-7608b2b3b90b

jarstelfox:
>I am encountering a bug on iPhone Safari 

Nice find! Do you have details of the iOS version?

kthaler:
> > I am encountering a bug on iPhone Safari
> 
> Nice find! Do you have details of the iOS version?

iPhone 12 
iOS Version 16.3.1

jarstelfox:
un_deploy_block 🌵 I think I fixed that issue

andyg0808 (approved):
CR :+1:

kthaler:
deploy_block 🦖 

I'm still encountering the issue, but I think I have some more understanding of the problem. I've noticed the issue only arrises when the TOC menu is scrolled all the way up. The issue will not occur if the TOC menu is scrolled down at all. 


https://github.com/iFixit/react-commerce/assets/145375039/00a80144-0f78-4ec0-9c72-e4c0abfea9fc

Also, the highlighting of options in the drop down can be a bit wonky. I am playing around with tapping and holding elements in the dropdown and some time it stays highlighted in blue or grey. Ex:

https://github.com/iFixit/react-commerce/assets/145375039/f962dce5-ed0d-4f03-bc0e-0fddd6961d27

jarstelfox:
oh! I misunderstood the bug you were reporting. (not your fault, it was mine from watching the video)

> I've noticed the issue only arises when the TOC menu is scrolled all the way up. The issue will not occur if the TOC menu is scrolled down at all.

I think that's acceptable. Basically, we do have control to close the mobile TOC if you scroll, but I'm not convinced it's bad enough behavior to require adding 10-15 lines to deal with this state.

Feel free to push back if you think it is annoying / a bad experience and I will totally take a look

ianrohde (approved):
CR ⚪

kthaler:
I went ahead and tested on multiple browsers again and I've confirmed everything noted in this comment: https://github.com/iFixit/react-commerce/pull/1980#issuecomment-1728566523

QA 👍  

Regarding the bug I mentioned here https://github.com/iFixit/react-commerce/pull/1980#issuecomment-1732127057 , I don't think the issue is serious enough for extra dev time, I believe the UI is still intuitive even when the issue occurs. 

deploy_block 🦖 

I'm now encountering the bug where the scroll on Android devices get interrupted. 
I've seen this issue on Pixel 6 + Chrome and Samsung Galaxy S23 + Chrome

https://github.com/iFixit/react-commerce/assets/145375039/cabd96d4-0072-4930-9a4c-327f24c189c1


I'm also seeing the issue noted here by @erinemay 
> Also, I'm starting to think it's worth it to override the browser's scrolling location. Every time I opened a page on mobile, my initial reaction was to click on a cause link. After reaching a cause link, I ended up using the TOC, but my first click always took me to a partially hidden cause. Not sure if it needs to be a blocker for release.

https://github.com/iFixit/react-commerce/assets/145375039/71c985e8-2699-4dd5-a8a1-a112711cad79

jarstelfox:
>I'm also seeing the issue noted here by @erinemay

I decided to tackle this. The causes links were not set to scroll in the exact same way before, but now they will.

>I'm now encountering the bug where the scroll on Android devices get interrupted.
>I've seen this issue on Pixel 6 + Chrome and Samsung Galaxy S23 + Chrome

I was able to repro, but it stopped after a "hard refresh" so I suspect this is something to due with vercel's dev preview.

@kthaler, I am going to claim it's fixed. Mind testing once more after clearing the browser cache if possible?

ianrohde (approved):
CR ⚪

kthaler:
> I decided to tackle this. The causes links were not set to scroll in the exact same way before, but now they will.

I tested this on iPhone and Samsung and confirm that this is fixed. 

> I was able to repro, but it stopped after a "hard refresh" so I suspect this is something to due with vercel's dev preview.

The issue also fixed for me after a hard reset, so I agree that this is an issue with the preview. 

After another pass through of all the tests, I confirm that everything is working as expected, and all the known bugs/issues are negligible and don't significantly effect the user experience.

QA 🦕   - Overall the TOC feels great and is very responsive!

jarstelfox:
>The issue also fixed for me after a hard reset, so I agree that this is an issue with the preview.

Sweet in that case un_deploy_block 🌵

PR #1977, opened 2023-09-14
Tests: Add login/logout test

batbattur:
## Summary
Added a new test that logs in and logs out as a test user.

Created an actual test user account on both the live site and
cominor with default user privileges to use for this test.

## QA notes
Passing CI is enough. 

closes #1742

Issue #1742, opened 2023-06-07
E2E: Add a test for login/logout


masonmcelvain:
Add a playwright test that verifies users can login and logout. We have tests for this in the PHP app, and we want to make sure that users can do it on Next.js too.

Consider checking that login state is synchronized with www.cominor.com.

PR #1981, opened 2023-09-18
Tests: Use dev-api-psk in CI workflows

batbattur:
## Summary
This adds `dev-api-psk` cookie in tests using the `beforeEach` method.
Previously, it appears that we were using `storageState` in [`generateBrowserConfig()`](https://github.com/iFixit/react-commerce/pull/1661)
to set the cookie but it wasn't working now, so dropped it.

Updated the test workflows to use the `DEV_API_AUTH_TOKEN` env. 
In github, the `DEV_API_AUTH_TOKEN` is added as a secret.

## QA 
Only modified the ci/testing code. Passing CI is enough. 

connects #1977 https://github.com/iFixit/react-commerce/issues/1742


mlahargou (approved):
CR 📱 

Nice job figuring this out!

batbattur:
QA 🚛 ci is still passing

PR #1982, opened 2023-09-21
Upgrade ts to 5.2.2

federicobadini:
This PR upgrade TS to 5.2.2.
The reason for this is that:
> To use an async Server Component with TypeScript, ensure you are using TypeScript 5.1.3 or higher and @types/react 18.2.8 or higher.

However, Chakra typings seems to be not 100% compliant with ts5.
See [here](https://github.com/chakra-ui/chakra-ui/pull/7506) and [here](https://github.com/chakra-ui/chakra-ui/issues/7459). The only working combination of chakra, react, react-dom etc. packages that I found was not producing compilation errors or a bunch of IDE errors is the one in this PR, which uses @types/react 18.2.0.

Moreover, we had to upgrade Chakra, and even though no new major was released, a [breaking change was shipped](https://github.com/chakra-ui/chakra-ui/issues/7269#issuecomment-1424868672) so also this was taken into account in this PR.

### QA
1. Visit vercel preview
2. Verify that nothing visible was changed in the ecommerce ui (Troubleshooting, Product, Productlist)
3. Verify that on local dev environment no new errors wrt the past are shown


federicobadini:
dev_block 🔴  on some e2e tests failing after merging main

federicobadini:
Tests were failing because `useBreakpointValue` had a few usages with option `{ ssr: false }` in a ssr context.
The previous chakra version was not failing due to differences in how it referenced window.

un_dev_block✅

deltuh-vee:
QA 🎬 
* Ecommerce UI still looks the same.
* No new errors on localdev.

frontend/templates/troubleshooting/index.tsx
```diff
@@ -836,7 +836,7 @@ function IntroductionSection({
    intro,
    ...headingProps
 }: { intro: Section } & HeadingProps) {
-   const bufferPx = useBreakpointValue({ base: -46, lg: -6 }, { ssr: false });
+   const bufferPx = useBreakpointValue({ base: -46, lg: -6 });
```
masonmcelvain, line 839: cc @jarstelfox since it seems like you intentionally added this setting in https://github.com/iFixit/react-commerce/commit/8b32868db2ac411f67e486cff399f8395d31d60f

masonmcelvain (approved):
CR 👏🏻  Thanks for taking this on!

deploy_block 🌵 on @jarstelfox getting a chance to look at the `useBreakpointValue` change

frontend/templates/troubleshooting/index.tsx
jarstelfox: Yea, if we generate the px serverside it defaults to the mobile value which means it the useEffect that checks the hash and scrolls to the correct ID on load will scroll to the incorrect position on first-page load.

Ex: Load https://www.ifixit.com/Troubleshooting/Television/TV+Has+Sound+But+No+Picture/493422#Section_Hardware_Issues in a new tab.

The scroll is not actually to the ID, but instead a custom scroll that takes into account the UI (padding above the text). It's not a huge deal, but this may break to nice scroll position behavior only on the first load

masonmcelvain: This is the difference I see on `main` vs this feature

Main
[main.webm](https://github.com/iFixit/react-commerce/assets/52104630/734deae5-c36f-42bd-a3b2-30cc967c9efd)

Feature
[feature.webm](https://github.com/iFixit/react-commerce/assets/52104630/e1dbdd90-f88c-41cc-977f-2504e9733918)

masonmcelvain:
un_deploy_block 👍🏻 since the potential UX fallout isn't very noticeable and we can iterate on the scroll feature as needed

frontend/templates/troubleshooting/index.tsx
jarstelfox: Yea, I agree the SSR may not have been needed here

PR #1987, opened 2023-09-22
Tests: skip product-page-and-cart.spec.ts

batbattur:
## Summary
The fake test products are not working when trying to checkout.  
- https://www.cominor.com/products/playwright-out-of-stock-product
- https://www.cominor.com/products/playwright-low-stock-product

Checking out actual products is working. It looks like the test products exist properly on Shopify
but something might be going wrong on iFixit/ifixit side.

qa_req 0
cc: @jarstelfox


jarstelfox (approved):
CR 🌵

Issue #1975, opened 2023-09-13
GA4: no admin traffic and debug_mode for development


addison-grant:
Please implement the following rules for GA4 events:

Updated spec (Sep 25)

- In **prod**: only enable `debug_view` with a URL query parameter.
- In **dev**: turn on tracking and toggle to `debug_mode = true` with a site setting (do this for any user type)
- In **GA4**: add a Data Filter to not log `debug_mode=true` traffic

@sterlinghirsh @danielcliu-ifixit

danielcliu-ifixit:
The goals of this issue have changed. After talking with @addison-grant about it

1.) Have production have ga4 on with debug_mode off by default
2.) Have development have ga4 off by default
3.) Add a URI parameter `ga4_debug` where if its set to true it will turn debug_mode on for that page (assuming ga4 is enabled for that site already)

sterlinghirsh:
Is this done now?

danielcliu-ifixit:
https://github.com/iFixit/ifixit/pull/50122
https://github.com/iFixit/shopify-v2/pull/1117

PR #1991, opened 2023-09-25
GA4: Configure to Log And Debug When We Want

danielcliu-ifixit:
## Overview
There are 2 major changes

1. Move the script back into `GoogleAnalytics.tsx` instead of being in `_document.tsx`. This is because we want to access the route for the `debug_mode` state. We don't use "beforeInteractive" to programatically ensure that `window.gtag` exists on pageload now; now we use nextjs Head tag with a script tag which should run before any body code runs.
2.  You can now turn on debug mode by including `ga4_debug=true` in your uri parameters.

## QA

Make sure that everything basically works the same.
Make sure that turning `process.env.NEXT_PUBLIC_GA_DEBUG` stops GA4 events from being logged in debug_mode (you may need someone with vercel access to change this.)
Make sure that adding `ga4_debug=true` as a uri parameter ensures that all GA4 events for that pageload are in debug_mode.

Connects: #1975


masonmcelvain (dismissed):
CR 👏🏻

frontend/components/analytics/GoogleAnalytics.tsx
```diff
@@ -24,7 +25,42 @@ export function GoogleAnalytics() {
...
+                  __html: `
+               window.dataLayer = window.dataLayer || [];
+               function gtag(){dataLayer.push(arguments);}
+               gtag('js', new Date());
+               gtag('config', '${GTAG_ID}', ${debugMode} ? '{ debug_mode: true }' : '{}'
+                  );
+               `,
+               }}
+            />
+         </Head>
```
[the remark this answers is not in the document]
danielcliu-ifixit: `<Script strategy="beforeInteractive">`

is only allowed by in `_document.js` and is processed server side. We want client side so we can access the url.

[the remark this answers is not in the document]
masonmcelvain: Ah I see thank you

danielcliu-ifixit:
deploy_block :v: gonna add some code to catch up the app router code. However, should still be fine to qa

danielcliu-ifixit:
un_deploy_block :v:

frontend/app/(defaultLayout)/components/GoogleAnalytics.tsx
```diff
@@ -1,53 +1,64 @@
...
+            ga('ifixit.send', 'pageview');
+         }
+      };
+      router.events.on('routeChangeComplete', handleRouteChange);
+      router.events.on('hashChangeComplete', handleRouteChange);
+      return () => {
+         router.events.off('routeChangeComplete', handleRouteChange);
+         router.events.off('hashChangeComplete', handleRouteChange);
+      };
+   }, [router?.events]);
```
masonmcelvain, line 25: Routing is [different](https://nextjs.org/docs/app/api-reference/functions/use-router) with the app router. I don't think any of this needs to change.

frontend/app/(defaultLayout)/components/GoogleAnalytics.tsx
```diff
@@ -1,53 +1,64 @@
...
-         {wantsUA && <UA />}
          {wantsGA4 && <GA4 />}
+         {wantsUA && <UA />}
       </>
    );
 }
 
 function GA4() {
+   const router = useRouter();
+   const { query } = router;
```
masonmcelvain, line 39: query is replaced by [useSearchParams](https://nextjs.org/docs/app/api-reference/functions/use-search-params)

frontend/app/(defaultLayout)/components/GoogleAnalytics.tsx
```diff
@@ -1,53 +1,64 @@
...
+                  __html: `
+               window.dataLayer = window.dataLayer || [];
+               function gtag(){dataLayer.push(arguments);}
+               gtag('js', new Date());
+               gtag('config', '${GTAG_ID}', ${debugMode} ? '{ debug_mode: true }' : '{}'
+                  );
+               `,
+               }}
+            />
+         </Head>
```
masonmcelvain, line 57: There is no `<Head>` component in the app router. I think we need to use `<Script beforeInteractive>` here

danielcliu-ifixit: unfortunately, `beforeInteractive` isn't allowed in this file as it is not app/layout.tsx

danielcliu-ifixit:
Theres a possibility that the app router code provided is not guaranteed to have run by page load.
I am not going to track down a possible Nextjs race condition on code we do not use yet. If this code becomes used we can check if everything operates as intended then

sterlinghirsh (dismissed):
CR :zap:

erinemay:
@danielcliu-ifixit just to be clear, with `process.env.NEXT_PUBLIC_GA_DEBUG` [set to true](https://ifixit.slack.com/archives/C90126AN4/p1696012204326119) and ?ga4_debug=true added to a url, should I see events going to the debug viewer?

I've gone to several urls, appended the param, reloaded, and I'm not seeing my device show up in the debug viewer. 
Examples:
- https://react-commerce-git-ga4-handle-prod-and-debug-ifixit.vercel.app/products/google-pixel-6-pro-screen-genuine?ga4_debug=true
- https://react-commerce-git-ga4-handle-prod-and-debug-ifixit.vercel.app/Parts/Google_Phone?ga4_debug=true

sterlinghirsh:
It looked to me that the way it was written, we could use either the param or the env, but shouldn't need both.

danielcliu-ifixit:
un_dev_block :v:

sterlinghirsh (approved):
CR :zap:

deltuh-vee:
What's the current status of `process.env.NEXT_PUBLIC_GA_DEBUG`?

danielcliu-ifixit:
dev_block :v:

jarstelfox (dismissed):
CR 🌵

deltuh-vee:
dev_block ✊ 
I'm still seeing GA4 events go through the debug viewer.
CleanShot 2023-10-02 at 16 04 43@2x

danielcliu-ifixit:
@deltuh-vee currently the way to not have debug mode on is to leave that config option blank. I think this is fine personally, but if you disagree i can change it

frontend/config/env.ts
```diff
@@ -35,7 +35,8 @@ export const GA_URL = process.env.NEXT_PUBLIC_GA_URL;
 
 export const GA_KEY = process.env.NEXT_PUBLIC_GA_KEY;
 
-export const GA_DEBUG = process.env.NEXT_PUBLIC_GA_DEBUG;
+export const GA_DEBUG =
+   process.env.NEXT_PUBLIC_GA_DEBUG?.toLowerCase() === 'true';
```
jarstelfox, line 39: We could add something like this:

```tsx
function getQueryBoolean(
   query: string | string[] | undefined | null,
   options: { considerEmptyAsTrue?: boolean }
): boolean {
   if (query === undefined || query === null) {
      return false;
   }

   if (Array.isArray(query)) {
      return query.some((q) => getQueryBoolean(q, options));
   }

   const boolValue = /true|1/i.test(query);

   const considerEmptyAsTrue = options.considerEmptyAsTrue || true;
   const isEmpty = query === '';

   if (considerEmptyAsTrue && isEmpty) {
       return true;
   } 

   return boolValue;
}
```

Here's some quick tests to show it's behavior:

```

var trueTests = [
'',
'true',
"True",
"TRUE",
"tRuE",
1,
];

console.log(trueTests.map(getQueryBoolean));

var falseTests = [
' ',
'false',
"False",
"FaLSe",
"FALSE",
0,
-2,
'Hello'
];

console.log(falseTests.map(getQueryBoolean));
```

danielcliu-ifixit:
un_dev_block :v:

frontend/config/env.ts
sterlinghirsh: I think this would work well for query params but for env vars I think a simple `=== 'true'` check is fine. If we ever change it, I could see switching to 1 vs 0, but the pattern is true/false right now so I'd rather stick to that. Being permissive makes sense in places where we don't control the input, but once you start allowing multiple values (`TRUE`, `true`, `1`) it gets a lot harder to make it strict later.

sterlinghirsh (approved):
CR :zap:

erinemay:
QA 🦖 
My events are getting sent to the GA4 debug console if I apply the `?ga4_debug=true` param, or change `NEXT_PUBLIC_GA_DEBUG` to true for this branch and re-deploy in Vercel.

davidrans:
Let's hold off on deploying this until we figure out what's going on with https://github.com/iFixit/ifixit/pull/50091

deploy_block :hand: (feel free to unblock if you think this is totally unrelated, I'm just concerned about being able to see the affect that #50091 has on analytics in relative isolation)

danielcliu-ifixit:
I think the issue addressed on ifixit/ifixit is unrelated, as it appears we are still getting events from react commerce in GA4. In the case there is something broken with our implementation, this change isn't a negative.

un_deploy_block :v:

PR #1758, opened 2023-06-14
Correct .npmrc env variable warnings

federicobadini:
closes #1757 

this also fixes `NEXT_PUBLIC_STRAPI_ORIGIN` in .env.test: it should point to the local Strapi instance

qa_req 0

PR #1994, opened 2023-09-26
Quick Fix: Verdaccio Token CLI Warnings

ianrohde:
## Issue

I noticed these warnings and did a quick search on Phind. One of the last recommendations was to remove the curly braces around the variable name. 

> When using environment variables in the .npmrc file, make sure to use the variable name without curly braces. For example, use `$NPM_TOKEN` instead of `${NPM_TOKEN}`. [stackoverflow.com](https://stackoverflow.com/questions/55543361/npm-config-file-not-reading-environment-variables)

I tried it. It worked.

## Details

<details>
<summary>Before</summary>

Screenshot 2023-09-26 at 8 11 18 AM

</details>

<details>
<summary>After</summary>
Screenshot 2023-09-26 at 8 11 57 AM

</details>

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -2,4 +2,4 @@ shamefully-hoist=true
 always-auth=false
 strict-peer-dependencies=false
 registry=https://verdaccio.ubreakit.com
-//verdaccio.ubreakit.com/:_authToken=${VERDACCIO_AUTH_TOKEN}
+//verdaccio.ubreakit.com/:_authToken=$VERDACCIO_AUTH_TOKEN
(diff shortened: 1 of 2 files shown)


federicobadini:
If you run on Vercel purging the cache you receive a 401.
We attempted the same on another `.npmrc` variable: https://github.com/iFixit/react-commerce/pull/1758
Since it was a minor I did not spent so much on it though. So maybe there is a slightly different solution that works. But I suspect that just removing the parenthesis doesn't

ianrohde:
I don't have Vercel access to react-commerce, so pardon me if this is already in use or tested, but it looks like these are supposed to be managed through the [Environment Variables UI](https://vercel.com/guides/how-to-add-vercel-environment-variables#environment-variables-ui), then called with `process.env`.

> Next.js has built-in support for loading environment variables from .env.local into process.env.

I've tried using process.env before [here](https://github.com/iFixit/react-commerce/pull/1636), but we had to revert. It was likely due to the same 401 issue you saw. It's odd to me that using a simple variable is such an issue when it's supposed to be supported 🤷

PR #2007, opened 2023-10-02
Sentry: Sample Next.js client errors

masonmcelvain:
[Motivation in slack](https://ifixit.slack.com/archives/C04C05FA3EH/p1696266414063559)

[Docs](https://vercel.com/docs/projects/environment-variables/system-environment-variables#system-environment-variables) on system environment variables

## QA

Make sure errors are not reported to sentry by default in dev

CC @jarstelfox @sctice-ifixit

### DIFF
--- a/frontend/sentry.client.config.ts
+++ b/frontend/sentry.client.config.ts
@@ -2,6 +2,7 @@
 // The config you add here will be used whenever a page is visited.
 // https://docs.sentry.io/platforms/javascript/guides/nextjs/
 
+import { VERCEL_ENV } from '@config/env';
 import { isCurrentProductionDeployment } from '@helpers/vercel-helpers';
 import * as Sentry from '@sentry/nextjs';
 import { BrowserTracing } from '@sentry/browser';
@@ -24,7 +25,7 @@ Sentry.init({
    },
    dsn: SENTRY_DSN,
    integrations: [new BrowserTracing()],
-   sampleRate: 1.0,
+   sampleRate: VERCEL_ENV === 'production' ? 0.05 : 1.0,
    normalizeDepth: 5,
    tracesSampleRate: 0.05,
    environment: process.env.NODE_ENV,
--- a/frontend/sentry.client.config.ts
+++ b/frontend/sentry.client.config.ts
@@ -1,13 +1,10 @@
-// This file configures the initialization of Sentry on the browser.
-// The config you add here will be used whenever a page is visited.
-// https://docs.sentry.io/platforms/javascript/guides/nextjs/
-
 import { VERCEL_ENV } from '@config/env';
 import { isCurrentProductionDeployment } from '@helpers/vercel-helpers';
 import * as Sentry from '@sentry/nextjs';
 import { BrowserTracing } from '@sentry/browser';
 
 const SENTRY_DSN = process.env.NEXT_PUBLIC_SENTRY_DSN;
+const sampleRate = VERCEL_ENV === 'production' ? 0.05 : 0;
 
 Sentry.init({
    async beforeSend(event) {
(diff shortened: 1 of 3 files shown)


frontend/sentry.client.config.ts
```diff
@@ -24,7 +25,7 @@ Sentry.init({
    },
    dsn: SENTRY_DSN,
    integrations: [new BrowserTracing()],
-   sampleRate: 1.0,
+   sampleRate: VERCEL_ENV === 'production' ? 0.05 : 1.0,
```
jarstelfox, line 28: Should we only report in dev?

That's what we currently do in the mono repo, but I am unsure if there is enough of a reason to keep reporting in non-prod envs


```suggestion
   sampleRate: VERCEL_ENV === 'production' ? 0.05 : 0,
```

jarstelfox (dismissed):
CR 🌵

frontend/sentry.client.config.ts
masonmcelvain: It's convenient to have error reporting turned on by default for QA and debugging/iterating. However, if we're worried about chipping away at our Sentry budget, I don't think QA would mind opting into error reporting like we do on the PHP side as long as we notified everyone about it.

jarstelfox: I think it's weird that local dev errors make it to Sentry and even cause Sentry's new PR feature to misreport errors:

Example:
https://github.com/iFixit/react-commerce/pull/1976#issuecomment-1730229461


https://ifixit.sentry.io/issues/4494886521/?project=6475315&referrer=github-pr-bot

Background:
https://github.com/getsentry/sentry/discussions/49996#discussioncomment-6807526

masonmcelvain: I [see](https://ifixit.sentry.io/settings/billing/overview/) our quota is 50k, and 4k is a significant chunk of that. I think turning off reporting in dev by default makes sense 👍🏻

jarstelfox: See: https://ifixit.sentry.io/dashboard/65897/?project=6475315&statsPeriod=30d

I'd say more than half of all errors are from local dev.

The query there is looking at URL like `localhost`

danielbeardsley: I agree. I almost *never* want to see dev errors and I practically only look at sentry to see prod errors. Seeing dev errors makes me generally less trusting of the stuff in sentry cause I'm constantly questioning... oh wait... did these happen in dev?

danielbeardsley: Maybe we should take in a `SAMPLE_RATE variable and have it default to `0`? Then in vercel env conf we can bump it for just the `main` branch deployments.

jarstelfox: Sounds like a great idea to me

frontend/sentry.client.config.ts
```diff
@@ -1,12 +1,10 @@
...
-// The config you add here will be used whenever a page is visited.
-// https://docs.sentry.io/platforms/javascript/guides/nextjs/
-
+import { VERCEL_ENV } from '@config/env';
 import { isCurrentProductionDeployment } from '@helpers/vercel-helpers';
 import * as Sentry from '@sentry/nextjs';
 import { BrowserTracing } from '@sentry/browser';
 
 const SENTRY_DSN = process.env.NEXT_PUBLIC_SENTRY_DSN;
+const sampleRate = VERCEL_ENV === 'production' ? 0.05 : 0;
```
danielbeardsley, line 7: As far as I remember, *all* our vercel deployments are "production", so this includes all pull request code.

masonmcelvain: The NODE_ENV is production, but not the VERCEL_ENV

image

danielbeardsley:
CR :+1:

jarstelfox (approved):

masonmcelvain:
QA ✌🏻 On main, I can produce an [error in sentry](https://ifixit.sentry.io/share/issue/de09d6e4572744af9b934ed2d0cfa74c/) in local dev. On this branch, I can't

PR #2008, opened 2023-10-02
Strapi: Allow scheduling content to be published / unpublished

jarstelfox:
I decided to go with https://github.com/ComfortablyCoding/strapi-plugin-publisher as it is slightly more robust: More devs, longer lived, more releases, newer releases, more downloads etc

QA
---

* I had to run the following to kick over strapi locally
	* `cd backend/ &&  rm -rf .cache/ node_modules/ dist/`
	* Then `cd ../ && pnpm install:all`
* After running `pnpm dev` strapi should allow scheduling
* Note: The scheduling UI does not appear until after the first save of an item

Before:
image

Click save once, then you should see: 
image

* Note: [You can not see if something is scheduled in the list view](https://github.com/ComfortablyCoding/strapi-plugin-publisher/issues/95)

Closes: https://github.com/iFixit/ifixit/issues/50009

### DIFF
--- a/backend/config/plugins.ts
+++ b/backend/config/plugins.ts
@@ -17,6 +17,18 @@ export default ({ env }) => {
             },
          },
       },
+      publisher: {
+         enabled: true,
+         // See: https://market.strapi.io/plugins/strapi-plugin-publisher
+         config: {
+            hooks: {
+               beforePublish: async ({ strapi, uid, entity }) => {},
+               afterPublish: async ({ strapi, uid, entity }) => {},
+               beforeUnpublish: async ({ strapi, uid, entity }) => {},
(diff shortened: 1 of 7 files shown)


masonmcelvain (approved):
CR 👏🏻 Nice work, glad it was simple! From a quick search I didn't see any other publishing plugins besides this one and the one mentioned in the issue, and I agree that this one looks better

jarstelfox:
QA 🌵 

I visited https://strapi-ability-to-schedule-posts.govinor.com/admin/content-manager/collectionType/api::product-list.product-list/2?plugins[i18n][locale]=en

Then saved a new item and it magically changed:

image

Note: You can add an item to be scheduled in the past (1 min ago) and the next cron tick will publish it.

PR #2010, opened 2023-10-03
Sentry: Add env var to disable sampling

masonmcelvain:
Followup to https://github.com/iFixit/react-commerce/pull/2007

I missed this conversation https://github.com/iFixit/react-commerce/pull/2007#discussion_r1343002422 and I think it would be a good idea to give devs a way to disable sampling for a specific preview via Vercel environment variable overrides. I allowed it to apply to any environment (production, preview, or dev) intentionally so that we could disable the sample rate in prod if we ever need to without a code change.

CC @danielbeardsley @jarstelfox 

qa_req 0 -- Blast radius is low, complexity of change is low, therefore value of QAing this is low. There are lots of other pulls where QA effort will be more impactful.


frontend/config/env.ts
```diff
@@ -31,6 +31,9 @@ export const MATOMO_URL = process.env.NEXT_PUBLIC_MATOMO_URL;
 
 export const PIWIK_ID = getPiwikIDForEnv(process.env.NEXT_PUBLIC_PIWIK_ENV);
 
+export const DISABLE_SENTRY_SAMPLING =
+   process.env.DISABLE_SENTRY_SAMPLING === 'true';
```
jarstelfox, line 35: Minor: I find it easier to read the positive form in all cases:

```suggestion
export const SENTRY_SAMPLING_ENABLED =
   process.env.SENTRY_SAMPLING === 'true';
```

danielbeardsley: How about `SENTRY_SAMPLING_OVERRIDE=1.0` instead? Clear and tunable.

masonmcelvain: What's the use case for a custom value, as opposed to "force it to be on" flag?

I really think that the production value should be hardcoded for visibility and history, so I'm worried that an env var custom value might get used in production. What do you think about only allowing the `SENTRY_SAMPLING_OVERRIDE` to be usable in preview and development environments?

sctice-ifixit: > I really think that the production value should be hardcoded for visibility and history

I think @masonmcelvain has a good point here. I've gone and looked up the error sample rate in the PHP app multiple times, and the code is where I expect to find an answer. Also, I don't really see utility in the area between "a small sample" and 100%.

masonmcelvain:
dev_block ✋🏻 on discussion above

danielbeardsley (approved):
CR :+1:

masonmcelvain:
un_dev_block ✌🏻

erinemay:
@masonmcelvain, I'm not seeing a difference either way. 

At first I tried setting the env var in the  Vercel ui for development only, but that had no effect on my local setup's errors getting Sent to sentry. I deleted it from the UI, and I edited the value of `SENTRY_SAMPLING_ENABLED` in frontend/.env.development and tried throwing errors. 

It didn't appear that any sampling was happening any way I tried it. Are there rules around the sampling, like only sample past the first so many? I reloaded my error pages and saw the sentry count for my errors increment with each page reload/throw of an error. 

With sampling turned on or off, I saw each instance of my error get sent to Sentry. 

My errors: [Vercel 1](https://ifixit.sentry.io/issues/4525873522/?project=6469069&project=6475315&query=&referrer=issue-stream&statsPeriod=14d&stream_index=3), [Vercel 2](https://ifixit.sentry.io/issues/4525872982/?project=6469069&project=6475315&query=&referrer=issue-stream&statsPeriod=14d&stream_index=1), [Vercel 3](https://ifixit.sentry.io/issues/4525855044/?project=6469069&project=6475315&query=&referrer=issue-stream&statsPeriod=14d&stream_index=0)

masonmcelvain:
> At first I tried setting the env var in the Vercel ui for development only

As far as I know, this won't have an effect because we don't use the Vercel CLI. It must be set in `.env.local` (or `.env.development` as you did) for the env var to be picked up by local dev.

> It didn't appear that any sampling was happening any way I tried it. Are there rules around the sampling, like only sample past the first so many? 

Sampling rates are configured in `sentry.client/server/edge.config`. There are no special rules, the defined sample rate will always apply.

If `SENTRY_SAMPLING_ENABLED=true`, then we expect errors to be sampled (therefore, errors should not be sent in development or previews).

If `SENTRY_SAMPLING_ENABLED=false`, then we expect errors to end up in Sentry for development and previews.

Does that answer your questions @erinemay?

erinemay:
Yeah. Then I think this isn't working as I test it, but I think this is probably in my setup. This seems fairly safe to me. If you think so, we can merge. 

Sampled: Each reload of the page causes a new instance of the error to show up in Sentry.

<details><summary>Details</summary>

SENTRY_SAMPLING_ENABLED=true and throwing an error

Each time I reload http://localhost:3000/Parts, the count on [this error](https://ifixit.sentry.io/issues/4525855044/?project=6469069&project=6475315&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=0) increments

```diff
diff --git a/frontend/pages/Parts/index.tsx b/frontend/pages/Parts/index.tsx
index dba48e36..9896551b 100644
--- a/frontend/pages/Parts/index.tsx
+++ b/frontend/pages/Parts/index.tsx
@@ -1,6 +1,8 @@
 import { ProductListType } from '@models/product-list';
 import { getProductListServerSideProps } from '@templates/product-list/server';

+throw new TypeError('Sampling set to true.');
+
 export { default } from '@templates/product-list';

 export const getServerSideProps = getProductListServerSideProps({

```

```bash
erin13mm@DESKTOP-LSSD59U:~/react-commerce$ cat frontend/.env.development | head
NEXT_PUBLIC_ALGOLIA_APP_ID=XQEP3AD9ZT
NEXT_PUBLIC_IFIXIT_ORIGIN=https://www.cominor.com
NEXT_PUBLIC_APP_ORIGIN=http://localhost:3000
NEXT_PUBLIC_STRAPI_ORIGIN=http://localhost:1337
NEXT_PUBLIC_SHOPIFY_STOREFRONT_VERSION=2022-10
STRAPI_IMAGE_DOMAIN=ifixit-dev-strapi-uploads.s3.us-west-1.amazonaws.com
NEXT_PUBLIC_SENTRY_DSN=https://95ab6917c0234f80b99bb0be8e720355@o186239.ingest.sentry.io/6475315
SENTRY_SAMPLING_ENABLED=true
NEXT_PUBLIC_MATOMO_URL=https://matomo.ubreakit.com
NEXT_PUBLIC_PIWIK_ENV=none

```


</details>


Unsampled: Each reload of the page causes a new instance of the error to show up in Sentry.


<details><summary>Details</summary>

SENTRY_SAMPLING_ENABLED=false and throwing an error

Each time I reload http://localhost:3000/Tools, the count on [this error](https://ifixit.sentry.io/issues/4525872982/?project=6469069&project=6475315&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=1) increments

```diff
diff --git a/frontend/.env.development b/frontend/.env.development
index c1b3e44f..37a5b05f 100644
--- a/frontend/.env.development
+++ b/frontend/.env.development
@@ -5,7 +5,7 @@ NEXT_PUBLIC_STRAPI_ORIGIN=http://localhost:1337
 NEXT_PUBLIC_SHOPIFY_STOREFRONT_VERSION=2022-10
 STRAPI_IMAGE_DOMAIN=ifixit-dev-strapi-uploads.s3.us-west-1.amazonaws.com
 NEXT_PUBLIC_SENTRY_DSN=https://95ab6917c0234f80b99bb0be8e720355@o186239.ingest.sentry.io/6475315
-SENTRY_SAMPLING_ENABLED=true
+SENTRY_SAMPLING_ENABLED=false
 NEXT_PUBLIC_MATOMO_URL=https://matomo.ubreakit.com
 NEXT_PUBLIC_PIWIK_ENV=none
 NEXT_PUBLIC_GA_KEY=UA-30506-1
diff --git a/frontend/pages/Tools/index.tsx b/frontend/pages/Tools/index.tsx
index 4be7497f..4e226a8b 100644
--- a/frontend/pages/Tools/index.tsx
+++ b/frontend/pages/Tools/index.tsx
@@ -3,6 +3,8 @@ import { getProductListServerSideProps } from '@templates/product-list/server';

 export { default } from '@templates/product-list';

+throw new TypeError('Sampling set to false');
+
 export const getServerSideProps = getProductListServerSideProps({
    productListType: ProductListType.AllTools,
 });

```



</details>

danielbeardsley:
Thanks @erinemay for your hard work!

I think the blast radius is pretty low here and the complexity of this change and the ease of fixing it are both low as well. I think your QA efforts will be more impactful on other pulls. I've dropped th QA requirement.

masonmcelvain:
I think I was able to QA ✌🏻 this

With `SENTRY_SAMPLING_ENABLED=true`, no errors showed up in sentry for local dev

With `SENTRY_SAMPLING_ENABLED=false`, I produced an [error](https://ifixit.sentry.io/issues/4527961185/?project=6475315&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=1h&stream_index=0)

PR #2013, opened 2023-10-04
ProductList: Add Option to Boost Skus

danielcliu-ifixit:
## Overview

We wanted to be able to highlight some products on product list pages by surfacing or boosting them to the top of the results. This pull allows us to do that by using algolia's optionalFIlters.

## QA

Go to a product list page and find a product that isn't at the top. Grab its sku. Go add that sku to that product list's `boostedSearchSkus` attribute in strapi. Reload the page. The desired product should now be the first thing on the list.

Closes: iFixit/ifixit#50010

### DIFF
--- a/backend/src/api/product-list/content-types/product-list/schema.json
+++ b/backend/src/api/product-list/content-types/product-list/schema.json
@@ -241,6 +241,16 @@
          "relation": "manyToMany",
          "target": "api::faq.faq",
          "inversedBy": "product_lists"
+      },
+      "boostedSearchSkus": {
+         "type": "component",
+         "repeatable": true,
+         "pluginOptions": {
+            "i18n": {
+               "localized": true
+            }
+         },
+         "component": "product-list.boosted-search-skus"
       }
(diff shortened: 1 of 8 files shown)


sterlinghirsh:
deploy_block :zap: Let's check with marketing (Kelsea? Or maybe Max?) to make sure this is what they want. I was imagining we'd have basically the same type of filter input that we use for the normal filter (i.e. a text area that lets you put in a string like `identifiers: "IF123-456" OR  ...`). I think the most common use so far has been to list specific skus, but I can imagine wanting to target sale items or Genuine OEM parts or something. A list of itemcodes is definitely simpler to enter, I just want to make sure it meets all the requirements.

danielcliu-ifixit:
in my defense the issue reads ` One or more products may be specified by their SKU.`

danielcliu-ifixit:
Closing in favor of a more robust optionalFiltering implemenation, one that doesn't limit people to just using SKUs to boost

PR #2014, opened 2023-10-05
ProductList: Add Optional Filters

danielcliu-ifixit:
## Overview

We wanted to be able to highlight some products on product list pages by surfacing or boosting them to the top of the results. This pull allows us to do that by using algolia's optionalFIlters.

## QA 

You'll want to find a productList to use. Go to it in strapi and add data to the optionalFilters field

For example I had something like for `/Parts/Google_Pixel_6_Pro`

```
category:Adhesive Strips
identifiers:356-324
identifiers:356-080
```

This will boost all products that match these conditions in algolia, and should boost the products that match more of these conditions more. So one of the items matches `356-080` `Adhesive Strips` so it should be at the top of the results.

If you want to read about option filter rules read https://www.algolia.com/doc/guides/managing-results/rules/merchandising-and-promoting/in-depth/optional-filters/

Note: you the strapi optionalFilters are split by endline characters, and in each line are split by `||` if you want to do a more complex filter. So

```
category:Adhesive Strips
identifiers:356-324 || identifiers:356-080
```
would result in the optional filters of

`['category:Adhesive Strips', ['identifiers:356-324', 'identifiers:356-080']]`

Closes: https://github.com/iFixit/ifixit/issues/50010


backend/src/api/product-list/content-types/product-list/schema.json
```diff
@@ -241,6 +241,14 @@
          "relation": "manyToMany",
          "target": "api::faq.faq",
          "inversedBy": "product_lists"
+      },
+      "optionalFilters": {
```
danielbeardsley, line 245: Sadly, we need to keep strapi pulls separate from the frontend pulls. This is because there are separate deploy processes. If the frontend is deployed first (which it usually is) then it will start asking for this field *before* it's actually available in strapi and strapi will respond with an error during the time between deployments.

It's possible this will only be for a few minutes, but it's also possible that it takes longer or that strapi deploy fails.

danielcliu-ifixit:
un_dev_block :v:  merged the backend in its own pull first

danielbeardsley (approved):
CR :+1:

erinemay:
Cool! When I add values to the Optional Filters field in a productlist, those products matching the filters are sorted to the top. 
|| splitting works -> I can filter several products to the top by sku/identifier
category works

This is great and it's working as described. QA 👍

PR #2022, opened 2023-10-06
Go Aff Pro: Add script to all next pages

jarstelfox:
Connects: https://github.com/iFixit/ifixit/issues/50128


jarstelfox:
deploy_block 🌵 on go-ahead on the issue

sterlinghirsh (approved):
CR :zap: but block remains on making sure this is what we want.

jarstelfox:
We decided to go a different route (a custom implementation) a new pull will be open shortly.

PR #2024, opened 2023-10-09
CI Fix: Pin Pnpm Version to 8.8.0

ardelato:
## Description

The Next.js Bundle Analysis action began failing. @batbattur  discovered that the default pnpm version was bumped to 8.9.0 while the prior version was 8.8.0. 

To address this let's pin the pnpm version to 8.8.0 instead of a just a major semver range of v8.

qa_req 0 CI should pass

**Reference:** https://ifixit.slack.com/archives/C01DBJK0H7U/p1696882385520809


batbattur (approved):
CR 🚛 LGTM!

Issue #2016, opened 2023-10-05
PPRO: Preferred Store custom dimension [react-commerce]


addison-grant:
Fill in value for the Preferred Store custom dimension in Piwik PRO

For the value you can use the store handle

- ifixit-us
- ifixit-australia
- ifixit-canada
- ifixit-de
- ifixit-eu
- ifixit-fr
- ifixit-uk
- ifixit-eu-pro
- ifixit-eu-test
- ifixit-eu-pro-test
- ifixit-test
- ifixit-dev
- ifixit-checkout
- ifixit-italy

PPRO docs

- [General docs on Custom Dimensions](https://help.piwik.pro/support/reports/custom-dimension/)
- [JS code required](https://help.piwik.pro/wp-content/uploads/2019/09/session-custom-dimension-value.png)

Custom dimensions settings

- https://ifixit.piwik.pro/analytics/#/6bf6bc54-82ca-4321-8620-b12d5c9b57b6/settings/custom-dimensions/

PR #2025, opened 2023-10-12
PPRO: Add Custom Dimension "preferredStore"

danielcliu-ifixit:
## Overview

We want to keep track of users preferred Store for their session. We can track that per user sessions by adding a custom dimension, in this case its `preferredStore`, to Piwik. This pull introduces the code to track that dimension in our code for each users session.

## QA

Make sure that users session has the `preferredStore` as a dimension in the analytics.

Connects: #2016


danielcliu-ifixit:
closing to reopen after #2026 to build off of that folder structure.

PR #2026, opened 2023-10-13
PPRO: Replace "Matomo" with "Piwik" In Most Instances

danielcliu-ifixit:
## Overview

Now that we are moving away from Matomo to Piwik, but many of our Matomo functions work for Piwik too, we are just going to change the function/variable/file names so that they are now "Piwik". Pretty straighforward.

We want to keep track of users preferred Store for their session. We can track that per user sessions by adding a custom dimension, in this case its preferredStore, to Piwik. This pull introduces the code to track that dimension in our code for each users session.

## QA

Make sure that the Piwik analytics still work.

Make sure that users session has the `preferredStore` as a dimension in the analytics.

Connects: #2016


packages/analytics/matomo/index.tsx
```diff
@@ -11,6 +15,14 @@ export function trackMatomoPageView(url: string) {
    matomoPush(['trackPageView']);
 }
 
+export function trackPiwikPreferredStore() {
+   if (typeof window !== 'undefined') {
+      const host = getShopifyStoreDomainFromCurrentURL();
```
aburke07, line 21: Isn't the react-commerce store US only? I think we can just hardcode this to `ifixit-us` instead of parsing the domain.

aburke07 (dismissed):
CR :alien: Dev_block :alien: on a comment

packages/analytics/matomo/index.tsx
danielcliu-ifixit: yes rn react-commerce is only for the us, but we would need to handle ifixit-test still so we cant just straight hardcode it

aburke07: I'm fine with leaving it as is then. No sense in simplifying to just US and the test stores only to have to expand it in the future.

aburke07:
Un_dev_block :alien:

kthaler:
dev_block 🟥 - I am clicking the preview link and navigation around react-commerce, but I am not seeing any corresponding analytics in the [Piwik Tracker debugger](https://ifixit.piwik.pro/analytics/#/748e3baf-bedc-4685-9f32-6b72ec349fd4/settings/debugger/)(ifixit development instance). Is there a step I'm missing?

danielcliu-ifixit:
@kthaler ok the issue is the vercel instances use `.env.production` to boot up, which sends their piwik events to the production by default. Ive verified these are sending. Ive implemented a new variable in vercel set to 'dev' so it should be working now. 

un_dev_block :v:

kthaler:
Verified that Piwik analytics remains functional, and navigating within Vercel correctly displays in the analytics data. Also ensured that users session has the preferredStore as a dimension in the analytics. 

QA 👍

aburke07 (approved):
CR :alien:

deltuh-vee:
Carry Over QA 🎬

Issue #2017, opened 2023-10-05
PPRO: Preferred Language custom dimension [react-commerce]


addison-grant:
Fill in value for the Preferred Language custom dimension in Piwik PRO.

For the value, use one of the 12 site languages:

- DE
- EN
- ES
- FR
- IT
- NL
- PT
- RU
- TR
- ZH
- JP
- KO

PPRO docs

- [General docs on Custom Dimensions](https://help.piwik.pro/support/reports/custom-dimension/)
- [JS code required](https://help.piwik.pro/wp-content/uploads/2019/09/session-custom-dimension-value.png)

Custom dimensions settings

- https://ifixit.piwik.pro/analytics/#/6bf6bc54-82ca-4321-8620-b12d5c9b57b6/settings/custom-dimensions/

PR #2030, opened 2023-10-17
PPRO: Add Custom Dimension Tracking for PreferredStore

danielcliu-ifixit:
## Overview

This pr introduces code to track the preferred store of a user in their piwik session. It first tries to use the users preferred langid from their ifixit account. If there isn't a user for the current session,  it just defaults to the current store's langid (which should always be EN now).

## QA

Make sure the `preferredStore` is correctly logged in piwik. Make sure that it uses the ifixit user's langid if they are logged in, and "EN" if not.

Closes: #2017

### DIFF
--- a/frontend/layouts/default/index.tsx
+++ b/frontend/layouts/default/index.tsx
@@ -59,7 +59,10 @@ import {
    UserMenuLink,
    WordmarkLink,
 } from '@ifixit/ui';
-import { trackPiwikPreferredStore } from '@ifixit/analytics';
+import {
+   trackPiwikPreferredStore,
+   trackPiwikPreferredLanguage,
+} from '@ifixit/analytics';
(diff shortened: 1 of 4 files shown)


aburke07 (approved):
CR :alien:

danielcliu-ifixit:
@erinemay i am unable to replicate your issue with the language not being reported correctly. Does your browser not allow third party cookies? If it does, I will need more information to debug the issue you ran into.

un_dev_block :v:

danielcliu-ifixit:
@deltuh-vee to be clear, the store preference is based on the store you are currently on for react commerce and shopify

deltuh-vee:
QA 🎬 
I was able to get the correct language to be set, so in combination with https://github.com/iFixit/react-commerce/pull/2030#issuecomment-1766806169 I think this counts as QA.

PR #2034, opened 2023-10-17
PPRO: Let Langid be optional.

danielcliu-ifixit:
Fix issue that we had with lang id where you would get
image_720


aburke07:
QA :alien: applying this diff to my localdev fixes the error.

aburke07 (approved):
CR :alien:

PR #2036, opened 2023-10-18
Error handling: Rework

jarstelfox:
We already override the `window.fetch` function to report to sentry.
This is so we hear about errors from other external services.

However, we know our API and don't want to hear about errors if we are explicitly handling them. An example is a 422 validation error. The fetch function would alert on it even if we correctly handle the error in code.

This pull addresses this by using a header from the API client to communicate to the overridden fetch function.

Now, all calls to our API effectively use the regular window.fetch functionality.

This means we needed to add sentry error handling inside our API client. This is done in the getJson function which assumes a 200 and json or throws.

If a dev wants to manage the response there is now a client.get function which returns the raw response.

### DIFF
--- a/frontend/app/(defaultLayout)/Troubleshooting/[device]/page.tsx
+++ b/frontend/app/(defaultLayout)/Troubleshooting/[device]/page.tsx
@@ -57,7 +57,7 @@ async function getTroubleshootingListData(
    const url = `Troubleshooting/Collection/${device}`;
 
    try {
-      return await client.get<TroubleshootingListProps>(
+      return await client.getJson<TroubleshootingListProps>(
          url,
          'troubleshootingList'
       );
--- a/frontend/app/(defaultLayout)/Troubleshooting/[device]/page.tsx
+++ b/frontend/app/(defaultLayout)/Troubleshooting/[device]/page.tsx
@@ -57,7 +57,7 @@ async function getTroubleshootingListData(
    const url = `Troubleshooting/Collection/${device}`;
 
    try {
-      return await client.getJson<TroubleshootingListProps>(
+      return await client.get<TroubleshootingListProps>(
          url,
          'troubleshootingList'
       );
--- a/frontend/lib/ifixit-api/international-buy-box.ts
+++ b/frontend/lib/ifixit-api/international-buy-box.ts
@@ -26,7 +26,7 @@ export function getBuyBoxForProduct(
    productcode: string
 ): Promise<BuyBoxResponse | null> {
    // TODO pull country override from browser url
-   return client.get(
+   return client.getJson(
       `internal/international_store_promotion/buybox?productcode=${productcode}`,
       'international-store-promotion-buybox'
    );
--- a/frontend/lib/ifixit-api/international-buy-box.ts
+++ b/frontend/lib/ifixit-api/international-buy-box.ts
@@ -26,7 +26,7 @@ export function getBuyBoxForProduct(
    productcode: string
 ): Promise<BuyBoxResponse | null> {
    // TODO pull country override from browser url
-   return client.getJson(
+   return client.get(
       `internal/international_store_promotion/buybox?productcode=${productcode}`,
       'international-store-promotion-buybox'
    );
--- a/frontend/lib/ifixit-api/productData.ts
+++ b/frontend/lib/ifixit-api/productData.ts
@@ -18,7 +18,10 @@ export async function fetchProductData(
          productHandle.length > 0,
          'productHandle cannot be a blank string'
       );
-      return await client.get(`store/product/${productHandle}`, 'product-data');
+      return await client.getJson(
+         `store/product/${productHandle}`,
+         'product-data'
+      );
    } catch (error: any) {
       return null;
    }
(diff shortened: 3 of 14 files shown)


jyee27 (approved):
CR ☕ through 7a91b2fbf1f57cb848f887ec52a92df0d419b31e

jyee27 (dismissed):
CR ☕

jarstelfox:
deploy_block 🌵 I'm unsure what QA looks like here. I've personally tested callsites while working on this plus tests caught some bugs.

sterlinghirsh (approved):
CR :zap:

jarstelfox:
dev_block 🌵 Everything mostly works but I found a really subtle behavior that explains a number of errors missing data.

It all centers around [this github issue](https://github.com/getsentry/sentry-javascript/issues/2948).

In short, this pattern does not correctly set the sentry context:

```
Sentry.withScope((scope) => {
      scope.setTag('foo', 'bar');
      throw error;
   });
```

We do this a number of places including [here](https://github.com/iFixit/react-commerce/blob/a616fbf83d6dbf979d05cf53b9281ed45a7230b3/frontend/lib/strapi-sdk/index.ts#L39-L44) and [here](https://github.com/iFixit/react-commerce/blob/a616fbf83d6dbf979d05cf53b9281ed45a7230b3/frontend/lib/shopify-storefront-sdk/index.ts#L44-L50).

The scope _will not_ have the new data set during the throw.

We actually ran into this problem in the monorepo and I worked around it by doing:

```
function reportError(error: Error, sentryDetails: SentryDetails){
   const reportFn = function (scope: Scope): void {
      setSentryDetails(sentryDetails);
      captureException($exception);
   };

   withScope(reportFn);
}
```

jarstelfox:
un_dev_block 🌵

jarstelfox:
I turned sentry to always be on and to always log the event then drop it beforesending:
```diff
diff --git a/frontend/sentry.client.config.ts b/frontend/sentry.client.config.ts
index a95a8832..64bd0172 100644
--- a/frontend/sentry.client.config.ts
+++ b/frontend/sentry.client.config.ts
@@ -5,11 +5,7 @@ import { BrowserTracing } from '@sentry/browser';
 import * as Sentry from '@sentry/nextjs';
 
 const SENTRY_DSN = process.env.NEXT_PUBLIC_SENTRY_DSN;
-const sampleRate = SENTRY_SAMPLING_ENABLED
-   ? VERCEL_ENV === 'production'
-      ? 0.05
-      : 0
-   : 1.0;
+const sampleRate = 1;
 
 Sentry.init({
    async beforeSend(event) {
@@ -23,7 +19,8 @@ Sentry.init({
             'Exception Checking Production Deployment': e,
          };
       }
-      return event;
+      console.log("Sentry Event", event);
+      return null;
    },
    dsn: SENTRY_DSN,
    integrations: [new BrowserTracing()],
diff --git a/frontend/sentry.edge.config.ts b/frontend/sentry.edge.config.ts
index 0ec0bad1..81f2cf1d 100644
--- a/frontend/sentry.edge.config.ts
+++ b/frontend/sentry.edge.config.ts
@@ -4,11 +4,7 @@ import { isCurrentProductionDeployment } from '@helpers/vercel-helpers';
 import * as Sentry from '@sentry/nextjs';
 
 const SENTRY_DSN = process.env.NEXT_PUBLIC_SENTRY_DSN;
-const sampleRate = SENTRY_SAMPLING_ENABLED
-   ? VERCEL_ENV === 'production'
-      ? 1.0
-      : 0
-   : 1.0;
+const sampleRate = 1;
 
 Sentry.init({
    async beforeSend(event) {
@@ -22,7 +18,8 @@ Sentry.init({
             'Exception Checking Production Deployment': e,
          };
       }
-      return event;
+      console.log("Sentry Event", event);
+      return null;
    },
    dsn: SENTRY_DSN,
    initialScope: {
```

I then made a bad request:

```diff
diff --git a/frontend/templates/product/sections/ProductOverviewSection/AddToCart/NotifyMeForm.tsx b/frontend/templates/product/sections/ProductOverviewSection/AddToCart/NotifyMeForm.tsx
index 9068bfdc..6f983711 100644
--- a/frontend/templates/product/sections/ProductOverviewSection/AddToCart/NotifyMeForm.tsx
+++ b/frontend/templates/product/sections/ProductOverviewSection/AddToCart/NotifyMeForm.tsx
@@ -45,6 +45,22 @@ export function NotifyMeForm({ sku }: NotifyMeFormProps) {
          setStatus(NotifyMeStatus.Error);
          return;
       }
+
+      const badResponseAPIClientRaw = await ifixitAPI.post(
+         'cart/product/notifyWhenSkuInStock',
+         'notify-when-sku-in-stock',
+         {
+            credentials: 'same-origin',
+            headers: {
+               'Content-Type': 'application/json',
+            },
+            body: JSON.stringify({
+               sku,
+               shop_domain: layoutProps.shopifyCredentials.storefrontDomain,
+               email: 'APIClientRaw@APIClientRaw',
+            }),
+         });
+
       const response = await ifixitAPI.postRaw(
          'cart/product/notifyWhenSkuInStock',
          'notify-when-sku-in-stock',
```

The tags and data are coming through as expected:

image


There is only one captured exception which is not via the global handler.

QA 🌵

jarstelfox:
QA 🐖 🔙

sterlinghirsh (approved):
CR :zap:

PR #2038, opened 2023-10-18
useLocalPreference: Handle Validation Failing More Cleanly

danielcliu-ifixit:
## Overview

Before, when grabbing cached data from localestorage, we were not handling validation errors well at all; it would error out the whole page. In the case of it failing on the user data, it was 500'ing every page. This is obviously not ideal, especially considering its supposed to be cached data thats just there to help make the front end faster.

When we merged 

we soon got a lot of errors and every react commerce page was erroring out
https://ifixit.sentry.io/issues/4554590951/?project=6469069&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=90d&stream_index=0

This was (likely) because we added a new key for the user data, but when the code went to grab the cached data it couldn't find the new key `langid` so it was `undefined`. 

We had to roll out https://github.com/iFixit/react-commerce/pull/2034 to allow this key to be undefined, but the same issue would rear its head in the future if we did nothing, meaning either we had to allow all `AuthenticatedUserSchema` to be `optional` or handle this better.

So this wraps the `useLocalePreference` validation code in a try catch so we wont run into this massive failure. We also will report user data api calls if they fail now. 

## QA

to test the `useLocalePreference` failure, you can apply this diff

<details><summary>Diff</summary>
<p>

```diff
diff --git a/packages/auth-sdk/user.tsx b/packages/auth-sdk/user.tsx
index 6710f081..b4fd6302 100644
--- a/packages/auth-sdk/user.tsx
+++ b/packages/auth-sdk/user.tsx
@@ -23,7 +23,7 @@ const AuthenticatedUserSchema = z.object({
    isAdmin: z.boolean(),
    teams: z.array(z.number()),
    links: z.record(z.any()),
-   langid: z.string().optional().nullable(),
+   langid: z.number(),
 });
 
 export function useAuthenticatedUser() {
 ```
</p>
</details> 

To test the user fetching/verification errors getting to sentry you can apply this diff. Obviously also have sentry enabled and all that jazz.

<details><summary>Diff</summary>
<p>

```diff
diff --git a/packages/auth-sdk/user.tsx b/packages/auth-sdk/user.tsx
index 6710f081..7022fe54 100644
--- a/packages/auth-sdk/user.tsx
+++ b/packages/auth-sdk/user.tsx
@@ -74,7 +74,7 @@ const UserApiResponseSchema = z.object({
    privileges: z.array(z.string()),
    teams: z.array(z.number()),
    links: z.record(z.any()),
-   langid: z.string().optional().nullable(),
+   langid: z.number(),
 });
 
 async function fetchAuthenticatedUser(
```

</p>
</details>


deltuh-vee:
dev_block ✊ 
I'm still not seeing the errors end up in Sentry. Here's my setup steps.
1. Switched my subdomain to master
2. Edited NeverSendSentry.php to send errors to Sentry 
3. Ran `make`
4. Checked out this branch on react-commerce in my subdomain
5. Applied the diffs
6. Visited react-commerce pages

danielcliu-ifixit:
@deltuh-vee the reporting for me is working fine on localhost, so i suspect this is an issue with your config

see the expected error to show up
https://ifixit.sentry.io/issues/4557182133/?project=6475315&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=90d&stream_index=0

deltuh-vee:
QA 🎬
Errors are sent to Sentry with the diffs applied.

danielcliu-ifixit:
QA :v:  carry-over

packages/ui/hooks/index.ts
```diff
@@ -206,12 +206,13 @@ export function useLocalPreference<Data = any>(
...
-         if (data !== null) {
-            setData(data);
+         try {
+            const parsedData = validator(JSON.parse(serializedData));
+            if (parsedData !== null) {
+               setData(parsedData);
+            }
+         } catch (e) {
+            safeLocalStorage.removeItem(key);
          }
```
masonmcelvain, line 216: The try/catch here is defensive, but probably a good idea. I think we should do the same in `useExpiringLocalPreference` above

packages/auth-sdk/user.tsx
```diff
@@ -36,7 +37,10 @@ export function useAuthenticatedUser() {
       userKeys.user,
```
masonmcelvain, line 37: Why are we using `z.parse()` instead of `z.safeParse()` above? Seems like we wouldn't have had the issue in https://github.com/iFixit/react-commerce/pull/2034 in the first place

packages/auth-sdk/user.tsx
```diff
@@ -36,7 +37,10 @@ export function useAuthenticatedUser() {
...
       () => {
          const responsePromise = fetchAuthenticatedUser(apiClient).catch(
-            () => null
+            (e) => {
+               Sentry.captureException(e);
+               return null;
+            }
```
masonmcelvain, line 43: The ifixit api client already does error handling for us. The only way we'd catch this exception is if the schema parsing threw.

Why not use `safeParse` in `fetchAuthenticatedUser` and return `null` if the parsing failed? We could also report a more specific exception like

```ts
   Sentry.captureMessage('User schema parsing failed', {
      level: 'debug',
      extra: {
         userSchema,
         payload,
      },
   });
```

masonmcelvain (approved):
CR 👏🏻 Thanks for taking this on! I'm glad we're trying to prevent these kinds of errors in the future

dev_block 👍🏻 on discussing some ideas

packages/auth-sdk/user.tsx
danielcliu-ifixit: im not sure why we don't use safeParse, im assuming, like me, the writer forgot about it. However even with safeParse there would have been an issue with malformed data.

packages/auth-sdk/user.tsx
danielcliu-ifixit: I don't think there is a significant difference between your proposal and what is implemented, but we can swap in safeParse for the sake of less try catches.

masonmcelvain: I agree, I don't think there's a functional difference. My intent was to reduce try/catch's where we can, and to suggest that we report the debug error in fetchAuthenticatedUser

packages/auth-sdk/user.tsx
masonmcelvain: For sure 👍🏻

danielcliu-ifixit:
un_dev_block :v:

masonmcelvain (approved):
CR 👏🏻  Thank you for the changes!

kthaler:
QA 💠 - Followed the steps [here](https://github.com/iFixit/react-commerce/pull/2038#issuecomment-1771720780) and got [this error](https://ifixit.sentry.io/issues/4575009744/?project=6475315&query=&referrer=issue-stream&statsPeriod=14d&stream_index=0) in sentry.

PR #2040, opened 2023-10-19
Upgrade sentry

federicobadini:
Upgrade to v 7.74.0.

This PR removes @sentry/tracing since [it is deprecated](https://www.npmjs.com/package/@sentry/tracing) and it is [not necessary anymore](https://github.com/getsentry/sentry-javascript/blob/7.47.0/MIGRATION.md/#remove-requirement-for-sentrytracing-package-since-7460).

> With 7.46.0 you no longer require the @sentry/tracing package to use tracing and performance monitoring with the Sentry JavaScript SDKs. The @sentry/tracing package will be removed in a future major release, but can still be used in the meantime.

A bundle size increase that was advertised in [7.73.0](https://github.com/getsentry/sentry-javascript/blob/develop/CHANGELOG.md#7730). So if any increase is shown this might be the case.

We are upgrading also because version of Sentry have been associated to errors like the one we are tracking in #1808 - see [here](https://github.com/getsentry/sentry-javascript/issues/5978). In those case they were build errors, in ours we have runtime errors but the similarity and the fact that the fix was deployed in a newer version wrt the one we are using is worth the try.

### QA

1. Visit Vercel preview
2. Verify that the website is working as before


jordycosta:
dev_block 🟥 on merge conflicts

Otherwise, the site seems to be working as before 👍

federicobadini:
Conflict resolved
un_dev_block 👍

deltuh-vee:
Carry Over QA 🎬

danielbeardsley (approved):
CR :+1:

PR #2041, opened 2023-10-19
Add pre-commit package.json linting

### LINKED ISSUE #1346: CI Feature: Add a package.json linter to ensure static versions are used

federicobadini:
closes #1346 

This PR enforce a package json linting at pre-commit stage. 
For the moment I decided to keep this step at pre-commit stage to reduce the round-trip time given by having this kind of errors in CI. 
A save-exact rule is also specified at `.npmrc` level to help reduce the amount of non-absolute install in the project. 
At the moment we just want to enforce absolute dependencies - so we are not restricting dev dependencies

qa_req 0


ardelato:
Ahh gotcha! And it's all good 👍.

PR #1946, opened 2023-08-29
Unify on a single `prerendered` component

ianrohde:
## Issue
Augment our usage of `dangerouslySetInnerHTML` with our prerendered component when it makes sense. Prerendered takes into account the majority of styling issues we will face, and will help us avoid bugs by having a single source of truth.

Closes https://github.com/iFixit/ifixit/issues/49482

PR #2045, opened 2023-10-23
Fix prerendered html component ref

dhmacs:
This fixes a regression introduced by #1946 where the rich text component has been replaced with one that does not support refs:

Screenshot 2023-10-23 at 16 42 00

This adds refs support to the `PrerenderedHTML` component

## QA

Verify that rich text is still correctly rendered (check product list page hero description, or look at wiki text in troubleshooting pages)

### DIFF
--- a/frontend/components/common/PrerenderedHTML.tsx
+++ b/frontend/components/common/PrerenderedHTML.tsx
@@ -2,9 +2,34 @@
 // is broken up into `shared`, `troubleshooting`, and `commerce` variables to control what
 // is included, and where via the styleMap.
 
-import { useEffect } from 'react';
-import { Box, chakra, SystemStyleObject } from '@chakra-ui/react';
+import { Box, SystemStyleObject, forwardRef } from '@chakra-ui/react';
 import 'lite-youtube-embed/src/lite-yt-embed.css';
+import { useEffect } from 'react';
+
+interface PrerenderedHTMLProps {
+   html: string;
+   template: 'troubleshooting' | 'commerce';
+   className?: string;
+}
+
+export const PrerenderedHTML = forwardRef<PrerenderedHTMLProps, 'div'>(
+   function PrerenderedHTML({ html, template, className }, ref) {
+      useEffect(() => {
+         if (template === 'troubleshooting') {
+            import('lite-youtube-embed');
+         }
+      }, [template]);
+
+      return (
+         <Box
+            ref={ref}
+            className={`prerendered ${className}`}
+            sx={styleMap[template]}
+            dangerouslySetInnerHTML={{ __html: html }}
+         />
+      );
+   }
+);
 
 const constrainStandardWidth = '282px'; // pulled from PHP app for legacy image sizing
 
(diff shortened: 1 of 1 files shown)


frontend/components/common/PrerenderedHTML.tsx
```diff
@@ -2,9 +2,34 @@
 // is broken up into `shared`, `troubleshooting`, and `commerce` variables to control what
 // is included, and where via the styleMap.
 
-import { useEffect } from 'react';
-import { Box, chakra, SystemStyleObject } from '@chakra-ui/react';
+import { Box, SystemStyleObject, forwardRef, BoxProps } from '@chakra-ui/react';
```
sctice-ifixit, line 5: It was there before, but it looks like `SystemStyleObject` is unused. Did you try dropping it?

sctice-ifixit (approved):
CR :flags: Thanks for fixing this. Was there a particular problem that brought it up? Any suggestions for how we might have noticed this sooner?

dhmacs:
@sctice-ifixit 

> Was there a particular problem that brought it up? 

I was working on something else and noticed the console error (screenshot above) in the dev tools.

> Any suggestions for how we might have noticed this sooner?

My recommendation is to keep dev tools open when making changes, these errors usually appear immediately in the console upon page load

frontend/components/common/PrerenderedHTML.tsx
dhmacs: As a side note I use quite often VSCode [organize imports](https://code.visualstudio.com/docs/languages/javascript#_organize-imports) which is quite handy to get rid of unused imports, and can be trigger with shortcuts:

Screenshot 2023-10-23 at 17 59 58

sctice-ifixit: Sorry, I wrote "it was there before", but then I only looked for usages in the diff. This `SystemStyleObject` is definitely used.

ianrohde:
@dhmacs so you recall what page this error was on? I would not have proceeded if there were any console errors, so I likely need to expand my store testing.

masonmcelvain:
In the QA section, he mentions "check product list page hero description, or look at wiki text in troubleshooting pages", so maybe those were the culprits?

ianrohde:
I absolutely checked the wiki text on multiple pages and did not observe any console errors. I'll keep the product list page Hero in mind as an addition.

dhmacs:
@ianrohde This was visible on the product list page, because in the hero section we're passing a ref to that component to calculate its height (so we know if we have to display the "show more" button). I don't think it affected wiki pages (QA instructions were to make sure that I didn't mess up anything).

masonmcelvain:
Just so everyone here is aware, we unintentionally shipped a styling bug in this PR that is being fixed here https://github.com/iFixit/ifixit/issues/51907. We live and we learn 😄

dhmacs:
Didn't expect `chakra` factory to merge `sx` property rather than override it. 🤔 TIL.

PR #2049, opened 2023-10-24
Upgrade Zod to fix vulnerability

dhmacs:
Zod has patched a Vulnerable email regex: https://github.com/colinhacks/zod/pull/2824 so let's upgrade the dependency in our codebase. It's unlikely though that we're impacted by this as it doesn't seem like we're using `.email` validation.

qa_req 0


masonmcelvain (dismissed):
CR 👏🏻

masonmcelvain (approved):
CR 👏🏻

PR #2051, opened 2023-10-24
Eslint: Add a package for custom rules

jarstelfox:
Based on #2036, I wanted to ensure we don't run into sentry data loss again.

I made a simple eslint rule to tell us not to use Error, but default to SentryError.
The Sentry details are optional, so there is no harm in forcing us to only use this error type.

I did not audit call sites to add extra data when running the eslint fix command, so feel free to suggest improvements.

qa_req 0 just passing tests


jarstelfox:
Woah, that merge was weird. I am going to force push.

packages/eslint/rules/no-new-error.js
```diff
@@ -0,0 +1,34 @@
...
+               node.callee.type === 'Identifier' &&
+               node.callee.name === 'Error'
+            ) {
+               context.report({
+                  node,
+                  message:
+                     'Avoid using new Error; consider using SentryError instead to attach contexts, tags, and more.',
+                  fix: function (fixer) {
+                     // Suggest replacing 'new Error' with 'new SentryError'
+                     return fixer.replaceText(node, 'new SentryError');
```
danielcliu-ifixit: im not super familiar with eslint, but `node` is `new Error` not just `Error` correct?

danielcliu-ifixit: op i see you address this

packages/eslint/rules/no-throw-in-sentry-scope.js
```diff
@@ -0,0 +1,51 @@
...
+               node.callee.type === 'MemberExpression' &&
+               node.callee.object &&
+               node.callee.object.name === 'Sentry' &&
+               node.callee.property &&
+               node.callee.property.name === 'withScope'
+            ) {
+               isSentryWithScope = true;
+            }
+         },
+         'CallExpression:exit': function (node) {
```
danielcliu-ifixit, line 26: I don't understand why we would need this, the expression either has scope or not no? Why do we need to check at the end of the expression?

danielcliu-ifixit (approved):
CR :v:

packages/eslint/rules/no-throw-in-sentry-scope.js
jarstelfox: Ah, this is to allow checking nested functions.

An example:

```
const func1 = () => {
   throw new Error();
};

const func2 = () => {
   func1();
};


Sentry.withScope(scope => {
  func2();
});
```

I am looking for the withScope the marking it as such. We also have to manually unmark it in eslint once we leave the withScope function call.

Issue #2046, opened 2023-10-24
Remove Google UA


danielcliu-ifixit:
## Overview

Based on [this comment,](https://ifixit.slack.com/archives/C051KPSGP19/p1698079977401639) we can now start removing Google Universal Analytics

CC @sctice-ifixit @addison-grant

PR #2054, opened 2023-10-25
Drop Google Universal Analytics Code

danielcliu-ifixit:
## Overview

We don't want to use Google UA anymore, so lets remove the code that supports it. These changes just drop GUA related code and replaces GUA usage with just using piwik now.

## QA

Make sure GA4 still works and GUA doesn't, i guess?

Closes: #2046


packages/analytics/index.tsx
```diff
@@ -1,54 +1,2 @@
...
-/**
- * @param trackData trackData.eventName will default to the page path if not provided
- */
-export const trackInPiwikAndGA = (trackData: TrackEventPiwik) => {
-   if (typeof window === 'undefined') {
-      return;
-   }
-   const eventName =
-      trackData.eventName ||
-      `${window.location.origin}${window.location.pathname}`;
```
jarstelfox, line 18: This is the only thing I'm sad to see go.

Previously, this was optional and set to the default Max asked for. Now you must repeat it at each callsite

jarstelfox (approved):
CR 🌵

packages/analytics/index.tsx
danielcliu-ifixit: Yeah its a little messier but i felt it was the best decision. Do you think theres is a cleaner solution?

jarstelfox: Yea, I think a small wrapper around pushToPiwik would be nice:

```
export type TrackEvent = {
   /**
    * Describes the type of events you want to track.
    * For example, Link Clicks, Videos, Outbound Links, and Form Events.
    */
   eventCategory: string;
   /**
    * The specific action that is taken.
    * For example, with a Video category, you might have a Play, Pause and Complete action.
    */
   eventAction: string;
   /**
    * Usually the title of the element that is being interacted with, to aid with analysis.
    * For example, it could be the name of a Video that was played or the specific
    * form that is being submitted.
    */
   eventName?: string;
   /**
    * A numeric value and is often added dynamically. It could be the cost of a
    * product that is added to a cart, or the completion percentage of a video.
    */
   eventValue?: number;
};


export const trackEventInPiwik = (trackData: TrackEvent) => {
   pushToPiwik([
      'trackEvent',
      trackData.eventCategory,
      trackData.eventAction,
      trackData.eventName,
      trackData.eventValue,
   ]);
};
```

The main benefit is that we get some extra type info.

I don't know if this holds true for Piwik, but Matomo actually responded with a 200 (data sent) but threw out any data that did not match the types.

Example:

```
pushToPiwik([
      'trackEvent',
      '',
      '',
      '',
      "1234",
   ]);
```

The fact we send a string caused Matomo to toss the whole event and not be recorded.

deltuh-vee:
QA 🎬 
GA4 is still working. Pretty sure GUA has stopped processing events as of July 1, 2023.

packages/analytics/index.tsx
danielcliu-ifixit: i might do this in my consolidation pull

PR #2056, opened 2023-10-26
User SDK: Don't validate null user data

masonmcelvain:
We got thousands of sentry events overnight notifying us that the value "null" failed the user schema validation. Let's bail early with no validation in that case.

Slack: https://ifixit.slack.com/archives/C04C05FA3EH/p1698329545488009

Sentry: https://ifixit.sentry.io/share/issue/d7d55c09325f4e55880ffdeea3c7e202/

Followup to: https://github.com/iFixit/react-commerce/pull/2038

qa_req 0 when this deploys, we can validate that it's working be observing whether the sentry events start trending down


dhmacs (approved):
CR ⚡

masonmcelvain:
Looks promising so far

image

image

PR #2057, opened 2023-10-26
Update FooterLink

jordycosta:
### Overview:

There were instances where the _Help translate_ instance of the FooterLink was being styled from some other stylesheet
(read https://github.com/iFixit/ifixit/issues/50320#issuecomment-1781328352)

In order to resolve this issue, we now assign the _customColor_ property to the wrapped **Text** component so it has its own color styling

---

**QA:** Make sure there isn't any change to the footer

---

Connects https://github.com/iFixit/ifixit/issues/50320

### DIFF
--- a/packages/footer/components/Shared.tsx
+++ b/packages/footer/components/Shared.tsx
@@ -72,7 +72,7 @@ export const FooterLink = forwardRef<FooterLinkProps, 'a'>(
             onClick={trackedOnClick}
             {...otherProps}
          >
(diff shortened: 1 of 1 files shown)


jordycosta:
dev_block 🟥

deltuh-vee:
QA 🎬 
Footer is unchanged.

batbattur (approved):
CR 🚛 makes sense to me

PR #2058, opened 2023-10-26
Pagination: Don't Use Search Params On First Page

danielcliu-ifixit:
## Overview

If pagination page is the first page dont add search params. We don't need p=1 as a url parameter when its the same page as the url by itself. Lets just always use the canonical url.

## QA
You can read the issue and find it pretty straightforward, but basically for pagination on collection list pages we don't need search params for the first page. 

Closes: https://github.com/iFixit/ifixit/issues/50321


davidrans (approved):
CR :coffee:

deltuh-vee:
QA 🎬 
Links to the first page of a paginated page go to the canonical page.

PR #2059, opened 2023-10-27
Worksin: Support passing it via the url

jarstelfox:
I chose not to include it in the query params filter on URL generation

This pull still supports the old URL style as well.

QA
---

The works-in filter should be correct for:
* http://localhost:3000/Parts/PlayStation_4:CUH-11XXA 
* http://localhost:3000/Parts/PlayStation_4?filter%5Bworksin%5D=PlayStation%204%20CUH-11XXA
* http://localhost:3000/Parts/PlayStation_4?filter%5Bfacet_tags.Capacity%5D%5B0%5D=1%20TB 
* http://localhost:3000/Parts/PlayStation_4:CUH-11XXA?filter%5Bfacet_tags.Capacity%5D%5B0%5D=1%20TB
* http://localhost:3000/Parts/PlayStation_4
* http://localhost:3000/Parts/iPod_Mini?filter%5Bworksin%5D=iPod%20Mini%20A1051%202nd%20gen%204%20GB

For clarity both `Parts/PlayStation_4:CUH-11XXA` and `PlayStation_4?filter...` should work. We will prefer the `:` version of the works in value.

Connects: https://github.com/iFixit/ifixit/issues/50436

### DIFF
--- a/frontend/components/common/InstantSearchProvider.tsx
+++ b/frontend/components/common/InstantSearchProvider.tsx
@@ -4,7 +4,7 @@ import { getClientOptions } from '@helpers/algolia-helpers';
 import {
    destylizeDeviceItemType,
    destylizeDeviceTitle,
-   destylizeDeviceTitleAndWorksIn,
+   destylizeDeviceTitleAndVariant,
    getFacetWidgetType,
    isValidRefinementListValue,
    stylizeDeviceItemType,
@@ -97,7 +97,7 @@ export function InstantSearchProvider({
          const devicePath = isDevicePartsPage
             ? getDevicePath(deviceHandle, routeState)
             : deviceHandle;
-         const devicePathIsWorksIn =
+         const devicePathHasVariant =
             isDevicePartsPage && devicePath.includes(':');
 
          let path = `/${firstPathSegment}`;
@@ -119,7 +119,7 @@ export function InstantSearchProvider({
             // Item Type is the slug on device pages, not in the query.
             delete filterCopy['facet_tags.Item Type'];
          }
-         if (devicePathIsWorksIn) {
+         if (devicePathHasVariant) {
             delete filterCopy['worksin'];
          }
          const { q, p, filter, ...otherParams } = qsModule.parse(
@@ -156,7 +156,7 @@ export function InstantSearchProvider({
          const deviceHandle = pathParts.length >= 2 ? pathParts[1] : '';
          const itemType = pathParts.length >= 3 ? pathParts[2] : '';
          const isDevicePartsPage = firstPathSegment === 'Parts' && deviceHandle;
-         const deviceHandleIsWorksIn =
+         const devicePathHasVariant =
             isDevicePartsPage && deviceHandle.includes(':');
 
          const { q, p, filter } = qsModule.parse(location.search, {
@@ -199,9 +199,9 @@ export function InstantSearchProvider({
             ).trim();
          }
 
-         if (deviceHandleIsWorksIn) {
+         if (devicePathHasVariant) {
             filterObject['worksin'] =
-               destylizeDeviceTitleAndWorksIn(deviceHandle);
+               destylizeDeviceTitleAndVariant(deviceHandle);
          }
 
          return {
@@ -334,17 +334,17 @@ function getBaseOrigin(location: Location): string {
 }
 
 function getDevicePath(handle: string, routeState: RouteState): string {
-   const [device, worksin] = handle.split(':');
-   const worksinFromRouteState = routeState.filter?.['worksin'];
+   const [device, variant] = handle.split(':');
+   const variantFromRouteState = routeState.filter?.['worksin'];
 
-   if (worksin && !worksinFromRouteState) {
+   if (variant && !variantFromRouteState) {
       return device;
    }
 
-   if (!worksin && worksinFromRouteState) {
+   if (!variant && variantFromRouteState) {
       const title = destylizeDeviceTitle(handle);
-      const newWorksin = worksinFromRouteState.replace(title, '').trim();
-      return `${handle}:${newWorksin}`;
+      const varaintToAdd = variantFromRouteState.replace(title, '').trim();
+      return `${handle}:${varaintToAdd}`;
    }
 
    return handle;
--- a/frontend/components/common/InstantSearchProvider.tsx
+++ b/frontend/components/common/InstantSearchProvider.tsx
@@ -3,6 +3,7 @@ import { ALGOLIA_APP_ID, IFIXIT_ORIGIN } from '@config/env';
 import { getClientOptions } from '@helpers/algolia-helpers';
 import {
    destylizeDeviceItemType,
+   destylizeDeviceTitleAndWorksIn,
    getFacetWidgetType,
    isValidRefinementListValue,
    stylizeDeviceItemType,
@@ -92,6 +93,8 @@ export function InstantSearchProvider({
          const firstPathSegment = pathParts.length >= 1 ? pathParts[0] : '';
          const deviceHandle = pathParts.length >= 2 ? pathParts[1] : '';
          const isDevicePartsPage = firstPathSegment === 'Parts' && deviceHandle;
+         const deviceHandleIsWorksIn =
+            isDevicePartsPage && deviceHandle.includes(':');
 
          let path = `/${firstPathSegment}`;
          if (deviceHandle) {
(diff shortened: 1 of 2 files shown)


deltuh-vee:
dev_block ✊ 
If the filter is removed via the Clear all filters button in the sidebar, I think it should also be cleared from the url. I noticed that the worksin filter reappears if I select an item type.
CleanShot 2023-10-27 at 15 28 09@2x

sterlinghirsh (approved):
CR :zap:

This looks fine to me. I'm ok to merge as-is to get it going, but maybe in a future pull I think we should standardize the terminology. The "worksin" name here is a little misleading. The field is called worksin on Algolia so it's not crazy for us to pick that name, but we've mostly been using the term "variant" or "topic variant" (though I think variant is sufficient in this context). So for the url `/Parts/ABC:123`, the device title is ABC and the variant is 123.

I say this because in the php app, both variants and worksins exist and are separate but related topics. A `worksin` (identified by worksinid) is a way to connect one or more products to one or more topic variants. The worksinid is associated with the product by entering the worksinid in an Akeneo field. Another way of looking at is is that the worksin represents a set of topic variants, and a product can optionally be associated with a single worksinid. In Algolia's product_group_en index, worksin is an array field that contains a list of variants, so I think if we referred to it as a variant here, it would make sense that a variant would be a member of the worksin array.

That said, it's extra confusing because product variants are also a thing (a product group is a collection of 1 or more product variants aka product options), so having a field on product_group_en called variants could also be confusing (maybe compatible_variants or compatible_topic_variants? but that's more verbose and worksin is probably fine). So anyway it's a pretty nitpicky point but I figured I'd mention it.

sterlinghirsh:
@deltuh-vee is right that we should update the url to remove the variant if the filter is cleared.

jarstelfox:
Thanks for the catch. I agree and should have fixed it.

In the process, it was easy to for URLs like this:

`http://localhost:3000/Parts/PlayStation_4?filter%5Bworksin%5D=PlayStation%204%20CUH-11XXA`

To change to `http://localhost:3000/Parts/PlayStation_4:CUH-11XXA` on load effectively updating old URLs to the new format.

un_dev_block 🌵

kthaler:
QA 💠 - The `/Device_Name:variant_title` URL format works correctly. The `worksin` query parameter still works, but redirects to the new format. Clicking "Clear all filters" removes the `:variant_title` portion of the URL. Overall, the new URL format works as expected!

frontend/helpers/product-list-helpers.ts
```diff
@@ -57,16 +57,35 @@ export function stylizeDeviceItemType(itemType: string): string {
...
-export function stylizeDeviceTitle(deviceTitle: string): string {
-   return deviceTitle.replace(/\s/g, '_');
+export function stylizeDeviceTitle(
+   deviceTitle: string,
+   variant?: string
+): string {
+   const urlTitle = deviceTitle.replace(/\s/g, '_');
+   const urlVariant = variant ? `:${variant}` : '';
+   return urlTitle + urlVariant;
 }
```
masonmcelvain, line 89: Will the variant ever have spaces? If so, do we want those spaces to be converted to `_`?

frontend/components/common/InstantSearchProvider.tsx
```diff
@@ -315,3 +332,20 @@ function getBaseOrigin(location: Location): string {
...
+   const variantFromRouteState = routeState.filter?.['worksin'];
+
+   if (variant && !variantFromRouteState) {
+      return device;
+   }
+
+   if (!variant && variantFromRouteState) {
+      const title = destylizeDeviceTitle(handle);
+      const varaintToAdd = variantFromRouteState.replace(title, '').trim();
+      return `${handle}:${varaintToAdd}`;
```
masonmcelvain, line 347: Very small typo, not worth changing unless you're making other changes
```suggestion
      const variantToAdd = variantFromRouteState.replace(title, '').trim();
      return `${handle}:${variantToAdd}`;
```

masonmcelvain (dismissed):
CR 👏🏻 but deploy_block ❓ on a question

frontend/helpers/product-list-helpers.ts
jarstelfox: Ah, good point. I am having a hard time finding works-in which would have spaces in it.

Take this example:

image

`Google Pixel 7 Pro GP4BC` would be the works in, but the base page is `/Parts/Google_Pixel_7_Pro` This means the URL will end up `/Parts/Google_Pixel_7_Pro:GP4BC`

Give me some time to dig into our data and I'll have a concrete answer

masonmcelvain: Spaces exist in the topic_variants table, but I'm not sure where if that's where the variant title comes from

```
mysql> select variant from topic_variants where variant like "% %" limit 20;
+--------------------------------------+
| variant                              |
+--------------------------------------+
| A1181 (EMC 2092 MacBook1,1) 1.83 GHz |
| A1181 (EMC 2092 MacBook1,1) 2 GHz    |
| A1051 1st gen 4 GB                   |
| A1051 2nd gen 4 GB                   |
| A1051 2nd gen 6 GB                   |
| A1241 16GB                           |
| A1241 8GB                            |
| A1137 1GB                            |
| A1137 2GB                            |
| A1137 4GB                            |
| A1199 2GB                            |
| A1199 4GB                            |
| A1199 8GB                            |
| A1236 4GB                            |
| A1236 8GB                            |
| A1285 16GB                           |
| A1285 8GB                            |
| A1112 1 GB                           |
| A1112 512 MB                         |
| A1204 1 GB                           |
+--------------------------------------+
20 rows in set (0.00 sec)
```

masonmcelvain: Spaces exist in the topic_variants table, but I'm not sure where if that's where the variant title comes from

```sql
select variant from topic_variants where variant like "% %"
```

jarstelfox:
deploy_block 🌵 on allowing spaces in the variants and also this thread: https://ifixit.slack.com/archives/C01GMQXJC95/p1698697001116009

frontend/components/common/InstantSearchProvider.tsx
```diff
@@ -315,3 +332,20 @@ function getBaseOrigin(location: Location): string {
    }
    return location.origin;
 }
+
+function getDevicePath(handle: string, routeState: RouteState): string {
+   const [device, variant] = handle.split(':');
+   const variantFromRouteState = routeState.filter?.['worksin'];
```
danielbeardsley, line 338: What's with this property accessor syntax? It seems like `['worksin']` was done intentionally, instead of the more idiomatic `.worksin` but I don't see why. It makes me think that there's something I'm missing or some odd feature used here that I'm not aware of.

danielbeardsley: Pretty sure it's not needed and they are equivalent:

image

frontend/helpers/product-list-helpers.ts
danielbeardsley: > I'm not sure where if that's where the variant title comes from

You can be sure. You are correct:

https://github.com/iFixit/ifixit/blob/d628373ca823190783ad8909cbc0b3aacd29f04c/Objects/Search/Algolia/Handlers/AlgoliaProductGroupHandler.php#L169-L186

18% of variants have spaces:
image

frontend/components/common/InstantSearchProvider.tsx
jarstelfox: Totally. I just made a mistake by tab completing copilot.

jarstelfox:
un_dev_block 🌵 un_deploy_block 🌵 

We now handle spaces and have chosen that the current implementation trade-offs are reasonable.

QA - I added another test case: http://localhost:3000/Parts/iPod_Mini?filter%5Bworksin%5D=iPod%20Mini%20A1051%202nd%20gen%204%20GB

which is a case where the variant and device both have spaces 👍🏻

frontend/components/common/InstantSearchProvider.tsx
```diff
@@ -315,3 +333,18 @@ function getBaseOrigin(location: Location): string {
    }
    return location.origin;
 }
+
+function getDevicePath(handle: string, routeState: RouteState): string {
+   const [device, variant] = handle.split(':');
```
danielbeardsley, line 338: I just realized this should have been coded to handle a `:` in the variant text. Right now, it throws away the rest. We'll need to use `indexOf` or something cause `split(':', 2)` also throws away the rest of the string.

There are only 2 variants with colons in them, but I expect that could grow.

image

PR #2062, opened 2023-10-27
NetSuite: Add new partner codes

masonmcelvain:
Connects https://github.com/iFixit/ifixit/issues/50282

Adds a map of netsuite partner values to icon components. When we migrate to the NetSuite ERP, we'll no longer have OEM partner codes, just display values.

These partner values come from the proposed values in this [spreadsheet](https://docs.google.com/spreadsheets/d/1R0l4vuSCNepE68od0hsO6gfaxEzzF9U5vjF2U248-1Q/edit#gid=508051636).

## QA

Make sure genuine parts still have the correct partner icon, as on `main`.

### DIFF
--- a/frontend/templates/product/sections/ProductOverviewSection/GenuinePartBanner.tsx
+++ b/frontend/templates/product/sections/ProductOverviewSection/GenuinePartBanner.tsx
@@ -42,6 +42,24 @@ const partnerCodeToComponentMap: { [key: string]: React.FC } = {
    valve: Valve,
 };
 
+const netsuitePartnerToComponentMap: { [key: string]: React.FC } = {
+   'Genuine Google Pixel Part': Google,
+   'Genuine Crucial Drive': () => <Crucial px="2" />,
+   'Genuine HP Part': () => <Hp py="2" />,
+   'Genuine HTC Vive Part': Vive,
+   'Genuine Lenovo Part': Lenovo,
+   'Genuine Logitech Part': Logitech,
+   'Genuine Micron Drive': () => <Micron px="2" />,
+   'Genuine Microsoft Surface Part': Microsoft,
+   'Genuine Microsoft Surface Tool': Microsoft,
+   'Genuine Motorola Part': Motorola,
+   'Genuine Nokia Part': Nokia,
+   'Genuine Samsung Galaxy Part': Samsung,
+   'Genuine Steam Deck Part': Steam,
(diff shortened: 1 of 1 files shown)


deltuh-vee:
QA 🎬 
Genuine parts still have the correct partner icon.

danielbeardsley:
It worries me where we write conditional logic based on display values. Should we "report to sentry" if we see a value *not* in the list?

I have a different idea I'd like us to consider. In this case, our frontend code is expecting the metafield to look like: `{text, code, url}`. I think I'd rather keep the same structure here and do this lookup upstream when we send data to shopify. Or perhaps use the "codes" for the netsuite values and look up the display values in our php app.

Effectively, this app still needs both display values and unchanging codes, so I'd rather leave this interface alone and deal with it upstream.

What do you think?

deploy_block :+1:

masonmcelvain:
@danielbeardsley see the backend work here https://github.com/iFixit/ifixit/pull/50429

My understanding is that option codes are going away when we move to NetSuite, hence the series of pulls I put out for https://github.com/iFixit/ifixit/issues/50282. Are you suggesting we map display values to "virtual" option codes in the backend in order to keep the interface unchanged for consumers like react-commerce and shopify-v2?

danielbeardsley:
> Are you suggesting we map display values to "virtual" option codes in the backend in order to keep the interface unchanged for consumers like react-commerce and shopify-v2?

That's one way of doing it.

We have to translate values that come out of netsuite into some behaviors in at least two places (react-commerce and shooify-v2).

I propose to do that mapping in our php app (one place) and communicate it by exposing the value using the same set of codes that we do now for Akeneo. This means consumers wouldn't have to change and it means that our php app would have the opportunity to fail to sync products with values outside of the known set. Changing a value in netsuite could at worst prevent the product from further syncing and show an error in ProductErrors.

Since we're here, I'd propose to put the "codes" in netsuite i.e. short, lower-case, non-display values and keep the important display values in our php app as a lookup with the same "report values outside the known set" logic as above.

Whether we have codes in Netsuite and display values in php or the reverse doesn't really matter that much, though it'll be easier to have translated values in php than in Netsuite.

masonmcelvain:
Ok well @sterlinghirsh nudged me in this direction here https://github.com/iFixit/ifixit/pull/50429#discussion_r1373914316. I'm happy to go back to using the codes, telling @andymdj that we need the codes to be in NetSuite, and closing these consumer pulls

PR #2068, opened 2023-10-31
Simplify FaIcon types

andyg0808:
FaIcon had its types changed and a cast added as part of the TS 5 upgrade. (fe1ebadfb0109db9e65205da3b07e413a8a042c7). This removes some of the changes, in an attempt to support the full range of possible types which the `color` prop supports (e.g., ResponsiveArray types).

qa_req 0
This only affects types, so no QA should be needed.


danielbeardsley (approved):
CR :+1:

PR #2069, opened 2023-10-31
Upgrade to Next.js 14

dhmacs:
We have a few errors (https://github.com/iFixit/react-commerce/issues/1808, https://github.com/iFixit/react-commerce/issues/1913) that seems to be related to Next.js. Let's update to the latest version to see if they get fixed.

WIP


dhmacs:
I will open this again when we finish porting pages to app router

PR #2070, opened 2023-10-31
Vulcan: Sidebar: Update Image Aspect Ratio

ianrohde:
## Issue

Update Vulcan sidebar image aspect ratio to match our images. Previously set at 1:1, but images are 4:3.

Mattia may have updates to the design at a later time ([Figma discussion](https://www.figma.com/file/QVLUEpDVoyiMM4MXEGq1L9?node-id=201:65526&mode=dev#600241124)), but this resolves the issue for now.

## CR/QA

Verify Image ratio has been updated and looks good on mobile/desktop.

`/Troubleshooting/Mac_Laptop/MacBook+Will+Not+Turn+On/483799`

### DIFF
--- a/frontend/templates/troubleshooting/Problem.tsx
+++ b/frontend/templates/troubleshooting/Problem.tsx
@@ -42,12 +42,14 @@ export default function ProblemCard({ problem }: { problem: Problem }) {
                outline="1px solid"
                outlineColor="gray.300"
                borderRadius="md"
-               boxSize={{ base: '56px', md: '32px' }}
+               width={{ base: '75px', md: '43px' }}
+               height={{ base: '56px', md: '32px' }}
                mr={2}
+               aspectRatio="4 / 3"
             >
                <Image
                   height="100%"
-                  htmlWidth={56}
+                  htmlWidth={75}
                   htmlHeight={56}
                   objectFit="cover"
                   src={imageUrl}
--- a/frontend/templates/troubleshooting/index.tsx
+++ b/frontend/templates/troubleshooting/index.tsx
@@ -955,13 +955,15 @@ function RelatedProblems({
                      <Image
                         src={imageUrl}
                         alt={displayTitle}
-                        boxSize={{ base: '56px', md: '96px' }}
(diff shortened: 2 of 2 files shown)


jordycosta:
dev_block 🟥 on receiving a 500 Internal Server Error on [this page](https://react-commerce-git-vulcan-sidebar-update-image-as-050a2e-ifixit.vercel.app/Troubleshooting/Mac_Laptop/MacBook+Will+Not+Turn+On/483799#solution-4)

jordycosta:
QA 🟢 

I think this looks much better!

---

Here is a before/after on both mobile and desktop views:

https://github.com/iFixit/react-commerce/assets/95656772/daad2053-a7cf-4bfc-806d-d60e4e65a73c

frontend/templates/troubleshooting/Problem.tsx
```diff
@@ -42,12 +42,14 @@ export default function ProblemCard({ problem }: { problem: Problem }) {
                outline="1px solid"
                outlineColor="gray.300"
                borderRadius="md"
-               boxSize={{ base: '56px', md: '32px' }}
+               width={{ base: '75px', md: '43px' }}
+               height={{ base: '56px', md: '32px' }}
                mr={2}
+               aspectRatio="4 / 3"
```
aburke07, line 48: Just curious, do we need to specify width _and_ height if we have an aspect ratio? It seems redundant and not extremely 'safe' in the case that we specify the wrong WxH that doesn't match the aspectRatio.

aburke07 (approved):
CR :alien:

frontend/templates/troubleshooting/Problem.tsx
ianrohde: I added `aspectRatio` as a guard in case sizes are changed later.

ianrohde: Width/height overwrite aspectRatio, so you're absolutely right. TIL.

Do you think it's worth correcting here?

PR #2072, opened 2023-11-01
ifixit-api-client: Use peer version of `@ifixit/sentry`

andyg0808:
This gives us a way to easily get access to the `@ifixit/sentry` version that's being used, rather than needing to override a dependency of a dependency to configure that version.

qa_req 0
As long as this builds, it shouldn't affect functionality, as we're already using `@ifixit/sentry` in `frontend`.

CC @jarstelfox
Connects https://github.com/iFixit/ifixit/issues/50513


danielbeardsley (approved):
CR :+1:

PR #2073, opened 2023-11-01
Revalidation reqs: Log info on success 

jarstelfox:
qa_req 0

I could not figure out how to test this change locally. Any advice is appreciated

Closes https://github.com/iFixit/ifixit/issues/49898

### DIFF
--- a/frontend/lib/swr-cache/index.ts
+++ b/frontend/lib/swr-cache/index.ts
@@ -111,6 +111,12 @@ export const withCache = <
          req.write(postData);
          req.end(() => {
             const elapsed = performance.now() - start;
+            logger.info('revalidation request end', {
+               path: req.path,
+               method: req.method,
+               postData,
+               elapsed,
+            });
             logger.success.event(`${statName}.revalidation.request_end`);
             logger.success.timing(
                `${statName}.revalidation.request_end`,
--- a/frontend/lib/swr-cache/index.ts
+++ b/frontend/lib/swr-cache/index.ts
@@ -111,12 +111,7 @@ export const withCache = <
          req.write(postData);
          req.end(() => {
             const elapsed = performance.now() - start;
-            logger.info('revalidation request end', {
-               path: req.path,
-               method: req.method,
-               postData,
-               elapsed,
-            });
+            logger.info('revalidation request end', postData);
             logger.success.event(`${statName}.revalidation.request_end`);
             logger.success.timing(
                `${statName}.revalidation.request_end`,
--- a/packages/helpers/logger.ts
+++ b/packages/helpers/logger.ts
@@ -1,4 +1,4 @@
-type Log = (message: string) => void;
+type Log = typeof console.log;
 type ExtendedLog = Log & {
    timing: (statName: string, ms: number) => void;
    event: (statName: string) => void;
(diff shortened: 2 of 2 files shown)


frontend/lib/swr-cache/index.ts
```diff
@@ -111,6 +111,12 @@ export const withCache = <
          req.write(postData);
          req.end(() => {
             const elapsed = performance.now() - start;
+            logger.info('revalidation request end', {
+               path: req.path,
+               method: req.method,
```
masonmcelvain, line 116: This is always going to be 'POST' right?

masonmcelvain (dismissed):
CR 👏🏻 Nice! Hopefully we can see some patterns in the path or variables

frontend/lib/swr-cache/index.ts
jarstelfox: Yes. I should probably remove it then.

masonmcelvain:
> I could not figure out how to test this change locally. ANy advice is appreciated

These show up in the Vercel logs when you visit a page that uses the cache (I think product pages and product list pages currently).

e.g. here's a [log in vercel](https://vercel.com/ifixit/react-commerce-prod/logs?page=1&timeline=past30Minutes&startDate=1698883640926&endDate=1698885440926&selectedLogId=1698885435747126941116755990&selectedLogTimestamp=1698885435747&forceEndDate=1698885437615)

image

I think these logs also appear in your console when you're running a development server and you've setup a redis backend with [upstash](https://upstash.com/) (we should probably add how to use upstash to the readme if we plan to keep using redis)

https://github.com/iFixit/react-commerce/blob/013df08f0a54215a65f01ca6ebff1bc861be9de3/frontend/.env.local.example#L3

frontend/lib/swr-cache/index.ts
masonmcelvain: Up to you, I just wanted to make sure I understood correctly

frontend/lib/swr-cache/index.ts
```diff
@@ -111,6 +111,12 @@ export const withCache = <
...
+               postData,
```
danielbeardsley, line 117: It seems like this (`postData`) is the only data we aren't already logging, so I'd vote for just logging that.

The lines below log this:
`cache.findProduct.stale => 12.97ms`  -- This shows *which cache* and *how long it took*

danielbeardsley (changes requested):
dev_block :+1: on just logging the data we're not already logging.

packages/helpers/logger.ts
```diff
@@ -1,4 +1,4 @@
-type Log = (message: string) => void;
+type Log = typeof console.log;
```
dhmacs, line 1: This change appears to tie the type to a specific implementation. We defined the type that way for adaptability to various implementations (e.g., file system, Slack, etc). Are there particular reasons for binding this type to the `console.log` implementation?

danielbeardsley (approved):
CR :+1:

packages/helpers/logger.ts
dhmacs: Ah I see, I like the idea of extending behaviour, improving the DX of the logger 👍  Imo it's safer if you explicitly tell what use cases we intend to support instead of saying "whatever console.log supports". For example, newer versions of node.js might introduce new stuff into `console.log` thus changing the contract. If you think that's not a concern because we'll always use only the `console.log` implementation then it's fine 👍

jarstelfox:
un_dev_block 🌵

danielbeardsley (approved):
CR :+1:

PR #2077, opened 2023-11-02
Fix `@ifixit/helpers` imports

andyg0808:
This switches to only doing an `import type` in `@ifixit/helpers` for NextJS, because we don't actually need the module itself. It also adds a peer dependency on `next`, because there isn't one yet. (I'm assuming that somehow the `react-commerce` build happens to have `next` available all over the place, which is why that hasn't broken yet.)

qa_req 0
This shouldn't make any actual difference, since it's type-only.

Connects https://github.com/iFixit/ifixit/issues/50513

### DIFF
--- a/packages/helpers/nextjs.tsx
+++ b/packages/helpers/nextjs.tsx
@@ -1,4 +1,4 @@
-import { GetServerSidePropsContext } from 'next';
+import type { GetServerSidePropsContext } from 'next';
 
 export function urlFromContext(context: GetServerSidePropsContext): string {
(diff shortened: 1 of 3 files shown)


danielbeardsley (approved):
CR :+1:

danielbeardsley:
Can you explain the most recent commit? Why the large diff in the lock file?

andyg0808:
Good question; I'm not sure. I'm still fighting this; I'm not sure that I need to make this change.

PR #2081, opened 2023-11-04
Add React keys to meta tags

dhmacs:
This PR improves rendering of meta tags by introducing React keys to deduplicate duplicate meta tags

qa_req 0


masonmcelvain (approved):
CR 👍🏻

PR #2083, opened 2023-11-07
Correct bit table section spacing

federicobadini:
The bit table section code included an additional extra margin that was not necessary.
Here the section spacing were reconciled to the mockup values:

Screenshot 2023-11-07 at 18 08 51

qa_req 0


jarstelfox (dismissed):
CR 🌵

jarstelfox (approved):
CR 🌵

PR #2086, opened 2023-11-07
Variants: Support variants with colons

jarstelfox:
I am not sure we want URls like this long-term, but we know we have at least two examples where we have variants that could work easily with this one-line fix.

Caused by https://github.com/iFixit/react-commerce/pull/2059#discussion_r1385419955

CC @danielbeardsley 

qa_req 0

### DIFF
--- a/frontend/components/common/InstantSearchProvider.tsx
+++ b/frontend/components/common/InstantSearchProvider.tsx
@@ -335,7 +335,9 @@ function getBaseOrigin(location: Location): string {
 }
 
 function getDevicePath(handle: string, routeState: RouteState): string {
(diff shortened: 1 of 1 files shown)


masonmcelvain (approved):
CR 👏🏻

danielbeardsley (approved):
CR :+1:

masonmcelvain:
Womp womp, the test store is empty on govinor so the react-commerce build is failing.

https://split-fix.govinor.com/admin/content-manager/collectionType/api::store.store/5

image

This code is not the cause of the failure, so let's merge as is

PR #2089, opened 2023-11-08
Item Types: add new collection type

danielbeardsley:
Meant to store a picture associated with item types.

Connect iFixit/ifixit#49086

### DIFF
--- a/backend/src/api/item-type/content-types/item-type/schema.json
+++ b/backend/src/api/item-type/content-types/item-type/schema.json
@@ -1,0 +1,27 @@
+{
+   "kind": "collectionType",
+   "collectionName": "item_types",
+   "info": {
+      "singularName": "item-type",
+      "pluralName": "item-types",
+      "displayName": "Item Type"
+   },
+   "options": {
+      "draftAndPublish": true
+   },
+   "pluginOptions": {},
+   "attributes": {
+      "akeneo_code": {
+         "type": "string",
+         "required": true,
+         "unique": true,
+         "regex": "[a-z][a-z0-9]*"
+      },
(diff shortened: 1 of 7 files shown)


backend/src/api/item-type/content-types/item-type/schema.json
```diff
@@ -0,0 +1,27 @@
...
+   "options": {
+      "draftAndPublish": true
+   },
+   "pluginOptions": {},
+   "attributes": {
+      "akeneo_code": {
+         "type": "string",
+         "required": true,
+         "unique": true,
+         "regex": "[a-z][a-z0-9]*"
```
sterlinghirsh, line 18: Not sure if this is getting used, but we need to support underscores.

sterlinghirsh:
CR :zap: dev_block :zap: on making sure underscores will be allowed contrary to the regex in the code.

backend/src/api/item-type/content-types/item-type/schema.json
danielbeardsley: Fixed.

danielbeardsley:
un_dev_block :+1:

masonmcelvain (changes requested):
We need to add `api::item-type.item-type` to `backend/src/bootstrap/permissions/config.json`, otherwise we'll Next.js will get a 401 when trying to fetch these resources

danielbeardsley:
Fixed! Thanks for the tip!

masonmcelvain (approved):
CR 👏🏻 Nice!

danielbeardsley:
QA :+1: 

Strapi on govinor has the new type:
image

PR #2091, opened 2023-11-08
Piwik: Update Ecommerce Events to V2

danielcliu-ifixit:
## Overview

We want to move away from piwik's ecommcer events and use their [ecommerce v2 events](https://developers.piwik.pro/en/latest/data_collection/web/javascript_tracking_client/api.html#e-commerce-v2), which are cleaner, easier, and more analogous to ga4 events. Also cleans up our analytics tracking by wrapping some of our analytics events in a joint function to call through to piwik and ga4. This in turns consolidates call spots for our analytics, making it simpler throughout the codebase.

Also makes sure the cart total price is accurate, as we weren't updating it previously.

## QA

This is a bit of a tedious one.
We changed some call sites for GA4 events, so we want to make sure view item, add to cart, and remove from cart are working still.

Then make sure that piwik events Product view, add to cart, remove from cart, cart update, and special events like frequently bought together and cross sells are all firing correctly.

Connects: https://github.com/iFixit/shopify-v2/issues/1139


erinemay:
@danielcliu-ifixit merge conflicts 😢

jordycosta:
QA 🟢 

Product detail view, add to cart, and remove from cart events are all being tracked
[
Piwik Pro Tracker Debugger](https://ifixit.piwik.pro/analytics/#/748e3baf-bedc-4685-9f32-6b72ec349fd4/settings/debugger/)

firefox_rIHGd8zgZA

---

deploy_block 🟨 on a question

Are the Custom events for adding to cart expected? It seems to be a duplicate of the **Add to cart** event

danielcliu-ifixit:
@jordycosta yes those are expected. Not sure their necessary but theyre expected.

un_deploy_block :v:

danielcliu-ifixit:
QA :v:  carry-over

deltuh-vee:
Carry Over QA 🎬

packages/analytics/index.tsx
```diff
@@ -1,2 +1,84 @@
...
+   }
+}
+
+export function convertCartLineItemsToAnalyticsItem(
+   items: CartLineItem[]
+): AnalyticsItem[] {
+   return items.map((item) => {
+      const shopifyVariantId = item.shopifyVariantId.startsWith('gid://')
+         ? getVariantIdFromVariantURI(item.shopifyVariantId)
+         : getVariantIdFromEncodedVariantURI(item.shopifyVariantId);
```
masonmcelvain, line 75: Oof, why can't we be certain which form we're dealing with? Does the iFixit API send us different formats depending on which endpoint we use?

masonmcelvain (dismissed):
CR 👍🏻 This looks right, I'm glad the piwik interface is now a lot simpler! dev_block 👍🏻 on a concern about parsing product id's

frontend/helpers/path-helpers.ts
```diff
@@ -102,7 +101,7 @@ export function shopifyStoreAdminProductUrl({
    product,
    storeCode,
 }: ShopifyAdminProductUrlProps) {
-   const adminProductId = getProductIdFromGlobalId(product.id);
+   const adminProductId = getVariantIdFromVariantURI(product.id);
```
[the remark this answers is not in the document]
danielcliu-ifixit: oops. nope good catch, but they should both be in the same file anyway

danielcliu-ifixit:
un_dev_block :v:

packages/helpers/product-helpers.ts
```diff
@@ -65,6 +65,15 @@ export function getVariantIdFromVariantURI(variantURI: string): string {
    return variantURI.replace(/^gid:\/\/shopify\/ProductVariant\//, '');
 }
 
+export function getProductIdFromProductUrl(productURI: string) {
+   if (!productURI.startsWith('gid://')) {
+      throw new Error(
+         'Variant URI must be a global shopify product variant id uri'
```
masonmcelvain, line 81: This error message is wrong, we are dealing with product uri's here

packages/analytics/index.tsx
masonmcelvain: I still don't understand which service gives us an encoded variant id vs a decoded one

danielcliu-ifixit: for some reason we encode cross sell item's variant url. I'm not entirely sure why, but I'm going to look through if we can just remove that in a future pull

kthaler:
dev_block 🦖 - I am not seeing Remove from Cart events in Piwik and GA4. I am seeing Add to Cart in GA4 and Piwik and Product detail view in Piwik.

I do notice that when I remove an item from my cart or try to change the quantity of an item in my cart in react-commerce, I get an error `Error: Variant URI must be a global shopify product variant id uri`. I am not getting this error in main.

Screenshot 2023-11-20 at 10 11 51 AM

Also, I don't think the prices in GA4 are correct, looks like they're being multiplied by a million.

Screenshot 2023-11-20 at 9 57 18 AM

erinemay:
> Also, I don't think the prices in GA4 are correct, looks like they're being multiplied by a million.
> 
> Screenshot 2023-11-20 at 9 57 18 AM

It's just the debug viewer, so we're ok there. 
https://stackoverflow.com/questions/67224841/ga4-price-is-appended-with-multiple-zeros#comment118847345_67224841

packages/analytics/index.tsx
masonmcelvain: Ok thanks! That's all I wanted to know

danielcliu-ifixit:
un_dev_block :v:

packages/analytics/index.tsx
danielcliu-ifixit: ok this turned out to be a little bit inaccurate. basically we've been mix and matching encoded vs not encoded.

The basis for this seems to be our ifixit cart api returning the shopify variant id encoded, so all cart items have their variants encoded. basically everything else didn't, except for the mini cart cross sell items.

erinemay:
QA 🦖 
Piwik:
- Product list view
- Product detail view
- Cart adds from the product page
- Cart drawer:
  - Add to cart
  - Remove from cart
  - Add cross/up sell
  - Check out
- Price and item quantity all look good.

GA4:
- Add to cart
- Remove from cart
- Page view


deploy_block 🌵 Add to cart, remove from cart, and page view are pretty much all I'm seeing on the GA4 side. If that's all you touched, that's ok. I'm not seeing special events in GA4. 

When I add one of the cross/up sells from the drawer, or frequently bought together, GA4 shows it as an add to cart and nothing special. GA4 also doesn't seem to pick up when I click checkout.

danielcliu-ifixit:
un_deploy_block :v: all of those are expected. so this is fine


QA  :v: carry-over just from main merge

masonmcelvain (approved):
CR 👏🏻 Great work team!

It would be nice if the iFixit API didn't return encoded variant id's so we didn't have to deal with decoding/encoding at all. But, let's ship this and deal with that another day.

PR #2096, opened 2023-11-09
Add feature flag package

andyg0808:
This is an attempt at an ultra-simple feature flag package which could be used across the monorepo and this NextJS repo.

See the readme for rationale and design notes.

The goal is to add a minimal package which we can use to consistently have feature flags available.

qa_req 0
This code is unused.

### DIFF
--- a/packages/feature_flags/index.ts
+++ b/packages/feature_flags/index.ts
@@ -14,7 +14,7 @@ export function checkFlag(flagName: keyof typeof raw_flags) {
 
    if (typeof flagValue === 'object') {
       if (flagValue.userToggle) {
-         return localStorage.getItem(enableFlag) === 'true';
+         return window?.localStorage?.getItem(enableFlag) === 'true';
       }
    }
 
--- a/packages/feature_flags/index.ts
+++ b/packages/feature_flags/index.ts
@@ -1,0 +1,35 @@
+import * as raw_flags from './flags.json';
+import type { FlagList as FormalFlagList } from './flag_schema';
+
+export const flags: FormalFlagList = raw_flags;
+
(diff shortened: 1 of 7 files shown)


packages/feature_flags/index.ts
```diff
@@ -0,0 +1,35 @@
...
+
+   if (flagValue === true) {
+      return true;
+   }
+
+   const enableFlag = 'enable-' + flagName;
+
+   if (typeof flagValue === 'object') {
+      if (flagValue.userToggle) {
+         return localStorage.getItem(enableFlag) === 'true';
```
aburke07, line 17: Do we need to check if localstorage is available? Also, should this be `window.localstorage`?

```suggestion
         return window.localStorage.getItem(enableFlag) === 'true';
```

packages/feature_flags/index.ts
```diff
@@ -0,0 +1,35 @@
...
+   return false;
+}
+
+function getHashBoolean(param: string): boolean {
+   const hash = window?.location?.hash;
+   if (!hash) {
+      return false;
+   }
+   return hash.indexOf(param) !== -1;
+}
```
aburke07, line 35: Do we also want to support query params to enable flags?

aburke07 (approved):
CR :alien:

Dev_block :alien: on two comments

packages/feature_flags/index.ts
andyg0808: No, possibly not-yet. I don't have a strong opinion one way or the other, and I want this minimal. Using the hash guarantees that nothing different gets sent to the server.

aburke07:
Un_dev_block :alien:

Issue #1702, opened 2023-05-22
Use local strapi instance by default


ardelato:
## Description

This came up recently as there was an error attempting to run `playwright` tests locally. This is most likely due to `strapi.cominor.com` being out-of-date.

![image](https://github.com/iFixit/react-commerce/assets/22064420/577f3f52-4bce-4413-8d8e-ac8eba30984d)

We've configured `playwright` to use the `.env.test` config https://github.com/iFixit/react-commerce/pull/1661/files#diff-1ed7245e0c7c634928b52839600c3d493c9d29150d369e9ab20979e61e7f8f97R14 but the `NEXT_PUBLIC_STRAPI_ORIGIN` is always overwritten in CI https://github.com/iFixit/react-commerce/blob/main/.github/workflows/main.yml#L52. This is only an issue when running the tests locally as we don't overwrite this variable. 

That said we should always use the local instance of strapi, i.e. `localhost:1337`, to ensure we are using a consistent environment across dev and CI.

### To Do:

- [ ] Change `.env*` and configuration files to use the local strapi instance

**Reference:**
- https://ifixit.slack.com/archives/C01GMQXJC95/p1684792927308459?thread_ts=1684791302.193119&cid=C01GMQXJC95

mlahargou:
How important is this? Will this be relevant once we have react-commerce in the monorepo?

ardelato:
>Will this be relevant once we have react-commerce in the monorepo?

I am not sure what the plan is for how we intend to use Strapi in the monorepo.

dhmacs:
@ardelato What's the next action here? Imo having tests run against local Strapi would be useful to debug tests locally 👍

PR #2097, opened 2023-11-10
Tests: Use Local NEXT_PUBLIC_STRAPI_ORIGIN

ardelato:
## Description

When we run tests locally we use `.env.development` which points to the local Strapi url `localhost:1337`. In CI, we use `.env.test` which points to `https://strapi.cominor.com` but we still override it to point to the local Strapi url. Instead of overriding it, let's just use the local Strapi url in `.env.test`.

Closes: https://github.com/iFixit/react-commerce/issues/1702


batbattur (approved):
CR 🚛

PR #2099, opened 2023-11-14
Troubleshooting: Split related-problems render on feature flag

andyg0808:
This adds a second component for the related-problems change and 
 ensures we're not accidentally exposing our new component to the world. It should be pretty hard to mess up, because it checks that the test attribute doesn't exist without the feature flag and does exist with the feature flag.

qa_req 0
There's a test for this!

### DIFF
--- a/frontend/tests/playwright/troubleshooting/loading.spec.ts
+++ b/frontend/tests/playwright/troubleshooting/loading.spec.ts
@@ -8,6 +8,29 @@ test.describe('Troubleshooting Page Content and SEO', () => {
       ).toBeVisible();
    });
 
+   test('Related Problems hidden by feature flag', async ({ page }) => {
+      await page.goto('/Troubleshooting/Dryer/Not+Spinning/478872');
+      // Select the element with the ID 'related-problems'
+      const element = await page.locator('#related-problems');
+
+      // Assert that the element's 'data-test' attribute is 'related-problems-v2'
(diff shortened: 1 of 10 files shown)


frontend/tests/playwright/troubleshooting/loading.spec.ts
```diff
@@ -8,6 +8,29 @@ test.describe('Troubleshooting Page Content and SEO', () => {
       ).toBeVisible();
    });
 
+   test('Related Problems hidden by feature flag', async ({ page }) => {
+      await page.goto('/Troubleshooting/Dryer/Not+Spinning/478872');
+      // Select the element with the ID 'related-problems'
+      const element = await page.locator('#related-problems');
+
+      // Assert that the element's 'data-test' attribute is 'related-problems-v2'
+      await expect(element).not.toHaveAttribute(
```
aburke07, line 17:
```suggestion
      // Assert that the element's 'data-test' attribute is not 'related-problems-v2'
      await expect(element).not.toHaveAttribute(
```

aburke07 (dismissed):
CR :alien:

Dev_block :alien: on updating a comment

andyg0808:
un_dev_block :+1: 
Thanks @aburke07, that makes good sense.

aburke07 (approved):
CR :alien:

PR #2100, opened 2023-11-14
Troubleshooting: Include extra problems from device & ancestors

andyg0808:
6 was determined by counting a screenful on mobile. I'm mostly just picking a reasonable value to start with.

This is behind a feature flag.
qa_req 0

### DIFF
--- a/frontend/templates/troubleshooting/index.tsx
+++ b/frontend/templates/troubleshooting/index.tsx
@@ -69,6 +69,7 @@ import {
    TOCContextProvider,
    useTOCBufferPxScrollOnClick,
 } from './tocContext';
+import { uniqBy } from 'lodash';
 
 const RelatedProblemsRecord = {
    title: 'Related Problems',
@@ -985,8 +986,12 @@ function RelatedProblemsV2({
    wikiData: TroubleshootingData;
    hasRelatedPages?: boolean;
 }) {
-   const { linkedProblems, deviceGuideUrl, countOfAssociatedProblems } =
-      wikiData;
+   const {
+      linkedProblems,
+      relatedProblems,
+      deviceGuideUrl,
+      countOfAssociatedProblems,
+   } = wikiData;
    const { displayTitle, imageUrl, description } = wikiData.category;
 
    const bufferPx = useBreakpointValue({ base: -40, lg: 0 });
@@ -994,6 +999,8 @@ function RelatedProblemsV2({
       RelatedProblemsRecord.uniqueId,
       bufferPx
    );
+   const problems = linkedProblems.concat(relatedProblems);
+   const uniqProblems = uniqBy(problems, 'title');
 
    return (
       <>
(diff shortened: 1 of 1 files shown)


frontend/templates/troubleshooting/index.tsx
```diff
@@ -985,15 +986,24 @@ function RelatedProblemsV2({
...
    const bufferPx = useBreakpointValue({ base: -40, lg: 0 });
    const { ref } = LinkToTOC<HTMLHeadingElement>(
       RelatedProblemsRecord.uniqueId,
       bufferPx
    );
+   const problems = linkedProblems.concat(relatedProblems);
+   // We don't want to omit any linked problems, but we also don't want to add
+   // an unlimited number of additional problems from the device.
+   const maxProblems = Math.max(6, linkedProblems.length);
+   const uniqProblems = uniqBy(problems, 'title').slice(0, maxProblems);
```
jarstelfox, line 1006: I think js always turns the full array when you slice more than the index:

```
[1,2,3].slice(0,6); // [1,2,3]
```

andyg0808: That's exactly what I want.

jarstelfox: I think you can remove the max problems for a hardcoded numbe
```suggestion
   const maxProblems = 6; // make a const at the top of the file
   const uniqProblems = uniqBy(problems, 'title').slice(0, maxProblems);
```

andyg0808: Oh, I think you're missing a detail here: the `linkedProblems` array isn't the same as `problems`. The goal is to always include all the `linkedProblems`, plus enough of the `relatedProblems` to bring the total up to 6. But we don't want to _remove_ any of the `linkedProblems` if there are more than 6. So we choose the max of those two numbers.

jarstelfox: Thanks, now I see the point. I missed that

danielcliu-ifixit:
CR :v:

frontend/templates/troubleshooting/index.tsx
sctice-ifixit: I think this misunderstanding indicates that there might be room for the code to be more self-explanatory. This is slightly odd behavior (we don't just want to show a maximum number of problems—we only want to limit the number of "padded" problems), so I think it might be worth making the code more explicit.

Here's my attempt:

```js
const minDesiredProblems = 6;
const maxPaddedProblems = 6;
let problems = linkedProblems;
if (problems.length < minDesiredProblems) {
   const paddedProblems = uniqBy(linkedProblems.concat(relatedProblems), 'title');
   problems = paddedProblems.slice(0, maxPaddedProblems);
}
```

It's just a suggestion, but I think we have some evidence that the current implementation is still a bit confusing despite the comment, so I think another pass here is justified.

And just for posterity: I'm assuming that the `uniqBy()` is just to handle duplicates between the directly-linked problems and the related problems, not to handle duplicates within those independent lists (otherwise there's a bug with the current implementation because we take the count of `linkedProblems` before `uniqBy()`). Also, it's an important property of `uniqBy()` here that it maintains the original order of the list.

danielcliu-ifixit (approved):
CR :v:

PR #2101, opened 2023-11-14
Sentry: Receive details on the server

jarstelfox:
qa_req 0 I added tests!

Closes: https://github.com/iFixit/ifixit/issues/50671

### DIFF
--- a/frontend/tests/jest/tests/sentry.test.tsx
+++ b/frontend/tests/jest/tests/sentry.test.tsx
@@ -1,13 +1,12 @@
 import {
    SentryDetails,
    SentryError,
-   injectSentryErrorHandler,
+   SentryErrorIntegration,
 } from '@ifixit/sentry';
 import * as Sentry from '@sentry/nextjs';
 
-describe('Sentry beforeSend', () => {
-   it('should add extra information to the event', () => {
-      const mockBeforeSend = initSentry();
+describe('Sentry SentryErrorIntegration', () => {
+   it('should have extra information to the event', async () => {
       const sentryDetails: SentryDetails = {
          contexts: {
             contextName: {
@@ -22,8 +21,17 @@ describe('Sentry beforeSend', () => {
             tag2: 'Goodbye',
          },
       };
+
+      const mockBeforeSend = initSentry();
+
+      // Trigger an exception to invoke the beforeSend callback
       Sentry.captureException(new SentryError('message', sentryDetails));
 
+      // Flush the event to Sentry now
+      const didFlush = await Sentry.flush();
+      expect(didFlush).toBe(true);
+
+      // Assert
       expect(mockBeforeSend).toHaveBeenCalledTimes(1);
 
       const event = mockBeforeSend.mock.calls[0][0];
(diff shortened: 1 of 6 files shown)


packages/sentry/index.tsx
```diff
@@ -2,6 +2,8 @@ import * as Sentry from '@sentry/nextjs';
 import { urlFromContext } from '@ifixit/helpers/nextjs';
 import { GetServerSidePropsContext } from 'next';
 import { Scope } from '@sentry/nextjs';
+import { SentryErrorIntegration as Integration } from './SentryErrorIntegration';
```
danielcliu-ifixit, line 5: why do we import and export this as `Integration`?

frontend/tests/jest/tests/sentry.test.tsx
```diff
@@ -33,20 +41,18 @@ describe('Sentry beforeSend', () => {
...
-const SENTRY_DSN = process.env.NEXT_PUBLIC_SENTRY_DSN;
-function initSentry() {
+function initSentry(debug: boolean = false) {
    const mockBeforeSend = jest.fn();
 
    Sentry.init({
-      sampleRate: 1,
-      dsn: SENTRY_DSN,
+      debug,
+      dsn: 'https://00000000000000000000000000000000@o000000.ingest.sentry.io/0000000',
```
danielcliu-ifixit, line 49: should this be an env variable?

danielcliu-ifixit:
CR :v: 

Light dev_block :v: on questions just because qa_req 0

frontend/tests/jest/tests/sentry.test.tsx
jarstelfox: I thought so too, but locally I did not have it set. Since this was the case undefined was pass in. Sentry (rightfully) just doesn't setup if it has nowhere to ping. FYI this value is from their own tests

packages/sentry/index.tsx
jarstelfox: I probably should have used export *

Then main reason was that I am skeptical that devs know what is exported. Both eslint and TS do know but at times its unapparent.

frontend/tests/jest/tests/sentry.test.tsx
danielcliu-ifixit: do jest tests not have access to env variables?

jarstelfox: They do, but the local dev default does not specify sentry being needed. This bypasses the need for checking the env in favor of hardcoding a value sentry will happily use and process errors under test regardless of a users env

masonmcelvain: We don't specify NEXT_PUBLIC_SENTRY_DSN in `.env.test` because we don't want tests reporting to Sentry. We could achieve this a different way, but that was the intent

packages/sentry/SentryErrorIntegration.tsx
```diff
@@ -0,0 +1,38 @@
...
+         const newScope = Sentry.Scope.clone(currentScope);
+
+         newScope.update(exception.sentryDetails);
+         newScope.applyToEvent(event, hint);
+      }
+
+      return event;
+   }
+
+   isDefaultInstance = true;
```
masonmcelvain, line 37: I don't know what this property does, based on the [docs](https://docs.sentry.io/platforms/javascript/configuration/integrations/custom/?original_referrer=https%3A%2F%2Fwww.google.com%2F) and the [interface](https://github.com/getsentry/sentry-javascript/blob/master/packages/types/src/integration.ts)

masonmcelvain (approved):
CR 👍🏻 Good job getting to the bottom of this

packages/sentry/SentryErrorIntegration.tsx
jarstelfox: It's used to ensure we don't overwrite it with another version of the extension: https://github.com/getsentry/sentry-javascript/blob/ad4f2d1be6fb7ae4ca7f3e1da09bb8e0de2fde84/packages/core/src/integration.ts#L34-L38

It's a small safety net. I figure it couldn't hurt to set it.

PR #2105, opened 2023-11-16
Refactoring bit table layout

federicobadini:
This PR introduce an even more compact design proposal for the bit table section.
In this reviewed design we can merge bits from various screwdriversize in the same row, but still preserving readability.
Bit types without size variations are pushed to the end of the list to keep the more dense information in the top left corner.

### QA

1. Visit Vercel preview
2. Verify that the new implementation is aligned with the updated mockup visible [here](https://www.figma.com/file/BolmFORO6ZfHXePOSyEGSS/iFixit---Page-sections?node-id=1766%3A10707&mode=dev)


deltuh-vee:
QA 🎬 
Bit table matches the updated mockup.

frontend/components/sections/BitTableSection.tsx
```diff
@@ -111,55 +100,140 @@ function Bit({ type, driverSizeColor, sizes }: GroupedBit) {
...
+               {Object.keys(sizes).map((driverSize) => {
+                  const sizesForDriver = sizes[driverSize];
+                  return (
+                     <Flex
+                        key={driverSize}
+                        flexWrap="wrap"
+                        gap={1}
+                        justifyContent="flex-end"
+                     >
+                        {sizesForDriver.map((size) => (
```
andyg0808, line 150: If `size` was null for all bits with this `driverSize`, we could have `sizesForDriver === []`. Do we still want to be rendering the parent `Flex` in that case? If not, we might be able to simplify the grouping logic, since we don't need to deal with quite as many `null` checks.

frontend/components/sections/BitTableSection.tsx
```diff
@@ -111,55 +100,140 @@ function Bit({ type, driverSizeColor, sizes }: GroupedBit) {
...
+         groupItem.sizes[type.driverSize].push(size);
+         groupItem.totalBitCount++;
+      }
+   });
+
+   return {
+      groupedBits: Object.values(grouped),
+      driverSizeColors,
+   };
+};
```
andyg0808, line 239: I took a crack at simplifying this logic; here's my try: https://github.com/iFixit/react-commerce/compare/bit-table-grouping-rewrite

frontend/components/sections/BitTableSection.tsx
```diff
@@ -85,18 +64,28 @@ export function BitTableSection({
...
+   groupedBit,
+   driverSizeColors,
+}: {
+   groupedBit: GroupedBit;
+   driverSizeColors: Record<string, string>;
+}) {
+   const { type, sizes, totalBitCount } = groupedBit;
    return (
       <Flex
+         display="inline-flex"
```
andyg0808, line 77: CSS Grid is actually better-supported in terms of absolute numbers than `inline-flex` ([97.79%](https://caniuse.com/css-grid) rather than [97.38%](https://caniuse.com/mdn-css_properties_display_inline-flex)); would this layout be easier to implement with CSS Grid? (This may be a stupid question; I've not loaded the page to check what the layout actually looks like.)

frontend/components/sections/BitTableSection.tsx
```diff
@@ -111,55 +100,140 @@ function Bit({ type, driverSizeColor, sizes }: GroupedBit) {
...
-                        {size}
+                        {driverSize}
                      </Text>
                   </Flex>
                ))}
             </Flex>
          </Flex>
+         <Flex flexGrow={1} alignItems="center" gap="3" p="3">
+            <Flex gap="1">
+               {totalBitCount > 0 && (
```
andyg0808, line 131: Would
```suggestion
               {totalBitCount && (
```
be reasonable?

andyg0808 (approved):
CR :+1: 
deploy_block :+1: on possibly simplifying the grouping code and on considering CSS Grid.

frontend/components/sections/BitTableSection.tsx
federicobadini: I tend to prefer explicit numeric comparison if the variable is numeric.
My reasoning is that, if the data type is changed in the upper layers from a numeric 0 to a `"0"` string, the outcome of that check would be different.
The `>` operator would raise an error while the simple variable check would not.

frontend/components/sections/BitTableSection.tsx
federicobadini: It is not necessary to have this element inline. This PR was discussed and refactored several times before shipping it, so it was probably a typo from one of the previous iterations

frontend/components/sections/BitTableSection.tsx
andyg0808: I don't have a strong opinion. I think I'd prefer it changed just slightly, and we can type `totalBitCount` as numeric if you're concerned.

frontend/components/sections/BitTableSection.tsx
federicobadini: I'm not a fan of lodash but your version is good to read so I'm modifying the implementation to match it.
I'll just tweak a little bit the function in order to handle edge cases like those of two driver sizes without bit size correctly:
Screenshot 2023-12-05 at 16 20 25

frontend/components/sections/BitTableSection.tsx
federicobadini: We can omit the div

deltuh-vee:
Carry Over QA 🎬

federicobadini:
Have you inserted the slug? Even though it is not needed it seems that strapi fails if it is missing

deltuh-vee:
un_deploy_block 👍 
I did but it seems to be working now

andyg0808 (approved):
CR :+1:

frontend/components/sections/BitTableSection.tsx
```diff
@@ -111,55 +101,121 @@ function Bit({ type, driverSizeColor, sizes }: GroupedBit) {
...
       </Flex>
    );
 }
+
+const groupScrewdriverBits = (bits: ScrewdriverBit[]): GroupedBit[] => {
+   const groups = Object.values(groupBy(bits, (bit) => bit.type.name));
+   return groups.map((group) => {
+      const sizes: Record<string, string[]> = mapValues(
+         groupBy(group, (bit) => bit.type.driverSize),
+         (bits) => bits.map((bit) => bit.size).filter(isNotNull)
```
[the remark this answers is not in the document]
federicobadini: To have duplicated bit one have to create a duplicated `ScrewdriverBit` in Strapi and also insert it twice in the `BitTableSection`.
It is not impossible but in case it happen I'd prefer the error to pop up so that the duplicate data can be identified and the strapi data source can be cleaned up.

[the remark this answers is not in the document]
andyg0808: Works for me for now. I subscribe to the "make errors unrepresentable" course; I'd suggest we consider whether there's a better data model which avoids this possibility. If not, I might not worry too hard about it, because I'm not sure that we want invisible restrictions on what the folks editing Strapi can do.

PR #2111, opened 2023-11-17
Add product list hreflangs

masonmcelvain:
Satisfies https://github.com/iFixit/ifixit/issues/50649 for the headless store.

As mentioned here https://github.com/iFixit/ifixit/issues/50649#issuecomment-1816735791, if the customer has an item type selected on a product list page that should have hreflangs, we exclude the hreflangs (unless we specifically have hreflangs set for that device/item type combination).

## CR

Please carefully review the commit that prunes redundant mappings (0d55b0f9546fd7c06571122647cdc219ca0aa14c), as I did this by eye. My intent was to reduce the bundle size increase and improve readability.

## Bundle size increase

The product list pages have a ~1 kb bundle size increase due to the size of the hreflang mappings.

## QA

Compare the [spec spreadsheet](https://docs.google.com/spreadsheets/d/1bFai7oXL1HJtgFm69taKA_VjKN8WtXyrJfMyB6UQ7xY/edit#gid=1714817208) with the hreflang tags present in the `<head>` on a variety of product list pages.

Make sure to cover the categories of `/Tools/*`, `/Parts/*`, `/Parts`, and the only item type hreflang (`/iMac/Kits`). Also make sure that the hreflangs do not show up when they aren't supposed to.


masonmcelvain:
dev_block 👍🏻 on de-duping the `/collections/` text

masonmcelvain:
un_dev_block 👍🏻

masonmcelvain:
dev_block 👍🏻 once more on simplifying the us mappings

masonmcelvain:
un_dev_block 👍🏻

kthaler:
URL hreflangs look good for `/Tools/*`, `/Parts/*`, `/Parts`, and `/iMac/Kits`. Could not find any weird cases of hreflangs not being where they're supposed to be.

QA 💎

frontend/seo/product-list/hreflangs.ts
```diff
@@ -0,0 +1,260 @@
...
+   'apple-watch': 'apple-watch',
+   'iphone-5c': 'iphone-5c',
+   macbook: 'macbook',
+   'apple-macbook-parts': 'mac-parts',
+   parts: 'parts',
+};
+
+const collections = Object.keys(
+   usHreflangsPaths
+) as unknown as Array<Collection>;
```
danielbeardsley, line 98: I'm surprised this was needed.

frontend/seo/product-list/hreflangs.ts
```diff
@@ -0,0 +1,260 @@
...
+   'ipad-air-2': 'ipad-air-2',
+   'ipad-6': 'ipad-6',
+   'iphone-6-plus': 'iphone-6-plus',
+   'ipad-air-1': 'ipad-air-1',
+   'ipad-5': 'ipad-5',
+   'iphone-4s': 'iphone-4s',
+   'ipad-2': 'ipad-2',
+   'iphone-5': 'iphone-5',
+   'ipad-mini-retina': 'ipad-mini-retina',
+   'imac-upgrade-kits': 'imac-upgrade-kits',
```
danielbeardsley: Are any of these different? Maybe it'd be clearer to only include the entries where the key is different than the value?

danielbeardsley: Ha! You did this in a later commit!

danielbeardsley:
> The product list pages have a ~1 kb bundle size increase due to the size of the hreflang mappings.

This strikes me as something that I would expect us to do server-side on process boot, is there a reason we've opted to do it client-side? I think the 1KB of js isn't terrible; I think it's worse that we're instantiating and parsing `65 * 7 = 455` urls as the module is included when we know we only need 7.

sterlinghirsh (approved):
CR :zap: Looks ok to me.

sterlinghirsh:
@danielbeardsley I agree but I'd like to get this out today so I'm going to let it slide. I talked to Max and we are already going to have some followup changes on Monday, but this is better than nothing.

frontend/seo/product-list/hreflangs.ts
masonmcelvain: Apparently this is by design in TS https://github.com/Microsoft/TypeScript/issues/12870, `Object.keys()` returns `string[]` not `(keyof Obj)[]`

PR #2112, opened 2023-11-17
Index Variants Instead Of Device: Implement

jarstelfox:
Closes: https://github.com/iFixit/ifixit/issues/50437

### DIFF
--- a/frontend/helpers/path-helpers.ts
+++ b/frontend/helpers/path-helpers.ts
@@ -18,7 +18,8 @@ type ProductListPathAttributes = Pick<
 >;
 
 export function productListPath(
-   productList: ProductListPathAttributes
+   productList: ProductListPathAttributes,
+   variant?: string | undefined
 ): string {
    switch (productList.type) {
       case ProductListType.AllParts: {
@@ -30,7 +31,7 @@ export function productListPath(
             'device product list does not have device title'
          );
          const deviceHandle = encodeURIComponent(
-            stylizeDeviceTitle(productList.deviceTitle)
+            stylizeDeviceTitle(productList.deviceTitle, variant)
          );
          return `/Parts/${deviceHandle}`;
       }
--- a/frontend/templates/product-list/MetaTags.tsx
+++ b/frontend/templates/product-list/MetaTags.tsx
@@ -108,9 +108,11 @@ function useCanonicalUrl(productList: ProductList): string {
    const itemTypeHandle = itemType
       ? `/${encodeURIComponent(stylizeDeviceItemType(itemType))}`
       : '';
+   const variant = useVariant();
 
    return `${appContext.ifixitOrigin}${productListPath(
-      productList
+      productList,
+      productList.indexVariantsInsteadOfDevice ? variant : undefined
    )}${itemTypeHandle}${page > 1 ? `?${PRODUCT_LIST_PAGE_PARAM}=${page}` : ''}`;
 }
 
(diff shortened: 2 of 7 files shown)


jarstelfox:
dev_block 🌵 on implementing the spec:

* The variant pages should be their own canonical.
* The main device page should be noindex and have no canonical

deltuh-vee:
QA 🎬 
When `indexVariantsInsteadOfDevice` is set to true:
* The variant page is indexed if the regular page would be indexed prior to this change
* Variant pages have their own canonical 
* The main page for a device has `noindex` set and has no canonical link.

frontend/helpers/path-helpers.ts
```diff
@@ -22,6 +22,7 @@ type ProductListPathAttributes = Pick<
 
 export function productListPath(
    productList: ProductListPathAttributes,
+   itemType?: string,
    variant?: string | undefined
 ): string {
```
masonmcelvain, line 27: Since we now have two arguments that are `string | undefined`, it might make sense to have this function accept an object instead of arguments to prevent mix ups

i.e.
```ts
function productListPath({ productList, itemType, variant}: {
   productList: ProductListPathAttributes,
   itemType?: string,
   variant?: string | undefined
}) { ... }
```

frontend/templates/product-list/MetaTags.tsx
```diff
@@ -104,9 +121,12 @@ function useCanonicalUrl(
...
       ? `/${encodeURIComponent(stylizeDeviceItemType(itemType))}`
       : '';
+   const variant = useVariant();
 
    return `${appContext.ifixitOrigin}${productListPath(
-      productList
+      productList,
+      undefined,
+      productList.indexVariantsInsteadOfDevice ? variant : undefined
    )}${itemTypeHandle}${page > 1 ? `?${PRODUCT_LIST_PAGE_PARAM}=${page}` : ''}`;
```
masonmcelvain, line 130: I don't see a reason that we don't pass the item type to `productListPath`, and instead compute it on our own. I'm just observing, not saying you need to change it in this pull.

masonmcelvain (approved):
CR 👏🏻

frontend/templates/product-list/MetaTags.tsx
```diff
@@ -89,6 +89,14 @@ function useFilters() {
    return filtersAndValues;
 }
 
+function useVariant(): string | undefined {
+   const variantRefinements = useFilters()['worksin'];
+   const variant = variantRefinements
+      ? variantRefinements[0]?.toString()
+      : undefined;
```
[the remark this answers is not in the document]
jarstelfox: I tried this, but I don't think `?.[0]?` is valid syntax

frontend/helpers/path-helpers.ts
jarstelfox: Yea, I was wondering about that. I built this branch before that code was merged. I did the quick fix on merge, but I thought this might make sense as well

Opinions on doing it now vs next time we touch it?

frontend/templates/product-list/MetaTags.tsx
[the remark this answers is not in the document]
masonmcelvain: This runs fine in my console `undefined?.[0]?.toString()`, but it's super minor, just thought it was a little more readable

frontend/helpers/path-helpers.ts
masonmcelvain: Next time

masonmcelvain:
I vote we ship, I can't help but write my comments but they are totally non-blocking

PR #2116, opened 2023-11-20
Product List: Add default H1 override

masonmcelvain:
## Problem

Original report in Slack: https://ifixit.slack.com/archives/C02KMFEHX62/p1700521827373909

Wrong H1:

image
https://www.ifixit.com/Parts/Samsung_Galaxy_S21/Batteries

Correct H1:

image
https://www.ifixit.com/Parts/Google_Pixel_4a/Batteries

## Fix

I noticed that the Samsung Strapi product list had an H1 set, while the Google Pixel Strapi page did not. When there are no item type overrides, the code picks a default title but not a default H1. This change picks a default H1.

### DIFF
--- a/frontend/templates/product-list/hooks/useItemTypeProductList.ts
+++ b/frontend/templates/product-list/hooks/useItemTypeProductList.ts
@@ -21,6 +21,7 @@ export function useItemTypeProductList(productList: ProductList) {
    if (!hasItemTypeOverrides) {
       return {
          ...productList,
+         h1: getDefaultH1Override(productList, selectedItemType),
          title: getDefaultTitleOverride(productList, selectedItemType),
       };
    }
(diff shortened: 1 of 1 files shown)


jarstelfox (approved):
CR 🌵 FYI we are going to do something similar for device variants, so it's possible worth us making this more robust at that time

masonmcelvain:
@jarstelfox I'm not 100% sure what you're suggesting, could you elaborate

jarstelfox:
Totally! I am talking about a bullet point [from this issue](https://github.com/iFixit/ifixit/issues/48261).

> When a variant is specified, include the variant as a suffix when using an auto-generated or templated h1, title, etc.

We will have something similar where an Algolia filter affects the generated H1

masonmcelvain:
Ah I see. I agree we could definitely make this more robust/reusable when we implement that feature 👍🏻

kthaler:
QA 💎 - Item type collection pages now display the item type in the H1 rather than defaulting to "Parts".

PR #2117, opened 2023-11-21
Refactor product list path helper

masonmcelvain:
Noticed in my CR here https://github.com/iFixit/react-commerce/pull/2112#discussion_r1399848133

## QA

Make sure the product list links throughout the Next.js app still work normally.

In particular, make sure that item type canonical urls in the `<head>` are still correct, and check the new variant syntax too (`DeviceTitle:Variant`)

### DIFF
--- a/frontend/helpers/path-helpers.ts
+++ b/frontend/helpers/path-helpers.ts
@@ -19,12 +19,17 @@ type ProductListPathAttributes = Pick<
    ProductList,
    'type' | 'handle' | 'deviceTitle'
 >;
+type ProductListPathProps = {
+   productList: ProductListPathAttributes;
+   itemType?: string;
+   variant?: string | undefined;
+};
 
-export function productListPath(
-   productList: ProductListPathAttributes,
-   itemType?: string,
-   variant?: string | undefined
-): string {
+export function productListPath({
+   productList,
+   itemType,
+   variant,
+}: ProductListPathProps): string {
    switch (productList.type) {
       case ProductListType.AllParts: {
          return '/Parts';
--- a/frontend/templates/product-list/MetaTags.tsx
+++ b/frontend/templates/product-list/MetaTags.tsx
@@ -123,11 +123,13 @@ function useCanonicalUrl(
       : '';
    const variant = useVariant();
 
-   return `${appContext.ifixitOrigin}${productListPath(
+   return `${appContext.ifixitOrigin}${productListPath({
       productList,
-      undefined,
-      productList.indexVariantsInsteadOfDevice ? variant : undefined
-   )}${itemTypeHandle}${page > 1 ? `?${PRODUCT_LIST_PAGE_PARAM}=${page}` : ''}`;
+      itemType: undefined,
+      variant: productList.indexVariantsInsteadOfDevice ? variant : undefined,
+   })}${itemTypeHandle}${
+      page > 1 ? `?${PRODUCT_LIST_PAGE_PARAM}=${page}` : ''
+   }`;
 }
 
 function useShouldNoIndex(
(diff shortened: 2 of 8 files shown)


jarstelfox (approved):
CR 🌵

kthaler:
QA 💠 - It seems like the product list links are still working normally and the item type canonical urls in the `<head>` are correct.

deploy_block 🦖 - 
For URLs that use the variant syntax, should the canonicals point to the specific variant URL, or should they refer to the main device URL instead? Currently, the canonicals seem to point to the main device URL. EX: For https://react-commerce-git-refactor-product-list-path-ifixit.vercel.app/Parts/PlayStation_4:CUH-11XXA, this is the canonical:
Screenshot 2023-11-20 at 4 44 43 PM

masonmcelvain:
> For URLs that use the variant syntax, should the canonicals point to the specific variant URL, or should they refer to the main device URL instead?

According to this issue, they should canonical to the device handle by default, and the variant handles if `indexVariantsInsteadOfDevice` is set in Strapi.

> Currently, the canonicals seem to point to the main device URL

That sounds like the right behavior, since this setting is new and most/all product lists in Strapi have it set to false by default.

frontend/templates/product-list/MetaTags.tsx
```diff
@@ -115,19 +114,14 @@ function useCanonicalUrl(
...
 
-   return `${appContext.ifixitOrigin}${productListPath(
+   return `${appContext.ifixitOrigin}${productListPath({
       productList,
-      undefined,
-      productList.indexVariantsInsteadOfDevice ? variant : undefined
-   )}${itemTypeHandle}${page > 1 ? `?${PRODUCT_LIST_PAGE_PARAM}=${page}` : ''}`;
+      itemType: itemType ?? undefined,
+      variant: productList.indexVariantsInsteadOfDevice ? variant : undefined,
+   })}${page > 1 ? `?${PRODUCT_LIST_PAGE_PARAM}=${page}` : ''}`;
```
dhmacs, line 124: I suspect that the increased complexity of `productListPath` function can (at least in part) be pushed into the product list model

> `itemType: itemType ?? undefined,`

`itemType` could be integrated into the product list model as an optional/nullable attribute. We already have `useItemTypeProductList` taking care of integrating item type override directly into the model (note that this happens to be a hook just because using Algolia hooks forces us to model it this way, otherwise that logic could be moved in the model).

> `variant: productList.indexVariantsInsteadOfDevice ? variant : undefined`

It looks like this ternary could be implemented directly inside `productListPath`.
Looking at the implementation though, it looks like that this kind of path that includes the variant is going to be used only in meta tags (btw, what's the reason for this behaviour?). If that's the case, we can create a specific function to compute the path and colocate it within this file, so that this logic is not made general purpose if it belongs only to a specific use case

masonmcelvain:
un_deploy_block 👍🏻

frontend/helpers/path-helpers.ts
```diff
@@ -19,12 +19,17 @@ type ProductListPathAttributes = Pick<
...
-export function productListPath(
-   productList: ProductListPathAttributes,
-   itemType?: string,
-   variant?: string | undefined
-): string {
+type ProductListPathProps = {
+   productList: ProductListPathAttributes;
+   itemType?: string;
+   variant?: string | undefined;
+};
```
dhmacs, line 26: Adding to my previous comment, it seems to me that this helper is becoming increasingly complex, and some of that complexity could be simplified if pushed into the model.
Ideally one should be able to call `productListPath(productList)` and get back the product list path. 
`itemType` is just a specific type of product list, not an additional attribute I need to think about when I'm trying to get the product list path.
Maybe we can conceive a similar simplification for `variant`? When should I pass this param to `productListPath`? Can we make it so when using `productListPath` we don't have to think about this?

dhmacs (commented):
Added a few non-blocking suggestions

masonmcelvain:
I just missed your comments @dhmacs! We will address them here: https://github.com/iFixit/react-commerce/issues/2119

PR #2118, opened 2023-11-21
Sentry/index Exports All of SentryErrorIntegration

danielcliu-ifixit:
## Overview

While deving I kept seeing the following in my console

```
instrument.js:123 ./sentry.client.config.ts
Attempted import error: 'SentryErrorIntegration' is not exported from '@ifixit/sentry' (imported as 'SentryErrorIntegration').
```

So this fixes that.

## QA

The above import error shouldn't show up anymore.

@jarstelfox Is this ok? Or am i missing something?
qa_req 0


masonmcelvain (approved):
CR 👍🏻

jarstelfox (approved):
CR 🌵

PR #2122, opened 2023-11-21
Parts Collection Feature Flag: Don't mutate the server response

aburke07:
1. This code is being SSR'ed without the feature flag. That mutates the solutions array and sets the parts collection to an empty array.
2. Then when the client passes the feature flag, it sets the partsCollection array to the empty array of solutions.partsCollection that we just mutated.
3. Setting this to a const instead of mutating the original array fixes the issue.

Qa_req 0 - this is still behind the feature flag.

Connects https://github.com/iFixit/ifixit/issues/50601


andyg0808 (approved):
CR :+1:

PR #2128, opened 2023-11-27
Troubleshooting: Part collections card: styling fixes

aburke07:
## Overview
Fixes image styling, centers the text on the card to match the Figma, and adds margin to keep buy button from touching the text.

Qa_req 0 - still behind a feature flag.

Connects https://github.com/iFixit/ifixit/issues/50601

### DIFF
--- a/frontend/templates/troubleshooting/Resource.tsx
+++ b/frontend/templates/troubleshooting/Resource.tsx
@@ -250,17 +250,20 @@ function Resource({
                   outline="inherit"
                   outlineColor="inherit"
                   borderRadius="inherit"
-                  objectFit="cover"
+                  objectFit="contain"
+                  width="100%"
+                  height="100%"
                   alt={title}
                   src={imageUrl}
                   position="relative"
                   zIndex="1"
+                  bgColor="white"
                />
             </Box>
          )}
          <Flex flex="1" flexDirection={{ base: 'column', sm: 'row' }}>
             <Stack
-               justify="flex-start"
+               justify="center"
                align="flex-start"
                spacing={spacing}
                alignSelf="stretch"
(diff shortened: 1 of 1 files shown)


frontend/templates/troubleshooting/Resource.tsx
```diff
@@ -305,6 +310,7 @@ function Resource({
                   openInNewTab={false}
                   url={href}
                   buyButtonText={buyButtonText}
+                  margin="8px 0 0 0"
```
hcutrone, line 313: It looks like the `BuyButton` has a `buttonStyling` prop that takes Chakra props, so I think you can do something like:
```suggestion
                  buttonStyling={{ mt: "8px" }}
```

Then you don't need the extra `margin` prop.

frontend/templates/troubleshooting/Resource.tsx
```diff
@@ -314,6 +320,7 @@ function Resource({
                   openInNewTab={openInNewTab}
                   url={href}
                   buyButtonText={buyButtonText}
+                  margin="0 0 0 12px"
```
hcutrone, line 323: Same thing here, if I'm correct.
```suggestion
                  buttonStyling={{ ml: "12px" }}
```

hcutrone (approved):
CR 🌯 with a suggestion. You should check me before implementing, I could be wrong.

Sorry I didn't realize this was qa_req 0. These aren't blocking changes so I don't think it's necessary to open a new pull for this.

frontend/templates/troubleshooting/Resource.tsx
aburke07: Sorry I didn't get to this. I like the suggestion. I'll add it to a future pull if there are more styling changes with this component.

PR #2129, opened 2023-11-27
Ifixit TroubleShooting API: Encode Parameters

danielcliu-ifixit:
## Overview

Before we werent encoding, which was resulting in bad requests that returned the doc endpoint and broke things in react-commerce. Now we encode so we at least hit the expected endpoint and don't redirect.

## QA

You can see the error on `/Troubleshooting/Television/TV+Has+Sound+But+No+Picture/493422` vs `/Troubleshooting/Television/TV+Has+Sound+But+No+Picture/..%2f..%2fpackage.json`. It should no longer 500

Closes: #50760


jarstelfox (approved):
CR 🌵

deltuh-vee:
QA 🎬 
Bad requests no longer crash react-commerce.

PR #2131, opened 2023-11-28
Product images: use fallback images

danielbeardsley:
Products may now have fallback images (only one right now, but using an
array made integration easier). These are only to be used when a variant
has no other image to represent it. They are identified by (altText ===
'*fallback')

Here we extract them into a new property and make use of them wherever
we are using product.images.

We can't inject copies of these fallback images into product.images
otherwise the gallery would show one copy of the fallback for each
variant that uses it.

fallback images won't always be present and we've dealt with that here.
Our schema validation on the cache should protect us against this
change.

Note: We are filtering out fallback images from the normal
product.images because in that array, variantId === null implies it's
valid for all variants and that's not true for fallback images.

### QA
Hand-edit the images for a product with multiple variants to ensure there are no group-level images and only one variant has images (alt text = sku). Upload a group level image and set the alt text to `*fallback`. When viewing this branch, this image should be used for any variants that don't have images but should not be used for variants that *do* have images.

Connect https://github.com/iFixit/ifixit/issues/49086 and https://github.com/iFixit/ifixit/pull/50859

### DIFF
--- a/frontend/models/components/image.ts
+++ b/frontend/models/components/image.ts
@@ -12,6 +12,8 @@ export const ImageSchema = z.object({
    thumbnailUrl: z.string().nullable().optional(),
 });
 
+export const ImageAltFallback = '*fallback';
+
 export type Image = z.infer<typeof ImageSchema>;
 
 interface ImageOptions {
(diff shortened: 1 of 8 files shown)


aburke07 (approved):
CR :alien;

masonmcelvain (approved):
CR 👍🏻

danielbeardsley:
QA :+1: 

I've validated that images with `altText="*fallback"` operate as expected and are used only for variants that have no images.

PR #2135, opened 2023-11-29
`next.config.js`: Drop legacy product list redirects

masonmcelvain:
Closes https://github.com/iFixit/ifixit/issues/50811

Drops redirects for legacy device product list slugs. We're not confident that these urls are still seeing traffic, and these redirects are beginning to conflict with new item types which is a problem. There's some debate about whether we should continue supporting the legacy tool categories, so they were excluded from this change.


kthaler:
QA 💎 - Item Types in legacy-filter-slugs.json no longer redirect

sterlinghirsh (approved):
CR :zap:

PR #2138, opened 2023-11-30
Variants: Add to text on page

danielcliu-ifixit:
## Overview

We are rolling out variant product list pages and we want to have that variant show up on the page. This changes makes sure that happens.

It doesn't overwrite itemTypeOverrides. @sterlinghirsh @ifixitmaxb is this how we want it?

If there is a variant, it adds should replace any text on the page where the product display title is with product display title + variant. It should also check if there is a `model_number_pattern` associated with the device's wiki and if it does replace the pattern in the display title where appropriate with the variant.

## CR
This pull has kinda become a mess, commits wise, so it probably best to just go through the files changed.

## QA

Does the above behavior seem to be correct? 
Heres some decent behavior to check

http://localhost:3000/Parts/PlayStation_4:CUH-11XXA
http://localhost:3000/Parts/PlayStation_4?filter%5Bworksin%5D=PlayStation%204%20CUH-11XXA
http://localhost:3000/Parts/PlayStation_4?filter%5Bfacet_tags.Capacity%5D%5B0%5D=1%20TB
http://localhost:3000/Parts/PlayStation_4:CUH-11XXA?filter%5Bfacet_tags.Capacity%5D%5B0%5D=1%20TB
http://localhost:3000/Parts/PlayStation_4
http://localhost:3000/Parts/iPod_Mini?filter%5Bworksin%5D=iPod%20Mini%20A1051%202nd%20gen%204%20GB

Closes: [#50866](https://github.com/iFixit/ifixit/issues/50866)


frontend/templates/product-list/hooks/useVariantProductList.ts
```diff
@@ -0,0 +1,43 @@
...
+) {
+   if (!deviceHandle) return productList;
+   const { variant } = splitDeviceAndVariant(deviceHandle);
+   if (!variant) return productList;
+
+   const modelNumberPattern = productList.wikiInfo?.find(
+      (info) => info.name === 'model_number_pattern'
+   )?.value;
+   const modelNumberRegex = modelNumberPattern
+      ? new RegExp(modelNumberPattern?.replace(/[*⎵]/g, '.'), 'g')
```
sterlinghirsh, line 20: I think using `replaceAll(modelNumberPattern, variant)` would be simpler than constructing a global regex to pass to replace.

frontend/templates/product-list/hooks/useVariantProductList.ts
```diff
@@ -0,0 +1,43 @@
...
+      : null;
+   const replaceModelNumber =
+      modelNumberRegex && variant.match(modelNumberRegex);
```
sterlinghirsh, line 23: I don't think we need to match the variant against the pattern. We already know the variants belong to the device which is enough. We just need to do the replacement i.e. `deviceDisplayTitle.replaceAll(modelNumberPattern, variant)` and if the pattern doesn't exist in the deviceTitle then we do `deviceDisplayTitle variant` with a space. This should be used everywhere that the device title gets used (including itemType pages).

danielcliu-ifixit: So to be clear, whenever a device title contains a model number pattern in it **all** of its variants will match that pattern? 

What about if someone types in the variant in their url? It should show an inaccurate replace because its nonsense anyway?

> pattern doesn't exist in the deviceTitle then we do deviceDisplayTitle variant with a space

there is no point to the pattern then. lets just delete those

dhmacs:
Hey, I'm curious about `PlayStation_4:CUH-11XXA` path encoding using `:`. What's the reasoning behind it? What's preventing us from using a more conventional approach like this: `/Parts/PlayStation_4/variants/CUH-11XXA` ?

danielcliu-ifixit:
@dhmacs based on https://github.com/iFixit/ifixit/issues/48261#issuecomment-1599452144 i think its to avoid route collisions?

frontend/templates/product-list/hooks/useVariantProductList.ts
masonmcelvain: I agree that we don't need to use model number patterns at all in Next. We should use the variant from the url no matter what.

> What about if someone types in the variant in their url? It should show an inaccurate display title because the request is nonsense?

Yes, that's what we do for the item type suffix. The only way a user could end up in this state is if they url hack

https://www.ifixit.com/Parts/Phone/asdf

image

frontend/templates/product-list/ProductListView.tsx
```diff
@@ -21,24 +21,28 @@ import {
...
+   deviceHandle,
 }: ProductListViewProps) {
    // This temporary hack allows to correctly populate the itemType facet during SSR
    // see: https://github.com/algolia/instantsearch/issues/5571
    const _ = useMenu({ attribute: 'facet_tags.Item Type' });
    const filters = computeProductListAlgoliaFilterPreset(productList);
    const optionalFilters =
       computeProductListAlgoliaOptionalFilters(productList);
+   productList = useVariantProductList(productList, deviceHandle);
    const itemTypeProductList = useItemTypeProductList(productList);
```
masonmcelvain, line 46: I think we should have a single hook that augments the product list. As implemented, calling these hooks out of order would introduce a bug ("Device Item Type Variant" versus "Device Variant Item Type").

frontend/templates/product-list/index.tsx
```diff
@@ -2,14 +2,21 @@ import { InstantSearchProvider } from '@components/common/InstantSearchProvider'
...
 }) => {
+   const router = useRouter();
+   const { deviceHandle } = getRouteData(router.asPath);
    return (
       <InstantSearchProvider {...appProps.algolia!}>
-         <ProductListView productList={productList} />
+         <ProductListView
+            productList={productList}
+            deviceHandle={deviceHandle}
+         />
```
masonmcelvain, line 19: Instead of reaching into the router everywhere we need the device/variant path component, it would be nice to have a hook like `useDevicePartsVariant` or something. Similar to what we with the item type hook [`useDevicePartsItemType`](https://github.com/iFixit/react-commerce/blob/main/frontend/templates/product-list/hooks/useDevicePartsItemType.ts).

We could probably pull the variant out of `useCurrentRefinements()['woksin']` instead of the router too.

frontend/templates/product-list/ProductListView.tsx
masonmcelvain: It would also make the code easier to understand if all the logic that changes the `h1`, `title`, etc was side by side in a single place.

frontend/templates/product-list/hooks/useVariantProductList.ts
danielcliu-ifixit: I still personally feel like the regex is the cleanest solution that is resilient to data mistakes IF we are using model number pattern matching

frontend/templates/product-list/ProductListView.tsx
danielcliu-ifixit: I like the separation where one function that handles variant overrides and one that handles item type overrides, personally. But yeah i should make it more clear the order matters.

danielcliu-ifixit:
un_dev_block :v:

deltuh-vee:
dev_block ✊ 
The `model_number_pattern` here should be replaced with the specific variant. 
CleanShot 2023-12-14 at 10 44 33

Govinor product list link: https://variants--add-to-title-h1-etc.govinor.com/admin/content-manager/collectionType/api::product-list.product-list/3052?plugins[i18n][locale]=en
Cominor device link: https://me.cominor.com/Device/Maytag_Refrigerator_GT19Y%E2%8E%B5

danielcliu-ifixit:
un_dev_block :v:

frontend/templates/product-list/index.tsx
jarstelfox: [I created a `useVariant()`](https://github.com/iFixit/react-commerce/blob/main/frontend/templates/product-list/MetaTags.tsx#L103-L109)

danielcliu-ifixit:
un_dev_block :v:

frontend/components/common/InstantSearchProvider.tsx
```diff
@@ -149,12 +149,9 @@ export function InstantSearchProvider({
...
       }) {
-         const pathParts = location.pathname
-            .split('/')
-            .filter((part) => part !== '');
-         const firstPathSegment = pathParts.length >= 1 ? pathParts[0] : '';
-         const deviceHandle = pathParts.length >= 2 ? pathParts[1] : '';
-         const itemType = pathParts.length >= 3 ? pathParts[2] : '';
+         const { firstPathSegment, deviceHandle, itemType } = getRouteData(
+            location.pathname
+         );
```
jarstelfox, line 154: This is definitely for later:

This "common" file has custom parsing for certain types of pages.
We should stop doing it his way and instead all each page to pass in a function.

Simplified Current code:

```
<InstantSearchProvider {...props}>
	<DefaultLayout>
		{children} /* Used on every page */
	</DefaultLayout>
</InstantSearchProvider>
```

My claim:

```
	<DefaultLayout>
		{children} /* Used on every page */
	</DefaultLayout>
```

MyPage.tsx
```jsx
const createUrl = () => {...};
const parseUrl = () => {...};
const props = useInstantSearchProps({createUrl, parseUrl});

<InstantSearchProvider {...props}>
	<Text>
		Hello
	</Text>
</InstantSearchProvider>
```

frontend/templates/product-list/hooks/useCurrentProductList.tsx
```diff
@@ -0,0 +1,75 @@
...
+// ): ProductList {
+//    const variantUpdatedProductList = useVariantProductList(
+//       productList,
+//       algoliaUrl
+//    );
+//    const itemTypeUpdatedProductList = useItemTypeProductList(
+//       variantUpdatedProductList
+//    );
+//    return itemTypeUpdatedProductList ?? variantUpdatedProductList;
+// }
```
jarstelfox, line 24:
```suggestion
```

frontend/templates/product-list/hooks/useCurrentProductList.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+type CurrentProductListContext = {
+   currentProductList: ProductList;
+   setCurrentProductList: (_: ProductList) => void;
```
jarstelfox, line 28: setState has more possibilities in the type than this. You can just do:

```suggestion
   setCurrentProductList: Dispatch<SetStateAction<ProductList>>
```

frontend/templates/product-list/hooks/useCurrentProductList.tsx
```diff
@@ -0,0 +1,75 @@
...
+};
+
+type CurrentProductListProviderProps = {
+   productList: ProductList; // Assuming ProductList is a type you've defined elsewhere
+   algoliaUrl?: string;
+   children: React.ReactNode;
+};
```
jarstelfox, line 35: Children can actually be more types than just a single node:

```suggestion
type CurrentProductListProviderProps = PropsWithChildren<{
   productList: ProductList; // Assuming ProductList is a type you've defined elsewhere
   algoliaUrl?: string;
}>;
```

frontend/templates/product-list/hooks/useCurrentProductList.tsx
```diff
@@ -0,0 +1,75 @@
...
+   );
+   console.log(variantUpdatedProductList);
+   const itemTypeUpdatedProductList = useItemTypeProductList(
+      variantUpdatedProductList
+   );
+   const alteredProductList =
+      itemTypeUpdatedProductList ?? variantUpdatedProductList;
+   const [currentProductList, setCurrentProductList] =
+      React.useState(alteredProductList);
+   //console.log(currentProductList);
```
jarstelfox, line 57:
```suggestion
```

frontend/templates/product-list/hooks/useCurrentProductList.tsx
```diff
@@ -0,0 +1,75 @@
...
+export const CurrentProductListProvider = ({
+   productList,
+   algoliaUrl,
+   children,
+}: CurrentProductListProviderProps) => {
+   const variantUpdatedProductList = useVariantProductList(
+      productList,
+      algoliaUrl
+   );
+   console.log(variantUpdatedProductList);
```
jarstelfox, line 49:
```suggestion
```

frontend/templates/product-list/hooks/useCurrentProductList.tsx
jarstelfox: It looks like this value is never used.

```suggestion
```

frontend/templates/product-list/hooks/useCurrentProductList.tsx
```diff
@@ -0,0 +1,75 @@
...
+//    return itemTypeUpdatedProductList ?? variantUpdatedProductList;
+// }
+
+type CurrentProductListContext = {
+   currentProductList: ProductList;
+   setCurrentProductList: (_: ProductList) => void;
+};
+
+type CurrentProductListProviderProps = {
+   productList: ProductList; // Assuming ProductList is a type you've defined elsewhere
```
jarstelfox, line 32:
```suggestion
   productList: ProductList;
```

kthaler:
dev_block 🦖 - The part title should update when the `Workin: <variant>` filter is removed.
EX: 

https://github.com/iFixit/react-commerce/assets/145375039/370e015c-0964-425a-8e49-6850b679c516

frontend/templates/product-list/hooks/useCurrentProductList.tsx
danielcliu-ifixit: oops

danielcliu-ifixit:
un_dev_block :v:

frontend/components/common/InstantSearchProvider.tsx
danielcliu-ifixit: hmm seems reasonable but i just want to get this pull out rn

frontend/helpers/path-helpers.ts
```diff
@@ -167,3 +167,28 @@ export function ifixitOriginWithSubdomain(subdomain: string) {
...
+   const deviceHandle = pathParts.length >= 2 ? pathParts[1] : '';
+   const itemType = pathParts.length >= 3 ? pathParts[2] : '';
+   return {
+      firstPathSegment,
+      deviceHandle,
+      itemType,
+   };
+}
+
+export function getLocation(url: string) {
```
jarstelfox, line 184: I think you can replace https://github.com/iFixit/react-commerce/pull/2138/files#diff-f344de4d26499cec01f1344a51b8d041d21e27a457b17cae0113e03cb876caa4R72-R79

danielcliu-ifixit: I did attempt this, maybe i did something wrong, but the next hydration started complaining for some reason

jarstelfox (dismissed):
CR 🌵

deltuh-vee:
dev_block ✊ 
I don't think we want to render the variant as is without some extra processing. I think the underscores for this example should be replaced with spaces.
CleanShot 2023-12-20 at 10 36 16@2x
https://react-commerce-git-variants-add-to-title-h1-etc-ifixit.vercel.app/Parts/iPod_Mini:A1051_2nd_gen_4_GB

danielcliu-ifixit:
dev_block :v: on trying to make the text update faster on instantsearch filter changes

danielcliu-ifixit:
un_dev_block :v:

jarstelfox (approved):
CR 🌵

deltuh-vee:
dev_block ✊ 
The item type is duplicated in the No matching products result.
CleanShot 2023-12-20 at 15 11 52@2x
1. Go to https://react-commerce-git-variants-add-to-title-h1-etc-ifixit.vercel.app/Parts/PlayStation_4/Antennas
2. Type in a nonsensical search to get no results.
3. Notice the duplicated item type.

danielcliu-ifixit:
un_dev_block :v:

kthaler:
Here are some of the URLs that I've tested that I feel cover most if not all possible cases:

Part: `/Parts/PlayStation_4`
Part + Item Type: `/Parts/PlayStation_4/Brackets`
Part + filter: /Parts/PlayStation_4?filter%5Bfacet_tags.Capacity%5D%5B0%5D=1%20TB
Part + variant: `/Parts/PlayStation_4:CUH-11XXA`
Part + variant + filter: `/Parts/PlayStation_4:CUH-11XXA?filter%5Bfacet_tags.Capacity%5D%5B0%5D=1%20TB`
Part + variant with old syntax: /Parts/PlayStation_4?filter%5Bworksin%5D=PlayStation%204%20CUH-11XXA
Part + variant with spaces: `/Parts/OnePlus_7:GM1900_%28China_%2F_HK%29`
Part + variant with underscores: `/Parts/iPod_Mini:A1051_2nd_gen_4_GB`
Part + variant with spaces + Item Type: `/Parts/OnePlus_7:GM1900_%28China_%2F_HK%29/Batteries`

QA 💎 - All of these URL displayed the correct title 

deploy_block 🦖 - On a few issues:

1. This page, which is a variant device with a model_number_patten of `RF28HM*LB**`, has a few issues: 

Screenshot 2023-12-21 at 9 43 18 AM

Notice that the header replaces the `*`s with spaces, expect in the summary where it just removes the spaces entirely. Also, the section with the "Search all of iFixit for" link doesn't include the item type as normal.

Parts page: https://react-commerce-git-variants-add-to-title-h1-etc-ifixit.vercel.app/Parts/Samsung_Refrigerator_RF28HM_LB/Drawers_Shelves_Trays_Bins_and_Racks
Device page: https://me.cominor.com/Device/Samsung_Refrigerator_RF28HM_LB?revisionid=HEAD#Section_Parts

2. The search box no longer includes the variant or item type.

Screenshot 2023-12-21 at 9 53 09 AM

danielcliu-ifixit:
@kthaler 
1.) thats just how the data is. There no bug here, the title is formated so internal * are spaces and trailing *s are dropped.

2.) This behavior isn't on master, and wasn't specifically requested, so im going to ignore it for now. Nothing has broken, so i think its fine

un_deploy_block :v:

danielcliu-ifixit:
just merged main, QA :v:  carry-over

jarstelfox (approved):
CR 🌵

PR #2139, opened 2023-12-01
Fix Checkout

danielcliu-ifixit:
## Overview

Basically the shopify graphql was giving us back a domain to the shopify that wasn't right. We want to use the domain the user is on, not the official shopify domain.

This fixes that.

### DIFF
--- a/packages/cart-sdk/hooks/use-checkout.ts
+++ b/packages/cart-sdk/hooks/use-checkout.ts
@@ -167,9 +167,12 @@ function useStandardCheckout() {
          const checkout = await createCheckout(lineItems);
          checkoutUrl = checkout.webUrl;
       }
+
+      const checkout = new URL(checkoutUrl);
+      checkout.hostname = shopifyClient.shopDomain;
       const ssoUrl = new URL(ssoRoute);
-      ssoUrl.searchParams.set('return_to', checkoutUrl);
-      return isUserLoggedIn ? ssoUrl.href : checkoutUrl;
+      ssoUrl.searchParams.set('return_to', checkout.toString());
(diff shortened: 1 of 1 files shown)


packages/cart-sdk/hooks/use-checkout.ts
```diff
@@ -167,9 +167,12 @@ function useStandardCheckout() {
...
          checkoutUrl = checkout.webUrl;
       }
+
+      const checkout = new URL(checkoutUrl);
+      checkout.hostname = shopifyClient.shopDomain;
       const ssoUrl = new URL(ssoRoute);
-      ssoUrl.searchParams.set('return_to', checkoutUrl);
-      return isUserLoggedIn ? ssoUrl.href : checkoutUrl;
+      ssoUrl.searchParams.set('return_to', checkout.toString());
+      return isUserLoggedIn ? ssoUrl.href : checkout.toString();
```
masonmcelvain, line 175: If shopDomain is the wrong thing, this should work

```ts
      const correctedCheckoutUrl = new URL(checkoutUrl);
      correctedCheckoutUrl.host = ssoUrl.host;
      ssoUrl.searchParams.set('return_to', correctedCheckoutUrl.href);
      return isUserLoggedIn ? ssoUrl.href : correctedCheckoutUrl.href;
```

masonmcelvain: Nvm ignore

danielbeardsley (approved):
CR :+1:

masonmcelvain:
Confirmed this is the right value in strapi

image

PR #2142, opened 2023-12-05
Fix Strapi seeding

dhmacs:
closes https://github.com/iFixit/ifixit/issues/50982

When Strapi announced a native way to import/export data from another instance I though: "Bingo! We can now just use that and ditch our custom implementation". For this reason I've attempted to integrate that into our workflow in https://github.com/iFixit/react-commerce/pull/1866. Beside not having to maintain our custom seed algorithm, the change would have fixed a couple of outstanding issues (one of which described by https://github.com/iFixit/ifixit/issues/50982).
Unfortunately that never quite worked out due to Strapi ongoing issues https://github.com/strapi/strapi/issues/18841.

So after many attempts to make it work, I decided that it's worth just fixing our custom plugin for the time being. What you see here is a redesigned algorithm that can import nested fields from another instance.
The algorithm is used only during the Strapi server bootstrap phase for now (the manual import interface is still using the Strapi native import/export under the hood, which is currently broken).

Note: Current implementation does not yet fix image importing, which remains a work-in-progress to avoid scope creep in this PR.

## QA

1. Open the [Strapi preview instance](https://fix-broken-seed-services.govinor.com/admin/content-manager/collectionType/api::menu.menu/3?plugins[i18n][locale]=en)
2. Verify that US Main menu has submenu set
3. Verify that you can create new records


deltuh-vee:
QA 🎬 
* US Main menu has submenu set
* New records can be created

sterlinghirsh (approved):
CR :zap:

PR #2144, opened 2023-12-06
Move 404 to app router

dhmacs:
closes https://github.com/iFixit/ifixit/issues/50983

We move the 404 page to the app router. The pages 404 is being dropped because it's not needed anymore. We've delete also the `app/(defaultLayout)` 404, as it's just a duplicate of the root 404.

## QA

1. Open the Vercel preview
2. Check a non-existent product page and verify that the 404 is shown
3. Check a non-existent product list page and verify that the 404 is shown


danielbeardsley:
This can be re-opened (if desired) in this code's new home

PR #2145, opened 2023-12-07
Product: Follow Shopify redirects

masonmcelvain:
Closes https://github.com/iFixit/ifixit/issues/50841

Our PHP [backend](https://github.com/iFixit/ifixit/blob/master/iFixit/Shopify/ShopifyUrlHandleRedirectJob.php) adds product redirects to Shopify whenever a product handle is changed.

## QA

Shopify redirects should be followed, unless the given handle corresponds to an enabled product that exists.

For example, `/products/t5-torx-screwdriver-1` should redirect to `/products/t5-torx-screwdriver`.
image

For a specific path or target, you can use
https://admin.shopify.com/store/ifixit-us/redirects.json?path=/products/macbook-pro-13-a2159-2019-trackpad

Existing product redirects are viewable at https://admin.shopify.com/store/ifixit-test/redirects.json

### DIFF
--- a/frontend/models/product/server.ts
+++ b/frontend/models/product/server.ts
@@ -5,6 +5,7 @@ import { Product, getProduct } from '.';
 import { findStoreByCode } from '../store';
 import { fetchProductData } from '@lib/ifixit-api/productData';
 import { IFixitAPIClient } from '@ifixit/ifixit-api-client';
+import { z } from 'zod';
 
 export type FindProductArgs = {
    handle: string;
@@ -12,11 +13,16 @@ export type FindProductArgs = {
    ifixitOrigin: string;
 };
 
+type ShopifyProductRedirect = z.infer<typeof ShopifyProductRedirectSchema>;
+export const ShopifyProductRedirectSchema = z.object({
+   target: z.string(),
+});
+
 export async function findProduct({
    handle,
    storeCode,
    ifixitOrigin,
-}: FindProductArgs): Promise<Product | null> {
+}: FindProductArgs): Promise<Product | ShopifyProductRedirect | null> {
    const store = await findStoreByCode(storeCode);
    const { storefrontDomain, storefrontDelegateAccessToken } = store.shopify;
    invariant(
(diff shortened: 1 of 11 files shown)


frontend/models/product/server.ts
```diff
@@ -45,13 +52,14 @@ export async function findProduct({
...
 
+   const urlRedirect = shopifyQueryResponse.urlRedirects.edges[0]?.node?.target;
    const product = await getProduct({
       shopifyProduct: shopifyQueryResponse.product,
       strapiProduct: strapiQueryResponse.products?.data[0],
       iFixitProduct: iFixitQueryResponse,
    });
    if (product == null) {
-      return null;
+      return urlRedirect ? { target: urlRedirect } : null;
```
danielbeardsley, line 62: I appreciate this will only return the redirect if a product isn't found.

danielbeardsley:
CR :+1: 

Though I'd consider changing the property name to something a little more explicit / obscure to prevent accidental overlaps and add clarity: `redirectToPath` or something similar.

jordycosta:
QA 🟢 

Re-directs look good! I crawled all paths from https://admin.shopify.com/store/ifixit-test/redirects.json they appear to be re-directing to their target path

[outlinks.csv](https://github.com/iFixit/react-commerce/files/13593845/outlinks.csv)

frontend/models/product/server.ts
masonmcelvain: That's true. We could add a `__typename` key or something.

masonmcelvain: cc4c4b084b1212bb2db62374eda606dcd8b9f060

frontend/templates/product/server.tsx
```diff
@@ -38,6 +38,15 @@ export const getServerSideProps: GetServerSideProps<ProductTemplateProps> =
          { forceMiss }
       );
 
+      if (product && 'target' in product) {
```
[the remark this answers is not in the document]
masonmcelvain: Yes I'm sure Shopify would 422 `source == target`. Now that you mention it, though, we do not check for redirect loops on the `product.redirectUrl` metafield (the one we use to redirect one product to another).

jordycosta:
QA 🟢 

Re-directs still working as before

PR #2148, opened 2023-12-11
VSCode: Update eslint Setting Value

ianrohde:
## Issue
I noticed an odd automatic update to our `source.fixAll.eslint` setting, dig a little digging, and found that the values have changed from boolean: https://code.visualstudio.com/updates/v1_83#_code-actions-on-save-and-auto-save 

This updates our setting, so autosave still functions as it currently does.

## CR
Confirm formatting on VSCode still functions as expected with the new `always` value.

qa_req 0


davidrans (approved):
CR :coffee:

PR #2149, opened 2023-12-11
GitHub Actions: Enable masking option for ECR Login

ardelato:
## Description

There is a warning about the 'mask-password' option for the "amazon-ecr-login" action. If ran in debug mode, the password will be outputted. By default this option is not enabled as the expectation is to run the service after the deployment but as a secondary job. As such the password cannot be masked as it is needed to run the service.

https://github.com/aws-actions/amazon-ecr-login#run-an-image-as-a-service

However, this is not the workflow we have set up, therefore we need to enable the option to mask the password.

Reference: https://github.com/iFixit/vigilo/pull/19

### DIFF
--- a/.github/workflows/deploy-strapi.yml
+++ b/.github/workflows/deploy-strapi.yml
@@ -31,7 +31,7 @@ jobs:
            id: login-ecr
            uses: aws-actions/amazon-ecr-login@v1
            with:
-               mask-password: true
+              mask-password: true
 
          - name: Build, Tag, and Push Base Image to Amazon ECR Private
            env:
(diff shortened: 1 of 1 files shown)


djmetzle (dismissed):
CR :closed_lock_with_key:

batbattur (approved):
CR 🚛 TIL

Should we update it to `v2`? From https://github.com/aws-actions/amazon-ecr-login#new-v2-release
> New v2 Release
In the new major version for this action, the default value of the mask-password input has changed from false to true.

ardelato:
>Should we update it to v2? From [aws-actions/amazon-ecr-login#new-v2-release](https://github.com/aws-actions/amazon-ecr-login#new-v2-release)

Yeah that would be a lot better and easier

ardelato:
> Should we update it to v2? From [aws-actions/amazon-ecr-login#new-v2-release](https://github.com/aws-actions/amazon-ecr-login#new-v2-release)

> Yeah that would be a lot better and easier

On second thought, I'll do that in a followup pull so we can spend a bit more time viewing the change log.

PR #2161, opened 2023-12-16
Product option: Manually handle output text

jarstelfox:
The selected state is being set in aria, but somehow not transfering.

This gets us the correct readout at least

Connects: https://github.com/iFixit/ifixit/issues/51161


danielcliu-ifixit (approved):
CR :v:

erinemay:
QA 🦖 this did it. The screen reader experience is much better. It's clear which is selected and it's easy to toggle one off and on. When the selection changes, the screen reader appropriately notes which is selected. This is a good experience now.

kwiens:
Thanks everyone for fixing all of these today, it's very helpful having it done before Monday. Appreciate it!

PR #2163, opened 2023-12-18
Product Pages: Support encoded slashes

jarstelfox:
See: https://github.com/vercel/next.js/issues/49646

We have slashes in some variants.

You can see the bug by visiting:

 * https://www.ifixit.com/Parts/iPhone_XR:A1984_US%2FCanada
 * Click Any item type and the page loads fine
 * Refresh and the page 404s

This works around the issue by manually handling the url parsing for this page. We only needed the path segments.

### DIFF
--- a/frontend/templates/product-list/server.tsx
+++ b/frontend/templates/product-list/server.tsx
@@ -285,11 +285,23 @@ async function getSafeServerState({
 function getDevicePathSegments(
    context: GetServerSidePropsContext<ParsedUrlQuery>
 ) {
-   const { deviceHandleItemType } = context.params || {};
-   if (Array.isArray(deviceHandleItemType)) {
-      return deviceHandleItemType;
-   }
-   return [];
+   /**
+    * We should be able to use `context.params` here,
+    * but see: https://github.com/vercel/next.js/issues/49646
+    *
+    * Basically something like device:variant%2FWithASlash
+    * Will be parsed into:
+    *   ['device', 'variant', 'WithASlash']
+    * Instead of:
+    *  ['device', 'variant/WithASlash']
+    */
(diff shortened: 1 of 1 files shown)


frontend/templates/product-list/server.tsx
```diff
@@ -285,11 +285,23 @@ async function getSafeServerState({
...
+    * We should be able to use `context.params` here,
+    * but see: https://github.com/vercel/next.js/issues/49646
+    *
+    * Basically something like device:variant%2FWithASlash
+    * Will be parsed into:
+    *   ['device', 'variant', 'WithASlash']
+    * Instead of:
+    *  ['device', 'variant/WithASlash']
+    */
+   const path = context.resolvedUrl.split('?')[0];
```
danielbeardsley, line 298: I forget which is which, but will this be the right url when the client side is hitting the next.js `data` api to just get the props?

I'm not entirely sure we'll run into instances where that will happen for this case, but it seems feasible.

sterlinghirsh (approved):
CR :zap: Shame that we have to do this, but it looks ok to me!

frontend/templates/product-list/server.tsx
jarstelfox: https://nextjs.org/docs/pages/api-reference/functions/get-server-side-props#context-parameter

>resolvedUrl
>
>A normalized version of the request URL that strips the _next/data prefix for client transitions and includes original query values.

The docs don't indicate much clarity here, but it seemed safe enough to me.

danielbeardsley: Sounds like the right one!

deltuh-vee:
QA 🎬 
No longer getting 404  when refreshing with an item type selected.

Issue #2067, opened 2023-10-30
Add Strapi product list redirect field


masonmcelvain:
## Problem

From time to time we get requests to redirect one product list to another. Let's add a field in Strapi for us to do that.

Edit: We need research/decide if we'd like to add these redirects at build time or at request time (see https://github.com/iFixit/react-commerce/issues/2067#issuecomment-1786809974)

Example requests:
https://github.com/iFixit/react-commerce/issues/2065
https://github.com/iFixit/ifixit/issues/45749

dhmacs:
@masonmcelvain should this be a field on the product list to redirect to another product list? Is this something that we want to use at build time to compile `next.config.js` redirects or should this happen dynamically at request time?

masonmcelvain:
Good question, I hadn't considered that we could fetch them at build time. That sounds like a more performant solution, though if I understand correctly, new redirects in Strapi would not go live until the next build.

I think users will be hitting these redirects far more often than we will be editing them, so I'd vote for fetching them all at build time and injecting them into `next.config.js`.

dhmacs:
> if I understand correctly, new redirects in Strapi would not go live until the next build.

Correct 👍 

> That sounds like a more performant solution

I'm not 100% sure it's more performant. Need to check this, but maybe there's a way to cache redirect at the CDN level? 🤔

sterlinghirsh:
I'm nervous about making it rely on a rebuild before the redirects go into effect. Seems like if we load the product list and then detect a redirect, we should follow it.

masonmcelvain:
> I'm nervous about making it rely on a rebuild before the redirects go into effect. Seems like if we load the product list and then detect a redirect, we should follow it.

I'm going to implement it at request time, like we did for products in https://github.com/iFixit/ifixit/issues/50841. So we won't need to rebuild.

ifixitmaxb:
Do these redirects have to be 308s? Is it possible to make them 301s? Also, would it be possible to have the option between 301 and 302? For example, the Black Friday, Holiday Gift Guide, and Cyber Monday pages should ideally be "temporary" (~10 month) redirects until we want to resurrect them every year around late October.

masonmcelvain:
[...]

PR #2165, opened 2023-12-20
Add Strapi product list redirect fields

masonmcelvain:
Connects https://github.com/iFixit/react-commerce/issues/2067

The intent is to grab `redirectTo` at request time (cached with Redis) and issue a redirect if appropriate.

`redirectFrom` isn't necessary for the Next.js app, but offers a convenient way for Strapi admin users to create redirects from a source to a target from the source product list page. (e.g. If I am on product list A and want to cause product list B to redirect to product list A, I can do so without visiting product list B in Strapi admin).

https://docs.strapi.io/dev-docs/backend-customization/models#relations

qa_req 0


jarstelfox (approved):
CR 🌵

danielbeardsley:
> e.g. If I am on product list A and want to cause product list B to redirect to product list A, I can do so without visiting product list B in Strapi admin

I feel like this is complexity that isn't really needed. We have to add another query to the graphQL query and it allows there to be two places where the same data is stored and thus there can be conflicts:

    X.redirectFrom = Y 
    Y.redirectTo = Z

Where do we go when visiting Y?

masonmcelvain:
> I feel like this is complexity that isn't really needed

You're right, it's not necessary, but it could be nice to see which product lists redirect to the one you're looking at. It is also a visual cue to help Strapi admin users prevent redirect chains.

> there can be conflicts

I don't think Strapi allows there to be conflicts. When you add a `redirectFrom` via the UI, the `redirectTo` is automatically set on the other product list, even if `redirectTo` is already set to something. Try it yourself: https://main.govinor.com/admin/content-manager/collectionType/api::product-list.product-list?page=1&pageSize=100&sort=title:ASC&plugins[i18n][locale]=en

We do not have to query for `redirectFrom` in Next.js, and I do not in the implementation here https://github.com/iFixit/react-commerce/pull/2166.

danielbeardsley:
> I don't think Strapi allows there to be conflicts.

Oh! I didn't understand the `"inversedBy": "redirectFrom"` before. I assumed this was two separate relationships, but instead it's more like giving a name to the other side of the relationship for easier reference / visibility. Nice!

Thanks for explaining!

PR #2167, opened 2023-12-21
Piwik: Add User Privilege Custom Dimension Tracking

danielcliu-ifixit:
## Overview

We want to track what users are doing what based on user privilege or role (admin, moderator, user, etc). This change will send piwik a userPrivilege dimension if a user is logged in. If the user is not logged in, it will not update or set the dimension.

## QA

Do users have their userPrivilege logged correctly in piwik? If you log out does that userPrivilege change (it shouldn't, we want to track the last privilege you had). 

CC @addison-grant 
Connects: [#51217](https://github.com/iFixit/ifixit/issues/51217)

### DIFF
--- a/frontend/layouts/default/index.tsx
+++ b/frontend/layouts/default/index.tsx
@@ -28,6 +28,7 @@ import {
 import {
    trackPiwikPreferredLanguage,
    trackPiwikPreferredStore,
+   trackPiwikUserPrivilege,
 } from '@ifixit/analytics';
 import { useAppContext } from '@ifixit/app';
 import { useAuthenticatedUser } from '@ifixit/auth-sdk';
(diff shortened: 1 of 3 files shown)


masonmcelvain (approved):
CR 👍🏻

kthaler:
QA ⚙️ - Checked all avaialable user priveledge types and confirmed that they all appear on piwik as the `userPriviledge` custom dimension. Also checked that if the user is not logged in, it will not update or set the dimension.

small deploy_block 🦖 - Should we remove the `userPriviledge` custom dimension if the user logs out? I logged out of my Teacher account and noticed that the dimension persisted while I clicked around. 
Also, is there a way to keep the ordering the same for all user types? For example, the ordering is different when its a Student vs Author+Mod:

Screenshot 2023-12-21 at 11 28 39 AM

Screenshot 2023-12-21 at 11 28 06 AM

danielcliu-ifixit:
@kthaler no, user privilege should remain the last value that was valid.

Im not concerned about the dimension order in the debugger.
un_deploy_block :v:

danielcliu-ifixit:
dev_block :v:

danielcliu-ifixit:
un_dev_block :v:

danielcliu-ifixit:
Just fixed a capitalization preference and then merged master, i think QA :v: carryover is safe

masonmcelvain (approved):
CR 👍🏻

PR #2170, opened 2023-12-21
Extended Related Problems: Switch feature flag to true

andyg0808:
Does what it says on the tin.

## QA Notes
This should now include problems from the parent device, along with any ancestor devices which have inheritance enabled. Please see the issue for the precise spec; if you find anything wonky, please let me know!

### DIFF
--- a/frontend/tests/playwright/troubleshooting/loading.spec.ts
+++ b/frontend/tests/playwright/troubleshooting/loading.spec.ts
@@ -8,22 +8,8 @@ test.describe('Troubleshooting Page Content and SEO', () => {
       ).toBeVisible();
    });
 
-   test('Related Problems hidden by feature flag', async ({ page }) => {
+   test('New Related Problems visible', async ({ page }) => {
       await page.goto('/Troubleshooting/Dryer/Not+Spinning/478872');
-      // Select the element with the ID 'related-problems'
-      const element = await page.locator('#related-problems');
-
-      // Assert that the element's 'data-test' attribute is not 'related-problems-v2'
-      await expect(element).not.toHaveAttribute(
-         'data-test',
-         'related-problems-v2'
-      );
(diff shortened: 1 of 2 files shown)


ianrohde (approved):
CR ⚪ 

Should we remove the old version function?

andyg0808:
@ianrohde Yes, but I was thinking of doing that in a future PR, so this could be as little change as possible.

ianrohde (approved):
CR ⚪

deltuh-vee:
The issue mentions
> Target N "Related Problems" on the problem page.

Is there a specific number for this?

deltuh-vee:
dev_block ✊ 
This troubleshooting page I created has a circular reference back to itself:
CleanShot 2023-12-22 at 13 14 07
Troubleshooting page: https://react-commerce-git-extended-related-problems-release-ifixit.vercel.app/Troubleshooting/Test_Device_2/test+troubleshooting/536566
Device page: https://www.cominor.com/Device/Test_Device_2

andyg0808:
@deltuh-vee Interesting. That definitely seems like a blocker. I'll work on keeping these cleaner in the new year.

andyg0808:
Okay, @deltuh-vee I think that's fixed. un_dev_block :+1:

deltuh-vee:
QA 🎬
- Related pages pulls from direct links to /Troubleshooting in the page text first, followed by directly linked Related Problems.
- If there are fewer than 6 problems to pull from, Related Pages pulls from problems related to the same device. If inheritance is enabled, problems are pulled from ancestor devices.
- Problems from each device are sorted by view count descending.
- When pulling from indirectly linked problems, a maximum of 6 problems total will be shown.
- When there are more than 6 problems that are directly linked, all of the directly linked problems are shown.

PR #2171, opened 2023-12-22
Expand tool category by default

federicobadini:
Closes https://github.com/iFixit/ifixit/issues/51222

### QA

1. Visit [Vercel preview](https://react-commerce-git-expand-tool-category-facet-acc-749d34-ifixit.vercel.app/Tools/)
2. Verify that the tool category facet accordion is open by default

### DIFF
--- a/frontend/templates/product-list/sections/FilterableProductsSection/facets/accordion/index.tsx
+++ b/frontend/templates/product-list/sections/FilterableProductsSection/facets/accordion/index.tsx
@@ -12,7 +12,10 @@ type FacetsAccordianProps = {
    productList: ProductList;
 };
 
-const initialExpandedFacets = ['facet_tags.Item Type'];
+const initialExpandedFacets = [
+   'facet_tags.Item Type',
+   'facet_tags.Tool Category',
+];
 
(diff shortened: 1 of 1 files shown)


masonmcelvain (approved):
CR 👍🏻

deltuh-vee:
QA 🎬 
Tool Category is expanded by default.

deploy_block ☝️ 
The issue didn't mention this, but the Slack discussion mentions fully expanding Tool Category: 
> yeah I'd be open to fully expanding and skipping Show more

Do we still want this?

federicobadini:
Do you have any update on this?

masonmcelvain:
Quoting Mattia from [Slack](https://ifixit.slack.com/archives/C01K32B4FPX/p1703835502251639?thread_ts=1702945253.007279&cid=C01K32B4FPX):

> Yes it makes sense to open the first filter if there are not many filters that can be selected. However, I would continue to keep the "show more" to make it obvious to the user that it is not the only filter possible, the price filter may not be immediately visible to the user.
We could also always keep the first one open

deltuh-vee:
un_deploy_block 👍

PR #2172, opened 2023-12-26
Tools section backend

federicobadini:
Connects https://github.com/iFixit/ifixit/issues/49326

This PR introduces the necessary backend changes to port the old PTT tools section to react-commerce.
Here is the mockup for the new section: https://www.figma.com/file/BolmFORO6ZfHXePOSyEGSS/iFixit---Page-sections?node-id=1839%3A1795&mode=dev

Currently the only scenario in which this section is going to be used is inside the pro tech toolkit product page.

Strapi is currently providing only the section positioning. The section content is static.
This is due to the fact that right now each image requires custom margins and scaling to fit the SVG proportions.
In the future, if we want to make the content of this section dynamic and provide data from Strapi, we will have to consider creating SVGs sized exactly as the tool images. 

1. Visit Vercel [preview](https://react-commerce-git-tools-section-backend-ifixit.vercel.app/products/repair-business-toolkit)
2. Verify that the product page is fully operational - no visible change should be present

PR #2173, opened 2023-12-26
Tools section frontend

federicobadini:
Close https://github.com/iFixit/ifixit/issues/49326

This PR introduces the necessary frontend changes to port the old PTT tools section to react-commerce.
Here is the mockup for the new section: https://www.figma.com/file/BolmFORO6ZfHXePOSyEGSS/iFixit---Page-sections?node-id=1839%3A1795&mode=dev

Currently the only scenario in which this section is going to be used is inside the pro tech toolkit product page.

Strapi is currently providing only the section positioning. The section content is static.
This is due to the fact that right now each image requires custom margins and scaling to fit the SVG proportions.
In the future, if we want to make the content of this section dynamic and provide data from Strapi, we will have to consider creating SVGs sized exactly as the tool images. 

1. Visit Vercel [preview](https://react-commerce-git-tools-section-frontend-ifixit.vercel.app/products/repair-business-toolkit)
2. Verify that the product page is fully operational
3. At the bottom of the page you should find the new section
4. Test its scroll, scroll-on-click and the center lock on page resize

https://github.com/iFixit/react-commerce/assets/7805759/97ff07c7-f374-40e9-9cba-c42138c70868

deploy_block 🔴 on https://github.com/iFixit/react-commerce/pull/2172


deltuh-vee:
QA ✊ 
* Scrolling, scrolling-on-click, and the center lock on page resize are working for the Tools Section.
* The product page is still working.

Minor nitpick:
* Could we stop the scrolling before it gets too far along the edge so that the text doesn't look off-center?
CleanShot 2023-12-26 at 14 09 49
* There seems to be a bit of extra white-space at the top of the tools section. Could we decrease some of the margin on the component? 
IMG_A896C5EECDF6-1

federicobadini:
> Could we stop the scrolling before it gets too far along the edge so that the text doesn't look off-center?

Yes I'm going to ship a commit for this

> There seems to be a bit of extra white-space at the top of the tools section. Could we decrease some of the margin on the component?

This is a bit more complex.
It seems too spaced for several reasons:

1. The previous section has grey background too
2. The central element is not as tall as the metal-spudger which sets the section height
3. Vertical paddings needs to be reduced across several section in the mobile layout of the product page

For point 3 we will work on a separate PR to make margins as uniform as possible even on mobile, so expect a reduction from there. However point 1 and 2 might still give the impression of having some extra space.

Given that this section is mostly designed for being used in a single place, as soon as it will be placed in the page we can eventually review those paddings to have the best look and feel possible.

federicobadini:
The track doesn't scroll past the first and last image anymore.
This is an additional static portion of code that needs to be taken into account in case of refactoring.
I'm still convinced that in the long term the best strategy would be to have images and svg traces that corresponds to each other sizes, allowing to reduce most of the code dedicated to image positioning contained in this PR

un_dev_block 🟢

deltuh-vee:
QA 🎬 
un_deploy_block 👍 
* [Earlier](https://github.com/iFixit/react-commerce/pull/2173#issuecomment-1869801216) QA still stands.
* Scrolling is appropriately bounded

ianrohde:
On the mobile spacing issue, there is 64px of bottom padding applied to ComponentProductBitTable and 64px of top padding applied to ComponentSectionTools. If we remove those and center the SVG's in ToolsSection [here](https://github.com/iFixit/react-commerce/pull/2173/files#diff-9fdaa63f8e72d175dfa0fa542e4c68aa75a1e904a52e828efa64c0ef4f1b5adcR132), we can improve the mobile layout without too much trouble:

<details>

https://github.com/iFixit/react-commerce/assets/1634505/2df17d0f-f556-4515-8a22-59c6e479ec88

</details>

Are we controlling scrolling with javascript here? I tried playing around with [CSS scroll-snap](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Scroll_Snap), but it's not working. Seems like it's pretty much built for this type of UI.

masonmcelvain (approved):
CR 👍🏻 The code looks good and it seems like the feature works as intended on mobile and desktop.

It's unfortunate that the product page bundle size went up by 30%, I guess it's from all the SVG files? I don't think it's a blocker because this is a big feature, but worth considering.

Other stakeholders will be excited to see this back on the pro tech page, so I think we should ship this and optimize in future PR's (margin, css scroll snap, bundle size maybe, etc).

Issue #2175, opened 2023-12-28
Strapi deploy: Move to github actions


mlahargou:
This is the last thing we use cimpler for. Let's switch it to actions.

Currently, cimpler runs `backend/strapi-deploy`. That builds and pushes
the docker image to the ECR registry. It also does a `cap prod strapi:deploy`.
It also tags the deploy. 

We already have a workflow that does some of this: https://github.com/iFixit/react-commerce/blob/1303b0def289d79355d0357dc3193057cebcadba/.github/workflows/deploy-strapi.yml

Let's finish the job and get it on actions. We may have to use a self-hosted runner
as we use a cap task to deploy strapi.

cc @masonmcelvain @danielbeardsley @djmetzle

danielbeardsley:
> Let's finish the job and get it on actions. We may have to use a self-hosted runner
as we use a cap task to deploy strapi.

There's a bit of nuance here that this repo doesn't contain the capistrano task (those are in the monorepo). So, we have this deploy process actually invoke code from another repo. This is unusual but when the repos are merged it won't be a big deal.

Maybe we should wait for project [Vestibule](https://github.com/iFixit/ifixit/issues/47724) and then this change will be relatively straight forward. Thoughts?

mlahargou:
I think we could go with either approach. I think we may be able to add a self-hosted runner and then we can still use the capistrano tasks to deploy.

PR #2176, opened 2023-12-28
Strapi: Deploy with actions

mlahargou:
We previously used cimpler to deploy strapi. We are trying to remove all
instances of cimpler, and this is the last one. This creates a copy of the
strapi-deploy script that cimpler uses. It's slightly modified because some
of the checks it did can be more easily expressed in the workflow file.

CR:
--
I think we eventually will want to transform the bash script into actual
workflow steps, but for now a copy of the existing/working bash script
should be enough.

QA:
--
Idk how we want to test this. Might have to just test this on deploy.

On Deploy:
--
Deactivate the webhook that sends the push event to ubreakit. Deploy
and verify that all went well.

Closes: https://github.com/iFixit/react-commerce/issues/2175

qa_req 0
QA on deployment


mlahargou:
deploy_block 📱 on the runner being created.

jordycosta:
I will mark this as qa_req 0 for now, feel free to remove that if it needs to be tested before deployment

mlahargou:
Runner is created. Keeping deploy_blocked b/c I need to be careful when deploying this.

batbattur (approved):
CR 🚛 

Since the deploy logs will be public, is there anything we should check?

Looks like this was the latest strapi deploy: [logs](https://www.ubreakit.com/Misc/cimpler-logs/strapi-deploy/main--5d7c686e03eaa54daff017929ae1ac553e77e04d--1704307511490.log). I don't see anything sensitive but just wanted to make sure.

ardelato:
Heads up, I just found this line of related code in `server-templates` that we should drop once this goes out.
https://github.com/iFixit/server-templates/blob/master/terraform/iam/github/actions_assume_role.tf#L7

danielbeardsley (approved):
Looks good!

After this goes out, we can drop the IAM permissions stuff from server-templates.

CR :+1:

PR #2181, opened 2024-01-03
Strapi: Add redirect type to product list

masonmcelvain:
Connects https://github.com/iFixit/ifixit/issues/51356

I decided to make `redirectToType` an enum instead of a boolean because I found it more readable as an admin user. If folks disagree I welcome their feedback.

One thing I noticed is that existing product list records do not have a default value populated. It seems like a Strapi bug (or "missing feature") https://github.com/strapi/strapi/issues/18636.

qa_req 0


danielbeardsley (approved):
CR :+1:

PR #2184, opened 2024-01-03
Strapi Deploy: Drop old deploy script

mlahargou:
This script used to be used by cimpler to deploy strapi. We have 
turned off the webhook, so this script is no longer used. I will also 
remove the cimpler config in a follow up pr on ifixit/ifixit.

qa_req 0

Connects: https://github.com/iFixit/react-commerce/issues/2175


ardelato (approved):
CR 👍

PR #2185, opened 2024-01-03
Strapi Deploy: Add concurrency

mlahargou:
We don't want two of these running at the same time. Let's make sure
only one can run at a time.

Connects: https://github.com/iFixit/react-commerce/issues/2175

qa_req 0


batbattur (approved):
CR 🚛

PR #2189, opened 2024-01-08
Footer: Fix HTML Semantics Warning

ianrohde:
## Issue

Quick fix. I'm seeing this warning on dev:

`Warning: validateDOMNesting(...): <p> cannot appear as a descendant of <p>.`

<details>

Screenshot 2024-01-08 at 11 42 02 AM

</details>

We're passing in children, which looks to have another Text component or paragraph element, so let's convert the parent to a Box instead.

## CR/QA

Confirm footer is still functioning as expected.


ardelato (approved):
CR 👍

kthaler:
QA 💎 - Footer still works as expected on all pages

PR #2191, opened 2024-01-09
Refactor app router page frame

dhmacs:
closes https://github.com/iFixit/ifixit/issues/51442

qa_req 0


frontend/app/(defaultLayout)/not-found.tsx
```diff
@@ -1,13 +1,20 @@
...
-   const layoutProps = await layoutPropsPromise;
-
-   return <NotFoundPage {...layoutProps} />;
+   return (
+      <PageFrame
+         layoutData={layoutData}
+         regionMenu={<RegionMenu regions={layoutData.stores} />}
+      >
+         <NotFoundPage />
+      </PageFrame>
```
masonmcelvain, line 18: This `not-found` component currently serves as the global 404 component and as the 404 component for routes in `(defaultLayout)`. Based on the [component hierarchy](https://nextjs.org/docs/app/building-your-application/routing#component-hierarchy) described in the docs, it seems like this implementation would result in 2 page frames for 404s on routes within `(defaultLayout)`.

```tsx
<Layout>
  <DefaultLayout>
    <PageFrame>
      <ErrorBoundary fallback={
        <NotFound>
          <PageFrame>
            <NotFoundPage />
          </PageFrame>
        </NotFound>
      }>
        <Page />
      </ErrorBoundary>
    </PageFrame>
  </DefaultLayout>
</Layout>
```

frontend/app/global-error.tsx
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export default function GlobalError({ error }: GlobalErrorProps) {
+   useEffect(() => {
+      Sentry.captureException(error, {
+         tags: { error_boundary: true },
+      });
+   }, [error]);
+   return (
+      <html>
```
masonmcelvain, line 21:
```suggestion
      <html lang="en">
```

frontend/app/(defaultLayout)/components/page-frame/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <AppProviders>
+         <ShopifyStorefrontProvider
+            shopDomain={shopifyCredentials.storefrontDomain}
+            storefrontAccessToken={shopifyCredentials.storefrontAccessToken}
+            apiVersion={SHOPIFY_STOREFRONT_VERSION}
+         >
+            <Box>
+               <Flex direction="column" minH="100vh">
+                  <Header menu={header.menu} />
+                  {children}
```
masonmcelvain, line 40: I notice that we lost the admin message in this refactor. That's ok, but we should make an issue to re-imagine this feature or implement it in the app router later on.

masonmcelvain (approved):
CR 👏🏻 Nice work!

deploy_block 🔍 on making sure that we don't accidentally render two page frames on some 404 routes.

frontend/app/(defaultLayout)/not-found.tsx
dhmacs: You're right, but there's a catch. My refactor is not changing the current behaviour (if you check main branch we're rendering the page frame in the not found) and I think the reason is that the version of Next.js we're using now, not found pages are not affected by nested layouts. When we will upgrade Next.js this will behave as you are suggesting (I've tested this during my work on #2144).
I know it's a silly breaking change, but I don't make the rules 😂

masonmcelvain: > the version of Next.js we're using now, not found pages are not affected by nested layouts

Wow that is surprising! Thanks for explaining 👍🏻

masonmcelvain:
un_deploy_block 👍🏻

frontend/app/(defaultLayout)/components/page-frame/index.tsx
dhmacs: Yes, this is part of a longer refactor to move the product page to the app router. I've done much of the work during these holidays, now I'm splitting it up in chunks to make it easier to review and ship

PR #2193, opened 2024-01-09
Setup new @vercel/speed-insights package

dhmacs:
closes https://github.com/iFixit/ifixit/issues/51443

qa_req 0

I think the only way to test this is to ship it and check that speed insights still collect data after the deploy

Screenshot 2024-01-09 at 17 09 10


frontend/app/layout.tsx
```diff
@@ -105,7 +106,10 @@ export default function RootLayout({
                strategy="beforeInteractive"
             />
          </head>
-         <body>{children}</body>
+         <body>
+            {children}
+            <SpeedInsights sampleRate={0.2} />
```
davidrans, line 111: 20% of requests seems like a lot, but it looks like that's what we were doing before

davidrans (approved):
CR :coffee:

PR #2194, opened 2024-01-09
Defensively parse fetch request body

federicobadini:
The [errors spike](https://ifixit.sentry.io/alerts/rules/details/180935/?end=2024-01-09T18%3A25%3A00&start=2024-01-09T16%3A40%3A29) that appeared today seems to be related to the way in which the sentry fetch config parses the input body.
This PR handles the process more defensively, not giving for granted that the body is JSON encoded.

qa_req 0

### DIFF
--- a/frontend/lib/shopify-storefront-sdk/generated/sdk.ts
+++ b/frontend/lib/shopify-storefront-sdk/generated/sdk.ts
@@ -2654,7 +2654,7 @@ export enum CountryCode {
    Tn = 'TN',
    /** Tonga. */
    To = 'TO',
-   /** Turkey. */
+   /** Türkiye. */
    Tr = 'TR',
    /** Trinidad & Tobago. */
    Tt = 'TT',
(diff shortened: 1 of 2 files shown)


jarstelfox (dismissed):
CR 🌵

jarstelfox (approved):
CR 🌵

jarstelfox:
FYI these sentry issues are still occurring at a high rate

dhmacs:
> FYI these sentry issues are still occurring at a high rate

@jarstelfox I'm seeing this for the past 24 hours. Where are you seeing high rate issues?

Screenshot 2024-01-10 at 22 13 08

federicobadini:
@jarstelfox and I discussed this yesterday and determined that the remaining error points we are still observing are likely attributable to users who continue to run the code without the fix. 
It could be related to caching, but a significant number of those sessions indicate that users remained on the page for an extended duration. The possibility exists that users still have tabs that were opened on those pages since before the fix. 
We expect a gradual reduction in these errors over time.

PR #2198, opened 2024-01-10
Add app router product page feature flag

dhmacs:
Since I intend to merge things into main gradually to avoid a huge refactor PR, let's setup a feature flag. The app router product page will be hosted under `/products/app-router/[handle]` to avoid conflicts with the existing page. When the whole refactor has been completed we can delete the `/pages` implementation and move the app router implementation into the current route.

ℹ️ I've enabled the flag for `react-commerce`

qa_req 0


masonmcelvain (approved):
CR 👍🏻

PR #2199, opened 2024-01-11
Seeding - check 404 before failing

federicobadini:
This PR switches two checks that were performed during Strapi seeding.
In detail we should check for 404 before checking the generic `response.ok`, otherwise new models introduced locally will make the seeding process fail.

qa_req 0


masonmcelvain (approved):
CR 👍🏻

PR #2201, opened 2024-01-11
Drop deprecated cart api coupon fields

masonmcelvain:
Connects https://github.com/iFixit/ifixit/issues/51463

These fields are going away, so let's drop them from the type.

qa_req 0


davidrans (approved):
CR :coffee:

PR #2202, opened 2024-01-12
Show Update Cart Errors In Cart Drawer

danielcliu-ifixit:
## Overview

We sometimes can have things error out when trying to update the cart. We would roll back the cart but not change anything, and not tell our user that anything wrong happened. Now we give some visual feedback that there was an error updating the cart in the cart drawer.

## QA

When we have an error updating the cart (adding a bad product, product is out of inventory, etc) we now display an error message in the cart drawer. The error message is closable, and also disappears after an alloted time.

Also, the checkout errors also still display as expected.

Closes: https://github.com/iFixit/ifixit/issues/51213


dhmacs (commented):
This is a really good idea, didn't realize we weren't handling errors. This implementation seems to address only errors that are happening during add to cart action, but there are also other cart mutations that might fail, so we might want to address those as well.

On the implementation side, I'd like to suggest an alternative approach. The current solution is coupling the mutation errors with the cart query, and I think long term it might be a little bit confusing to have both `cart.isError` (for queries) and `cart.data.error` (for mutations) on the cart query object.
My suggestion is to use the cart drawer context to keep the error state for mutations and expose methods to show/clear the errors. For example you can expose a method named `showError` that can be called from the mutation `onError` handler:

```tsx
const drawer = useCartDrawer();
useMutation(
  () => { ... },
  onError: (error) => {
    drawer.showError(error);
  }
)
```

this is similar to the approach described [here](https://tkdodo.eu/blog/react-query-error-handling#the-onerror-callback) by a React Query maintainer.

I should note also that if we upgrade to React Query v5, the implementation would probably be easier thanks to new [useMutationState](https://tanstack.com/query/latest/docs/react/reference/useMutationState). I didn't test it, but it seems that this method would enable us to retrieve mutations errors from the cart drawer, without having to store it ourselves.

dev_block 🔢 on these considerations

danielcliu-ifixit:
@dhmacs that is a cleaner implementation, thanks for the advice!

Closing in favor of using hte cart drawer context instead.

PR #2203, opened 2024-01-12
Add product page store select slot

dhmacs:
This adds a [parallel route](https://nextjs.org/docs/app/building-your-application/routing/parallel-routes) to override the default store select (inside the footer) when we navigate to a product page.
Region menu is also renamed as store select to better align with model name.

## QA

1. Visit [app router product page preview](https://react-commerce-git-add-product-page-store-select-slot-ifixit.vercel.app/products/app-router/iphone-11-screen)
2. Verify that the store select links in the footer are the same of the ["pages" product page](https://react-commerce-git-add-product-page-store-select-slot-ifixit.vercel.app/products/iphone-11-screen)


masonmcelvain (approved):
CR 👍🏻

deltuh-vee:
QA 🎬 
The Store select links are still the same in the footer.

PR #2207, opened 2024-01-15
Show Errors in Cart Drawer

danielcliu-ifixit:
## Overview

We sometimes can have things error out when trying to update the cart. We would roll back the cart but not change anything, and not tell our user that anything wrong happened. Now we give some visual feedback that there was an error updating the cart in the cart drawer.

## QA

When we have an error updating the cart (adding a bad product, product is out of inventory, removing an item and it fails) we now display an error message in the cart drawer. The error message is closable, and also disappears after an alloted time.

Also, the checkout errors also still display as expected.

CC @dhmacs 
Closes: https://github.com/iFixit/ifixit/issues/51213

### DIFF
--- a/packages/cart-sdk/hooks/use-remove-line-item.ts
+++ b/packages/cart-sdk/hooks/use-remove-line-item.ts
@@ -7,6 +7,7 @@ import { useIFixitApiClient } from '@ifixit/ifixit-api-client';
 import { useMutation, useQueryClient } from '@tanstack/react-query';
 import { Cart, CartLineItem } from '../types';
 import { cartKeys } from '../utils';
+import { useCartDrawer } from '@ifixit/ui';
 
 interface DeleteLineItemInput {
    item: CartLineItem;
@@ -18,6 +19,7 @@ interface DeleteLineItemInput {
 export function useRemoveLineItem() {
    const client = useQueryClient();
    const iFixitApiClient = useIFixitApiClient();
+   const { addErrorMessage } = useCartDrawer();
    const mutation = useMutation(
       async (input) => {
          await iFixitApiClient.delete(
@@ -74,6 +76,9 @@ export function useRemoveLineItem() {
             client.setQueryData<Cart | undefined>(
                cartKeys.cart,
                context?.previousCart
+            );
+            addErrorMessage(
+               (error as string) || 'An error occurred updating the cart.'
             );
          },
          onSuccess: (data, variables) => {
--- a/packages/cart-sdk/hooks/use-remove-line-item.ts
+++ b/packages/cart-sdk/hooks/use-remove-line-item.ts
@@ -77,9 +77,7 @@ export function useRemoveLineItem() {
                cartKeys.cart,
                context?.previousCart
             );
-            addErrorMessage(
-               (error as string) || 'An error occurred updating the cart.'
-            );
+            addErrorMessage(error);
          },
          onSuccess: (data, variables) => {
             const cart = client.getQueryData<Cart>(cartKeys.cart);
--- a/packages/cart-sdk/hooks/use-remove-line-item.ts
+++ b/packages/cart-sdk/hooks/use-remove-line-item.ts
@@ -19,7 +19,7 @@ interface DeleteLineItemInput {
 export function useRemoveLineItem() {
    const client = useQueryClient();
    const iFixitApiClient = useIFixitApiClient();
-   const { addErrorMessage } = useCartDrawer();
+   const { addError } = useCartDrawer();
    const mutation = useMutation(
       async (input) => {
          await iFixitApiClient.delete(
@@ -77,7 +77,7 @@ export function useRemoveLineItem() {
                cartKeys.cart,
                context?.previousCart
             );
-            addErrorMessage(error);
+            addError(error);
          },
          onSuccess: (data, variables) => {
             const cart = client.getQueryData<Cart>(cartKeys.cart);
--- a/packages/cart-sdk/hooks/use-update-line-item-quantity.ts
+++ b/packages/cart-sdk/hooks/use-update-line-item-quantity.ts
@@ -21,7 +21,7 @@ interface UpdateLineItemQuantityInput {
 export function useUpdateLineItemQuantity() {
    const client = useQueryClient();
    const iFixitApiClient = useIFixitApiClient();
-   const { addErrorMessage } = useCartDrawer();
+   const { addError } = useCartDrawer();
    const mutation = useMutation(
       async (input) => {
          return iFixitApiClient.post(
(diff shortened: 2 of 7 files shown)


deltuh-vee:
dev_block ✊ 
I added a product that didn't have enough stock and the error ended up being overlapped by the Cart icon.
CleanShot 2024-01-15 at 15 15 23

Could you also make the error last a little longer? I feel like it goes away really quickly.

danielcliu-ifixit:
un_dev_block :v:

packages/ui/cart/drawer/CartDrawer.tsx
```diff
@@ -152,14 +169,7 @@ export function CartDrawer() {
...
                      )}
-                     <Fade
-                        show={isCartEmpty}
-                        disableExitAnimation
-                        position="absolute"
-                        w="full"
-                        top="0"
-                        left="0"
-                     >
+                     <Fade show={isCartEmpty} disableExitAnimation w="full">
```
danielcliu-ifixit, line 172: @dhmacs is this ok to change? https://github.com/iFixit/react-commerce/commit/7d2a8bbf70bf878da4844912672fcf9bce1c936e

I couldn't tell a difference animation wise personally, but i also don't know what i was looking for exactly

deltuh-vee:
QA 🎬 
* Errors with cart updates are shown to the user in the cart drawer.
* Checkout errors are still shown.

packages/ui/cart/drawer/CartDrawer.tsx
dhmacs: I think the purpose was that the cart empty state would animate in place rather than slide up from below, but I've tested it and you're right, doesn't seem to make any difference really. I think it's ok to drop it 👍

packages/cart-sdk/hooks/use-update-line-item-quantity.ts
```diff
@@ -91,6 +93,9 @@ export function useUpdateLineItemQuantity() {
                cartKeys.cart,
                context?.previousCart
             );
+            addErrorMessage(
+               (error as string) || 'An error occurred updating the cart.'
+            );
```
dhmacs, line 98: What do you think about passing an error object instead of a string? This would keep the solution open for extensions (by extending error one could pass more information if needed) and would align with the approach used for other error handling situation (e.g. [error boundaries](https://nextjs.org/docs/app/api-reference/file-conventions/error))

```suggestion
            if (error instanceof Error) {
               addErrorMessage(error);
            } else {
               addErrorMessage(new Error('An error occurred updating the cart.'));
            }
```

packages/ui/cart/drawer/CartDrawer.tsx
```diff
@@ -39,12 +39,23 @@ import { useCartDrawer } from './hooks/useCartDrawer';
...
    const isMounted = useIsMountedState();
    const cart = useCart();
    const checkout = useCheckout();
    const isCartEmpty = cart.isFetched && !cart.data?.hasItemsInCart;
-
+   const clearAddToCartErrors = () => clearErrorMessages();
+   const [hasErrors, setHasErrors] = React.useState(errorMessages.length > 0);
+   React.useEffect(() => {
+      setHasErrors(errorMessages.length > 0);
+   }, [errorMessages]);
```
dhmacs, line 58: Correct me if I'm missing something, but I think we don't need `useState` here since `hasErrors` is derived from `errorMessages`. Synchronizing states in React should be avoided as it frequently results in weird sync bugs https://kentcdodds.com/blog/dont-sync-state-derive-it

```suggestion
   const hasErrors = errorMessages.length > 0;
```

packages/ui/cart/drawer/CartDrawer.tsx
```diff
@@ -39,12 +39,23 @@ import { useCartDrawer } from './hooks/useCartDrawer';
...
+      onViewCart,
+      errorMessages,
+      clearErrorMessages,
+   } = useCartDrawer();
    const isMounted = useIsMountedState();
    const cart = useCart();
    const checkout = useCheckout();
    const isCartEmpty = cart.isFetched && !cart.data?.hasItemsInCart;
-
+   const clearAddToCartErrors = () => clearErrorMessages();
```
dhmacs, line 54: Can we use `clearErrorMessages` directly without wrapping it? I think `clearAddToCartErrors` might be misleading as a name, since this would not just clear "add to cart" errors, but really any error displayed in the drawer

dhmacs (approved):
CR ✅ Looks great 👌 

deploy_block 🔢 on considering some minor changes

P.S.: It might be a good idea to clear error messages when the drawer is closed

danielcliu-ifixit:
un_deploy_block :v:  comments were addressed.

packages/ui/cart/drawer/hooks/useCartDrawer.tsx
```diff
@@ -6,17 +6,44 @@ import { useCart } from '@ifixit/cart-sdk';
...
+
+   const value = React.useMemo(
+      (): CartDrawerContext => ({
+         ...disclosure,
+         errors,
+         addError,
+         clearErrors,
+      }),
+      [disclosure, errors]
+   );
```
masonmcelvain, line 46: I think I misunderstand how the eslint `react-hooks/exhaustive-deps` check works, because I would expect it to complain that `addError` and `clearErrors` is not included in the dependency array.

If we were to include those two functions, we would need to wrap them in `useCallback`.

dhmacs (approved):
CR 👍 Looks great 👏

masonmcelvain (approved):
CR 👍🏻

packages/cart-sdk/hooks/use-remove-line-item.ts
```diff
@@ -75,6 +77,7 @@ export function useRemoveLineItem() {
                cartKeys.cart,
                context?.previousCart
             );
+            addError(error);
```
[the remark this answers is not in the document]
dhmacs: I think there is, but we need to upgrade to React Query v5 first:

https://github.com/iFixit/react-commerce/pull/2202#pullrequestreview-1818528934

kthaler:
QA 💎 - The following error is now displayed in the cart when the item is out of stock
Screenshot 2024-01-16 at 11 19 54 AM

deploy_block 🦖 - I am not getting an error when the item goes out of stock right before the user adds an item to their cart. Adding the item to the cart understandably doesn't display an error, since the quantity needs some time to update, but I think there should be an error when the user opens their cart again. EX:

https://github.com/iFixit/react-commerce/assets/145375039/81eff943-3a7b-47bf-88fb-dddce884e1b0

danielcliu-ifixit:
@kthaler i think this would be another issue, as technically we aren't receiving an error from our cart api. I think this is also a very unlikely edge case where we have manually set our inventory db out of sync with shopify's inventory intentionally and I'm unconvinced we should treat this as a case likely to occur in production.

PR #2208, opened 2024-01-16
Memoize app router find product

dhmacs:
Part of https://github.com/iFixit/ifixit/issues/51165

This PR uses React.cache to memoize product request, since we are fetching the product multiple times.

qa_req 0

### DIFF
--- a/frontend/app/(defaultLayout)/@storeSelect/products/app-router/[handle]/page.tsx
+++ b/frontend/app/(defaultLayout)/@storeSelect/products/app-router/[handle]/page.tsx
@@ -1,36 +1,32 @@
-import { DEFAULT_STORE_CODE, IFIXIT_ORIGIN } from '@config/env';
+import { DEFAULT_STORE_CODE } from '@config/env';
 import { getStoreList } from '@models/store';
-import Product from '@pages/api/nextjs/cache/product';
 import { ProductPageProps } from 'app/(defaultLayout)/products/app-router/[handle]/page';
-import { devSandboxOrigin, shouldSkipCache } from 'app/_helpers/app-helpers';
+import { findProduct } from 'app/_data/product';
+import { shouldSkipCache } from 'app/_helpers/app-helpers';
 import { StoreSelect } from '../../../store-select';
+import { invariant } from '@ifixit/helpers';
 
 export default async function ProductPageStoreSelect({
    params,
    searchParams,
 }: ProductPageProps) {
-   const [stores, product] = await Promise.all([getStoreList(), getProduct()]);
+   const [stores, product] = await Promise.all([
+      getStoreList(),
+      findProduct({
+         handle: params.handle,
+         noCache: shouldSkipCache(searchParams),
+      }),
+   ]);
 
    return <StoreSelect stores={product ? storesWithProductUrls() : stores} />;
 
-   async function getProduct() {
-      const result = await Product.get(
-         {
-            handle: params.handle,
-            storeCode: DEFAULT_STORE_CODE,
-            ifixitOrigin: devSandboxOrigin() ?? IFIXIT_ORIGIN,
-         },
-         { forceMiss: shouldSkipCache(searchParams) }
-      );
-      return result?.__typename === 'Product' ? result : null;
-   }
-
    function storesWithProductUrls() {
+      invariant(product);
       return stores.map((store) => {
          if (productEnabledFor(store.code)) {
             return {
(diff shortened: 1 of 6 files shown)


danielbeardsley:
> This PR uses React.cache to memoize product request, since we are fetching the product multiple times.

I was under the impression that Next (when using the app router) wraps the global `fetch()` in a cache which effectively does this same caching (though using the url as the cache key I presume).

I personally don't like the idea of depending on the global `fetch()` cache as the docs are rather hand-wavey and would rather it be an explicit cache (as is done here) or something similar. So in principle, I like this change. However I'm curious why we can't or don't want to rely on the next.js fetch() caching? i.e. what prompted this change?

dhmacs:
> I was under the impression that Next (when using the app router) wraps the global `fetch()` in a cache which effectively does this same caching (though using the url as the cache key I presume).
> I personally don't like the idea of depending on the global `fetch()` cache as the docs are rather hand-wavey and would rather it be an explicit cache (as is done here) or something similar. So in principle, I like this change. 
> However I'm curious why we can't or don't want to rely on the next.js fetch() caching? 

I agree, I don't love that they are overriding the `fetch` api either. I also don't love that this uses Vercel Data Cache which pricing is still unknown. Interestingly `next.config.js` allow us to provide a custom data cache implementation via `incrementalCacheHandlerPath` setting, but I've found out that when deploying on Vercel that setting is ignored 🤦‍♂️  (see discussion: https://github.com/vercel/next.js/discussions/52203#discussioncomment-7904110)

> what prompted this change?

First, this change is due to new data fetching paradigm brought by app router and RSC (read more [here](https://nextjs.org/docs/app/building-your-application/data-fetching/patterns#fetching-data-where-its-needed). We need to fetch the same product data in multiple places during a request (e.g. `generateMetadata` and page component) so memoizing expensive operations is useful.
Second, we're still using our Redis cache and that is not automatically memoized since it's not using `fetch`.

masonmcelvain (approved):
CR 👏🏻

masonmcelvain:
One thing that is unclear to me about [`React.cache`](https://react.dev/reference/react/cache) is how long the cached result is kept for. `cache()` does not accept any arguments to configure the TTL, is this configurable elsewhere? Do the results remain cached for the lifetime of the deployment?

If it is for the lifetime of the deployment, I am concerned that an admin user may make changes to a product in Shopify Admin, and those changes won't be made visible to end users until we redeploy.

I am looking at Vercel's [data cache](https://vercel.com/docs/infrastructure/data-cache) and it explains revalidation behavior for `fetch()` but does not seem to mention `cache()`.

dhmacs:
> One thing that is unclear to me about [`React.cache`](https://react.dev/reference/react/cache) is how long the cached result is kept for. 

Good question! `React.cache` is memoization scoped to the request. The purpose is just to be able to call the same data fetch method in different parts of the component tree without having to centralize it with React context or pass it around with props. It's a fundamental part of the RSC architecture, as it allows Server Components to be isolated even if they share data requirements.
You can read more [here](https://nextjs.org/docs/app/building-your-application/caching#request-memoization). Here's an excerpt:

Screenshot 2024-01-26 at 08 09 19

cc @masonmcelvain

danielbeardsley:
> You can read more here. Here's an excerpt:

Wat?! *React* overloads the global `fetch()` to add caching in certain circumstances *and* next.js overloads it to add its own caching in a different set of circumstances? Or has the global `fetch()` override always been a react thing and we've just *thought* it was nexts.js and they are actually describing the same feature?

masonmcelvain:
> React overloads the global fetch()

As far as I can tell from the React docs this is not true. React docs only _suggest_ using `cache` for memoizing fetches. 

> `cache` is also recommended for memoizing data fetches ([ref](https://react.dev/reference/react/cache#deep-dive-cache))

When the Next.js docs say "Request memoization is a React feature" ([ref](https://nextjs.org/docs/app/building-your-application/caching#request-memoization)), I think they are referring to how "React will invalidate the cache for all memoized functions for each server request" ([ref](https://react.dev/reference/react/cache#caveats)).

masonmcelvain:
Next.js does explicitly [say](https://nextjs.org/docs/app/building-your-application/caching#request-memoization) "React extends the [fetch API](https://nextjs.org/docs/app/building-your-application/caching#fetch) to automatically memoize requests that have the same URL and options" but I can't find that anywhere in React's documentation.

PR #2211, opened 2024-01-18
Bump sentry to 7.93.0

federicobadini:
[The Sentry issue](https://github.com/getsentry/sentry-javascript/issues/6726#issuecomment-1860473544) tracking Nextjs app router support was closed on Dec 18.
As far as we can read from there, at that time the majority of the work to support app router was completed.
Given that we are still on `7.80.0` which was released before Dec 18, here we are bumping Sentry to the last version available which is 7.93.0.

The [CHANGELOG](https://github.com/getsentry/sentry-javascript/blob/master/CHANGELOG.md) is not describing any breaking change for the transition

qa_req 0


masonmcelvain (approved):
CR 👏🏻

PR #2212, opened 2024-01-18
Product: Make tooltips visible to screen readers

masonmcelvain:
Closes https://github.com/iFixit/ifixit/issues/51384

Makes the shipping restriction and returns notice visible to screen readers. Also adds aria label text to each.

This is similar to what we do for the prop 65 warning.

Chakra's docs recommend that the popover trigger be focusable, hence the use of IconButton. https://chakra-ui.com/docs/components/popover/usage#basic-usage

## QA

Visit the Vercel preview and make sure that the shipping restrictions and returns info buttons are not skipped by screen readers.


jordycosta:
QA 🟢 

The shipping restriction and 30 day returns ℹ️ buttons are being read by screen readers 👍

davidrans (approved):
CR :coffee:

PR #2215, opened 2024-01-22
Troubleshooting: Add causes header

aburke07:
## Overview
We display a 'Causes' header in the TOC, but don't actually show one in the troubleshooting page content. This adds one in the same style as the other headers.

We don't want to be able to permalink to this, so it should just be an h2 without special link properties.

If there are no causes to display, we don't render the header.

Closes https://github.com/iFixit/ifixit/issues/51468

### DIFF
--- a/frontend/templates/troubleshooting/index.tsx
+++ b/frontend/templates/troubleshooting/index.tsx
@@ -551,14 +551,15 @@ function IntroductionSection({
 
 function CausesListHeader() {
    return (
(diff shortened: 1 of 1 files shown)


ianrohde (approved):
CR ⚪

kthaler:
QA 💎 - There is now a 'Causes' h2 header on all the troubleshooting pages, until there are no causes listed.

PR #2216, opened 2024-01-23
Add app router product page cache control headers

dhmacs:


jordycosta:
QA 🟢 

https://react-commerce-git-setup-app-router-caching-ifixit.vercel.app/products/app-router/surface-display-bonding-frame-m1260233-001

Without `?disableCacheGets` param:
**X-Vercel-Cache: HIT**

With `?disableCacheGets` param:
**X-Vercel-Cache: MISS**

danielbeardsley:
Good job figuring out a way to make this happen. It sure feels like this is getting a bit ridiculous though. We're being forced to do more and more weird things to be able to support what feels like normal functionality.

dhmacs:
@danielbeardsley Agree, I'm still puzzled by the design decisions that they made. They're trying to be overly clever about caching. What makes me mad is that it seems to be designed to push you toward relying on the Data Cache, which eventually they will charge you for. Having said that I don't think we have many other options right now

jarstelfox:
>What makes me mad is that it seems to be designed to push you toward relying on the Data Cache

Yep! You can use the [react cache function](https://react.dev/reference/react/cache) to cache the output of some component, but not cache the entire page in a user's browser.


>I don't think we have many other options right now

Agreed from what I have seen 😢

sterlinghirsh (approved):
CR :zap: I read the whole discussion and I think it's interesting. It seems like a lot of the restrictions come from using streaming. Some of the comments indicated streaming won't even play well with a CDN since the CDN caches the streamed version of the page which includes placeholders and extra script tags. We'll definitely want to check that out and make sure we're not caching the wrong thing. That said, since this affects the app-router product page only, I'm fine with merging it. We may have to make a CloudFront change before we can properly test this on the ifixit.com domain though.

dhmacs:
> Some of the comments indicated streaming won't even play well with a CDN since the CDN caches the streamed version of the page which includes placeholders and extra script tags

It seems a non-issue. This PR preview shows that when you get a CDN hit, you get the page back fast (~100ms) so I'm not sure what is really problematic about caching a stream. 

> We may have to make a CloudFront change before we can properly test this on the ifixit.com domain though.

Do we need to explicitly forward headers on cloudfront? If so we need to forward `CDN-Cache-Control`

PR #2229, opened 2024-02-02
Refactor store pages metadata to app router format

dhmacs:


masonmcelvain (approved):
CR 👍🏻

deltuh-vee:
QA 🎬 
* Meta title is the same.
* Meta description is the same.
* Open graph meta tags are otherwise the same with title and description added.

PR #2232, opened 2024-02-07
Troubleshooting: Match navtab style to /Parts page

aburke07:
## Overview
Reuses the components from the /Parts nav tabs on troubleshooting view pages. Now the styles are consistent across the tabs.

## CR
1. Do we want to extract this into a single component that we pass the links / names / current link into? Then we can use that component in both places.
2. Do we also need to make a change to [this file](https://github.com/iFixit/react-commerce/blob/dad3de2c0a0594932ce8c3ab198a13d63c3c87cf/frontend/app/(defaultLayout)/Troubleshooting/%5Bdevice%5D/components/NavBar.tsx#L159-L188)? I couldn't get this to affect the rendered page when editing it.

## QA
- The navtabs on /Troubleshooting should now match the tabs on /Parts.
- When there is no /Parts or /Guides for a device, we disable the link.

Closes https://github.com/iFixit/ifixit/issues/51876

### DIFF
--- a/frontend/templates/troubleshooting/components/NavBar.tsx
+++ b/frontend/templates/troubleshooting/components/NavBar.tsx
@@ -1,12 +1,10 @@
 import {
    Box,
-   BoxProps,
    Button,
    Flex,
    FlexProps,
    IconButton,
    Link,
-   LinkProps,
    Menu,
    MenuButton,
    MenuItem,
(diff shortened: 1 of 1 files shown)


ianrohde:
> Do we also need to make a change to [this file](https://github.com/iFixit/react-commerce/blob/dad3de2c0a0594932ce8c3ab198a13d63c3c87cf/frontend/app/(defaultLayout)/Troubleshooting/%5Bdevice%5D/components/NavBar.tsx#L159-L188)? I couldn't get this to affect the rendered page when editing it.

That's the Problems List component (e.g. `/Troubleshooting/dryer` ect), and the main trigger for setting up a shared component. There are differences though. Maybe we should punt that to a separate issue to scope and take it on as time permits.

jarstelfox:
Found the convo: https://github.com/iFixit/ifixit/discussions/46122#discussioncomment-4898813

aburke07:
Ah, bummer. Closing this because the changes are significantly different.

PR #2235, opened 2024-02-08
Add app router troubleshooting page under feature flags

dhmacs:
Part of https://github.com/iFixit/ifixit/issues/52003

qa_req 0


danielbeardsley:
I don't remember the specifics of the routing system priorities, do pages routes get matched first? i.e. does having this file exist end up making the pages router inaccessible?

CR :+1: otherwise

deploy_block :+1:  on the question

danielbeardsley (approved):
I don't remember the specifics of the routing system priorities, do pages routes get matched first? i.e. does having this file exist end up making the pages router inaccessible?

CR :+1: otherwise

deploy_block :+1:  on the question

danielbeardsley:
un_deploy_block :+1: I can still access troubleshooting urls and thus the pages router must take precedence.

danielbeardsley:
> must take precedence.

Hmm.. but if that's the case, what does the env var buy us as it's not changing the precedence order it seems?

danielbeardsley:
Turns out, I missed that this new route starts with `/app-router/Troubleshooting` and thus is a separate route from the existing page. The env var helps by just hiding it in prod so it doesn't get seen or indexed.

un_deploy_block :+1:

PR #2236, opened 2024-02-08
Sentry: temp disable production check

danielbeardsley:
This won't work the same way once this code is in a private repo.

Let's disable this while we move this code instead of trying to re-solve this problem beforehand.

qa_req 0


jarstelfox (approved):
CR 🌵 but you will need to solve the import as well

masonmcelvain (approved):
CR 👍🏻

danielbeardsley:
> but you will need to solve the import as well

Thanks! Fixed.

jarstelfox (approved):
CR 🌵

masonmcelvain (approved):
CR 👍🏻

